Issue #3475, opened 2019-09-13
Adapt update.sh for GitHub


gunchleoc:
Rewrite update.sh to pull from GitHub

PR #3476, opened 2019-09-14
Adapt update.sh for GitHub. #3475

gunchleoc:


klaus-halfmann (dismissed):
Mhhh, I am not sure if this will ork on a branch?
We still cn use it as a start for now, lets find out.

gunchleoc:
> Mhhh, I am not sure if this will ork on a branch?
> We still cn use it as a start for now, lets find out.

You can test if from this branch. It should check if the branch is clean and then switch to master before pulling.

I have just experimented another addition to add a safeguard to the update script if the user is already on master, but the working directory is not clean. Bad idea though, we better rely on git to do that for us.

klaus-halfmann:
Mhh, when we use such forked / private repos we must alway pull the coplete history.
If we use branches from master in the main project its easy to pull tem into a worktree.
Or is there a way to derive such a private branch from its origin?

klaus-halfmann (changes requested):
OK I had to:
git clone git@github.com:gunchleoc/widelands.git
git fetch origin update-script
git checkout update-script

> First check if this is an bzr checkout at all - only in that case
<  First check if this is an git checkout at all - only in that case

gunchleoc:
> OK I had to:
> git clone [git@github.com](mailto:git@github.com):gunchleoc/widelands.git

The original script also required you to `bzr branch lp:widelands` first, so no change there.

> git fetch origin update-script
> git checkout update-script

Yes, you had to do that because this branch hasn't been merged into master yet.

compile.sh
```diff
@@ -329,7 +331,10 @@ if ! [ -f src/wlapplication.cc ] ; then
   exit 1
 fi
 
-bzr pull
+# Checkout master and pull latest version
+git checkout master
```
stonerl: Shouldn't this be:
```
git checkout
```
We want to checkout the current branch we are working on and pull in the latest changes from master, won't we? Otherwise, we switch to the master branch.

compile.sh
```diff
@@ -329,7 +331,10 @@ if ! [ -f src/wlapplication.cc ] ; then
...
+git pull https://github.com/widelands/widelands.git master
```
stonerl, line 336: Here we should use
```
git pull --rebase https://github.com/widelands/widelands.git master
```

stonerl (changes requested):

compile.sh
gunchleoc: I actually don't know who the target audience is and how the script is being used. I'd be fine with not switching to master. I have changed this line.

compile.sh
gunchleoc: My experience with rebase is that sometimes you have to resolve the same conflict again and again and again during 1 rebase process if there are multiple commits involved. And if you don't understand them, you have to abort, which means learning how to abort this on the command line. This make the learning curve far too high, so I think merging would be easier here.

stonerl:
@gunchleoc I added a change to the revision detection to match it to the one in #3488

stonerl (approved):

gunchleoc:
I have tested your changes and it's working. So, let's have it :)

utils/detect_revision.py
```diff
@@ -44,14 +44,21 @@ def detect_debian_version():
...
+        cmd = subprocess.Popen(
+            ['git', 'rev-parse', '--abbrev-ref', 'HEAD'],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE, cwd=base_path
+        )
+        stdout, stderr = cmd.communicate()
+        git_abbrev = stdout.rstrip()
         if git_revnum:
-            return 'unofficial-git-%s' % git_revnum
+            return 'git-%s[%s]' % (git_revnum, git_abbrev)
```
SirVer, line 61: Late to the party, but I'd like to repeat here again that the branch name is a local construct. Including it in the revision is meaningless outside of your local checkout. I think it is a mistake to use it in the revision identifiers, since it is not unique. Let's stick to number of commits (for lexical sorting) & sha for identifying.

gunchleoc: Although it is meaningless, the branch name will help us when low-tech people make bug reports on what they have been testing. If they have been testing a branch, we will automatically know that fro the version info.

stonerl:
We could have to schemes.
For master
```git-count[sha1]```

For other branches
```git-sha1[branch-name]```

gunchleoc:
Sounds good to me :)

Issue #3430, opened 2019-09-09
Problem with --homedir command line option in Windows


bunnybot:
See https://wl.widelands.org/forum/topic/4513/

Homedir is fixed, but there is still a problem with the localedir (see #4)

------------------------------------
Imported from Launchpad using lp2gh.

 * date created: 2019-05-09T07:49:28Z
 * owner: gunchleoc
 * assignee: gunchleoc
 * the launchpad url was https://bugs.launchpad.net/bugs/1828355

bunnybot:
(by franku)
With linux the --homedir option works as expected.

bunnybot:
(by fuchur77)
The problem is the windows logger which is initialized in WLApplication::setup_homedir(). But there is a call of log() in WLApplication::WLApplication(int const argc, char const* const* const argv) before that happens. See source below

	if (commandline_.count("homedir")) {
		log("Adding home directory: %s\n", commandline_["homedir"].c_str());  // <-- this is it
		homedir_ = commandline_["homedir"];
		commandline_.erase("homedir");
	}
#ifdef REDIRECT_OUTPUT
	if (!redirect_output())
		redirect_output(homedir_);
#endif

	setup_homedir();

If the labeled line is changed to print the output using std::cout (maybe encapsuled by #ifdef _WIN32) or commented it works.

Windows doesn't have a stdout if you run a program using a shortcut. Therefore log() writes to a file stdout.txt which is saved in the homedir folder. I guess in Linux starting widelands from a shell stdout is always accessible -> no problem with log() before setting the homedir.

I assume in Linux one should see two messages about the home directory if one uses the homedir option. One stating "Adding home directory:" and the second one "Set home directory:".

bunnybot:
(by gunchleoc)
I have replaced the log command with cout in https://bazaar.launchpad.net/~widelands-dev/widelands/trunk/revision/9105

Is it working well enough now, or do we need to do more fixes?

bunnybot:
(by fuchur77)
Ok, I made a quick test, the --homedir option is recognised and the given directory is created. A config and stdout.txt are created as well. Starting a game and saving one works. The directories ai, save, replays and temp are then created.

But clicking on options results in an error:

Unexpected error during the game
FileSystem::create: could not find file or directory: F:\widelands\widelands-bzr\hgs-homedir\data/locale

[...]

PR #3477, opened 2019-09-14
Fix localedir on Windows. #3430

gunchleoc:
- Ensure that "locale" exists in the datadir
- Show an error message box and exit gracefully when it doesn't
- Shift creating of locale directory from compile.sh to CMake


gunchleoc:
We still need to call ensure_directory_exists for the localedir, and maybe add a warning box

gunchleoc:
I have now tested this on Windows and it works there.

stonerl (dismissed):
LGTM :-)

Can't test this, since I don't have a Windows machine.

gunchleoc:
Thanks for the review!

stonerl:
@klaus-halfmann why are you closing this PR?

klaus-halfmann:
Looks like a git push deletes branches in some cases, strange.

stonerl:
Only if you force push a locally deleted branch. And from the logs I can see you did a force push. Did you push your entire local repo?

TinoM:
@gunchleoc 
Sorry for the delay, will do resolve the conflict the next days and do some tests on my windows machine.

TinoM:
Hm, "cmake -E create_symlink" seems to be problematic on windows:
https://gitlab.kitware.com/crondog/cmake/commit/afb7f6e4ff8a0c1a68c8d8cf2d6cf72401cfb8ff

At least on my machine (Win 10 1903) cmake fails at this step.

CMakeLists.txt
```diff
@@ -388,6 +388,12 @@ install(
   PATTERN "CMakeLists.txt" EXCLUDE
 )
 
+# Create and link locale dir. We always need one even if it's empty, because Widelands is checking for it to prevent crashes.
+add_custom_target(link_locale_dir ALL
+    COMMAND ${CMAKE_COMMAND} -E make_directory locale
```
TinoM, line 393: @gunchleoc : Is this really necessary? "create_symlink" does only work on windows with administrative privileges.
And: Even if there is a symlink to the locale data, you'll still have to start the binary explicitely with "--datadir".

I think it would be better to add a "--localedir" parameter or just describe in the build documentation that (if the dev wants to run with locales) to copy the files or make the symlink manually.

The current solution with the symlink is not platform independant.

gunchleoc:
I still had a local copy of the branch, so I have created a new PR using my fork https://github.com/widelands/widelands/pull/3629 because this branch is destroyed.

Issue #3459, opened 2019-09-09
Open dropdowns don´t close when closing parent window


bunnybot:
To reproduce:
– Open the editor
– Press P to open the players menu
– Click on the dropdown to open it
– Press P again (while the dropdown is open)
Result: The window closes but the dropdown remains visible and can´t be removed

------------------------------------
Imported from Launchpad using lp2gh.

 * date created: 2019-08-26T20:41:54Z
 * owner: nordfriese
 * assignee: gunchleoc
 * the launchpad url was https://bugs.launchpad.net/bugs/1841471

gunchleoc:
PR: #3478

PR #3478, opened 2019-09-14
Fix deleting of dropdown lists in UniqueWindow and Player Menu. #3459

gunchleoc:


Notabilis:
I was able to reproduce the bug and the fix seems to work. Playing around a bit with various dropdown menus haven't showed any errors.
No objections to the code either, though I didn't delved too deep into it.

Notabilis (approved):

PR #3484, opened 2019-09-16
closes #3445

### LINKED ISSUE #3445: Rework programs for trainingssites
During my sleep time unification I figured out that there are some flaws with the programs for trainingssites.

Some sites have work animations some don't. That means sites without work animations have to wait 30s before they can fail because they miss food or a soldier has been removed. While those with a working animation, only have to wait 15 seconds.

Also in case of the barbarians battle arena. The working animation comes before the soldier training and after consume. This means that within a window of 15 seconds a soldier can be removed from the building without even being trained.

We should change it to a program flow like in the mines. Here a pseudo code example for the barbarians battle arena.

main prog
{ sleep 5000
  call=sleep
  call=train_evade_0
  call=animate
  call=sleep
  call=train_evade_1
  call=working
  return=no_stats
}

sleep prog {
   sleep 15000
   return=no_stats
}

working animate {
   animate=working 15000
   return=no_stats
}

evade0 prog {
   check soldier
   consum wares
   train soldier
}

I would add this to the unification branch
evade1 prog {
   check soldier
   consum wares
   train soldier
}

 * date created: 2019-05-30T11:44:58Z
 * owner: stonerl
 * the launchpad url was https://bugs.launchpad.net/bugs/1831066

stonerl:
Split animation times in half where possible to make the screen less "noisy".


Noordfrees:
For the frisian sites, I´d suggest to increase "working" and decrease "sleeping" animation times a bit so all "working" anim times are multiples of 1000 ms, which will look nicer.

Noordfrees (approved):
Good :)

Issue #3471, opened 2019-09-09
Automate utils/fix_formatting.py


gunchleoc:
Until now, bunnybot took care of running 'utils/fix_formatting.py` on every merge. We need to find a new solution for this.

If all else fails, have one designated person running it manually on their machine - `clang-format` results differ from machine to machine, so we can't just let everybody run it.

gunchleoc:
This script should reduce the fallout if people run it locally: https://github.com/barisione/clang-format-hooks/

stonerl:
Hmmm, maybe we can integrate this into a webhook?

stonerl:
I try to incorporate this into a git-hook.

stonerl:
Do we just want to add the scripts or add the repo as a git sub module?

gunchleoc:
We cannot have developers run this on their local machines. The clang-format results will change depending on the machine, resulting in useless merge conflicts :(

stonerl:
Why does this change depending on the machine? I'm currently testing this. It utilizes the .clang-format file in the root folder. The developers just need to add a pre-commit hook. I'll show you what I mean in my PR.

gunchleoc:
Because every time I ran it, the results differed from what Bunnybot had produced. It seems that different versions of clang-format interpret the .clang-format file differently.

gunchleoc:
I can't reproduce the problem with my machines any more, so I guess that they have fixed it.

The submodule makes it easy for maintenance, but it will be scary for newcomers, because now they have to learn how to deal with submodules too. So, I'd rather add this in a directory under utils with a link to the original repo.

The lua tabs formatting still produces differences, but clang-format doesn't any more.

We also run pyformat, which we could set up to run only if there are changed python files.

gunchleoc:
Fixed in b1f5bebf4010f2e41fe6bb92fd2c397ac6357f9c

PR #3485, opened 2019-09-16
#3471 add clang format hook

stonerl:
Added git sub module for clang format.
Added README.md and description for registering the hooks.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "utils/clang-format-hooks"]
+	path = utils/clang-format-hooks
+	url = git@github.com:barisione/clang-format-hooks.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "utils/clang-format-hooks"]
-	path = utils/clang-format-hooks
-	url = git@github.com:barisione/clang-format-hooks.git
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 Run at a terminal:
 
 ```sh
-$ ./git-pre-commit-format install
+$ ./utils/clang-format-hooks/git-pre-commit-format install
 ```
 
 This will add a hook (in `.git/hooks/`) which is run every time you commit.
--- a/README.md
+++ b/README.md
@@ -1,0 +1,31 @@
+[![Build status](https://api.travis-ci.org/widelands/widelands.svg?master)](https://travis-ci.org/widelands/widelands)
+
+### Registering the `git` hook [1]
+
+Run at a terminal:
+
+```sh
+$ ./git-pre-commit-format install
+```
+
+This will add a hook (in `.git/hooks/`) which is run every time you commit.
+
+###Using the pre-commit hook
+
+After registering the git hook, you don't need to do anything except for committing your changes.
+
+Every time you commit, the hook will check if your code matches the coding standard. If it doesn't, you get asked what to do and you can decide to:
+
+* Apply the fixes automatically (only to the code you are actually committing, not to unstaged code).
+* Commit anyway.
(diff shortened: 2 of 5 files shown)


gunchleoc:
I have done some testing. The problem is that it's interactive, which seems nice on the command line, but it will make GUI clients choke on the commit.

stonerl:
```
git config hooks.clangFormatDiffInteractive false
```
Solves this. I will add it to the Readme.

gunchleoc:
Making it non-interactive breaks committing on the command line (see below).

Maybe it would be easier to use our own algorithm from utils/fix-formatting.py and apply it only on changed files.

```
widelands (3471-add-clang-format-hook)$ git reset --hard origin/3471-add-clang-format-hook 
HEAD is now at 14894dfc7f add clang-format hooks
widelands (3471-add-clang-format-hook)$ git pull
Already up to date.
widelands (3471-add-clang-format-hook)$ ./utils/clang-format-hooks/git-pre-commit-format install
Pre-commit hook installed.
widelands (3471-add-clang-format-hook)$ git config hooks.clangFormatDiffInteractive false
widelands (3471-add-clang-format-hook)$ git diff
diff --git a/src/main.cc b/src/main.cc
index 820f6d2b42..81873deee6 100644
--- a/src/main.cc
+++ b/src/main.cc
@@ -76,5 +76,6 @@ int main(int argc, char* argv[]) {
 
                return 1;
        }
+       // xxxxxxxxxxxxxxxxx // xxxxxxxxxxxxxxxxx // xxxxxxxxxxxxxxxxx // xxxxxxxxxxxxxxxxx // xxxxxxxxxxxxxxxxx // xxxxxxxxxxxxxxxxx
 #endif
 }
widelands (3471-add-clang-format-hook)$ git commit -am test
The staged content is not formatted correctly.
You can fix the formatting with:
    $ ./utils/clang-format-hooks/apply-format --apply-to-staged

You can also make this script interactive (if you use git from a terminal) with:
    $ git config hooks.clangFormatDiffInteractive true
widelands (3471-add-clang-format-hook)$ git status
On branch 3471-add-clang-format-hook
Your branch is up to date with 'origin/3471-add-clang-format-hook'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   src/main.cc

no changes added to commit (use "git add" and/or "git commit -a")
widelands (3471-add-clang-format-hook)$ 

```

stonerl:
It's not breaking, but one has to choose one or the other. Committing form the command line needs the interactive mode, using a GUI needs it turned off, that's what the documentation says, anyway. I'll add this to our Readme. Since it is optional I'd say we keep it and work on our tools as well. In the end we want contributers to commit clean code, won't we?

gunchleoc:
I will definitely be using both the command line for small commits and a GUI tool for complicated commits, and I don't want to have to continually switch this on and off. -1 on that hassle.

We have a working algorithm in utils/fix-formatting.py that does not have this problem, and we should use that idea as a base. It is possible to get a list of changed files from git and then to run clang-format on those that end in .h or .cc.

stonerl:
That's true. From what I read, a git hook can be a python script as well so maybe we can adapt our script to work as a commit-hook.

gunchleoc:
I wouldn't use it as-is though, because it format everything. We will only want to format the changed C++ files, because:

* The Lua tab fixing produces different results on my machine from what bunnybot did, so I don't want to end up with an edit war on those files blowing up the repository
* People would have to install pyformat before they can commit anything
* Formatting only changed files means less time waiting for the commit to go through
* We will have changes to Lua and Python less often, so my formatting everything once a week when I update the translations should suffice.

stonerl:
Closing this one, for the time being.

PR #3486, opened 2019-09-16
Fix time display in General Statistics

### LINKED ISSUE #3457: Incorrect time on the X axis in General Statistics window
Continuation of bug #1243700 - sorry for the confusion

The time on the X axis in the General Statistics window is shown incorrectly if the slider is set to "game" and the game length is between 32 and 34 hours, or between 36 and 38 hours. See the attached screenshot (left side): the game has been going for over 32 hours but the leftmost label on the time axis shows -26 hours, much shorter. After applying my patch it correctly shows -34.

 * date created: 2019-08-13T22:14:26Z
 * owner: niektory
 * the launchpad url was https://bugs.launchpad.net/bugs/1840087

Niektory:
Fixes #3457
This fixes the problem that the time on the X axis in the General Statistics window is shown incorrectly if the slider is set to "game" and the game length is between 32 and 34 hours, or between 36 and 38 hours.
The fix decreases how_many_ticks until it is a divisor of max_x, because the label drawing code assumes it is the case and doesn't work properly if it isn't.

### DIFF
--- a/src/wui/plot_area.cc
+++ b/src/wui/plot_area.cc
@@ -231,6 +231,11 @@ void draw_diagram(uint32_t time_ms,
 	// Make sure we haven't more ticks than we have space for -> avoid overlap
 	how_many_ticks = std::min(how_many_ticks, calc_plot_x_max_ticks(inner_w));
 
+	// Make sure how_many_ticks is a divisor of max_x
+	while (max_x % how_many_ticks != 0) {
+		how_many_ticks--;
+	}
+
 	// Draw coordinate system
 	// X Axis
 	dst.draw_line_strip({Vector2f(kSpacing, inner_h - kSpaceBottom),


src/wui/plot_area.cc
```diff
@@ -231,6 +231,11 @@ void draw_diagram(uint32_t time_ms,
 	// Make sure we haven't more ticks than we have space for -> avoid overlap
 	how_many_ticks = std::min(how_many_ticks, calc_plot_x_max_ticks(inner_w));
 
+	// Make sure how_many_ticks is a divisor of max_x
+	while (max_x % how_many_ticks != 0) {
```
gunchleoc, line 235: This should be more efficient:

```
if (max_x % how_many_ticks != 0) {
    how_many_ticks -= max_x % how_many_ticks;
}
```

gunchleoc (changes requested):

src/wui/plot_area.cc
Niektory: No, this is wrong. Notice we're decreasing the divisor, not the dividend.

gunchleoc: I don't have my math hat on right now - is there no solution that avoids having to have a loop?

Niektory: I can't think of a simple way, and I see no need for it either. This is not performance critical code, shaving off a few machine instructions will make no difference.

gunchleoc: OK, let's keep it like it is then.

gunchleoc: OK, let's have it then. Travis failure is just the usual timeout, so I'll merge this.

PR #3487, opened 2019-09-16
use proper in-/decrement operators

### LINKED ISSUE #1681: Replace +=1 with increment operator
We have lots of lines of code that look like this:

    variable += 1;

I think these should be replaced with

    ++variable;

where appropriate, because that's easier to read.

Same for

    variable -= 1;

 * date created: 2017-06-22T06:35:06Z
 * owner: gunchleoc
 * the launchpad url was https://bugs.launchpad.net/bugs/1699684

stonerl:
Re: #1681


gunchleoc (dismissed):
Looks good :)

Please leave the issue open, because we will want a codecheck rule to prevent this from regressing.

gunchleoc (approved):

PR #3488, opened 2019-09-16
Update build_app.sh

stonerl:
version string for development builds contains <branch>-<commits>

I'd propose that we use this as a general version scheme for development builds.

### DIFF
--- a/utils/macos/build_app.sh
+++ b/utils/macos/build_app.sh
@@ -2,7 +2,7 @@
 
 set -e
 
-USAGE="Usage: $0 <clang|gcc|gcc6> <debug|release> <bzr_repo_directory>"
+USAGE="Usage: $0 <clang|gcc|gcc6> <debug|release> <git_repo_directory>"
 USE_ASAN="OFF"
 
 if [ ! -z "$3" ]; then
@@ -77,7 +77,7 @@ if [ ! -d "$SDK_DIRECTORY" ]; then
    fi
 fi
 
-REVISION=`bzr revno $SOURCE_DIR`
+REVISION=`git --git-dir=$SOURCE_DIR/.git rev-parse --abbrev-ref HEAD`-`git --git-dir=$SOURCE_DIR/.git rev-list --count HEAD`
 DESTINATION="WidelandsRelease"
 
 if [[ -f $SOURCE_DIR/WL_RELEASE ]]; then
--- a/utils/macos/build_app.sh
+++ b/utils/macos/build_app.sh
@@ -77,13 +77,14 @@ if [ ! -d "$SDK_DIRECTORY" ]; then
    fi
 fi
 
-REVISION=`git --git-dir=$SOURCE_DIR/.git rev-parse --abbrev-ref HEAD`-`git --git-dir=$SOURCE_DIR/.git rev-list --count HEAD`
+REVISION=`git rev-parse --short=5 HEAD`[`git --git-dir=$SOURCE_DIR/.git rev-parse --abbrev-ref HEAD`]
+COMMIT_COUNT=`git --git-dir=$SOURCE_DIR/.git rev-list --count HEAD`
 DESTINATION="WidelandsRelease"
 
 if [[ -f $SOURCE_DIR/WL_RELEASE ]]; then
    WLVERSION="$(cat $SOURCE_DIR/WL_RELEASE)"
 else
-   WLVERSION="r$REVISION"
+   WLVERSION="git-$REVISION"
 fi
 
 echo ""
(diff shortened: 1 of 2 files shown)


gunchleoc:
Looks good, but I can't test it, because I don't have a Mac.

BTW you don't need to request reviews from me personally, this only means that I will get a second e-mail about the same pull request ;)

utils/macos/build_app.sh
```diff
@@ -77,13 +77,15 @@ if [ ! -d "$SDK_DIRECTORY" ]; then
    fi
 fi
 
-REVISION=`bzr revno $SOURCE_DIR`
+REVISION=`git rev-parse --short=5 HEAD`[`git --git-dir=$SOURCE_DIR/.git rev-parse --abbrev-ref HEAD`]
```
SirVer: Not sure why you'd want the branch in there. The SHA already is unique and identifiable. Also, your 'master' branch could be my 'awesome' branch, git does not guarantee uniqueness of branch names across checkouts. I suggest dropping the branch name.

utils/macos/build_app.sh
```diff
@@ -77,13 +77,15 @@ if [ ! -d "$SDK_DIRECTORY" ]; then
...
 DESTINATION="WidelandsRelease"
 
 if [[ -f $SOURCE_DIR/WL_RELEASE ]]; then
    WLVERSION="$(cat $SOURCE_DIR/WL_RELEASE)"
+   COMMIT_COUNT=$WLVERSION
 else
-   WLVERSION="r$REVISION"
+   WLVERSION="git-$REVISION"
```
SirVer: `git-r12354` is not unique. You have to mention the SHA somehow. see above for my suggestion, here simply `$REVISION`. I also do not think `git` adds something here, since we already do not use revisions and shas for official tagged builds. Just drop `git-`?

utils/macos/build_app.sh
SirVer: My suggestion remains `"r$(git rev-list --count HEAD)_$(git rev-parse --short HEAD)"`. Having the number of commits in there makes lexicographical sorting of consecutive builds possible, the SHA though is the real identifier. I strongly suggest sticking with `short=7`, so that the numbers are similar than on github. Which is very useful for recognition.

utils/macos/build_app.sh
```diff
@@ -108,9 +110,9 @@ function MakeDMG {
 
    echo "Creating DMG ..."
    if [ "$TYPE" == "Release" ]; then
-      hdiutil create -fs HFS+ -volname "Widelands $WLVERSION" -srcfolder "$DESTINATION" "$UP/widelands_${OSX_MIN_VERSION}_$WLVERSION.dmg"
+      hdiutil create -fs HFS+ -volname "Widelands $WLVERSION" -srcfolder "$DESTINATION" "$UP/widelands_${OSX_MIN_VERSION}_r${COMMIT_COUNT}.dmg"
```
SirVer: keep unchanged as WLVERSION if you follow my above reasoning.

utils/macos/build_app.sh
gunchleoc: Although it is meaningless, the branch name will help us when low-tech people make bug reports on what they have been testing. If they have been testing a branch, we will automatically know that fro the version info. Branch names weren't forced to be the same on Bazaar either.

+1 on being similar to GitHub

stonerl:
Since we already merged the other branch, I made both changes here. I currently have two schemes. One for master one for other branches. As I suggested [here](https://github.com/widelands/widelands/pull/3476#issuecomment-535822947).

If we want a unified scheme, I made this suggestion in the forum:
```
git-count[sha1@branchname]
```
For this branch, it would be:
```
git-24238[c8f77ad@adapt-build_app.sh-for-GitHub]
```

for current master:
```
git-24233[f257b18@master]
```

We could strip the git part, though. And use an *r* or *v*.
```
r24233[f257b18@master]
```
```
v24233[f257b18@master]
```
I also think, keeping the branchname makes it easier for non-technical minded user. But I could live without it.

gunchleoc:
+1 for using r or v. It will save us 3 letters.

utils/macos/build_app.sh
SirVer: Okay, good arguments. I concur.

utils/macos/build_app.sh
SirVer: WLVERSION seems unused now. Remove?

SirVer:
I like `r24233[f257b18@master]` and came around to the idea of branch names. Let's ship this!

utils/macos/build_app.sh
```diff
@@ -77,13 +77,20 @@ if [ ! -d "$SDK_DIRECTORY" ]; then
    fi
 fi
 
-REVISION=`bzr revno $SOURCE_DIR`
+
+REVISION=`git --git-dir=$SOURCE_DIR/.git rev-parse --abbrev-ref HEAD`
```
SirVer: Why is this not calling out to utils/detect_revision.py?

utils/macos/build_app.sh
SirVer: Remember to remove the git-

SirVer (approved):
lgtm and works on my machine. Let's ship it and get nightlies out again!

Issue #3162, opened 2019-09-09
Implement support for Spritemaps


bunnybot:
Since we now have a zoom feature, it would be nice to have higher-resolution graphics. This will mean an increase in filesize though, so we should have support for spritemaps.

There have been 2 approaches to implement utils/make_spritemap.py this so far:

1. A grid that joins the images together (trunk)
2. A compact spritemap that analyzes the image content and saves a lot of space (lp:~widelands-dev/widelands/spritemaps_testing)

I have attached an image comparing the 2 approaches.

I have pretty much implemented the C++ part for the second approach, but the python script still needs work if we want to use that. I would be in favor of option 2 if we can fix the script.

Either script would need adjusting so they will write to Lua format rather than the old conf format.

# Implementation steps

- [x] Implement mipmaps https://en.wikipedia.org/wiki/Mipmap
- [ ] Implement simple spritesheets (option 1). Bob animations have big changes from frame to frame, so it won't be possible to make them more compact. #3489
- [ ] Convert graphics
- [ ] Write a test https://github.com/gunchleoc/widelands/tree/test-animations
- [ ] Implement compact spritesmaps (Option 2). These will be useful for immovables. https://github.com/gunchleoc/widelands/tree/spritemaps-experiment

------------------------------------
Imported from Launchpad using lp2gh.

 * date created: 2016-11-01T18:48:55Z
 * owner: gunchleoc
 * assignee: gunchleoc
 * the launchpad url was https://bugs.launchpad.net/bugs/1638356

bunnybot:
(by gunchleoc)

bunnybot:
(by sirver)
I am also in favor of 2, but instead of using the python script, I would rewrite it in c++. When this script was started, our build system did not easily allow multiple binaries to be built. And the code could benefit from having access to the graphics system and the lua dictionary parsing that is already implemented in c++.

bunnybot:
(by gunchleoc)
+1 to #2

bunnybot:
(by gunchleoc)
My current plan is the following:

1. Implement mipmaps https://en.wikipedia.org/wiki/Mipmap

2. Implement simple spritesheets (option 1). Bob animations have big changes from frame to frame, so it won't be possible to make them more compact.

3. Now we can start converting graphics

4. Implement compact spritesmaps (Option 2). These will be useful for immovables.

bunnybot:
(by gunchleoc)
See also https://bugs.launchpad.net/widelands/+bug/1121982

[...]

PR #3489, opened 2019-09-17
Add support for spritesheets #3162

gunchleoc:
- Add support for spritesheets and refactor animation classes
- New utility wl_create_spritesheet to create spritesheets for animations
- Animations are also cropped to save more space
- Spritesheet animations are kept in separate Lua tables for performance reasons
- New Lua function add_custom_worker to support writing tests
- Convert Barbarian Carrier, Wood Hardener and Flag as examples


gunchleoc:
Spritesheets :)

Note that the idle carrier gets a black background when zoomed out - this is a proof-of-concept that the correct mipmap files are being generated and used. The files at 0.5 scale can be deleted before merging.

Once this is in, we should look at exporting all 4 scales directly from Blender for maximum quality.

doc/sphinx/source/animations.rst
```diff
@@ -90,6 +100,71 @@ We also support mipmaps here -- name the files ``walk_ne_0.5_00.png``,
...
+to call ``./wl_create_spritesheet`` from the Widelands program directory.
+This will print the command line options to use for converting your animation.
+
+After conversion, you will need to delete the old files and copy over the new files,
+then delete the old animation code in the map object's ``init.lua`` file and add the new code.
+The tool will print the new code to the console for you, so all you need to do is copy/paste
+and then add any missing optional parameters back in.
+
+Note that except for flag and frontier animations, the table for spritesheets is
+called ``spritesheets``, not ``animations``!
```
Noordfrees: It is confusing that spritesheets are sometimes in a table that looks like a file anim table. Please rename the tribe´s "animations" table to "spritesheets"

src/graphic/animation/animation.cc
```diff
@@ -0,0 +1,195 @@
...
+}
+
+void Animation::add_available_scales(const std::string& basename, const std::string& directory) {
+	for (const auto& scale : kSupportedScales) {
+		add_scale_if_files_present(basename, directory, scale.first, scale.second);
+	}
+
+	if (mipmaps_.count(1.0f) == 0) {
+		// There might be only 1 scale
+		add_scale_if_files_present(basename, directory, 1.0f, "");
```
Noordfrees, line 128: Since we can assume that kSupportedScales always contains 1×, is there really a reason why it should work on the second try if it didn´t on the first?

src/graphic/animation/animation.h
```diff
@@ -0,0 +1,207 @@
...
+ * render itself at a given time to the given place.
+ *
+ * The dimensions of an animation is constant and can not change from frame to
+ * frame.
+ */
+class Animation {
+public:
+	/// Whether we have an animation consisting of multiple files or of 1 single spritesheet file
+	enum class Type {
+		kFile,
```
Noordfrees: Perhaps call it kFiles since the point here is that it will typically have more than one?

src/graphic/animation/nonpacked_animation.cc
```diff
@@ -0,0 +1,238 @@
...
+
+NonPackedAnimation::NonPackedAnimation(const LuaTable& table, const std::string& basename)
+   : Animation(table) {
+	try {
+		// Get image files
+		if (table.has_key("pictures")) {
+			// TODO(GunChleoc): Old code - remove this option once conversion has been completed
+			mipmaps_.insert(std::make_pair(
+			   1.0f, std::unique_ptr<NonPackedMipMapEntry>(
+			            new NonPackedMipMapEntry(table.get_table("pictures")->array_entries<std::string>()))));
```
Noordfrees: Suggestion: Print a warning for every old-format animation if --verbose output is enabled

src/graphic/animation/nonpacked_animation.cc
```diff
@@ -0,0 +1,238 @@
...
+				   "Mismatched existence of player colors in animation table for scales %.2f and %.2f",
+				   static_cast<double>(mipmaps_.begin()->first), static_cast<double>(mipmap.first));
+			}
+		}
+		if (mipmaps_.count(1.0f) != 1) {
+			throw Widelands::GameDataError(
+			   "All animations must provide images for the neutral scale (1.0)");
+		}
+	} catch (const LuaError& e) {
+		throw Widelands::GameDataError("Error in animation table: %s", e.what());
```
Noordfrees: Please add the basename to the message, such errors are always horrible to debug

src/graphic/animation/spritesheet_animation.cc
```diff
@@ -0,0 +1,222 @@
...
+		if (sheet->height() != playercolor_mask_sheet->height()) {
+			throw Widelands::GameDataError(
+			   "animation sprite sheet has height %d but playercolor mask sheet has height %d. The sheet's image is %s",
+			   sheet->height(), playercolor_mask_sheet->height(), sheet_file.c_str());
+		}
+	}
+
+	// Frame width and height
+	w = sheet->width() / columns;
+	h = sheet->height() / rows;
```
Noordfrees, line 101: Throw an error if (w × columns != sheet->width()) and same for the height

src/graphic/animation/spritesheet_animation.cc
```diff
@@ -0,0 +1,222 @@
...
+		if (representative_frame() < 0 || representative_frame() > nr_frames_ - 1) {
+			throw Widelands::GameDataError("Animation has %d as its representative frame, but the frame indices "
+			                 "available are 0 - %d",
+			                 representative_frame(), nr_frames_ - 1);
+		}
+
+		if (rows_ * columns_ < nr_frames_) {
+			throw Widelands::GameDataError("Animation has %d frames, which does not fit into %d rows x %d columns",
+										   nr_frames_, rows_, columns_);
+		}
```
Noordfrees, line 194: Also throw an error if we have blank rows?

doc/sphinx/source/animations.rst
Noordfrees: Ah, so this is mixed up in TribeDescr loading code. IMHO it would be cleaner if we have one table "spritesheets" for the spritesheets and one table "animations" for nonpacked anims (both optional) in the TribeDescr, like for the MapObjects.

Noordfrees:
Code looking good, some comments added

src/graphic/animation/animation.cc
gunchleoc: This is when the scale got omitted from the filename, so we can handle `idle.png` instead of `idle_1.png` if there is only 1 scale available.

src/graphic/animation/nonpacked_animation.cc
gunchleoc: I'll add the information to map_object.cc. Sample output:

```
Could not read world information: [../src/scripting/lua_errors.cc:22] [../src/scripting/lua_errors.cc:22] Error loading animation 'idle' for map object 'pebble1': NOCOM Trigger test exception
stack traceback:
	....
```

doc/sphinx/source/animations.rst
gunchleoc: Done.

src/graphic/animation/animation.h
gunchleoc: Done.

src/graphic/animation/nonpacked_animation.cc
gunchleoc: Done.

src/graphic/animation/spritesheet_animation.cc
gunchleoc: Done.

src/graphic/animation/spritesheet_animation.cc
gunchleoc: Done.

gunchleoc:
Thanks for your suggestions, they should all be implement now. Can you double-check and resolve the conversations?

Some of them can be seen in the last commit: https://github.com/widelands/widelands/pull/3489/commits/25a563a576b8acf75feeaa0671bd5e056461fdf6

Noordfrees (dismissed):
Code looks fine, untested

Issue #3493, opened 2019-09-17
Check Travis integration


TinoM:
The new Travis repository seems to enforce the normal 50 minutes limit for jobs.

Looking at the old (wideland-old) repository, it seems there was an exception for us, because there are many finished jobs > 50 minutes.

Maybe we can again ask for this as an OpenSource project?

stonerl:
From what I read it is now possible to use caches. E.g. we can use ccache for the builds this could speed up the builds.

stonerl:
Also, from what I read if we move our project from travis-ci.org to travis-ci.com we have the same execution time (120min) per job as private repos, since we are an open source project.

See here: [Open Source Accounts](https://docs.travis-ci.com/user/migrate/open-source-on-travis-ci-com/)

gunchleoc:
> From what I read it is now possible to use caches. E.g. we can use ccache for the builds this could speed up the builds.

I researched ccache a few years ago because my machine was slow and decided against it, because I couldn't be sure that the builds would reflect the current state of the project with 100% certainty. So, we should research this carefully to see if that is still a concern before using it.

stonerl:
I use ccache on my Mac without problems. Also my Travis changes use ccache for macbuilds. The build time is reduced tremendous, from 40 minutes to 16.

I'm currently looking into the Linux builds.

gunchleoc:
Since you have good experience with it, let's use it. Reducing the build time will be excellent :)

I found this: https://stackoverflow.com/questions/1815688/how-to-use-ccache-with-cmake

stonerl:
Another question regarding the regression test. Do we need to run them in debug and release builds or would it be sufficient to run them in release only?

gunchleoc:
It's better to run them in both modes. For example, if somebody adds an assert or other special debug code that has an unintended side effect, we could catch this via the tests if the code passes that point.

stonerl:
Do we need to run the documentation test in every build? Wouldn't it be enough to have one dedicated job just for this?

Also, wouldn't it be sufficient to have just one job for the codecheck? If it throws an error, we then don't start the others.

gunchleoc:
It would be sufficient to run them in 1 build, yes.

I just compared the old and the new list of builds, and we have lost some coverage for compiler versions under Linux:

```
CLANG_VERSION="3.4"
CLANG_VERSION="3.5"
CLANG_VERSION="7"
CLANG_VERSION="8"
clang 3.4
clang 7

[...]

PR #3502, opened 2019-09-18
Skip website binaries in gcc debug builds on Travis

gunchleoc:
This helps a bit with the job timeouts.
I'd say that your testing coverage will still be good enough.
Re. #3493

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -31,7 +31,12 @@ cd build
 
 if [ "$BUILD_TYPE" == "Debug" ]; then
    # We test translations only on release builds, in order to help with job timeouts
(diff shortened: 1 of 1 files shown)


stonerl:
I don't think this will work. Not building the website binaries gives us a minute at most.

gunchleoc:
Not completely - we're just on the edge though, so it made 2 of the 4 timed out builds go green. At this point, I think that the test suite is more important than another build of the website binaries.

We have to wait for SirVer to be available for talking to the Travis people, so this could still be an interim solution.

gunchleoc:
Superseded by #3508.

PR #3503, opened 2019-09-18
Default uuid to empty string rather than nullptr to prevent crashes.

### LINKED ISSUE #3501: Accessing multiplayer lobby as unregistered user crashes the game
The error says:

```
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_M_construct null not valid
```

Did not tried as a registered user.

See also: https://www.widelands.org/forum/post/29883/

gunchleoc:
Closes #3501


Notabilis (approved):
Change is looking good and works. Thanks!

PR #3504, opened 2019-09-18
Fix formatting when updating translations. #3471

gunchleoc:
Periodically format the C++, Lua & Python source code.


gunchleoc:
Since my machine's clang-format no longer produces an edit war wit what bunnybot produces, I can run this once a week on my machine. This will ensure C++ formatting when people have not installed the WIP clang-format githook, and also take care of Python and Lua.

There will be a difference with the Lua formatting when I first run this, because that doesn't produce the same results as bunnybot did for some nebulous reason.

stonerl (approved):

SirVer (approved):

PR #3508, opened 2019-09-21
revamp Travis-CI builds

stonerl:
Travis-CI builds are completely rewritten and don't require sudo anymore.

macOS: We build on 10.14(clang 7), 10.13(clang 5) & 10.12(clang3.9)
Linux: We build on 18.04 (clang 7 & gcc 7.4), 16.04 (gcc 5.4) & 14.04 (clang 3.4 & gcc 4.8)

On macOS all builds use ccache on Linux only the gcc Release builds. Since clang compiles
faster it is not as big of a problem and jobs are within the travis time-limit.

Regression suite is run on Linux jobs only if the compile-time didn't exceed 33 mins, to
avoid hitting the time-limit of 50 mins while executing the tests. This mainly affects
gcc debug builds.

For xvfb we utilize the new service stanza - only available on 18.04 and 16.04, though.
For 14.04 we use the old fashion way.

Only pushes to master and PR trigger builds.

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -1,50 +1,56 @@
 set -ex
 
-export START_TIME=$(date +%s)
-
 # Create build folder.
 mkdir build
 cd build
 
-if [ "$BUILD_TYPE" == "Debug" ]; then
-   # We test translations only on release builds, in order to help with job timeouts
-   cmake .. -DCMAKE_BUILD_TYPE:STRING="$BUILD_TYPE" -DOPTION_BUILD_TRANSLATIONS="OFF" -DOPTION_ASAN="OFF"
+case "$1" in
+build)
+   export START_TIME=$(date +%s)
+   
+   if [ "$BUILD_TYPE" == "Debug" ]; then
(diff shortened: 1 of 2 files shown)


gunchleoc:
Looking at this: https://travis-ci.org/widelands/widelands/builds/587774617

I can't see the compiler versions for each build any more, except for Mac builds.

gunchleoc:
Yep, that fixed it. Let's wait for Travis to finish and then merge if it looks OK.

stonerl:
Added DEBUG/RELEASE as a prefix.

stonerl:
Should be better now. Compiling doesn't start when the code isn't clean or building the documentation fails. Removing the code check from all debug builds saves roughly 2-3 minutes on each job.

gunchleoc (approved):
Perfect :)

I think we should switch off codecheck for the debug builds now rather than skipping the test suite. That requires some CMake change though, so let's do that in a follow-up branch.

PR #3516, opened 2019-09-23
Add 3rd test and translation switch

stonerl:
The the Compiler Errors test should avoid situations were the code would not compile with any compiler like it was in this case [were verything failed.](https://travis-ci.org/widelands/widelands/builds/585738263)

Also a translation switch has been added so translations can be en-/disabled on a per job basis.

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -32,12 +32,6 @@ codecheck)
       exit 1 # CodeCheck warnings.
    fi
    ;;
-compiler)
-   # Just run this to test whether there are any error during compiling.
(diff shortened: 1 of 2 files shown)


gunchleoc:
Good idea :)

Rather than having a dedicated compile job, we could shift 1 Release and 1 Debug build into a second stage, then build the rest in a third stage. This will save us 1 compiler run overall.

stonerl:
You're right :-) I used the macOS builds for the second stage.

gunchleoc:
Travis didn't trigger this. I'll merge it so that we can get a build.

PR #3523, opened 2019-09-26
Add script for installing dependencies

gunchleoc:
This script will try to detect the Linux distribution and install its dependencies.
Operating systems can also be specified manually on the command line.
The package lists were taken from the Wiki and might not be current for all distributions.

### DIFF
--- a/install-dependencies.sh
+++ b/install-dependencies.sh
@@ -1,0 +1,137 @@
+#!/bin/bash
+
+## Install the dependencies for building Widelands.
+## Note that this hasn't been tested on all distributions.
+## Patches welcome ;)
+## Linux distro detection taken from
+## https://superuser.com/questions/11008/how-do-i-find-out-what-version-of-linux-im-running
+
+DISTRO="$1"
+
+echo "Script for installing Widelands dependencies."
+echo "You can call this script with"
+echo " "
+echo "    ./install-dependencies.sh"
+echo " "
+echo "and it will try to autodetect your operating system."
+echo "You can also specify your operating system manually, like this:"
+echo " "
+echo "    ./install-dependencies.sh <operating_system>"
+echo " "
+echo "<operating_system> needs to be one of the following:"
+echo " "
+echo "Linux:"
+echo "* arch      Arch"
+echo "* fedora    Fedora/Red Hat/CentOS"
+echo "* gentoo    Gentoo"
+echo "* suse      SuSE"
+echo "* slackware Slackware"
+echo "* mandriva  Mageia/Mandriva"
+echo "* debian    Debian/Ubuntu/Mint"
+echo " "
+echo "BSD:"
+echo "* freebsd   FreeBSD"
+echo "* openbsd   OpenBSD"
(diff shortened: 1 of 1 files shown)


install-dependencies.sh
```diff
@@ -0,0 +1,144 @@
...
+    mingw-w64-i686-glbinding mingw-w64-i686-glew
+elif [ "$DISTRO" == "msys64" ]; then
+   echo "Installing dependencies for 64-bit Windows..."
+   pacman -S mingw-w64-x86_64-toolchain git mingw-w64-x86_64-cmake \
+    mingw-w64-x86_64-ninja mingw-w64-x86_64-boost mingw-w64-x86_64-SDL2_ttf \
+    mingw-w64-x86_64-SDL2_mixer mingw-w64-x86_64-SDL2_image \
+    mingw-w64-x86_64-glbinding mingw-w64-x86_64-glew
+elif [ "$DISTRO" == "homebrew" ]; then
+   echo "Installing dependencies for Mac Homebrew..."
+   brew install boost bzrtools cmake doxygen gettext glew graphviz icu4c jpeg \
```
stonerl: `bzrtools` isn't necessary anymore.

install-dependencies.sh
```diff
@@ -0,0 +1,144 @@
...
+# If distro was not given on the command line, try to autodetect
+if [ -z "$DISTRO" ]; then
+   echo "Autodetecting operating system..."
+   if [ -f /etc/fedora-release -o -f /etc/redhat-release -o -f /etc/centos-release ]; then
+      DISTRO="fedora"
+   elif [ -f /etc/gentoo-release ]; then
+      DISTRO="gentoo"
+   elif [ -f /etc/SuSE-release ]; then
+      DISTRO="suse"
+      echo "Installing dependencies for SuSE"
```
stonerl: Redundant since it is echoed during the installation.

install-dependencies.sh
```diff
@@ -0,0 +1,144 @@
...
+   elif [ -f /etc/slackware-release -o -f /etc/slackware-version ]; then
+      DISTRO="slackware"
+   elif [ -f /etc/mandriva-release ]; then
+      DISTRO="mandriva"
+   elif [ -f /etc/debian_version ]; then
+      DISTRO="debian"
+   fi
```
stonerl, line 83:
```bash
elif [ -f /usr/local/Homebrew/bin/brew ]; then
   DISTRO="homebrew"
fi
```

stonerl (changes requested):

install-dependencies.sh
gunchleoc: What's the name of the corresponding git package?

install-dependencies.sh
gunchleoc: I'll delete it.

install-dependencies.sh
gunchleoc: Will add.

stonerl (approved):

Issue #1032, opened 2019-09-09
Late AI seafaring due to late building of Weaving-Mill


bunnybot:
I was working on AI seafaring improvements and found AI players very late on building ships, although they build Ports and Shipyards rather early. On a regular basis, ship building + expedition only starts after 2 hours in the game, which is too late to make a difference! On seafaring maps ships/expeditions should be initiated / enabled latest after 1 h. 

The suspected culprit of this behaviour is the little attention AI gives to the building of Weaving-Mills. There is only the exemption of the Atlanteans, which build it early.

--> for Barbarians and Empire the building of Weaving-Mills is the bottleneck for building ships early
--> for Atlanteans something else (unrecognised) is the bottleneck, probably planks, but they usually come out earlier than the other races with a ship and expedition

I recommend to adjust AI to make better use of priorities on seafaring maps to speed up colony founding. Weaving-Mills should be built shortly after enough raw materials for weaving are available or when Cloth in stock (not in map) falls below a significant value!

------------------------------------
Imported from Launchpad using lp2gh.

 * date created: 2016-11-16T09:25:59Z
 * owner: 7010622-q-deactivatedaccount
 * the launchpad url was https://bugs.launchpad.net/bugs/1642196

bunnybot:
(by tiborb95)
I am currently modifying AI and this will also modify the priority of productionsites and it will work like a kind of black box, so code can not per se prescribe order of buildings.

However what is possible is to force a building in configuration file with "forced_after" flag, or eventually we can create new attribute "forced_after_if_seafaring" to address this.

bunnybot:
(by gunchleoc)
The reason that the Weaving Mill is early with the Atlanteans is that it's also needed for Spidercloth, which is a building material.

Empire needs cloth for Armor.

Barbarians only ever need Weaving Mills on seafaring maps.

How about: when you have a shipyard, start building a weaving mill if you don't have one already.

bunnybot:
(by 7010622-q-deactivatedaccount)
Perhaps there is a simple general solution on base of the stock value for Cloth. Like: if stock value sinks below the a minimum, or to zero, then build the Weaving-Mill. - Also: perhaps a Cloth "demand" needs to be set up for the Shipyard if the Cloth row isn't filled completely.

bunnybot:
(by gunchleoc)
The economy demand is already there, I think - so theoretically, the AI could track that.

Alternatively, add a new AI hint to the shipyards that requests building a weaving mill?

[...]

PR #3528, opened 2019-09-26
Ai supports seafaring

tibor95:
This introduces new ai_hint "supports_seafaring", that will be used to boost building score for particular building when on seafaring map. Should be followed by some AI training focused on seafaring...


gunchleoc:
Code LGTM. Should we add this hint to all buildings in the Ship production chain for all tribes?

tibor95:
I dont think it is necessary. But this is good question. Definitely not needed for buildings in basic economy...

gunchleoc:
The Atlanteans won't need it because they build their weaving mill early anyway, but I think that Empire and Frisians could benefit from this.

stonerl:
@tibor95 I ran fix_formatting.py on this branch and rebased it to master. Please pull before you intend to make any changes.

stonerl (approved):
LGTM :-)

@tibor95 can this go in or do you want to add hints for frisians and empire?

tibor95:
It can go, I dont know which other buildings need such hints, because this is intended for special cases only...

stonerl:
I just realized that the frisians' waving mill also need ```needs_seafaring = true,``` since it is only available on such maps, as it is with the barbarians'.

hessenfarmer:
fixes #1032

PR #3530, opened 2019-09-28
Handle high preciousness

### LINKED ISSUE #3275: Too high ware preciousness can hurt AI
Currently each ware has a preciousness defined in init.lua and AI is using it when making decisions what buildings to build.

However late bug 1767187 revealed that too hing number (254 in the case of bug) badly confused AI into insane building of metalworks. So the conclusion is that upper limit for preciousness is needed. 

I want to propose value 25, that would be hardcoded and if exceeded throw an exemption.

Note that normal players would never see this error, only if somebody is fiddling with init.lua files.

 * date created: 2018-04-30T09:22:46Z
 * owner: tiborb95
 * the launchpad url was https://bugs.launchpad.net/bugs/1767976

gunchleoc:
- Warn if ware/worker preciousness is > 25
- Throw an exception if it's > 200

Closes #3275.

I don't think we need anything more fancy at this point.

BTW this triggers a warning for Frisian coal, which has a preciousness of 40. Do we want to increase the warning threshold?

### DIFF
--- a/data/tribes/wares/armor/init.lua
+++ b/data/tribes/wares/armor/init.lua
@@ -41,7 +41,7 @@ dirname = path.dirname(__file__)
 --    for each tribe's economy. For example, ``{ atlanteans = 3, empire = 1 }``
 --
 --    **preciousness**: How precious this ware is to each tribe. For example,
---    ``{ atlanteans = 0, empire = 1 }``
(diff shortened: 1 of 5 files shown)


tibor95 (approved):
I am fine with this.

gunchleoc:
Great, thanks for the review! :)

PR #3531, opened 2019-09-28
Handle error while reading the shading language

gunchleoc:
Show error message box rather than crashing when reading the shading language version returns an error.

My Windows VM doesn't have a decent graphics driver, so it crashed with no shading language detectable.

This is the output produced by this branch:

no-shading-language-error-message

### DIFF
--- a/src/graphic/gl/initialize.cc
+++ b/src/graphic/gl/initialize.cc
@@ -207,6 +207,29 @@ SDL_GLContext initialize(
 		exit(1);
 	};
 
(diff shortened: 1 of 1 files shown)


klaus-halfmann (approved):
I dont know that syntax, but it looks reasonable.
I will not no an extra test on windows, though.
As of that Screenshot that shoud be proof enough.

gunchleoc:
Thanks for the review!

Yes, it's hard to reproduce. You'd need a system with a crappy graphics driver.

PR #3532, opened 2019-09-29
Allow empty password for unregistered accounts in internet gaming.

gunchleoc:


stonerl (approved):
Just one question. Why did this happen in the first place? When I implemtet the new internet game this was working without a problem. Was something changed along the way?

gunchleoc:
Yes, there were some changes with fetching the config in the XDG branch that probably influenced this, or it might have been the gameclient code refactoring.

PR #3533, opened 2019-10-01
return correct description instead of empty string

### LINKED ISSUE #3510: Wrong/Misleading ToolTipText
While playing first mission of the Atlanteans, wrong tooltiptext was visible. See screenshot.

OS: Win10
Version: Official Build20
Attachments: Savegame and screenshot
However, after restarting game and loading the savegame, it looks fine again
tooltip

[wrong tooltiptext.zip](https://github.com/widelands/widelands/files/3638569/wrong.tooltiptext.zip)

# To fix:
- [x] Allow more space in tooltips (problem does not reproduce)
- [ ] Fix German translation to shift the "ist" where it belongs

jmoerschbach:
Fixes #3510

### DIFF
--- a/src/logic/map_objects/tribes/production_program.cc
+++ b/src/logic/map_objects/tribes/production_program.cc
@@ -289,13 +289,13 @@ bool ProductionProgram::ActReturn::Negation::evaluate(const ProductionSite& ps) 
 }
 
 // Just a dummy to satisfy the superclass interface. Returns an empty string.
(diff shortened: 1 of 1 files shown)


gunchleoc (approved):
LGTM, thanks for the fix!

How do you want to be credited, nickname and/or username?

jmoerschbach:
both would be nice :)

gunchleoc:
Done! Thanks for the fix :)

PR #3540, opened 2019-10-08
Refactoring of the lower layer networking classes

Notabilis:
Refactoring and partial rewrite of the networking classes dealing with sockets.

Fixes a rare bug with slow or overwhelmed network connections. The problem was that sending too much data leaded to the buffers within the operating system running full. This in turn could result in corruption of the send data stream since I made a wrong assumption about the behavior of a function.

The low-level handling of network sockets is now mostly done within one class.

Added support for priorities when sending data over the network. Useless as of now but can be useful to let chat messages through while sending files (e.g., a save game on startup).

Note: The commits in this branch do not compile. For reviewing I suggest reading bufferedconnection.h / .cc first and then just going down in the diff.


src/network/bufferedconnection.cc
```diff
@@ -0,0 +1,374 @@
...
+	boost::asio::ip::tcp::endpoint remote = socket_.remote_endpoint(ec);
+	if (!ec) {
+		log("[BufferedConnection] Closing network socket connected to %s:%i.\n",
+		    remote.address().to_string().c_str(), remote.port());
+	} else {
+		log("[BufferedConnection] Closing network socket.\n");
+	}
+	// Stop the thread
+	io_service_.stop();
+	while (!io_service_.stopped()); // Not sure if that is required
```
klaus-halfmann: Not sure is this could result in some endless loop.

src/network/bufferedconnection.h
```diff
@@ -0,0 +1,321 @@
...
+	 * then start reading after the previous value. Only a successful peek will move the pointer.
+	 *
+	 * Note that a successful peek does not mean that the received bytes really are of the requested
+	 * type, it only means they could be interpreted that way. Whether the type matches is in the
+	 * responsibility of the caller.
+	 *
+	 * Using any method of this class will trigger an internal read from the network socket inside
+	 * the given BufferedConnection.
+	 *
+	 * \warning Calling any receive() method on the given connection will invalidate the Peeker.
```
klaus-halfmann, line 59: So there should be some `bool isValid()` Method?

src/network/bufferedconnection.h
```diff
@@ -0,0 +1,321 @@
...
+
+	/**
+	 * Receive a RecvPacket.
+	 * \warning Calling this method is only safe when peek_recvpacket() returned \c true.
+	 *          Otherwise the behavior of this method is undefined.
+	 * \param out The variable to write the value to.
+	 */
+	void receive(RecvPacket* out);
+
+	// Temporary method, will be removed.
```
klaus-halfmann: will be removed when / after ?

src/network/bufferedconnection.h
```diff
@@ -0,0 +1,321 @@
...
+	 decides which of the send_T_() methods have to be called. Since each of them takes a different
+	 first argument out of Fargs only one of them can match. The matching methods transforms the
+	 received first argument to an uint8_t string and appends it to the given vector. Then it
+	 calls send_T_() again, but with one argument less. This results in a "recursive" call
+	 until no more arguments are left.
+	 */
+	template<typename... Targs>
+	void send(NetPriority priority, const Targs&... Fargs) {
+
+		std::vector<uint8_t> v;
```
klaus-halfmann, line 222: TODO(k.khalfmann): think about some reasobale presizing?

src/network/network.h
```diff
@@ -243,4 +224,17 @@ struct ProtocolException : public std::exception {
...
 };
 
+/**
+ * The priorities for sending data over the network when using a BufferedConnection.
+ * Data with low priority values is send first even when data with high priority values
+ * have been passed to a BufferedConnection first
+ */
+// The values assigned to the entries are arbitrary, only their order is important
+// No "enum class" on purpose since this has to be interpreted as ints (I need a known ordering)
+enum NetPriority : uint8_t {
```
klaus-halfmann, line 234: Good idea, so we do not loose pings during filetransfers?

klaus-halfmann (commented):
Code is fine for me, even though I could not check all details.
Please check some inline comments.
As this is your private repository I had to `git clone --single-branch git@github.com:Notabilis/widelands.git net-refactor-blocking` but it takes ~20 Minutes still.
I will compile this branch now and try some local network games.
I will hang around in the lobby these days with this branch so we could so some
testgames.

klaus-halfmann (dismissed):
OK, I played this for about 1:30 now with this branch as Host.
I think this can go in now.

src/network/bufferedconnection.cc
Notabilis: Good point, I don't know either. I transformed it to a for-loop, just in case.

src/network/bufferedconnection.h
Notabilis: I could add one, but it would probably relatively complicated.
Currently the Peeker is only used for at most a few consecutively lines, so I would keep the warning as a "callers responsibility". Adding a isValid() method doesn't make it any better anyway I guess, since the caller would have to remember to call that method.

src/network/network.h
Notabilis: Indeed. Is not relevant at the moment but will be with my next branch.

src/network/bufferedconnection.h
Notabilis: When the data is used somewhere. I extended the comment.

src/network/bufferedconnection.h
Notabilis: Sounds sensible, done.

Notabilis:
Thanks for the review and the testing. I addressed your comments, when the checks are happy this can be merged as far as I am concerned.

src/network/bufferedconnection.h
klaus-halfmann: std::vector does not allow  presizing in the CTor? like `std::vector<uint8_t> v(kNetworkBufferSize);`
Indeed, so thats OK now.

Notabilis:
Argh, as it turns out I use too new and fancy features of libboost in my code, so the Travis checks fail. Seems I have to add some ugly workarounds after all. So... this is no longer ready for merge. :-/

Notabilis:
I replaced the not-yet-supported code, if the tests accept it this is ready for merged.

klaus-halfmann (commented):
Where does THE_FUTURE_IS_HERE come from? some name like BOOS_GT_XX might better indicate what this is about. Who defines our mininmal Boost version? maybe we can jus tbump it to the version you need?

I still may approve it as THE_FUTURE_IS_HERE is quit local.

Notabilis:
THE_FUTURE_IS_HERE is just some random name, I can change it if you want.
The boost version is defined in src/CMakeLists.txt. I don't think we can simply increase it there, since some of the "computers" used by Travis are installing older versions than I need from their apt repositories.

klaus-halfmann:
> THE_FUTURE_IS_HERE is just some random name, I can change it if you want.

Yes please use soething like OLD_BOOST_COMPATIBILITY or something even more descriptive.

stonerl (changes requested):
Please check for the installed boost-version and enable the new code for boost >=1.66. The minimum required version won't be set to 1.66 or greater until Ubuntu 18.04 is EOL. This will be in April 2023. The extended support will end in 2028 which would be another five years if we are willing to support Ubuntu 18.04 till then.

Basically this means the new code won't be used for at least another 5 years.

stonerl (approved):

stonerl:
LGTM :-)

Notabilis:
Hey, you are not supposed to say something until the checks are fine! ;-)
Travis is happy, though, so I guess both code versions compile fine.
Thanks for the reviews and the dates, I wasn't aware that it will take that long until we can drop support for old boost versions.

Notabilis:
GitHub says AppVeyor is still pending but in the detailed view it is seems to be done without problems. So I guess GitHub is hanging and this can be merged?

src/network/bufferedconnection.h
gunchleoc: @klaus-halfmann Checking out from a fork is a lot easier than that - see https://www.widelands.org/wiki/GitPrimer/#checking-out-a-branch-from-somebody-elses-fork

PR #3545, opened 2019-10-16
Refactor gamehost

klaus-halfmann:
Thi makes mostly makes the code in gamehost.cc more readable,
I intenden no change in semantics.

I finally found a missig '!' so this should be ready to go in.
I missed to set a valid upstream firstm so I acciddently deleted some branches when pushing

I got this one, but it happens in master, too:
Assertion failed: (&worker == working_positions_[main_worker_].worker), function get_building_work, file ../src/logic/map_objects/tribes/productionsite.cc, line 823.


klaus-halfmann:
travis-ci failed only for clang 9 on Ubuntu, worked for me on OSX

src/network/gamehost.h
```diff
@@ -145,9 +165,11 @@ struct GameHost : public GameController {
 
 	void broadcast(SendPacket&);
 	void write_setting_map(SendPacket&);
-	void write_setting_player(SendPacket&, uint8_t number);
+	void     write_setting_player(SendPacket&, uint8_t number);
```
Notabilis: Personally I don't like this whitespace, though I understand why you are adding it

src/network/gamehost.h
```diff
@@ -63,6 +66,7 @@ struct GameHost : public GameController {
 
 	// Pregame-related stuff
 	const GameSettings& settings();
+	/** return true in case all conditions for the game start are met */
```
Notabilis, line 69: Just as a hint: You can use /// for single line Doxygen comments.

src/logic/game_settings.h
```diff
@@ -57,7 +57,7 @@ struct PlayerSettings {
 	uint8_t shared_in;  // the number of the player that uses this player's starting position
 };
 
-struct UserSettings {
+struct UserSettings { // TODO(k.halfman): make this some const instead of calculating this every time
```
Notabilis: I don't quite understand what you mean here. Do you want to make the struct const or the following static functions?

src/network/gamehost.cc
```diff
@@ -1508,6 +1490,14 @@ void GameHost::write_setting_player(SendPacket& packet, uint8_t const number) {
 	Notifications::publish(NoteGameSettings(NoteGameSettings::Action::kPlayer, number));
 }
 
+void GameHost::broadcast_setting_player(SendPacket& packet, uint8_t const number) {
```
Notabilis: Why are you taking the packet as a parameter if it is cleared anyway? Why not simply create a local packet instance inside this function? Nearly everywhere the packet is created in the calling function just to pass it to this method...

src/network/gamehost.cc
```diff
@@ -1204,6 +1209,7 @@ void GameHost::set_player_tribe(uint8_t const number,
 
 	PlayerSettings& player = d->settings.players.at(number);
 
+	// TODDO(k.halfmann): check this logic, will tribe "survive" een when random is selected?
```
Notabilis: Typo "een"

src/network/gamehost.cc
```diff
@@ -1525,6 +1515,14 @@ void GameHost::write_setting_user(SendPacket& packet, uint32_t const number) {
 	   NoteGameSettings::Action::kUser, d->settings.users.at(number).position, number));
 }
 
+void GameHost::broadcast_setting_user(SendPacket& packet, uint32_t const number) {
```
Notabilis: See above comment.

src/network/gamehost.cc
```diff
@@ -1602,15 +1600,15 @@ void GameHost::welcome_client(uint32_t const number, std::string& playername) {
...
+    if (!d->game) { // just in case we allow connection of spectators/players after game start
+        for (uint32_t i = 0; i < d->settings.users.size(); ++i) {
 			if (d->settings.users[i].position == UserSettings::not_connected()) {
 				client.usernum = i;
 				d->settings.users[i].result = Widelands::PlayerEndResult::kUndefined;
 				d->settings.users[i].ready = true;
 				break;
-			}
-	if (client.usernum == -1) {
+    }}}
```
Notabilis: Personally, I would prefer to have those on three separate lines.

src/network/gamehost.cc
```diff
@@ -2025,256 +2019,258 @@ void GameHost::handle_network() {
...
+	std::string clientname = r.string();
+	client.build_id = r.string();
 
-	case NETCMD_SETTING_MAP:
-		if (!d->game) {
-			throw DisconnectException("NO_ACCESS_TO_SERVER");
-		}
-		break;
+	welcome_client(client_num, clientname);
+	return;
```
Notabilis: This (and similar) return's are no longer needed. Do they exist on purpose?

src/network/gamehost.cc
```diff
@@ -2025,256 +2019,258 @@ void GameHost::handle_network() {
...
-		break;
+	std::string clientname = r.string();
+	client.build_id = r.string();
 
-	case NETCMD_SETTING_MAP:
-		if (!d->game) {
-			throw DisconnectException("NO_ACCESS_TO_SERVER");
-		}
-		break;
+	welcome_client(client_num, clientname);
```
Notabilis, line 2097: Theoretically you could put welcome_client() in here. But that method is already very long already...

src/network/gamehost.cc
```diff
@@ -2025,256 +2019,258 @@ void GameHost::handle_network() {
...
+	    case NETCMD_SETSPEED:               return handle_speed(client, r);
+	    case NETCMD_NEW_FILE_AVAILABLE:     return handle_new_file(client);
+	    case NETCMD_FILE_PART:              return handle_file_part(client, r);
+
+	    case NETCMD_SETTING_MAP:
+	    case NETCMD_SETTING_PLAYER:
+	    case NETCMD_WIN_CONDITION:
+	    case NETCMD_PEACEFUL_MODE:
+	    case NETCMD_LAUNCH:
+		if (!d->game) { // not expected while game is in progress -> something is wrong here
```
Notabilis: Actually, it seems to me as if the state of the game has nothing to do with it anyway. These commands seem to only be send by the gamehost and consequently should never be received by the gamehost. Maybe add a comment mentioning this?

src/network/gamehost.cc
```diff
@@ -1121,9 +1130,9 @@ void GameHost::set_player_state(uint8_t const number,
 		return;
 
 	if (player.state == PlayerSettings::State::kHuman) {
-		//  0 is host and has no client
-		if (d->settings.users.at(0).position == number) {
-			d->settings.users.at(0).position = UserSettings::none();
+		// kSpectatorPlayerNum has no client
```
Notabilis: I think this change is wrong, see my comment about kSpectatorPlayerNum.
Of course it compiles and works fine since kSpectatorPlayerNum is 0, but I don't like the semantics of it. Naming kSpectatorPlayerNum as kHostPlayerNum would be okay by me, but I doubt that user.at(0) is a spectator.

src/network/gamehost.h
```diff
@@ -41,6 +41,9 @@ struct Client;
  * launch, as well as dealing with the actual network protocol.
  */
 struct GameHost : public GameController {
+	/** playernumber 0 identifies the spectators */
```
Notabilis, line 44: Are you really sure about this?? I think entry 0 in settings.users is the host player, but not the spectators!
In the constructor in line 534 it says:
`d->settings.users.push_back(hostuser);`

Notabilis (dismissed):
Overall I think the changes are okay and a short test worked without problems, so I guess nothing broke. Some minor suggestions are in the diff.
However, I have some doubts about the kSpectatorPlayerNum constant. From my understanding of the code it seems to me that player 0 is always the host player, and not a special constant denoting the spectators. Are you sure that it marks the spectators?

src/logic/game_settings.h
klaus-halfmann: The compiler (at least in my debugger) must recalculate the values every time, even though they are constant.

src/network/gamehost.cc
klaus-halfmann: There is one place where the SendPacket is reused, but your are right, give me some time for that one

src/network/gamehost.cc
klaus-halfmann: Leftover form original code in switch, removed it ..

src/network/gamehost.cc
klaus-halfmann: I still do not gasp that logic completly, some commands can be send any time (like chat, ping) some only during setup, some only during the actual game. So I hope my comment is correctz

src/network/gamehost.h
klaus-halfmann: I asked @gunchleoc about this and she told be its the specatator, but maybe the magic '0' user / player is used with overloaded semantics?

Notabilis: I haven't checked, but it could indeed well be that player 0 is a spectator in some parts of the codes. That doesn't seem to be the case for the settings.user structure, though.

stonerl:
@klaus-halfmann I ran fix_formatting.py and rebased to master. Please pull this branch again before you make any changes.

src/network/gamehost.h
stonerl: fixed by running fix_formatting.py

src/network/gamehost.cc
klaus-halfmann: OK, simplified that one

klaus-halfmann:
@gunchleoc now that you are back we could finally merge this. I have some spare time the next week and will try to review some simple PRs.

src/network/gamehost.cc
gunchleoc: I agree with Nordfrees, `d->settings.users` are the connected clients, not the player slots. So, better leave the 0 here until we're sure what the actual semantics are.

src/network/gamehost.cc
```diff
@@ -1276,10 +1276,7 @@ void GameHost::set_player_ai(uint8_t number, const std::string& name, bool const
 
 	// Broadcast changes
 	SendPacket packet;
```
gunchleoc: Delete this variable. Same for all instances where you call broadcast_setting_player.

src/network/gamehost.cc
```diff
@@ -1423,10 +1411,7 @@ void GameHost::switch_to_player(uint32_t user, uint8_t number) {
 
 	// Broadcast the user changes to everybody
 	SendPacket packet;
```
gunchleoc: Remove the packet variable from `broadcast_setting_user` too.

src/network/gamehost.cc
```diff
@@ -1509,6 +1491,15 @@ void GameHost::write_setting_player(SendPacket& packet, uint8_t const number) {
 	Notifications::publish(NoteGameSettings(NoteGameSettings::Action::kPlayer, number));
 }
 
+void GameHost::broadcast_setting_player(uint8_t const number) {
+	SendPacket packet;
+	packet.reset();
```
gunchleoc: You have just declared this variable, so no need to reset it.

src/network/gamehost.cc
```diff
@@ -2026,6 +2023,173 @@ void GameHost::handle_network() {
...
+	// Now we wait for the client to say Hi in the right language,
+	// unless the game has already started
+	if (d->game) {
+		throw DisconnectException("GAME_ALREADY_STARTED");
+	}
+	if (cmd != NETCMD_HELLO) {
+		throw ProtocolException(cmd);
+	}
+	uint8_t version = r.unsigned_8();
+	if (version != NETWORK_PROTOCOL_VERSION)
```
gunchleoc: Add {. Please go through the "if" statements in this pull request and add more wherever you find them missing.

src/network/gamehost.cc
gunchleoc: Host and client have some common protocol checks in their handle_hello functions. If we pulled these out, there would be very little left in this function and adding the code from welcome_client here instead would make sense then.

src/network/gamehost.h
gunchleoc: Player 0 is indeed the spectator for the player slots. This variable is used in client slots though, so it doesn't make sense there.

src/network/gamehost.h
```diff
@@ -161,7 +183,7 @@ struct GameHost : public GameController {
 	std::unique_ptr<NetTransferFile> file_;
 	GameHostImpl* d;
 	bool internet_;
-	bool forced_pause_;
+	bool forced_pause_;  // TODO(k.halfmann): this is never really used?
```
gunchleoc: Yes it is, triggered by the `forcePause` host chat command.

gunchleoc:
I have looked over the code and added some comments. No testing.

klaus-halfmann:
@gunchleoc I think I have adressed all your comments

src/network/gamehost.cc
gunchleoc: If you're not sure that your comment is correct, better remove it.

gunchleoc:
@klaus-halfmann I went through the conversations and resolved some, but some are still open and need a bit of work.

src/network/gamehost.cc
klaus-halfmann: reverted that one

src/network/gamehost.h
klaus-halfmann: Now I am finally confused about player slots, client slots, etc and have no idea what to do.

src/network/gamehost.h
klaus-halfmann: Sorry I come from the JavaDocs side :-)

src/network/gamehost.cc
klaus-halfmann: I am quite sure about that comment meanwhile

src/network/gamehost.h
gunchleoc: Found it!

```
int16_t GameHost::get_local_playerposition() {
	return d->settings.users.at(0).position;
}
```

So, client `0` is the host, and player `0` is the spectator. We're dealing with the host here.

klaus-halfmann:
Only "DEBUG: Ubuntu 18.04 with clang 9" fails at test_persistence.lua.

Can we put some final documentation somewhere whats slots numbers and ids are used for client/server savegame and so forth? we will have no chance to get this clean anytime soon.

gunchleoc:
Good idea - I think gamesettings.h could be a good spot for this.

gunchleoc (approved):
This can be merged as soon as The CI is through :)

klaus-halfmann:
No it failed at "RELEASE: Ubuntu 18.04 with gcc 7.4" scripting/test_persistence.lua ...  no valid video driver: No available video device. ...
Can I focre the PR via the GUI? or am I lacking some super power?

I will setup a new Bug to start a discussion for the docs in gamesettings.h

gunchleoc:
Only admins can force. Will merge now.

PR #3546, opened 2019-10-20
minimizing windows 

### LINKED ISSUE #3320: Building windows too easy to minimize (Ctrl-click)
Ctrl-clicking a window minimizes it so only the title bar is shown. The ware priority buttons are very small, so when Ctrl-clicking them it is easy to miss them by a pixel and accidentally minimize the window, which is annoying.
I propose to either enlarge the ware priority buttons significantly, or permit the minimizing of a window with Ctrl-click only if the mouse pointer is in the window´s title bar (middle mouse click could be left unchanged, as it has no other function), or both.

 * date created: 2018-09-16T08:24:09Z
 * owner: nordfriese
 * the launchpad url was https://bugs.launchpad.net/bugs/1792774

jmoerschbach:
Fixes #3320

window is only minimized via ctrl+left click when title bar is clicked upon
minimizing via middle click is left unchanged.

### DIFF
--- a/src/ui_basic/window.cc
+++ b/src/ui_basic/window.cc
@@ -375,7 +375,7 @@ bool Window::handle_mousepress(const uint8_t btn, int32_t mx, int32_t my) {
 	//  TODO(unknown): This code is erroneous. It checks the current key state. What it
 	//  needs is the key state at the time the mouse was clicked. See the
 	//  usage comment for get_key_state.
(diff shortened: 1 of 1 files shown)


stonerl:
Could you run fix_formatting here? I merged the latest changes from master. It should add the correct line ending now.

jmoerschbach:
hmm, all builds suceeded but one:  RELEASE: Ubuntu 18.04 with gcc 7.4 
rerun to try once again?^^

jmoerschbach:
Hi @stonerl @gunchleoc 

same here: Is something missing before merging into master? :)

stonerl (approved):

Issue #3542, opened 2019-10-10
Wrong OpenGL resolution with macOS 10.15 Catalina


klaus-halfmann:
I Upgraded my OSX Computer to 10.14 Catalina.

I use MacPorts but the people there did not adopt it to Catlina, yet.
https://www.macports.org/install.php
They actually fight the first bugs here: https://trac.macports.org/search?q=Catalina

The relesed build 20 is fine.

I noticed some problems with master though:
- Display shows only 1/4 of the actual resolution -> Issue with retina display
   we had this with R20, too. So maybe we "just" need a fixed SDL Library or such
   Found now bugs for Catalina in sdl for now, closest thing I found was:
   https://bugzilla.libsdl.org/show_bug.cgi?id=3281
   R20 uses SDL version: 2.0.9 I have 2.0.10 but I cannot downgrade until macports works work for catalina :\

Can anyone direct me where the initial resoultion and pixels are dertemined?

I will collect other Issues herer once I hit them.

For now I will try a complete recompile to check if this fixes the Issue

klaus-halfmann:
OK I installed MacPorts from the source, recompiled all my ports. 
But the Bug is still there :(. Dowgrading a Port ist not so easy:
https://trac.macports.org/wiki/howto/InstallingOlderPort. Maybe I try this tomorrow

kencu:
Downgrading a port takes some initial setup, but after the first one, is easy.

You need to clone the entire macports repo somewhere. I put it in ```/opt``` to avoid permissions errors etc, but that's not the only place it could go. 
```
cd /opt
sudo git clone https://github.com/macports/macports-ports.git
```
once you have that, you have everything in macports history.

Then go to the folder you're interested in (which hopefully still exists, and in the case of libsdl2, it does) 
```
cd /opt/macports-ports/devel/libsdl2
```
now, find the commit you're interested in. 
```
git log .
```
will list them all. Look for the last commit before the update to 2.0.10, for example. Once you have that commit, copy it and then
```
sudo git checkout THECOMMITID
```
now, you're back where you wanted to be. If it's not too long ago, the Portfile will still work with the current MacPorts installation and other supporting ports. If it's from a LONG time ago, you might need to do some work to bring the Portfile into a usable state with the current situation.

Now, you can just install it right from where you are. If you don't specify a port, MacPorts will build the one in the current folder:
```
sudo port -v install
```
[...]

PR #3555, opened 2019-11-03
Workaround for #3542

stonerl:
This is just a workaround until it is fixed in SDL. @SirVer could you test if this key poses any problems on your legacy system? @klaus-halfmann I already provided a link for a fixed build in #3542.


klaus-halfmann (approved):
AFAIK this should fix the Issue for now, but maybe the DSL team is faster, lets see.
Tried the dmg file @stonerl provided, work for me.

SirVer:
Will test this week and report back.

SirVer (approved):
Works no problem on my old machine. Let's have it!

PR #3561, opened 2019-11-10
Loading map menu

### LINKED ISSUE #3106: Subfolders in savegame folder
=> https://imgur.com/ADzGOMG

If you have subfolder in ~/.widelands/save, Widelands shows them as "incompatible" saves.

It should either hide subfolders here or make them browsable.

 * date created: 2016-02-08T12:50:45Z
 * owner: tino79
 * the launchpad url was https://bugs.launchpad.net/bugs/1543103

jmoerschbach:
Hi all,

this PR is not intended to be merged (yet), since it is still work in progress.
Some time soon, it will fix #3106 

However, I thought before adding new functionality, I'll do a cleanup/refactoring to improve code quality and readablity :)

I greatly appreciate feedback before proceeding (still, I'm a C++ novice, not really knowing what I'm doing... ;))

So, here is what I did:
* Separating 2 major responsibilities: reading/parsing files (LoadOrSaveGame) and displaying them (new class: SavegameTable)
* Introducing a small inheritance hierachy for the visualization to move further to the Single Responsibily Principle (no need to rely on boolean conditions all over the code)
* Extracting some functions in LoadOrSaveGame to make code more readable (there is still a lot of potential)

So, some questions from me:

* Is this more "object orientated" style welcome here? :)
* Is there anything terrible wrong (I'm thinking of memory leaks I am not aware of)?
* Why does LoadOrSaveGame not have a destructor in the first place?
 * Shouldn't at least UI::Box* table_box_ and UI::Button* delete_ be freed?


stonerl:
You can run the code check locally. Have a look in the travis.sh file.

jmoerschbach:
Hi,

so I guess no feedback means nothing is terrible wrong, yes? :D

Sub-directories are now working, see screenshots:
load menu
load menu subdir

What do you guys think about inserting directory/savegame icon (like in editor loading screen) to emphasize difference between file and directory?
load map

While there is still much refactoring potential, I think this might now be worth considering to merge...? :)

stonerl:
> What do you guys think about inserting directory/savegame icon (like in editor loading screen) to emphasize difference between file and directory?

+1

jmoerschbach:
with icons:
grafik

I'll stop here now. There is still room for refactoring already existing code, but I think this should be merged first and then I may revisit this code again.

fixes #3106

@gunchleoc do you have time for review and feedback? :)

jmoerschbach:
any chance this will be merged sometime soon? :)

stonerl (dismissed):
LGTM from my side.

@tibor95 @Notabilis could one of you also have a brief look?

src/wui/gamedetails.cc
```diff
@@ -52,6 +64,32 @@ void SavegameData::set_mapname(const std::string& input_mapname) {
...
+
+bool SavegameData::is_sub_directory() const {
+	return type_ == SavegameType::kSubDirectory;
+}
+
+// static
+SavegameData SavegameData::create_parent_dir(const std::string& current_dir) {
+	std::string filename = FileSystem::fs_dirname(current_dir);
+	if (!filename.empty()) {
+		// fs_dirname always returns a directory with a separator at the end.
```
Notabilis, line 120: Maybe add this as an assert?
assert(filename.back() == '/');
Okay, probably will crash on windows due to it being '\\'. Is the separator defined somewhere? Alternatively, never mind.

src/wui/gamedetails.h
```diff
@@ -33,6 +34,7 @@
  * Data about a savegame/replay that we're interested in.
  */
 struct SavegameData {
+	enum class SavegameType { kSavegame, kParentDirectory, kSubDirectory };
```
Notabilis, line 37: structs can have protected/private members. Maybe make this enum and type_ private? Or do you want other classes to modify it?

src/wui/load_or_save_game.cc
```diff
@@ -92,63 +61,37 @@ LoadOrSaveGame::LoadOrSaveGame(UI::Panel* parent,
...
-	                    "or the map’s name followed by the last objective achieved."),
-	                  UI::Align::kLeft, UI::TableColumnType::kFlexible);
-	table_.set_column_compare(
-	   0, boost::bind(&LoadOrSaveGame::compare_date_descending, this, _1, _2));
-	table_.set_sort_column(0);
-	fill_table();
 
-	table_box_->add(&table_, UI::Box::Resizing::kExpandBoth);
+	table_->set_column_compare(
+	   0, boost::bind(&LoadOrSaveGame::compare_date_descending, this, _1, _2));
```
Notabilis: You could also use std::bind() or lambda functions here. But feel free to leave it as it is.

src/wui/load_or_save_game.cc
```diff
@@ -162,25 +105,38 @@ const std::string LoadOrSaveGame::filename_list_string(const std::set<uint32_t>&
...
 }
-
 bool LoadOrSaveGame::compare_date_descending(uint32_t rowa, uint32_t rowb) const {
-	const SavegameData& r1 = games_data_[table_[rowa]];
-	const SavegameData& r2 = games_data_[table_[rowb]];
-
+	const SavegameData& r1 = games_data_[table_->get(table_->get_record(rowa))];
+	const SavegameData& r2 = games_data_[table_->get(table_->get_record(rowb))];
+	// parent directory always on top
+	if (r1.is_parent_directory())
```
Notabilis: Preferable always add curly brackets {}, even when they are not needed here. (Avoids potential problems/bugs if the code is worked on later.)

src/wui/load_or_save_game.cc
```diff
@@ -162,25 +105,38 @@ const std::string LoadOrSaveGame::filename_list_string(const std::set<uint32_t>&
...
+		return true;
+	// sub directories sort after name
+	if (r1.is_sub_directory() && r2.is_sub_directory())
+		return r1.filename > r2.filename;
+	// savegames sort after timestamp
 	return r1.savetimestamp < r2.savetimestamp;
 }
 
 std::unique_ptr<SavegameData> LoadOrSaveGame::entry_selected() {
+	log("entry selected\n");
```
Notabilis: I guess this is to-be-removed debug output? If not, add some more information so it is actually useful.

src/wui/load_or_save_game.cc
```diff
@@ -203,27 +159,27 @@ std::unique_ptr<SavegameData> LoadOrSaveGame::entry_selected() {
...
-			table_.select(idx);
+			table_->select(idx);
 			return;
 		}
 	}
 }
 
-UI::Table<uintptr_t const>& LoadOrSaveGame::table() {
-	return table_;
+SavegameTable& LoadOrSaveGame::table() {
```
Notabilis, line 195: Personally, I would have returned a pointer (either shared or raw) here. For me, this makes it clear(er) that the caller does not own the referenced object and it might be/become invalid at any time. With the reference, it seems as if the object will exist forever, which is not true (the caller might end up with a reference to unused memory if the reference is stored for some time).
But, well, just my point of view. Feel free to leave it as it is. It already was a reference in the old code, so someone must have decided on it for a reason.

src/wui/load_or_save_game.cc
```diff
@@ -361,213 +322,179 @@ UI::Button* LoadOrSaveGame::delete_button() {
 	return delete_;
 }
 
-void LoadOrSaveGame::fill_table() {
+bool LoadOrSaveGame::is_valid_gametype(SavegameData& gamedata) {
```
Notabilis: Since you won't modify it, you can take a "const SavegameData&" here. Or even only take a GameType (as value, not as reference), you don't need the rest of the SavegameData after all.
Also, the whole function can be marked as const.

src/wui/load_or_save_game.cc
```diff
@@ -203,27 +159,27 @@ std::unique_ptr<SavegameData> LoadOrSaveGame::entry_selected() {
...
 	game_details_.clear();
 }
 
 void LoadOrSaveGame::select_by_name(const std::string& name) {
-	table_.clear_selections();
-	for (size_t idx = 0; idx < table_.size(); ++idx) {
-		const SavegameData& gamedata = games_data_[table_[idx]];
+	table_->clear_selections();
+	for (size_t idx = 0; idx < table_->size(); ++idx) {
+		const SavegameData& gamedata = games_data_[table_->get(table_->get_record(idx))];
```
Notabilis: If you want, you could also use `(*table_)[idx]` here and at the similar places. ( `*` -> Get the object the unique_ptr is pointing to, `()` -> after that, `[]` -> call the operator[] on it.)

src/wui/load_or_save_game.cc
```diff
@@ -361,213 +322,179 @@ UI::Button* LoadOrSaveGame::delete_button() {
...
+		           (gamedata.gametype != GameController::GameType::kReplay)) {
+			return false;
+		}
+	}
+	return true;
+}
 
-	clear_selections();
-	table_.clear();
+void LoadOrSaveGame::add_time_info(SavegameData& gamedata, Widelands::GamePreloadPacket& gpdp) {
```
Notabilis: gpdp and the function itself can be made const.
`void LoadOrSaveGame::add_time_info(SavegameData& gamedata, const Widelands::GamePreloadPacket& gpdp) const {`.
Obviously this isn't needed here, but it makes it clearer what the function is doing with its parameters and may allow calling it in the future if we have a const LoadOrSaveGame for whatever reason.

src/wui/load_or_save_game.cc
```diff
@@ -361,213 +322,179 @@ UI::Button* LoadOrSaveGame::delete_button() {
...
-			return boost::algorithm::ends_with(fn, kReplayExtension);
-		});
-		// Update description column title for replays
-		table_.set_column_tooltip(2, show_filenames_ ? _("Filename: Map name (start of replay)") :
-		                                               _("Map name (start of replay)"));
-	} else {
-		gamefiles = g_fs->filter_directory(kSaveDir, [](const std::string& fn) {
-			return boost::algorithm::ends_with(fn, kSavegameExtension);
-		});
+void LoadOrSaveGame::add_general_information(SavegameData& gamedata,
```
Notabilis, line 415: const gpdp + method
Personally, I make every parameter and method const if possible. I am not aware of disadvantages of doing so, but it might be useful in the future (and if only so someone else understands the method interface easier).

src/wui/load_or_save_game.cc
```diff
@@ -361,213 +322,179 @@ UI::Button* LoadOrSaveGame::delete_button() {
...
+                                             Widelands::GamePreloadPacket& gpdp) {
+	gamedata.set_mapname(gpdp.get_mapname());
+	gamedata.set_gametime(gpdp.get_gametime());
+	gamedata.set_nrplayers(gpdp.get_number_of_players());
+	gamedata.version = gpdp.get_version();
+	gamedata.wincondition = gpdp.get_localized_win_condition();
+	gamedata.minimap_path = gpdp.get_minimap_path();
+}
+
+void LoadOrSaveGame::add_error_info(SavegameData& gamedata, std::string errormessage) {
```
Notabilis: Make this static? Or make it const.

src/wui/load_or_save_game.h
```diff
@@ -89,17 +93,31 @@ class LoadOrSaveGame {
 	/// Reverse default sort order for save date column
 	bool compare_date_descending(uint32_t, uint32_t) const;
 
+	/// To display if the directory is empty and has no parent
+	static SavegameData create_empty_dir(const std::string& current_dir);
+
+	/// Create a subdirectory
+	static SavegameData create_directory(const std::string& directory);
```
Notabilis: Are these two functions implemented anywhere?
(And I don't understand what the first one is supposed to be doing.)

src/wui/savegametable.h
```diff
@@ -0,0 +1,68 @@
...
+class SavegameTable : public UI::Table<uintptr_t> {
+public:
+	SavegameTable(UI::Panel* parent, UI::PanelStyle style, bool localize_autosave);
+
+	void fill(const std::vector<SavegameData>& entries);
+	virtual void set_show_filenames(bool show_filenames);
+
+protected:
+	const std::string map_filename(const std::string& filename, const std::string& mapname);
+	const std::string find_game_type(const SavegameData& savegame);
```
Notabilis: I think these functions can be made const.

src/wui/savegametable.cc
```diff
@@ -0,0 +1,261 @@
...
+
+void SavegameTable::create_directory_entry(UI::Table<const uintptr_t>::EntryRecord& te,
+                                           const SavegameData& savegame) {
+	size_t last_column_index = number_of_columns() - 1;
+	for (size_t col_index = 0; col_index < last_column_index; col_index++) {
+		te.set_string(col_index, "");
+	}
+	if (savegame.is_parent_directory()) {
+		te.set_picture(last_column_index, g_gr->images().get("images/ui_basic/ls_dir.png"),
+		               (boost::format("%s") % _("..")).str());
```
Notabilis: I am a bit unsure about the ".." label. Outside of the programmer/Linux world it might not be understandable, not sure. Maybe really use the "\<parent\>" label you had in your screenshot? Besides, the map selection dialog is using "\<parent\>" as well and I guess we should use the same option for both.

Notabilis (commented):
Thanks for the change, I like the feature! The code looks good overall and mostly it seems to work fine. I haven't really followed through which code has been refactored to where, but since it seems to work I guess you made no mistakes there. (Side note here: Doing the refactoring and the new feature in different branches makes it easier to review, but I end up with "mixed" branches often enough as well.)
I added a bunch of review comments to the diff of your changes. Most of them are only (not so important) code style changes, decide for yourself whether you want to address them.

Two things that should probably be changed:
- There seems to be a debug output with `log()` left, you probably intended to remove it?
- Two functions (`create_empty_dir()`, `create_directory()`) are declared but never implemented. I guess they can be removed?

One bug: The "Delete" button is broken for directories. Instead of printing the name of the directory to remove, it only displays an empty message. For savegames it displays the mapname, but of course that does not exist for directories.

I am unsure whether the sorting of the tables works as intended by you:
- Initial sorting after date works fine. If clicking the "Date" column again, the sorting is inverted so the parent directory ends up at the bottom. To fix this you probably have to modify the Table class, though. If you want to do so, maybe open up a new branch after this one has been merged.
- When sorting after the description/filename, the directories are no longer at the top but are sorted alphabetically somewhere in the middle. If you want to avoid this, I think you have to define a custom sort operator for this column as well.

And a feature request (for future branches): A "Create directory" option and maybe even a "Rename directory" option. ;-)

src/wui/gamedetails.cc
jmoerschbach: I'll leave it as is ;)

src/wui/load_or_save_game.cc
jmoerschbach: Since this is only copied from old/existing code, I'll leave it as is for the time being.

src/wui/load_or_save_game.cc
jmoerschbach: yes, thank you for finding! removed...

src/wui/load_or_save_game.cc
jmoerschbach: yes, everything which can be const should be const I think. Added.

src/wui/load_or_save_game.cc
jmoerschbach: const.

src/wui/load_or_save_game.h
jmoerschbach: they were implemented initially, but not anymore. Deleted :)

src/wui/savegametable.h
jmoerschbach: yes, done

src/wui/load_or_save_game.cc
jmoerschbach: yes. done. I wonder if this should be enforced by IDE/codecheck/clang-format/whatever? :)

jmoerschbach:
Hi @Notabilis 
thank you very much for your profound review! I addressed the minor issues already.
Thanks for testing the delete button. I never tried it Oo
For the time being I think I will disable it when a directory is selected. Deleting a directory with savegames in it relates somehow to your requested features (creating/renaming directories), so I'll have another look to that on a new branch when this one is merged.

I'll fix the sorting of the description column though, since this looks really weird with the subdirectories now being visible

src/wui/load_or_save_game.cc
jmoerschbach: yes, done

src/wui/savegametable.cc
jmoerschbach: Personally, I like the ".." better but I agree that it should be consistent. Changed to "\<parent\>"

src/wui/load_or_save_game.cc
jmoerschbach: Yeah, actually I *really* don't like exposing sophisticated objects like this at all! I think in OO design this is a code smell...
But as you said, it was the same in the old code and I didn't want to change too much code at once... ;)
In one of the posts above I already mentioned this piece of code should be revisted once again some time soon, but for now I'd like to leave it as is to get in merged and maybe tested by more people before proceeding the refactoring :) 

Hope this is ok for you

src/wui/load_or_save_game.cc
jmoerschbach: added const where applicable

src/wui/gamedetails.h
jmoerschbach: type_ is private now

src/wui/load_or_save_game.cc
jmoerschbach: done that, much shorter now, thanks :)

jmoerschbach:
Hi,
I addressed all the issues from @Notabilis review. Ready for merging finally? :)

Notabilis:
Sorry, I found a new bug. :(
When selecting multiple files and directories, a list of them is shown on the right side of the screen. For directories, they are only listed as "," (since they don't have a mapname nor a save date). This is the same behavior as in the (now disabled) delete dialog. I think you have to modify `LoadOrSaveGame::filename_list_string` to fix this for the listings.

Some `table_->get(table_->get_record(` are still present, could you please replace them as well? Admittedly it is old code, but then everything will "look the same".

Apart from that, I am fine with the changes and the functionality, thanks!

jmoerschbach:
Hi @Notabilis 
ok, next try ;)

thanks for testing! Hope everything is ok now.
Once this is merged I'll create another (refactoring only) branch since there are some issues I dislike, especially the misuse of SavegameData when multiple savegames are selected...

Notabilis:
Aaand back to you. Sorry! *duck and run*
It is looking good when selecting files and directories, but it crashes when only selecting multiple directories (but no files):
`#2  0x00007ffff640f40f in __assert_fail_base (fmt=0x7ffff6571ee0 "%s%s%s:%u: %s%sAssertion '%s' failed.\n%n", assertion=0x555557dad540 "filename.find('/') != std::string::npos",`
`    file=0x555557dad4e0 "src/wui/gamedetails.cc", line=214, function=<optimized out>) at assert.c:92`          
`#3  0x00007ffff641d102 in __GI___assert_fail (assertion=0x555557dad540 "filename.find('/') != std::string::npos", file=0x555557dad4e0 "src/wui/gamedetails.cc",`
`    line=214, function=0x555557daf5e0 <GameDetails::update(SavegameData const&)::__PRETTY_FUNCTION__> "void GameDetails::update(const SavegameData&)") at assert.c:101`
`#4  0x000055555760bf7a in GameDetails::update (this=0x7fffffff9480, gamedata=...) at src/wui/gamedetails.cc:214`
`#5  0x00005555576118c5 in LoadOrSaveGame::entry_selected (this=0x7fffffff9448) at src/wui/load_or_save_game.cc:162`
`#6  0x0000555556b4c98b in FullscreenMenuLoadGame::entry_selected (this=0x7fffffff8c60) at src/ui_fsmenu/loadgame.cc:165`

Suggestion: In wui/gamedetails.cc, change:
Line 186: `if (gamedata.filename_list.empty() && !gamedata.filename.empty()) {`
Line 236: `} else if (!gamedata.filename_list.empty()) {`
Not sure whether this is the best way to fix it, but at least it doesn't crash for me any longer.

Notabilis:
Just in case these questions are still open:

> Why does LoadOrSaveGame not have a destructor in the first place?
> Shouldn't at least UI::Box* table_box_ and UI::Button* delete_ be freed?

All UI elements are linked in a hierarchy of Panels (i.e., a rectangle that wants to be drawn). When creating a new Panel object (or one of its descendants, e.g., a button) a "higher" panel is given as parent in the constructor and the parent stores the newly created Panel as a child. When a Panel object is deleted, it also frees all it child Panels, so no explicit delete calls for the created buttons are required.
Consequently, we don't need a destructor in LoadOrSaveGame. The existing raw pointers are part of the panel hierarchy, so we don't have to care about deleting them. The other heap allocated objects are deleted by their unique_ptrs.

jmoerschbach:
Applied your proposals for fixing the crash. This must be refactored, though to make it more clearer and readable in general... :)

Notabilis (dismissed):
Thanks. Not tested again, but looks like my own modification. As far as I am concerned, this can be merged now.
I think before merging you have to merge the current master into your branch (either by clicking the "Update branch" button or manually), clear up any potential conflicts, test again and of course the automatic checks have to succeed.

src/wui/load_or_save_game.cc
```diff
@@ -246,23 +213,28 @@ void LoadOrSaveGame::clicked_delete() {
...
-		            (boost::format(ngettext("Do you really want to delete this %d replay?",
-		                                    "Do you really want to delete these %d replays?",
-		                                    no_selections)) %
-		             no_selections)
-		               .str();
+		header =
+		   no_selections == 1 ?
+		      _("Do you really want to delete this replay?") :
+		      /** TRANSLATORS: Used with multiple replays, 1 replay has a
+		                                                                                                                                                                                                                separate string. DO NOT omit the
```
gunchleoc: Please restore the TRANSLATORS comment

src/wui/load_or_save_game.cc
```diff
@@ -246,23 +213,28 @@ void LoadOrSaveGame::clicked_delete() {
...
-		            (boost::format(ngettext("Do you really want to delete this %d game?",
-		                                    "Do you really want to delete these %d games?",
-		                                    no_selections)) %
-		             no_selections)
-		               .str();
+		header =
+		   no_selections == 1 ?
+		      _("Do you really want to delete this game?") :
+		      /** TRANSLATORS: Used with multiple games, 1 game has a separate
+		                                                                                                   string. DO NOT omit the placeholder in your translation. */
```
gunchleoc: Please restore the TRANSLATORS comment

src/wui/load_or_save_game.cc
```diff
@@ -361,213 +333,180 @@ UI::Button* LoadOrSaveGame::delete_button() {
...
-			return boost::algorithm::ends_with(fn, kReplayExtension);
-		});
-		// Update description column title for replays
-		table_.set_column_tooltip(2, show_filenames_ ? _("Filename: Map name (start of replay)") :
-		                                               _("Map name (start of replay)"));
-	} else {
-		gamefiles = g_fs->filter_directory(kSaveDir, [](const std::string& fn) {
-			return boost::algorithm::ends_with(fn, kSavegameExtension);
-		});
+void LoadOrSaveGame::add_general_information(SavegameData& gamedata,
```
gunchleoc, line 415: This function is called only once and add_time_info as well, directly after each other. Unify into 1 function?

src/wui/load_or_save_game.cc
```diff
@@ -361,213 +333,180 @@ UI::Button* LoadOrSaveGame::delete_button() {
...
+	    % _("Error message:") % errormessage)
+	      .str();
+
+	gamedata.mapname = FileSystem::filename_without_ext(gamedata.filename.c_str());
+}
+
+void LoadOrSaveGame::add_sub_dir(const std::string& gamefilename) {
+	// Add subdirectory to the list
+	const char* fs_filename = FileSystem::fs_filename(gamefilename.c_str());
+	if (!strcmp(fs_filename, ".") || !strcmp(fs_filename, ".."))
```
gunchleoc: Please add the optional {} here, it's safer to always do that to prevent bugs.

We still have lots of there in the code base, because cleaning them all up at one time would be a huge undertaking. We don't want to add any new cases where they are missing though.

src/wui/load_or_save_game.h
```diff
@@ -77,29 +79,43 @@ class LoadOrSaveGame {
...
+	bool compare_map_name(uint32_t, uint32_t) const;
 
 	UI::Panel* parent_;
 	UI::Box* table_box_;
-	UI::Table<uintptr_t const> table_;
 	FileType filetype_;
-	bool show_filenames_;
-	bool localize_autosave_;
+
+	std::unique_ptr<SavegameTable> table_;
```
gunchleoc: The UI::Panel class does all the memory management for our UI classes, including tables.

So, although we want to move to unique_ptr everywhere else, you could simply use `SavegameTable table_`; here and your code should still be safe.

src/wui/savegametable.cc
```diff
@@ -0,0 +1,261 @@
...
+		te.set_string(col_index, "");
+	}
+	te.set_string(
+	   last_column_index, (boost::format(_("Incompatible: %s")) % savegame.mapname).str());
+}
+
+void SavegameTable::create_directory_entry(UI::Table<const uintptr_t>::EntryRecord& te,
+                                           const SavegameData& savegame) {
+	size_t last_column_index = number_of_columns() - 1;
+	for (size_t col_index = 0; col_index < last_column_index; col_index++) {
```
gunchleoc: C++ convention is to use `++col_index` instead of `col_index++`.

src/wui/savegametable.cc
```diff
@@ -0,0 +1,261 @@
...
+		te.set_string(col_index, "");
+	}
+	if (savegame.is_parent_directory()) {
+		te.set_picture(last_column_index, g_gr->images().get("images/ui_basic/ls_dir.png"),
+		               (boost::format("%s") % _("<parent>")).str());
```
gunchleoc: This should be identical to `MapData::parent_name()`, otherwise we'll have additional work for translators that can introduce inconsistencies in the translations.

src/wui/savegametable.cc
```diff
@@ -0,0 +1,261 @@
...
+	} else if (savegame.is_sub_directory()) {
+		te.set_picture(
+		   last_column_index, g_gr->images().get("images/ui_basic/ls_dir.png"),
+		   (boost::format("%s") % FileSystem::filename_without_ext(savegame.filename.c_str())).str());
```
gunchleoc: I think you can lose the `boost::format` call here, `filename_without_ext` already returns a `std::string`.

src/wui/savegametable.cc
```diff
@@ -0,0 +1,261 @@
...
+	}
+}
+
+void SavegameTable::set_show_filenames(bool) {
```
gunchleoc, line 112: Please add a comment here that this is empty on purpose.

src/wui/savegametable.cc
```diff
@@ -0,0 +1,261 @@
...
+   : SavegameTable(parent, style, localize_autosave), show_filenames_(false) {
+	add_columns();
+}
+
+void SavegameTableReplay::add_columns() {
+	add_column(130, _("Save Date"), _("The date this game was saved"), UI::Align::kLeft);
+	std::string game_mode_tooltip = "";
+	/** TRANSLATORS: Tooltip header for the "Mode" column when choosing a game/replay to
+	load.
+	            */
```
gunchleoc: The comment closing can go in the line above.

src/wui/savegametable.cc
```diff
@@ -0,0 +1,261 @@
...
+	/** TRANSLATORS: %s is a list of game modes. */
```
gunchleoc: There is no %s in the translatable string below, so this comment is confusing. Remove.

src/wui/savegametable.cc
```diff
@@ -0,0 +1,261 @@
...
+}
+
+void SavegameTableMultiplayer::add_columns() {
+	add_column(130, _("Save Date"), _("The date this game was saved"), UI::Align::kLeft);
+	std::string game_mode_tooltip = "";
+	/** TRANSLATORS: Tooltip header for the "Mode" column when choosing a game/replay to
+	load.
+	            */
+	/** TRANSLATORS: %s is a list of game modes. */
+	g_gr->styles().font_style(UI::FontStyle::kTooltipHeader).as_font_tag(_("Game Mode"));
```
gunchleoc, line 222: Fix translators comment (see above)

gunchleoc (commented):
I apologize for not chiming in earlier, I had to take a break from the project because I became very busy with some other stuff.

Thanks for taking on this feature! I also like the refactoring.

I have added some nitpicky inline comments with code style suggestions.

Some additional observations: 

1. I noticed that the diff shows changes to debian/translations, which should not have any changes in it from this branch.
2. When taking on more refactoring later on, also have a look at mapselect.cc. There is some table sorting and display code for directories used in there somewhere. We might also be able to pull out some common code from SavegameData/MapData.

gunchleoc:
I did a very quick test and found a bug: Uncompressed savegames show up as directories and not as savegames.

src/wui/load_or_save_game.cc
jmoerschbach: done

src/wui/load_or_save_game.cc
jmoerschbach: done

src/wui/load_or_save_game.cc
jmoerschbach: Sry, I disagree. Functions should do one thing only which then can be properly named (like add_time_info). Admittedly, "add_general_information is a rather poor name but still better to keep them seperated...

src/wui/load_or_save_game.cc
jmoerschbach: done

src/wui/savegametable.cc
jmoerschbach: done

src/wui/savegametable.cc
jmoerschbach: done

src/wui/savegametable.cc
jmoerschbach: done

src/wui/savegametable.cc
jmoerschbach: done

src/wui/savegametable.cc
jmoerschbach: done

src/wui/savegametable.cc
jmoerschbach: done

src/wui/savegametable.cc
jmoerschbach: done

jmoerschbach:
> 
> 
> I did a very quick test and found a bug: Uncompressed savegames show up as directories and not as savegames.

Oh, I wasn't aware of uncompressed savegames at all. I fixed that. Not sure if this is the most elegant way, though. What do you think. A possible alternative would be to try to actually load the savegame-directory and when that fails, treat it as a normal directory...

src/wui/load_or_save_game.h
jmoerschbach: Since SavegameTable is an abstract type, I need to use a pointer, right? Do you want me to use a raw pointer SavegameTable* table_?

src/wui/load_or_save_game.cc
gunchleoc: Let's keep it like it is then

src/wui/load_or_save_game.h
gunchleoc: Yes, this would be sufficient, because memory management for our UI toolkit is done by `UI::Panel`. Generally, we want to use `unique_ptr`, but the UI classes are a exception.

BTW when constructing the table, you can use a switch statement, that will be more efficient than the if/else.

gunchleoc:
Have a look at `FullscreenMenuMapSelect::fill_table()` on how uncompressed maps are detected. Maybe we can do something similar here?

jmoerschbach:
Hi @gunchleoc 
next try to address your review :)

FullscreenMenuMapSelect::fill_table() does something very similar to find out if it is a single file (by checking the filename's extension). If it is not considered as a regular file, it checks if it is a subdirectory and if so, adds a directory entry to the list.
The code here now tries to preload the savegame and if that fails, checks if it is a directory... Better?

gunchleoc:
Uncompressed save detection is working now on the load game screen :)

Found another bug though: When double-clicking on "example_folder" in the save game screen, Widelands saves the game to "example_folder.wfg" instead of entering the folder.

jmoerschbach:
> 
> 
> Uncompressed save detection is working now on the load game screen :)
> 
> Found another bug though: When double-clicking on "example_folder" in the save game screen, Widelands saves the game to "example_folder.wfg" instead of entering the folder.

fixed that

gunchleoc:
All good now - merged :)

PR #3566, opened 2019-11-12
Display seafaring buildings at the end of the list

### LINKED ISSUE #3565: Reorder buildings only available on seafaring maps
Some buildings like the shipyard and the weaving mill are only available when the map supports seafaring. These buildings are then placed directly before the military buildings which means, that the buildings following these buildings move one place to the right.

This is quite inconsistent. Icons & buttons should always be in the same position. Therefore, these buildings should always be displayed at the end of the list.

stonerl:
fixes #3565


hessenfarmer (dismissed):
Although we should have sought some minimal consensus on the forum this sounds reasonable and code looks good.

PR #3580, opened 2019-11-18
Fix Multiplayer Setup Window

### LINKED ISSUE #3578: Mulitplayer Setup Layout needs some fixing
The MPS-Layout needs some changes since the introduction of the Peaceful-mode ate up some screen space used for the client info. Also, there is a little too much space before and after the checkbox, especially on 800 × 600.

stonerl:
* Chatpanel is now properly aligned to the ok and back button
* Chatpanel only has focus when clicked on it (removes cursor when not active)
* Peaceful Mode checkbox has fixed spacing above and below
* Text for missing map/savegame files has been moved to map_info_
* client_info_ has been remove since it does not provide any information of value and only duplicates information already available.
* Full width button for map select.
* Start Game button has been moved below the back button.

fixes #3578


stonerl:
Here are some pictures:
shot0020
shot0021

jmoerschbach:
personally I would like the "start game" button to move down just above the "back" button (or maybe even below the "back button") 
Feels more natural... ;) 
The top and middle is all about configuration and the bottom would be about navigation ("back" and "next" aka "start game")

stonerl:
You're right. It does look way better. Here is a screenshot:
shot0023

stonerl:
I made the "Map/Save Game"-button wider and moved the map name into it. Looks much cleaner now.
Host:
shot0000
shot0001

gunchleoc:
LGTM. Let's merge when the CI finishes :)

PR #3583, opened 2019-11-19
Separate Armor Smithy from Weapon Smithy

kvilhaugsvik:
Armor Smithy was wrongly translated as Våpensmie. Våpensmie means Weapon Smithy. This made the two buildings hard to distinguish.

### DIFF
--- a/po/tribes/nn.po
+++ b/po/tribes/nn.po
@@ -525,7 +525,7 @@ msgstr "Demonter stad"
 #: ../../data/tribes/buildings/productionsites/atlanteans/armorsmithy/init.lua:94
 msgctxt "atlanteans_building"
 msgid "Armor Smithy"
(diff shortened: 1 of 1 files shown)


hessenfarmer (changes requested):
Widelands translation is done on transifex and only pulled in the files here. So this chanmge wouldn't be permanent. So please make an account and translate this string on transifex. We could always use translators there and would be very happy if you could find time to translate all the currently untranslated strings there (or at least some of them)

kvilhaugsvik:
I think I have changed it there too.

hessenfarmer:
If it is changed there it should go into the repo when our chieftain is pulling the translations the next time. Normally she does that weekly but at least once a month.
I think we should close this Pull request though.

stonerl:
Closing this one.

PR #3586, opened 2019-11-20
play sound on new messages when chat windows is closed

### LINKED ISSUE #1038: No sound at chat arrival when chat window is closed
When the chat window is closed in-game, no sound is played at chat message arrival.

 * date created: 2016-12-04T20:46:08Z
 * owner: gunchleoc
 * the launchpad url was https://bugs.launchpad.net/bugs/1647207

stonerl:
fixes #1038

### DIFF
--- a/src/wui/chat_overlay.cc
+++ b/src/wui/chat_overlay.cc
@@ -31,6 +31,7 @@
 #include "graphic/rendertarget.h"
 #include "graphic/style_manager.h"
 #include "graphic/text/rt_errors.h"
+#include "sound/sound_handler.h"
 #include "wlapplication_options.h"
 #include "wui/chat_msg_layout.h"
 #include "wui/logmessage.h"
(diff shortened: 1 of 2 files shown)


Notabilis (approved):
Oha, quite an old bug! Thanks for fixing. The diff is looking good and the bug is fixed for me.
A few things I noticed:
- In LAN games it works for me even without your fix. No idea why, I couldn't find the matching code path.
- If the chat window is open the sound might be placed twice, I think. I doesn't sound "wrong", but the sound might (not sure) be a bit louder when the window is open.
- The chat window plays the sound only for non-system messages, you are playing the sound always.

stonerl:
>     * In LAN games it works for me even without your fix. No idea why, I couldn't find the matching code path.

That for some reason never worked for me.

>     * If the chat window is open the sound might be placed twice, I think. I doesn't sound "wrong", but the sound might (not sure) be a bit louder when the window is open.
 
I tested this as well, but couldn't hear any difference. On the other hand I'm not gold standard, since my hearing isn't the best 😄 

>     * The chat window plays the sound only for non-system messages, you are playing the sound always.

To avoid this, we would have to check the messages' content. We'll see whether users complain about it or not and implement this later. 

Thanks for your review.

PR #3587, opened 2019-11-21
allow peaceful mode for territorial time

stonerl:
Since the game ends after 4 hours anyway, we can also allow peacful mode here.


Notabilis (approved):
Sounds sensible and testing works.
It might be frustrating if one player owns half the map after one hour and there is nothing the other player can do for the remaining three hours except for waiting for its defeat. Maybe the win condition should check for this case and end the game early?

stonerl:
> Maybe the win condition should check for this case and end the game early?

Sounds sensible. I'm going to open a bug report for it.

PR #3602, opened 2019-12-05
Ai fixes

### LINKED ISSUE #1514: Computer player does not evict experienced worker when he is needed somewhere else
See the yellow player of the attached replay. He has two deep marble mines, but an experienced worker (master miner) is missing in both of them. After a while, the workers in his other mines (gold and iron ore) become master miners. If they were sent away, they would probably walk to the deep marble mine and work there.
Note that the yellow player does nothing anymore because all of his constructionsites are lacking marble.

 * date created: 2014-03-28T14:13:27Z
 * owner: wl-zocker
 * the launchpad url was https://bugs.launchpad.net/bugs/1299011

hessenfarmer:
Implements an AI worker management
adds malus for unoccupied builfing type 
fixes crude_percent from overflowing
fixes #1514


src/ai/defaultai.cc
```diff
@@ -4260,6 +4262,8 @@ bool DefaultAI::check_productionsites(uint32_t gametime) {
 	// Get link to productionsite that should be checked
 	ProductionSiteObserver& site = productionsites.front();
 
+
+
```
stonerl: Any reason for these 2 empty lines?

stonerl (changes requested):
Could you run fix_formatting.py, please?

src/ai/defaultai.cc
hessenfarmer: no defintily no reason maybe a leftover of some logs.

hessenfarmer: funny thing is according to the log the script was already run by travis. The complaint in the log was about productionsites.cc where nothing has changed except 2 values in a formula.

stonerl: Travis runs the script only for testing but doesn't actually change anything. So, when it fails on the "Code Formatting" test it means you need to run fix_formatting.py locally and push the changes to the branch. On Launchpad bunny-bot took care of the formatting.

hessenfarmer: how do i do this on windows?

stonerl: Go to the base folder of your widelands git-repo.

Run ```python utils/fix_formatting.py```from there.

jmoerschbach: also consider installing provided git hooks which will take care of this automatically when comitting/pushing (execute <widelandsRepo>/install-githooks.sh)

hessenfarmer: had a look into the script for c++ clang-format is called which I don't have on windows. Travis should change the files imho. perhaps something went wrong with the git hook of travis.

src/ai/defaultai.cc
```diff
@@ -4369,7 +4391,11 @@ bool DefaultAI::check_productionsites(uint32_t gametime) {
 
 	// No upgrade without proper workers
 	if (considering_upgrade && !site.site->has_workers(enhancement, game())) {
-		considering_upgrade = false;
+		const BuildingDescr& bld = *tribe_->get_building_descr(enhancement);
+		BuildingObserver& en_bo = get_building_observer(bld.name().c_str());
+			if (get_stocklevel(en_bo, gametime, WareWorker::kWorker) < 1) {
+				considering_upgrade = false;
```
tibor95: Well, dont we need to have "considering_update = true" somewhere?

src/ai/defaultai.cc
```diff
@@ -4293,6 +4295,26 @@ bool DefaultAI::check_productionsites(uint32_t gametime) {
...
+	for (uint8_t i = 0; i <  site.site->descr().nr_working_positions(); i++) {
+		// get the pointer to the worker assigned to the actual position
+		const Worker* cw = site.site->working_positions()[i].worker;
+		if (cw) {  // a worker is assigned to the position
+			// get the descritpion index of the worker assigned on this position
+			DescriptionIndex current_worker = cw->descr().worker_index();
+			// if description indexes of assigned worker and normal worker differ
+			// (this means an experienced worker is assigned to the position)
+			// and we have none of the experienced workers on stock
+			if (current_worker != site.bo->positions.at(i) &&
```
tibor95, line 4309: Dont we need to check if we have junion for this position as well? If no junior is available, will the senior come back, will not he? - Just idea for consideration...

src/ai/defaultai.cc
```diff
@@ -705,7 +705,9 @@ void DefaultAI::late_initialization() {
 			}
 
 			for (const auto& temp_position : prod.working_positions()) {
-				bo.positions.push_back(temp_position.first);
+				for (uint8_t i = 0; i < temp_position.second; i++) {
```
tibor95, line 708: Oh, I am confused by this insertion, can you please explain in plain words - I admit I am not fully familiar with this worker stuff.

src/ai/defaultai.cc
hessenfarmer: we have. see line 4325. Mechanism is that we declare that to true at the beginning and then we perform a lot of checks why it should not be done setting this to false. In this case I just added the check if a worker is available in warehouses. so this gets only false if no experienced/needed worker in warehouse and no suitable worker in site. The check of warehouses is necessary as experienced workers will be kicked out now.

tibor95: OK then...

src/ai/defaultai.cc
hessenfarmer: that is a good question. normally if no junior exists and no tools exist this should be the case. However he might run into another building with an older request. Potetnially this might lead to a deadlock for barbarians and frisians with their smithes if not enough hammers could be produced in time. I wanted to keep it simple though and therefore I'd wait for some results if this ever happens. But it could be implemented if you deem it neccessary

src/ai/defaultai.cc
hessenfarmer: that was a nasty one, which led to a crash for the barbarian big inn for example. 
if we have a building with two workers of the same type (eg. lua declaration like "barbarians_inn_keeper=2" the old code just had only 1 position in BuildingObserver while the site observer had two which led to vector length mismatch error. 
This code now creates as much positions in the building observer as specified with the amount (temp_position.second)

src/ai/defaultai.cc
tibor95: Can this lead to endless loop of senior being kicked and returning and kicked again from a single site?

hessenfarmer:
ok I merged the format fixed files now thanks to stonerl for doing this

src/ai/defaultai.cc
tibor95: Oh, ok

src/ai/defaultai.cc
hessenfarmer: if no proper tool and no junior available and no other senior in any other site yes this is possible. But thats why I only kicked it out if no experinced worker on stock. 
So do you think I should add a check if junior or tool is available?

tibor95: Well, it is unambiguous. Also the fact whether the current site is or is not fully staffed can make difference. I dont have clear opinion what is best solution here :( Also, I am not that much into micromanagement... you can discuss it in forum, let other players say their opinion.

hessenfarmer: I am afraid they won't have one as this is not affecting their gameplay. it is just the AI kicking workers out normal players need to do this by themselves.

hessenfarmer: I think we should try with this solution and if we see AI's behaving weird on this I'll implement the check. this shouldn't affect training in any case I believe.

src/ai/defaultai.cc
hessenfarmer: merged the format fix branch so I believe this resolved

src/ai/defaultai.cc
tibor95: OK, after some thinking and providing that workers ends the current working cycle (thus no material is wasted) I agree with you and no fatal problem will happen

tibor95:
OK, not tested but from code point of view - this is fine for me.... do I need to do any special action?

hessenfarmer:
@tibor thanks for reviewing   
you could document your review as approved with the proper button. This is to be found top right in the "Files changed" tab

tibor95 (approved):
Code OK, not tested, but looks safe to me:)

stonerl (approved):

PR #3605, opened 2019-12-06
Fix scrolling

### LINKED ISSUE #3351: Keypresses affect the game window from the save menu
From menus that hide the game/editor, keypresses are often routed to the game/editor window in the background and then handled there normally, causing (usually unwanted) actions (e.g. pause/unpause game, regulate game speed, show minimap, show editor tools, etc). Whether this happens depends on which element is focussed, i.e., whether and how this element handles keys.

This works for example
- from the save menu if the focus is not on the edit box
- from the load menu in the editor
- from the map options menu in the editor if focus is on none of the edit boxes

Haven't checked what route the key handling takes to end up at the game/editor window.

On a related note: If a menu doesn't cover the whole window, mouse events might also get through to the window in the background. This might be okay for many things, but feels a bit weird for the map options menu in the editor for example.

Possible fixes:
1. Would it be enough (and feasible) to open the menus in modal mode? Is key/mouse handling automatically contained then?
2. The game/editor window could have some "state" variable indicating whether it is in background or not (and possibly other states). And depending on this state, key/mouse handling is suspended.
3. Deal with the handling in all the relevant the menus, but that seems to be the wrong way.

 * date created: 2018-11-07T00:21:34Z
 * owner: artydent
 * the launchpad url was https://bugs.launchpad.net/bugs/1802019

### LINKED ISSUE #3429: Increase speed of map move when using arrow keys
Moving the map with arrow keys is too slow.

Posted at the forum: https://wl.widelands.org/forum/topic/4512/

 * date created: 2019-05-07T05:58:17Z
 * owner: franku
 * the launchpad url was https://bugs.launchpad.net/bugs/1827992

### LINKED ISSUE #3603: Map scrolling with arrow key broken
**Describe the bug**
Since scrolling has been moved out of `InteractiveBase::think()` to `InteractiveBase::handle_key()`it does not work properly anymore. Diagonal scrolling isn't possible anymore and scrolling isn't as smooth as it was before

The choppy scrolling is caused by the fact that `InteractiveBase::handle_key()` doesn't get called as often as `InteractiveBase::think()`.

The diagonal scrolling problem is caused by `handle_key()` only reacting to the latest SDL key_press events. Pressing UP, scrolls up and pressing LEFT while holding UP scrolls left, since pressing left is the latest event handle_key() registers. Releasing UP while still holding left does nothing. Pressing UP again while holding LEFT results in scrolling upwards.

Implementing the old behavior (in think()) would cause other problems like hotkeys not being recognized since other UI Elements may have constant focus.

stonerl:
Implemented diagonal scrolling:
* Number pad: 1/3/7/9 scrolls diagonal. 
* 5 works like pressing HOME.
* Arrow keys: LEFT/UP, LEFT/DOWN, RIGHT/UP & RIGHT/DOWN scrolls diagonal

Scrolling distance is 1/8, with SHIFT 1/4 and with CTRL 1/16 of the screen resolution in x/y-direction and set to smooth.
fixes #3351 
fixes #3429


stonerl:
Scrolling is working again. I moved it back to InteractiveBase::think(). Also, Windows don't focus by default and focused windows loose focus when minimized. 

Windows now explicitly need `set_can_focus(true)`if they need focus.

Still, to do:
- [ ] Regain focus on Window::restore() for Seafaring Statistics Window
- [ ] Regain focus on Window::restore() for Messages Window
- [ ] Fix scrolling after clicking on dropdown button

src/wui/mapview.cc
```diff
@@ -587,10 +586,61 @@ Widelands::NodeAndTriangle<> MapView::track_sel(const Vector2i& p) {
 	return node_and_triangle;
 }
 
+bool MapView::scroll_map() {
```
Noordfrees, line 589: There are lots of similar-looking lines in this function. How about simplifying it like:
```
int32_t distance_to_scroll_x = 0;
int32_t distance_to_scroll_y = 0;
if (…left key down…) distance_to_scroll_x -=scrollDistanceX;
if (…right key down…) distance_to_scroll_x +=scrollDistanceX;
if (…up key down…) distance_to_scroll_y -=scrollDistanceY;
if (…down key down…) distance_to_scroll_y +=scrollDistanceY;
if (distance_to_scroll_x == 0 && distance_to_scroll_y == 0) return false;
pan_by(…distance_to_scroll_x, distance_to_scroll_y…)
return true
```
Also please use under_scores instead of CamelCase for the names of `ScrollDistanceX/Y` and make them `const`, and cache the value of `!(SDL_GetModState() & KMOD_NUM)` in a `const bool`.

Noordfrees (commented):
Code looking good, one comment inline. Will test this today…

Noordfrees:
Tested scrolling with the arrow keys and it works. (My keyboard doesn´t have a numpad so I can´t test that.) But the scrolling speed is much too high now IMHO. Please half the scrolling distance; and perhaps add Shift/Ctrl modifiers for faster/slower scrolling?

stonerl:
@Noordfrees I addressed your comments. Since I don't have a keypad myself we might need someone to test this. Although it should be working 😄

src/ui_basic/editbox.cc
```diff
@@ -265,11 +260,6 @@ bool EditBox::handle_key(bool const down, SDL_Keysym const code) {
...
 			return true;
 
-		case SDLK_KP_6:
-			if (code.mod & KMOD_NUM) {
-				break;
-			}
-			FALLS_THROUGH;
 		case SDLK_RIGHT:
 			if (m_->caret < m_->text.size()) {
 				while ((m_->text[++m_->caret] & 0xc0) == 0x80) {
```
ghost, line 265: Not sure if the prein-/decrements are confusing here.

src/wui/mapview.cc
```diff
@@ -587,51 +587,46 @@ Widelands::NodeAndTriangle<> MapView::track_sel(const Vector2i& p) {
...
-	return false;
+	// arrow keys
+	const bool kUP = get_key_state(SDL_SCANCODE_UP);
+	const bool kDOWN = get_key_state(SDL_SCANCODE_DOWN);
+	const bool kLEFT = get_key_state(SDL_SCANCODE_LEFT);
+	const bool kRIGHT = get_key_state(SDL_SCANCODE_RIGHT);
+
+	// numpad keys
+	const bool kNumlockOff = !(SDL_GetModState() & KMOD_NUM);
+	const bool kNP1 = get_key_state(SDL_SCANCODE_KP_1);
```
Noordfrees: Just one more suggestion (feel free to ignore):
```
#define kNP(x) const bool kNP##x = kNumlockOff && get_key_state(SDL_SCANCODE_KP_##x);
kNP(1) kNP(2) … kNP(9)
#undef kNP
```
This saves some repetitive code and you can get rid of the many `kNumlockOff` checks in the `if`s below

src/wui/mapview.cc
```diff
@@ -587,51 +587,46 @@ Widelands::NodeAndTriangle<> MapView::track_sel(const Vector2i& p) {
...
+	const bool kNP6 = get_key_state(SDL_SCANCODE_KP_6);
+	const bool kNP7 = get_key_state(SDL_SCANCODE_KP_7);
+	const bool kNP8 = get_key_state(SDL_SCANCODE_KP_8);
+	const bool kNP9 = get_key_state(SDL_SCANCODE_KP_9);
+
+	// set the scrolling distance
+	uint8_t denominator =
+	   ((SDL_GetModState() & KMOD_SHIFT) ? 4 : (SDL_GetModState() & KMOD_CTRL) ? 16 : 8);
+	uint16_t scroll_distance_y = g_gr->get_yres() / denominator;
+	uint16_t scroll_distance_x = g_gr->get_xres() / denominator;
```
Noordfrees: Please make these three variables `const`

src/ui_basic/editbox.cc
stonerl: Haven't really touched that part of the code because it is not in scope of this PR. But you can open a separate PR for this if you want to.

src/wui/mapview.cc
```diff
@@ -587,51 +587,46 @@ Widelands::NodeAndTriangle<> MapView::track_sel(const Vector2i& p) {
...
+	const bool kNP3 = get_key_state(SDL_SCANCODE_KP_3);
+	const bool kNP4 = get_key_state(SDL_SCANCODE_KP_4);
+	const bool kNP6 = get_key_state(SDL_SCANCODE_KP_6);
+	const bool kNP7 = get_key_state(SDL_SCANCODE_KP_7);
+	const bool kNP8 = get_key_state(SDL_SCANCODE_KP_8);
+	const bool kNP9 = get_key_state(SDL_SCANCODE_KP_9);
+
+	// set the scrolling distance
+	uint8_t denominator =
+	   ((SDL_GetModState() & KMOD_SHIFT) ? 4 : (SDL_GetModState() & KMOD_CTRL) ? 16 : 8);
```
Noordfrees: Suggestion: Exchange the semantics of Ctrl and Shift (in your gamespeed branch, Shift makes the effect of the keystroke less and Ctrl makes it stronger; here it´s the other way around)

Noordfrees (dismissed):
Tested and working well.
Three more suggestions inline.

src/wui/mapview.cc
stonerl: Usually we use SHIFT for a big in-/decrease, don't we? So, I would change it in the other branch.

stonerl: We use Shift as a modifier in other parts of the game. Is it used for small or big in-/decreases there?

Noordfrees: For InputQueueDisplays, Ctrl is the big change (min/max) and Shift is the change of everything by 1. In the attack box Ctrl is also more powerful than Shift. So personally I associate Shift with small and Ctrl with big deltas

src/wui/mapview.cc
```diff
@@ -587,10 +586,56 @@ Widelands::NodeAndTriangle<> MapView::track_sel(const Vector2i& p) {
...
+	if (kRIGHT || (kNumlockOff && (kNP3 || kNP6 || kNP9)))
+		distance_to_scroll_x += scroll_distance_x;
+
+	// do the actual scrolling
+	if (distance_to_scroll_x == 0 && distance_to_scroll_y == 0)
+		return false;
+	pan_by(Vector2i(distance_to_scroll_x, distance_to_scroll_y), Transition::Smooth);
+	return true;
+}
+
```
ghost, line 631: Not needed, but I wouldn't omit the block brackets on ifs, as its then easier to add code, etc.

ghost (changes requested):
Functionality-wise this changes are working well. But take a look at my code suggestions.

src/wui/mapview.cc
stonerl: Thanks, for this suggestion.

stonerl:
Addressed your comments. @nifker do you have a keypad on your keyboard?

Noordfrees (approved):

stonerl:
Thank you for your review.

ghost:
> Addressed your comments. @nifker do you have a keypad on your keyboard?

Yes I do and it worked - I just couldn't test the HOME button thingy.

stonerl:
> > Addressed your comments. @nifker do you have a keypad on your keyboard?
> 
> Yes I do and it worked - I just couldn't test the HOME button thingy.

Pressing 5 on the Keypad or the HOME button, should center your headquarter.

ghost:
Ah ok - yes thats working too.

PR #3611, opened 2019-12-09
make game speed more fine grained

stonerl:
SHIFT + PAGE UP/DOWN = +/- .25x
CTRL + PAGE UP/DOWN = +/-  10x
This also works with the menus buttons.

SHIFT + PAUSE Resets the speed to 1x

### DIFF
--- a/src/wui/interactive_gamebase.cc
+++ b/src/wui/interactive_gamebase.cc
@@ -301,14 +301,41 @@ void InteractiveGameBase::gamespeed_menu_selected(GameSpeedEntry entry) {
 
 void InteractiveGameBase::increase_gamespeed() {
 	if (GameController* const ctrl = get_game()->game_controller()) {
-		ctrl->set_desired_speed(ctrl->desired_speed() + 1000);
+		uint32_t const speed = ctrl->desired_speed();
+		switch (speed) {
+		case 0:
+			FALLS_THROUGH;
+		case 500:
+			FALLS_THROUGH;
+		case 1000:
+			FALLS_THROUGH;
+		case 1500:
+			ctrl->set_desired_speed(speed + 500);
+			break;
+		default:
+			ctrl->set_desired_speed(speed + 1000);
+			break;
+		}
 	}
 }
 
 void InteractiveGameBase::decrease_gamespeed() {
 	if (GameController* const ctrl = get_game()->game_controller()) {
 		uint32_t const speed = ctrl->desired_speed();
-		ctrl->set_desired_speed(1000 < speed ? speed - 1000 : 0);
+		switch (speed) {
+		case 0:
+			break;
+		case 500:
+			FALLS_THROUGH;
+		case 1000:
+			FALLS_THROUGH;
+		case 1500:
+			ctrl->set_desired_speed(speed - 500);
+			break;
+		default:
+			ctrl->set_desired_speed(speed - 1000);
+			break;
+		}
 	}
 }
 
--- a/src/wui/interactive_gamebase.cc
+++ b/src/wui/interactive_gamebase.cc
@@ -282,12 +282,24 @@ void InteractiveGameBase::rebuild_gamespeed_menu() {
 void InteractiveGameBase::gamespeed_menu_selected(GameSpeedEntry entry) {
 	switch (entry) {
 	case GameSpeedEntry::kIncrease: {
-		increase_gamespeed();
+		if (SDL_GetModState() & KMOD_SHIFT) {
+			increase_gamespeed(250);
+		} else if (SDL_GetModState() & KMOD_CTRL) {
+			increase_gamespeed(10000);
+		} else {
+			increase_gamespeed(1000);
+		}
 		// Keep the window open so that the player can click this multiple times
 		gamespeedmenu_.toggle();
 	} break;
 	case GameSpeedEntry::kDecrease: {
-		decrease_gamespeed();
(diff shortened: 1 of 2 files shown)


src/wui/interactive_gamebase.cc
```diff
@@ -282,12 +282,24 @@ void InteractiveGameBase::rebuild_gamespeed_menu() {
 void InteractiveGameBase::gamespeed_menu_selected(GameSpeedEntry entry) {
 	switch (entry) {
 	case GameSpeedEntry::kIncrease: {
-		increase_gamespeed();
+		if (SDL_GetModState() & KMOD_SHIFT) {
```
Noordfrees: Personally, I would find this more readable:
`increase_gamespeed(SDL_GetModState() & KMOD_SHIFT? 250 : SDL_GetModState() & KMOD_CTRL ? 10000 : 250);`
Also please make the three values `constexpr`.

src/wui/interactive_gamebase.cc
```diff
@@ -299,16 +311,54 @@ void InteractiveGameBase::gamespeed_menu_selected(GameSpeedEntry entry) {
...
+		case 1000:
+			if (current_speed < 1000) {
+				ctrl->set_desired_speed(1000);
+			} else {
+				ctrl->set_desired_speed(current_speed - current_speed % speed + speed);
+			}
+			break;
+		case 10000:
+			ctrl->set_desired_speed(
+			   current_speed >= speed ? current_speed - current_speed % speed + speed : speed);
```
Noordfrees: What does this line do? Please add () so it´s apparent which calculations the % operator refers to, and add a comment explaining the maths here.

src/wui/interactive_gamebase.cc
```diff
@@ -299,16 +311,54 @@ void InteractiveGameBase::gamespeed_menu_selected(GameSpeedEntry entry) {
...
+			break;
+		default:
+			break;
```
Noordfrees: Can this happen? Just increase the desired speed by the requested value and add a log() call for the unexpected argument.

Noordfrees: And use the `constexpr`s I requested in a another comment above as case labels here

src/wui/interactive_gamebase.cc
```diff
@@ -299,16 +311,54 @@ void InteractiveGameBase::gamespeed_menu_selected(GameSpeedEntry entry) {
...
-		uint32_t const speed = ctrl->desired_speed();
-		ctrl->set_desired_speed(1000 < speed ? speed - 1000 : 0);
+		uint32_t const current_speed = ctrl->desired_speed();
+		switch (speed) {
+		case 250:
+			ctrl->set_desired_speed(current_speed > speed ? current_speed - speed : 0);
+			break;
+		case 1000:
+			if (current_speed >= 2000) {
+				ctrl->set_desired_speed(current_speed - current_speed % speed - speed);
```
Noordfrees: add brackets and a comment explaining the math

src/wui/interactive_gamebase.cc
```diff
@@ -299,16 +311,54 @@ void InteractiveGameBase::gamespeed_menu_selected(GameSpeedEntry entry) {
...
+			} else if (current_speed > 1000) {
+				ctrl->set_desired_speed(1000);
+			} else {
+				ctrl->set_desired_speed(0);
+			}
+			break;
+		case 10000:
+			ctrl->set_desired_speed(
+			   current_speed > speed ? current_speed - current_speed % speed - speed : 0);
```
Noordfrees: here as well

src/wui/interactive_gamebase.cc
```diff
@@ -299,16 +311,54 @@ void InteractiveGameBase::gamespeed_menu_selected(GameSpeedEntry entry) {
...
+			break;
+		default:
+			ctrl->set_desired_speed(1000);
```
Noordfrees: Setting the speed to 1× does not belong in a function named `decrease`, please move the resetting call to a new function and make this `default` clause here behave like I requested for the `default` in `increase_gamespeed()`

Noordfrees (changes requested):
A number of comments inline. Will test today…

Noordfrees:
Tested, and found the semantics of the various modifiers quite unintuitive. If I press PageUp, I expect the game speed to be increased by 1×. Instead, it is sometimes 1×, sometimes 0.75× or 0.5× or 0.25×, depending on the previous speed. This should be changed so that pressing PageUp/PageDown without modifier, it always changes the speed by 1000, with Shift always by 250, and with Ctrl always by 10000, regardless of the speed before the keystroke.

src/wui/interactive_gamebase.cc
stonerl: Since I'm going to make it increase 1x and 10x regardless of the previous values (regarding your testing) These calculations aren't necessary anymore.

src/wui/interactive_gamebase.cc
stonerl: done

Noordfrees (approved):
Great :)

Noordfrees:
Do you want to explain the modifiers in the tutorial and/or the game tips?

stonerl:
> Do you want to explain the modifiers in the tutorial and/or the game tips?

Hmm, would be a good idea, wouldn't?

stonerl:
I'm merging this. The explanation can be handled in another branch.

PR #3613, opened 2019-12-11
results of about 100 rounds of training

tibor95:
Just some training after latest changes in AI related code


hessenfarmer (approved):
new AI files should be always ok. 
will test them though

hessenfarmer:
tested the values. looked good what Ai was doing

stonerl (approved):
LGTM

PR #3614, opened 2019-12-11
add some settings

### LINKED ISSUE #3507: ASan crashes reset sound settings
- Start Widelands, choose Options -> Sound, disable all five sound checkboxes. Also change some other settings.
- Close and restart Widelands. All settings were saved correctly, sound is disabled.
- Do something that causes an ASan error (tested with multiple development branches), or try force-quitting while the editor is loading
- Start Widelands. The sound settings are reset to the defaults; other changed options are unaffected.

Of course a crash means almost by definition undefined behaviour, so one can´t really expect everything to work correctly when the game crashes, but since only this one option section is affected (and I believe the sound settings were also not affected by crashes in b20), I consider this worth fixing.

stonerl:
* add option to only zoom when Ctrl is pressed
* add option to disable in-game clock
* read/write sound settings to config on start
fixes #3507


stonerl:
some pictures
shot0000
shot0001

hessenfarmer (approved):
did some testing works as desired

gunchleoc:
Code LGTM, merging :)

PR #3616, opened 2019-12-12
minimap overhaul

### LINKED ISSUE #528: Minimap can become too big at 2x Zoom
The size of the minimap of a 512x512 map is very big. I guess this is normal... but if i click 2x zoom the minimap gets a size where the bottom buttons aren't reachable anymore (screen resolution 1680x1050). In this situation it is not possible to get the minimap into normal zoom again. Closing it with right click and reopen it again, shows the minimap in 2x zoom status. The only way to get to a normal status is to exit the game and reload a saved game.

Steps to reproduce:
1. Start editor
2. Create a new map with 512x512 dimension
3. Open minimap
4. click 2x zoom on the bottom of the minimap

Depending on your resolution of the screen (my resolution: 1680x1050) the buttons at the bottom of the minimap switch below the bottom edge of the screen.

Maybe the 2x Zoom button should be disabled when the resulting height would be greater than the resolution of the screen?

 * date created: 2017-04-16T21:17:20Z
 * owner: franku
 * the launchpad url was https://bugs.launchpad.net/bugs/1683185

stonerl:
* buttons for the minimap are in one row
* minimap zoom2 limited to 600px
* minimap tries to scale up to 300px/400px in normal mode
* when map has a width of more than 300, zoom is disabled
* disable zoom if the minimap height gets greater than the game's
* check size when switching into window mode

fixes #528

### DIFF
--- a/src/graphic/minimap_renderer.cc
+++ b/src/graphic/minimap_renderer.cc
@@ -266,7 +266,11 @@ int scale_map(const Widelands::Map& map) {
 	const uint16_t map_h = map.get_height();
 	if (!(map_w > 300)) {
 		int longest_axis = (map_w >= map_h ? map_w : map_h);
-		return (300 - (300 % longest_axis)) / longest_axis;
+		if (longest_axis <= 150) {
+			return (300 - (300 % longest_axis)) / longest_axis;
+		} else {
+			return (600 - (600 % longest_axis)) / longest_axis;
+		}
 	}
 	return 1;
 }
--- a/src/graphic/minimap_renderer.cc
+++ b/src/graphic/minimap_renderer.cc
@@ -111,8 +111,9 @@ void draw_view_window(const Map& map,
 	}
 	}
 
-	const int width = zoom ? map.get_width() * 2 : map.get_width();
-	const int height = zoom ? map.get_height() * 2 : map.get_height();
+	const int width = zoom ? map.get_width() * 2 * scale_map(map) : map.get_width() * scale_map(map);
+	const int height =
+	   zoom ? map.get_height() * 2 * scale_map(map) : map.get_height() * scale_map(map);
 	const auto make_red = [width, height, &texture](int x, int y) {
 		if (x < 0) {
 			x += width;
@@ -160,15 +161,16 @@ void do_draw_minimap(Texture* texture,
 	const int32_t mapwidth = map.get_width();
 
 	for (uint32_t y = 0; y < surface_h; ++y) {
-		Widelands::Coords coords(
-		   Widelands::Coords(top_left.x, top_left.y + ((layers & MiniMapLayer::Zoom2) ? y / 2 : y)));
-		map.normalize_coords(coords);
-		Widelands::FCoords f = map.get_fcoords(coords);
-		Widelands::MapIndex i = Widelands::Map::get_index(f, mapwidth);
 		for (uint32_t x = 0; x < surface_w; ++x) {
-			if (x % 2 || !(layers & MiniMapLayer::Zoom2)) {
-				move_r(mapwidth, f, i);
-			}
+			Widelands::Coords coords(Widelands::Coords(
+			   top_left.x +
+			      ((layers & MiniMapLayer::Zoom2) ? x / (2 * scale_map(map)) : x / scale_map(map)),
+			   top_left.y +
+			      ((layers & MiniMapLayer::Zoom2) ? y / (2 * scale_map(map)) : y / scale_map(map))));
+			map.normalize_coords(coords);
+			Widelands::FCoords f = map.get_fcoords(coords);
+			Widelands::MapIndex i = Widelands::Map::get_index(f, mapwidth);
+			move_r(mapwidth, f, i);
 
 			uint16_t vision = 0;  // See Player::Field::Vision: 1 if seen once, > 1 if seen right now.
 			Widelands::PlayerNumber owner = 0;
@@ -215,7 +217,7 @@ Vector2f minimap_pixel_to_mappixel(const Widelands::Map& map,
 		break;
 	}
 
-	const float multiplier = zoom ? 2.f : 1.f;
+	const float multiplier = zoom ? 2.f * scale_map(map) : 1.f * scale_map(map);
 	Vector2f map_pixel = top_left + Vector2f(minimap_pixel.x / multiplier * kTriangleWidth,
 	                                         minimap_pixel.y / multiplier * kTriangleHeight);
 	MapviewPixelFunctions::normalize_pix(map, &map_pixel);
@@ -231,8 +233,10 @@ std::unique_ptr<Texture> draw_minimap(const EditorGameBase& egbase,
 	//       necessary. The created texture could be cached and only redrawn two
 	//       or three times per second
 	const Map& map = egbase.map();
-	const int16_t map_w = (layers & MiniMapLayer::Zoom2) ? map.get_width() * 2 : map.get_width();
-	const int16_t map_h = (layers & MiniMapLayer::Zoom2) ? map.get_height() * 2 : map.get_height();
+	const int16_t map_w = (layers & MiniMapLayer::Zoom2) ? map.get_width() * 2 * scale_map(map) :
+	                                                       map.get_width() * scale_map(map);
+	const int16_t map_h = (layers & MiniMapLayer::Zoom2) ? map.get_height() * 2 * scale_map(map) :
+	                                                       map.get_height() * scale_map(map);
 
 	std::unique_ptr<Texture> texture(new Texture(map_w, map_h));
 
@@ -255,4 +259,14 @@ std::unique_ptr<Texture> draw_minimap(const EditorGameBase& egbase,
 	texture->unlock(Texture::Unlock_Update);
 
 	return texture;
+}
+
+int scale_map(const Widelands::Map& map) {
+	const uint16_t map_w = map.get_width();
+	const uint16_t map_h = map.get_height();
+	if (!(map_w > 300)) {
+		int longest_axis = (map_w >= map_h ? map_w : map_h);
+		return (300 - (300 % longest_axis)) / longest_axis;
+	}
+	return 1;
 }
--- a/src/graphic/minimap_renderer.cc
+++ b/src/graphic/minimap_renderer.cc
@@ -273,7 +273,7 @@ int scale_map(const Widelands::Map& map, const bool zoom) {
 			return (600 - (600 % longest_axis)) / longest_axis;
 		} else if (longest_axis > 150) {
 			return (400 - (400 % longest_axis)) / longest_axis;
-		} else if (longest_axis < 300) {
+		} else {
 			return (300 - (300 % longest_axis)) / longest_axis;
 		}
 	}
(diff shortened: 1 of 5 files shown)


stonerl:
Some pictures:
shot0001
shot0002
shot0003
shot0004
shot0005

src/graphic/minimap_renderer.cc
```diff
@@ -90,7 +90,7 @@ void draw_view_window(const Map& map,
                       const MiniMapType minimap_type,
                       const bool zoom,
                       Texture* texture) {
-	const float divider = zoom ? 1.f : 2.f;
+	const float divider = zoom ? scale_map(map, false) : scale_map(map, true);
```
Notabilis: Is this "if-clause" really required? Can't you simply pass `zoom` as parameter to `scale_map()`?

src/graphic/minimap_renderer.cc
```diff
@@ -105,14 +105,16 @@ void draw_view_window(const Map& map,
...
+		center_pixel = Vector2i(origin.x / kTriangleWidth, origin.y / kTriangleHeight) *
+		               (zoom ? scale_map(map, true) : scale_map(map, false));
 		break;
 	}
 	}
 
-	const int width = zoom ? map.get_width() * 2 : map.get_width();
-	const int height = zoom ? map.get_height() * 2 : map.get_height();
+	const int width =
+	   zoom ? map.get_width() * scale_map(map, true) : map.get_width() * scale_map(map, false);
```
Notabilis: Same comment as above, you should be able to pull out `map.get_width()`:
`const int width = map.get_width() * scale_map(map, zoom);

src/graphic/minimap_renderer.cc
```diff
@@ -160,15 +162,16 @@ void do_draw_minimap(Texture* texture,
 	const int32_t mapwidth = map.get_width();
 
 	for (uint32_t y = 0; y < surface_h; ++y) {
-		Widelands::Coords coords(
-		   Widelands::Coords(top_left.x, top_left.y + ((layers & MiniMapLayer::Zoom2) ? y / 2 : y)));
-		map.normalize_coords(coords);
-		Widelands::FCoords f = map.get_fcoords(coords);
-		Widelands::MapIndex i = Widelands::Map::get_index(f, mapwidth);
 		for (uint32_t x = 0; x < surface_w; ++x) {
-			if (x % 2 || !(layers & MiniMapLayer::Zoom2)) {
```
Notabilis, line 169: Disclaimer: I have no idea how this function `do_draw_minimap` is working. But: This if() clause doesn't appear in your version. Is this correct? If you say so, that is good enough for me. (Wild guess: The if() makes sure that no useless pixels are drawn by small zoom?)

src/graphic/minimap_renderer.cc
```diff
@@ -256,3 +261,19 @@ std::unique_ptr<Texture> draw_minimap(const EditorGameBase& egbase,
 
 	return texture;
 }
+
+int scale_map(const Widelands::Map& map, const bool zoom) {
+	// If a map is wider than 300px we don't scale. Otherwise we fit as much as possible into a 300px
```
Notabilis: Is map.get_width() the map width in pixels or in triangles? Or is this the same?
Lots of magic numbers here. Where do they come from? My guess: A result of trial-until-it-looks-nice.

src/wui/minimap.cc
```diff
@@ -70,11 +70,18 @@ bool MiniMap::View::handle_mousepress(const uint8_t btn, int32_t x, int32_t y) {
...
-void MiniMap::View::set_zoom(int32_t z) {
+void MiniMap::View::set_zoom(const bool zoom) {
 	const Widelands::Map& map = ibase_.egbase().map();
-	set_size((map.get_width() * z), (map.get_height()) * z);
+	set_size(map.get_width() * scale_map(map, zoom), map.get_height() * scale_map(map, zoom));
 }
 
+bool MiniMap::View::can_zoom() {
+	const Widelands::Map& map = ibase_.egbase().map();
+	if (scale_map(map, true) == 1 || map.get_height() * scale_map(map, true) > ibase_.get_h() - 40) {
```
Notabilis: 40 is the size of the window border + minimap buttons?

src/wui/minimap.cc
```diff
@@ -205,21 +212,25 @@ void MiniMap::toggle(MiniMapLayer const button) {
 }
 
 void MiniMap::resize() {
-	view_.set_zoom((*view_.minimap_layers_ & MiniMapLayer::Zoom2) ? 2 : 1);
+	view_.set_zoom((*view_.minimap_layers_ & MiniMapLayer::Zoom2) ? true : false);
```
Notabilis: Please, just use `view_.set_zoom(*view_.minimap_layers_ & MiniMapLayer::Zoom2);` here... o_O

Notabilis (changes requested):
Overall testing worked good, nice to have a usable minimap on small maps.
Some observations:
- The "red box" that should display the current viewport is too small, no idea what it is showing. This is also visible in your demo-screenshots, e.g., the first one: The red box is a small rectangle in the middle of the player's territory but the viewport even shows the borders. Please fix this one, my further observations are "nice to have".
- The zoom button is still disabled after switching to fullscreen. Also, the map might be "lost" outside the screen when switching back to window mode, but this is a problem of all dialogs and most likely out of scope of this branch.
- The "is zoomed" state is stored and restored when re-opening the window, even when the resolution of the game window is now too small to contain the zoomed minimap.
- Just a thought: The minimap window might be too large for some players, in case they prefer to keep it always open in a corner. But we can deal with that if anyone complains.
- For small map, the minimap does look ugly/pixelated. But well, so be it. Personally, I like it, but no idea what younger players think of it. ^^
- An old bug, already in trunk: After switching to fullscreeen mode, the red box is only recalculated when the map is moved (even though the game window now displays more of the game map)

The code seems to be okay, though I haven't understood everything and was too lazy to dig deeper. But it is UI code and the result looks good, so I guess it is alright.
You have a lot of `?:` clauses in the code, I think all of them can be replaced by passing the condition directly as parameter to `scale_map()`. Or is there a reason that you still have them in the code?

src/graphic/minimap_renderer.cc
stonerl: That got me my had scratching as well, for a while. It only fills every second pixel when zooming. This cannot be used anymore, since we have a variable zoom factor now. This would also result in a vertical interlaced picture.

src/graphic/minimap_renderer.cc
stonerl: It is the map width in pixel. `map.get_width()` on a 400×400 map would give you 400.

The numbers are easily explained.

The max width for the minimap is 600px. This is because the 6 buttons used in the minimap have a width 100 each. That is the maximum we can scale a map to. if we have more than 600 pix the minimap layout completely breaks, as it does ATM.

Without zoom, we try to fit the map in a 300px wide window. And for maps with more than 150 in a 400px wide window. This makes sure that we don't end up with a minimap window that is too small (e.g. 64px or 151px).

src/wui/minimap.cc
stonerl: 40px are the window borders. The buttons would be an additional 20. But I reckoned we could sacrifice 10px of the top and 10px of the bottom border to a zoomed map.

src/wui/minimap.cc
stonerl: Thank you for your review. I addressed some of your comments.


> Overall testing worked good, nice to have a usable minimap on small maps.
> Some observations:
> 
>     * The "red box" that should display the current viewport is too small, no idea what it is showing. This is also visible in your demo-screenshots, e.g., the first one: The red box is a small rectangle in the middle of the player's territory but the viewport even shows the borders. Please fix this one, my further observations are "nice to have".
> 

I'll dig into this.

>     * The zoom button is still disabled after switching to fullscreen. Also, the map might be "lost" outside the screen when switching back to window mode, but this is a problem of all dialogs and most likely out of scope of this branch.
> 
>     * The "is zoomed" state is stored and restored when re-opening the window, even when the resolution of the game window is now too small to contain the zoomed minimap.
> 

That's why I don't disable the zoom button when the map is zoomed. So, the player can still resize the  map. Afterwards, the zoom button gets disabled. Haven't really looked into alternatives though.

>     * Just a thought: The minimap window might be too large for some players, in case they prefer to keep it always open in a corner. But we can deal with that if anyone complains.
> 

I think so, too.

>     * For small map, the minimap does look ugly/pixelated. But well, so be it. Personally, I like it, but no idea what younger players think of it. ^^
> 

I like it as well. But it is the way the map gets drawn that makes it pixelated.

>     * An old bug, already in trunk: After switching to fullscreeen mode, the red box is only recalculated when the map is moved (even though the game window now displays more of the game map)
> 

I'll look into it. 

> The code seems to be okay, though I haven't understood everything and was too lazy to dig deeper. But it is UI code and the result looks good, so I guess it is alright.
> You have a lot of `?:` clauses in the code, I think all of them can be replaced by passing the condition directly as parameter to `scale_map()`. Or is there a reason that you still have them in the code?

No, just me having not slept enough 😄. It is fixed.

stonerl:
@Notabilis The viewport problem should be solved. All the other issues need to be fixed otherwhere. Since this is a general problem with window resizing, that happens to all ingame windows.

Notabilis (dismissed):
Thanks for the changes, looks good now.
Can you please add your explanations about the pixel sizes as a comment to the code?
Beside that, feel free to merge as soon as the automatic checks are fine.

stonerl:
@Notabilis would you test again.

>     * The zoom button is still disabled after switching to fullscreen. Also, the map might be "lost" outside the screen when switching back to window mode, but this is a problem of all dialogs and most likely out of scope of this branch.
> 
fixed
>     * The "is zoomed" state is stored and restored when re-opening the window, even when the resolution of the game window is now too small to contain the zoomed minimap.
> 

fixed
>     * An old bug, already in trunk: After switching to fullscreeen mode, the red box is only recalculated when the map is moved (even though the game window now displays more of the game map)
> 
 Currently, no idea how to fix it.

Notabilis:
Nice, I didn't expect Widelands to offer such a notification. Its looking good!

A small, most likely old bug I noticed: When opening the window, the state of the buttons does not represent their logical state (i.e., they are not pressed down but the terrain is drawn anyway). This can be fixed by adding `update_button_permpressed();` in wui/minimap.cc after line 233, but I haven't checked for side effects. Also, I haven't checked why they become "unpressed" at all, so there might be a cleaner fix for it.

For the third bug you cited ("An old ..."), it seems to be enough to add `mainview_move();` after wui/interactive_base.cpp line 229 within the GraphicResolutionChanged notification handler.

stonerl:
@Notabilis should be fixed now.

Notabilis (approved):
Thanks, works fine. I wasn't able to find any further bugs, so feel free to merge when the tests are done.

src/wui/minimap.cc
gunchleoc: The border thickness can be calculated by using `get_w() - get_inner_w()` from the UI::Panel class. I'd prefer to keep this dynamic to be stable whenever we change something in the basic Window layout.

PR #3627, opened 2019-12-29
index.html

ericbucag:


gunchleoc (changes requested):
I'm sorry, but >I don't understand why we would have an html file in our C++ code.

PR #3629, opened 2019-12-29
WIP: Fix localedir on Windows

gunchleoc:
* Ensure that "locale" exists in the datadir
* Show an error message box and exit gracefully when it doesn't
* Shift creating of locale directory from compile.sh to CMake

Supersedes https://github.com/widelands/widelands/pull/3477, see also discussion there.


gunchleoc:
Comment from the other PR:

Is this really necessary? "create_symlink" does only work on windows with administrative privileges.
And: Even if there is a symlink to the locale data, you'll still have to start the binary explicitely with "--datadir".

I think it would be better to add a "--localedir" parameter or just describe in the build documentation that (if the dev wants to run with locales) to copy the files or make the symlink manually.

The current solution with the symlink is not platform independant.

hessenfarmer:
1. is there something specific to test here?    
2. Would it be possible to have the translations compiled and pushed to master with a Github action so we would always have actual translations in the datadir without the need to compile them each time. So we could gain a lot of time in our CI environments.

gunchleoc:
Seems like it doesn't work on Windows, so this needs to be revisited.

gunchleoc:
With CMake 3.14, there's a universal link command available: https://cmake.org/cmake/help/v3.14/command/file.html#create-link

My own system is still on cmake version 3.10.2, and we'll need to wait until everybody has been upgraded. So, I'll make a separate branch just for the warnings and leave the build system changes until later.

PR is up: https://github.com/widelands/widelands/pull/3757

PR #3631, opened 2019-12-30
Format C++ before Lua

gunchleoc:
Formatting Lua takes longer than C++, so let's do C++ first. Although it is fastest, I kept formatting Python at the bottom, because changes are infrequent and not everybody might have pyformat installed.


stonerl (approved):
LGTM

gunchleoc:
Thanks for the review!

PR #3645, opened 2020-01-10
Productionsite performance

hessenfarmer:
Changes Productionsite Productivity from attempt based to time based
after positive feedback this could be merged however it might be good to rename some things to reflect new status. (e.g. calc_statisitcs to format_statistics and get rid of the crude_ as it is no more crude.) I believe there is a redundant funtion in AI with this change (get_statistics_percent is redundant to get_crude_percent now)


src/map_io/map_buildingdata_packet.cc
```diff
@@ -744,10 +744,20 @@ void MapBuildingdataPacket::read_productionsite(
...
+			if (packet_version >= 8) {
+				productionsite.crude_percent_ = fr.unsigned_32();
+			} else {
+				uint16_t const stats_size = fr.unsigned_16();
+				uint8_t ok = 0;
+				for (uint32_t i = 0; i < stats_size; ++i) {
+					if (fr.unsigned_8()) {
+						ok++;
+					}
+				}
```
Noordfrees, line 757: Suggestion for needing one variable less:
```
uint16_t ok = 0;
for (uint16_t i = fr.unsigned_16(); i; --i) {
  if (fr.unsigned_8)) {
    ++ok;
  }
}
```

Noordfrees (commented):
Code looking good, one codestyle suggestion inline. Will test soon

Noordfrees:
Okay the numbers behaviour is intuitive and useful. However there is one strange thing: When I set the building´s input wares to 0, the productivity falls gradually in 1% steps. But the trend jumps between '–' and '='. That should be fixed.

hessenfarmer:
> 
> 
> Okay the numbers behaviour is intuitive and useful. However there is one strange thing: When I set the building´s input wares to 0, the productivity falls gradually in 1% steps. But the trend jumps between '–' and '='. That should be fixed.

okay this is a rounding issue. I will use the unrounded value for the trend this should fix it. (in fakt it is falling not by 1% but 0.5% if you are under 20%

src/map_io/map_buildingdata_packet.cc
hessenfarmer: okay will try this

hessenfarmer: doesn't work as I need to divide by the length of the old stats vector

src/logic/map_objects/tribes/productionsite.cc
```diff
@@ -1020,8 +1018,8 @@ void ProductionSite::train_workers(Game& game) {
 }
 
 void ProductionSite::notify_player(Game& game, uint8_t minutes, FailNotificationType type) {
-	if (last_stat_percent_ == 0 ||
-	    (last_stat_percent_ <= descr().out_of_resource_productivity_threshold() &&
+	if (last_stat_percent_ / 10 == 0 ||
+	    (last_stat_percent_ / 10 <= descr().out_of_resource_productivity_threshold() &&
```
tibor95: What is this '/10' ?
EDIT: OK ignore this :)

src/logic/map_objects/tribes/productionsite.cc
```diff
@@ -421,15 +421,13 @@ InputQueue& ProductionSite::inputqueue(DescriptionIndex const wi, WareWorker con
 /**
  * Calculate statistic.
  */
-void ProductionSite::calc_statistics() {
-	// TODO(sirver): this method does too much: it calculates statistics for the
-	// last few cycles, but it also formats them as a string and persists them
-	// into a string for reuse when the class is asked for the statistics
-	// string. However this string should only then be constructed.
+void ProductionSite::format_statistics_string() {
+	// TODO(sirver): this method does too much: it formats the actual statistics
```
tibor95, line 425: There was an idea of "normalizing" percent like from range 3-97 to 0-100, this was not implemented, correct? If you are going to do this, this function would be the best place I think...

hessenfarmer: I just wrote this morning in the forum that my tests have shown it reaches 0% but not 100% (mostly 98% although I have seen 99% as well. this is due to having an integer division in here which discards the remainder resulting in a floor function basically. so as we reach 0% by this I beleive we can live with 98% max productivity. But i will await feedback from the users on this.

tibor95: simple 'actual_statistics+5/10' would solve the rounding issue...

hessenfarmer: I believe this would result in exactly the same as 5/10 is zero in integer division

tibor95: I dont uderstand, we would change
999/10 = 99
to 
(999+5) /10 = 100
Oh now I see, I should had put the brackets to my initial suggestion

src/logic/map_objects/tribes/productionsite.cc
gunchleoc: It's better to use `get_statistics_percent()` and lose the `/ 10`, because I asked myself that question too.

gunchleoc (commented):
Added 1 nit, rest of the code LGTM. Not tested.

src/logic/map_objects/tribes/productionsite.cc
hessenfarmer: I used get_actual_statistics now, because I believe get_statistics_percent might be superfluous now. But I have to check whether it is used somewhere else.

gunchleoc:
Thanks! Will merge :)

hessenfarmer:
@Gunchleoc 
Basically I wanted to wait for feedback, and I hadn't implemented the normalisation yet. But I will add the normalisation code and then we can wait if somebody complains to eventually undo this. Thanks for taking care anyway.

Issue #3633, opened 2019-12-30
Multiplayer savegame distribution very slow


hbeni:
**Describe the bug**
The distribution of saved multiplayer games is very slow. It takes minutes to transfer some 100k files between two machines in a 100Mb/s wired LAN.

**To Reproduce**
- Play in multiplayer
- save the game
- reload

**Expected behavior**
I expect the transmission to complete in few seconds, depending on the connection speed.

**Version:**
 - OS: Debian testing
 - Widelands Version: build-20 (Release)

mahendra9911:
hu

gunchleoc:
@Notabilis Any idea where the bottleneck is?

Notabilis:
Yes, it is in our code. My last branch was a preparation for fixing this, when the branch "Refactor gamehost" is merged I will prepare a branch that fixes this bug.

gunchleoc:
refactor-gamehost has been merged https://github.com/widelands/widelands/commit/3a428bd93712294d6f09634541a01d053e2b7da9

Notabilis:
Thanks for the notification, I pushed a branch proposing a fix for this.

hessenfarmer:
fixed by #3649

PR #3649, opened 2020-01-11
Faster file transfers

Notabilis:
Speeding up file transfers by sending the whole file at once instead of waiting for acknowledges of every part. Waiting for acknowledges is not required, since the underlying TCP connection takes care of completeness and order of the packets. This reduces the transmission time over the metaserver from over 3 minutes to 15 seconds for my test savegame.

A problem with sending the complete file at once is that chat messages become delayed while the file is transmitted. The underlying problem is that the buffer in the operating system gets filled and it takes some time until it is emptied and the chat message packet is send. To fix this, two approaches are currently present in the code for testing, one of them should be removed before merging. The currently active variant (see the `#ifdef __linux__` block) only passes data to the operating system when the sending buffer for the TCP connection is more or less empty. The other approach (see function `reduce_send_buffer()`) reduces the size of the buffer. The first variant performed a bit better in my tests, but is platform dependent.

Fixes issue #3633.
Also, fixing an unrelated buffer overflow with LAN game promotions.


src/network/bufferedconnection.cc
```diff
@@ -304,6 +326,23 @@ void BufferedConnection::start_receiving() {
 	   });
 }
 
+void reduce_send_buffer(boost::asio::ip::tcp::socket& socket) {
```
stonerl: `../src/network/bufferedconnection.cc:329:6: warning: no previous prototype for function 'reduce_send_buffer' [-Wmissing-prototypes]`
I get this warning during compiling.

stonerl:
`../src/network/bufferedconnection.cc:329:6: warning: no previous prototype for function 'reduce_send_buffer' [-Wmissing-prototypes]`
I get this warning during compiling.

stonerl:
@Notabilis the `#ifdef __linux__` part also works on macOS without any problems.

Notabilis:
Thanks for testing, good to know that the code works under macOS as well.
The missing function protoype is fixed.

stonerl:
@hessenfarmer You have a Windows machine, haven't you? Could you test whether the `#ifdef __linux__` part also works there?

hessenfarmer:
don't know how to test this. in fact it would be the else tree of the #ifdef statement containing the sleep which should be executed. Will download and compile the branch though

stonerl:
@hessenfarmer change the `#ifdef __linux__`to `#ifndef __linux__`

Notabilis:
It will probably fail under Windows, but feel free to try. Might work under MinGW, but I am unsure whether we should rely on that one being present.
I am leaning towards using the (currently out commented) alternative, though (the one in `reduce_send_buffer()` instead of the `#ifdef ... #else` clause). Its platform independent and should be (nearly?) as performant as the Linux specific code. Compared to current trunk both are pretty fast anyway.
I will push a commit enabling that code. Can you both please start a multiplayer host with it (has to be an online game / over the relay, no players required) and check the log output for a line saying something about "[BufferedConnection] Reduced send buffer size"? It should either be present or a "Failed to reduce..." line should exist.

stonerl:
Looks good.

Host:

```
[Host]: starting up.
[BufferedConnection] Trying to connect to *:*:*:*:*:*:*:*:* ... success.
[BufferedConnection] Reduced send buffer size
[BufferedConnection] Starting networking thread
```

Client:

```
InternetGaming: Client tries to join a game with the name test
InternetGaming: Received ips of the game to join: *:*:*:*:*:*:*:*:* *.*.*.*.
InternetGaming: Client update on metaserver.
[BufferedConnection] Trying to connect to *:*:*:*:*:*:*:*:* ... success.
[BufferedConnection] Reduced send buffer size
[BufferedConnection] Starting networking thread
[Client] SETTING_MAP 'Three Fish 1.0' 'save/wl_autosave_nethost_00.wgf'
[Client] Pong!
```

hessenfarmer:
I tried it on my Windows machine and I saw the line in question when I entered the lobby and opened a game. couldn't try to actually send a game however. so it seems to work

Notabilis:
Thanks for testing again!
If you don't object, I will remove the OS specific code and prepare the branch for merging.

stonerl (approved):

PR #3664, opened 2020-02-01
Automatically run fix_formatting.py on master and pull requests

Noordfrees:
Formatting your code
## Automatically run fix_formatting.py on master and pull requests

On every push to master and any branch with an open pull request, GitHub runs the formatting suite on it. If there are changes, they are pushed back to the branch.

There is one problem with this feature: GitHub forbids pushing to the head branch of a PR if it is in a different fork, i.e. PRs from `Noordfrees:branchname` to `widelands:master` could not be formatted using the default workflow. 
To work around this, I instead launch the formatting suite on *every* push to *every* branch, and if it is not the master branch and has no open PR, the changes will be logged but not committed or pushed. 
As a side effect, you will see a red X or green checkmark next to every commit you push, which indicates whether fix_formatting is happy with your code.

**In order for this feature to work on branches in your fork,** you need to enable *local and 3rd party actions* in `github.com/(username)/widelands/settings/actions`.
Since the opening of a pull request does not trigger the action workflow, your code will only be formatted the first time you push *after* you opened the PR.
Formatting takes about 3-4 minutes. Don't forget to `git pull` the changes.

GitHub's runner installs clang-format version 6.0, and I tried in vain to get a newer version to run. Therefore this PR's diff is pretty big due to the formatting, but the only files that need review are `format.yaml` and `.travis.*`. (I removed the fix_formatting check from travis as it is now no longer needed). Formatting of `master` is untested of course.

You can test this branch by pushing commits to it that mess up the source code formatting, and watch as the changes are automagically undone four minutes later :)


Noordfrees:
@stonerl could you please check whether 3rd party actions are enabled in https://github.com/widelands/widelands/settings/actions ? Only admins can view and change this setting.

stonerl:
> @stonerl could you please check whether 3rd party actions are enabled in https://github.com/widelands/widelands/settings/actions ? Only admins can view and change this setting.

Activated it.

Noordfrees:
Thanks :)

Now I´m only working on some performance cosmetics, and then this branch is ready for merging

hessenfarmer:
Looks good for me However we should seek approval of GunChleoc probably as she is still our chieftain.

stonerl:
Before this can be merged, all test-files need to be reverted to their original state.

Only:

* .github/workflows/format.yaml
* .travis.sh
* .travis.yml

should be merged in this PR.

hessenfarmer:
> 
> 
> Before this can be merged, all test-files need to be reverted to their original state.
> 
> Only:
> 
>     * .github/workflows/format.yaml
> 
>     * .travis.sh
> 
>     * .travis.yml
> 
> 
> should be merged in this PR.

I believe they will be changed to the current state afterwards again as the changes to these files are the result of the script. That has already been run on this PR. It is a bit like a hen / egg problem. 
The differences in the files are caused by the script running a different version of clangformat than that they were originally checked by.

gunchleoc:
It's strange that the results are so different from what I get on my machine, since I have clang-format 6.0.0. I am in favor of this feature though, because it will make our workflow easier. So, we can live with some additional merge conflicts in our branches for this. Painful once and then plain sailing going forward :)

I think we should run it on master only. I already see myself forgetting to pull my branches when I have a PR open an ending up with annoying push failures and having to force push all the time.

Noordfrees:
Personally I'd prefer to have my PRs formatted automatically. You can easily turn it off for branches in your fork by disabling actions in your fork settings, wouldn't that suffice?

gunchleoc:
> You can easily turn it off for branches in your fork by disabling actions in your fork settings, wouldn't that suffice?

A valid point. We'll need to document this somewhere. Gotta run now, can be a few days until I check back in.

gunchleoc:
Thanks for the documentation! We still need to get rid of the following documentation and the corresponding files, otherwise we'll have a code formatting edit war on our hands:

> To reduce the amount of changes introduced by automatic formatting, you can install our githook that will automatically format any changed C++ files when you commit:
> 
>     Install clang-format
>     Run ./install-githooks.sh
>

gunchleoc (approved):
Looking good now, thanks!

Please don't forget to squash when you merge, to keep the master branch history tidy.

PR #3668, opened 2020-02-03
added savegame compat code

hessenfarmer:
fixed broken savegame compatibility due to 3 programs productivity change


Noordfrees:
IMHO having a specialised lookup table is a bit overkill. Just rename one of the programs back to the original name.

hessenfarmer:
Basically you are right however I implemented it the way other savegame issues were solved too. So I believe as it is working this way it would help us with future compatibility issues if necessary. And as this is only used in loading at startup it shouldn't cost much.

Noordfrees:
But if we now add a fifth tribe which has a building with a program named "smelt_iron", this will also be looked up and changed to "smelt_iron_1", which does not exist. So looking up program names always implicitly introduces program name restrictions and program substitution bugs. 
If you want to keep this code, you need to increase the packet version and only call the lookup for savegames with the older packet version.

src/map_io/map_buildingdata_packet.cc
```diff
@@ -681,6 +681,7 @@ void MapBuildingdataPacket::read_productionsite(
 			//  skipped programs
 			uint32_t const gametime = game.get_gametime();
 			for (uint8_t i = fr.unsigned_8(); i; --i) {
+
```
Noordfrees: Please remove the newline again

hessenfarmer:
> 
> 
> But if we now add a fifth tribe which has a building with a program named "smelt_iron", this will also be looked up and changed to "smelt_iron_1", which does not exist. So looking up program names always implicitly introduces program name restrictions and program substitution bugs.
> If you want to keep this code, you need to increase the packet version and only call the lookup for savegames with the older packet version.

Very valid point so I'll keep the lookup code in (for possible future use, but empty the table while implementing it like you suggested.

Noordfrees (approved):
Code looking good now :)
untested

gunchleoc:
How about resetting to the start if a program isn't found? The game state won't be perfect, but we won't need to keep a table.

hessenfarmer:
> 
> 
> How about resetting to the start if a program isn't found? The game state won't be perfect, but we won't need to keep a table.

Could be done as well. However this should go in as it fixes the break in savegame compatibility I accidentally introduced with the 3 programs performance improvement.

gunchleoc:
Can you open a new issue before we merge this, so we won't forget?

hessenfarmer:
done see #3671

gunchleoc:
Thanks! Let's squash & merge when the CI is ready.

PR #3672, opened 2020-02-06
Cleanup luainterface include

gunchleoc:
Remove superfluous include and dependency from wl_library logic_map_objects


Notabilis (approved):
The checks say it compiles, so I guess this can go in.

PR #3674, opened 2020-02-06
Made some functions const to help the compiler optimize

gunchleoc:


src/logic/map_objects/tribes/worker.h
```diff
@@ -77,6 +77,10 @@ class Worker : public Bob {
 	explicit Worker(const WorkerDescr&);
 	~Worker() override;
 
+	Player& owner() const {
```
Notabilis: Why have you added this method?
Are you adding it intentionally (okay by me) or is it "added" because it has been removed from master since starting this branch?

Notabilis (dismissed):
I haven't bothered compiling and testing it locally, but based on the diff it looks okay.
Please take a look at my comment, though.

src/logic/map_objects/tribes/worker.h
gunchleoc: Removed.

gunchleoc:
Compiling + test suite should be sufficient for this one. Thanks for the review!

AppVeyor failure is the usual timeout.

PR #3681, opened 2020-02-07
removed 2 unused properties

hessenfarmer:
removed some unused parameters from Ai Code


tibor95 (commented):
I hope attributes like BuildingAttribute::kRanger + kLumberjack and the one for quarry are not affected, It seems they are not, but we need to be sure

hessenfarmer:
> 
> 
> I hope attributes like BuildingAttribute::kRanger + kLumberjack and the one for quarry are not affected, It seems they are not, but we need to be sure

They are not as they are defined by the collects ware from map Ai hint.
However this change unveiled another problem. I'll discuss this in the forum.

src/ai/defaultai.cc
```diff
@@ -5662,18 +5644,20 @@ BuildingNecessity DefaultAI::check_building_necessity(BuildingObserver& bo,
 			}
 		} else if (bo.type == BuildingObserver::Type::kMine) {
 			bo.primary_priority = bo.max_needed_preciousness;
+			// NOCOM the following could be addressed as mines_per_type[bo.mines].total_count as well
+			// NOCOM any reason for leaving this with in_construction + finished
```
tibor95: I see no reason not to use total_count here, so feel free to change it...

hessenfarmer: In the following there are multiple instances similar to this. Shall I go over them as well?

tibor95: You can, I think the total_count() was introduced later, this is why it is this way

hessenfarmer: Ok. Thanks for the explanation.

hessenfarmer:
> 
> 
> I hope we will not accumulate unoccupied and non-working mines...

That is one issue where this needs polish. Basically I could reproduce the behaviour of the old state with 1 coal mine and 2 crystal/marble mines and 2 iron/gold mines with the corrected definition of KBuildingMatProducer. However AI lacked in all tests I made from severe coal shortage so I'Ll try to get more coal mines build. Furthermore as build material is short in the beginning I'll try to make it finish the first mine before building another.

src/ai/defaultai.cc
```diff
@@ -4887,14 +4861,14 @@ bool DefaultAI::check_mines_(uint32_t const gametime) {
 	bool single_critical = false;
 	if ((site.bo->is(BuildingAttribute::kBuildingMatProducer) ||
 	     site.bo->mines == iron_resource_id) &&
-	    mines_per_type[site.bo->mines].total_count() == 1) {
+	    mines_per_type[site.bo->mines].finished == 1) {
```
tibor95, line 4860: the logic changes here, total vs finished - was it intentional?

src/ai/defaultai.cc
```diff
@@ -4887,14 +4861,14 @@ bool DefaultAI::check_mines_(uint32_t const gametime) {
...
 		single_critical = true;
 	}
 
 	// first get rid of mines that have been  missing workers for some time (6 minutes),
```
tibor95: you might change  6 -> 10

src/ai/defaultai.cc
```diff
@@ -5662,18 +5632,16 @@ BuildingNecessity DefaultAI::check_building_necessity(BuildingObserver& bo,
...
 			    site_needed_for_economy != BasicEconomyBuildingStatus::kDiscouraged) {
 				// unless a mine is prohibited, we want to have at least one of the kind
 				bo.max_needed_preciousness = bo.max_preciousness;
 				return BuildingNecessity::kNeeded;
-			} else if (((mines_per_type[bo.mines].in_construction +
-			             mines_per_type[bo.mines].finished) == 1) &&
-			           bo.is(BuildingAttribute::kBuildingMatProducer) &&
+			} else if (mines_per_type[bo.mines].finished == mines_per_type[bo.mines].total_count() &&
+			           bo.current_stats >
+			              (85 + std::abs(management_data.get_military_number_at(173)) / 10) &&
```
tibor95: you remove number at 129 (few lines below) and add 173. Intentional?

tibor95 (commented):
Few comments left, generally OK. We are waiting for testing, correct?

src/ai/defaultai.cc
hessenfarmer: yes before a single critical mine could be dismanteled when another was in construction this led to constant rebuild and could lead to a stuck situation if not enough wares for the construction of the mine

src/ai/defaultai.cc
hessenfarmer: ok thanks did not notice that

src/ai/defaultai.cc
hessenfarmer: no that was 2 single steps. First in the deletion I was not aware that I freed a number. So I'll reuse 129 for this.  
Always good to have a reviewer

tibor95 (commented):
for me OK.
Can I give approve or we wait for testing?

hessenfarmer:
You may approve and we wait for some test to merge. or you wait with the approval to prevent unwanted merge. Maybe it is save for the latter. 
One question do you have any benchmark from your training runs to compare the result with? that might be helpful as my judgement is only based on visible results in a well known environment.

tibor95:
If I approve it will be merged immediately - so let wait.. 

No, there is no benchmark.There is some score generated by AI training, but it is more-less zero sum game, so I see no way how to use it...
However I can imagine something like reference wai files and use them to fight different wai files... but this is not applicable here...

gunchleoc:
If there anything specific you want this tested for?

hessenfarmer:
basically you could run a 4 player local lan game with 4 AI one time with current trunk and one time with this branch and evaluate there are no shortfalls or in best case performance is improved.  
improvement should be recognizeable in mines handling (in trunk Ai does not build more then 2 of a type if at all)  
the other changes were only a cleanup that needed less extra definition on lua side.

src/ai/defaultai.cc
```diff
@@ -5663,18 +5633,16 @@ BuildingNecessity DefaultAI::check_building_necessity(BuildingObserver& bo,
...
 			    site_needed_for_economy != BasicEconomyBuildingStatus::kDiscouraged) {
 				// unless a mine is prohibited, we want to have at least one of the kind
 				bo.max_needed_preciousness = bo.max_preciousness;
 				return BuildingNecessity::kNeeded;
-			} else if (((mines_per_type[bo.mines].in_construction +
-			             mines_per_type[bo.mines].finished) == 1) &&
-			           bo.is(BuildingAttribute::kBuildingMatProducer) &&
+			} else if (mines_per_type[bo.mines].finished == mines_per_type[bo.mines].total_count() &&
+			           bo.current_stats >
+			              (85 + std::abs(management_data.get_military_number_at(129)) / 10) &&
```
gunchleoc: This line is producing a compiler warning:

```
../src/ai/defaultai.cc: In member function ‘Widelands::BuildingNecessity DefaultAI::check_building_necessity(Widelands::BuildingObserver&, DefaultAI::PerfEvaluation, uint32_t)’:
../src/ai/defaultai.cc:5642:32: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
               bo.current_stats >
               ~~~~~~~~~~~~~~~~~^
                  (85 + std::abs(management_data.get_military_number_at(129)) / 10) &&
                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

gunchleoc:
I have done some testing now, all good :)

This can go in once the compiler warning is fixed.

gunchleoc (dismissed):
All good now :)

I merged master again because of the big include changes there.

gunchleoc (approved):
Changes LGTM. Let's have it :)

PR #3682, opened 2020-02-07
Tribes fixes

### LINKED ISSUE #1651: Atlantean dark root flour economy limit not considered
The Atlantean mill will grind every dark root into flour, whatever limit is set in the economy settings.

 * date created: 2016-10-30T01:04:41Z
 * owner: tmk
 * the launchpad url was https://bugs.launchpad.net/bugs/1637788

### LINKED ISSUE #3270: Add "unless economy needs log" to charcoal kilns
Add "unless economy needs log" to charcoal kilns and playtest how gameplay will handle, especially when logs get short.

is might save us some of the not so interesting micromanagement.

https://wl.widelands.org/forum/topic/4212/?page=5#post-24277

 * date created: 2018-04-21T14:41:35Z
 * owner: gunchleoc
 * assignee: kxq
 * the launchpad url was https://bugs.launchpad.net/bugs/1765963

hessenfarmer:
fix #1651 
fix some issues from the forum see https://www.widelands.org/forum/topic/4755/


data/tribes/buildings/productionsites/frisians/charcoal_burners_house/init.lua
```diff
@@ -78,18 +78,18 @@ tribes:new_productionsite_type {
          -- TRANSLATORS: Completed/Skipped/Did not start making a charcoal stack because ...
          descname = _"making a charcoal stack",
          actions = {
-            "return=skipped unless economy needs coal",
+            "return=skipped when economy needs log",
             "return=failed unless site has log:3",
-            "callworker=make_stack",
```
gunchleoc, line 83: Is make_stack still being called somewhere?

data/tribes/workers/frisians/charcoal_burner/init.lua
```diff
@@ -29,6 +29,10 @@ tribes:new_worker_type {
          "callobject=with_stack",
          "animate=stacking_2 1000",
          "return"
+      },
+      find_pond = {
```
gunchleoc, line 33: Do older savegames still load?

gunchleoc (commented):
2 questions.

Still needs testing to see how it performs.

data/tribes/buildings/productionsites/frisians/charcoal_burners_house/init.lua
hessenfarmer: yes 2 lines below. Find_pond is a helper function.  
It was suggested to first consume and then make a stack in an empty pond. To ensure we don't consume if there isn't the possibility on the map to really produce I added this helper just to check whether there is a free pond to burn coal in. by this I could move make stack after consuming the wood.

data/tribes/workers/frisians/charcoal_burner/init.lua
hessenfarmer: they should load in any case as adding a new program never broke anything. Only deleting and renaming is tricky as the old status is saved in the savegame. By the way this is the same mechanism as in the last turn withthe 3 Programs performance turn. There we finally added a second program as well. But compatibility only broke cause I renamed the old program which I then reverted.   
I have no savegame to test this though.

Noordfrees:
Do I understand the intended charcoal kiln change correctly: It will produce coal if and only if no logs are needed; so it will produce even if no coal is needed and there are enough logs, but it won't produce coal if coal and logs are both needed. Correct? If so, this will make it impossible to stockpile logs and will result in unresolvable coal shortages when logs are scarce, so **-1** from me.

hessenfarmer:
yes that was the idea. Reason was it should have needed less micromanagement, and constantly relief the food insustry by producing some extra coal. 
However it doesn't work as intended and is bad for frisian AI as well
so -1 from my side as well.

data/tribes/workers/frisians/charcoal_burner/init.lua
```diff
@@ -29,6 +29,10 @@ tribes:new_worker_type {
...
+         "findobject=attrib:pond_dry radius:8",
+         "return"
```
Noordfrees: This doesn't need a `return`, the `findobject` alone is enough

data/tribes/buildings/productionsites/frisians/charcoal_burners_house/init.lua
```diff
@@ -80,16 +80,16 @@ tribes:new_productionsite_type {
          actions = {
             "return=skipped unless economy needs coal",
             "return=failed unless site has log:3",
```
Noordfrees: No need to fail here, since the worker won't actually do anything until the consume statement is reached

Noordfrees (commented):
Changes looking good now, two comments inline. Not tested.

data/tribes/workers/frisians/charcoal_burner/init.lua
hessenfarmer: ok
thanks overlooked that

data/tribes/buildings/productionsites/frisians/charcoal_burners_house/init.lua
hessenfarmer: basically there is no difference in skipping and failing anymore. However if no logs available I'd rate this a fail. anyhow this is only to exclude the search from having a positive on statistics.

Noordfrees: With this implementation, the program will fail immediately if there are no logs. If there are logs but no dry ponds, the program will then fail.

If you remove the "return=failed unless site has log:3", then the program will first check for ponds availability, and if there are no ponds it fails. If there are ponds, "consume" will fail unless there are enough logs.

In either case, the consumption of logs and placing of a stack are carried out only after both checks have succeeded.
So the behaviour would be exactly the same but with one statement less needed

Noordfrees (dismissed):

Noordfrees (approved):

PR #3684, opened 2020-02-09
Fixing dialog for non attackable big buildings

### LINKED ISSUE #1592: Attack on big buildings is not started when the entrance is under the fog of war
tried to send soldiers to Barbarian HQ by pressing the "Start attack" button but nothing happened.  restarting the game didn't help.
widelands-build18-win32

 * date created: 2015-03-04T16:23:40Z
 * owner: arenaaz
 * the launchpad url was https://bugs.launchpad.net/bugs/1428201

Notabilis:
Fixes #1592.
Different checks were used for opening the attack window and starting an attack. This resulted in the player opening the attack window for a big building but being unable to execute the attack when the entrance is not visible.

Also fixing a sporadic resizing of the attack dialog when newly opened. Depending on the timing of clicking and the next tick / think() call, the window was first drawn much bigger than needed which sometimes was visible for a frame.


gunchleoc:
Thanks for the fix!

Travis failure is xcode not installing

PR #3688, opened 2020-02-13
Refactor loader_ui

gunchleoc:
* ProgressWindow and GameTips are now managed by EditorGameGase.
  This avoids ownership conflicts and crashes when we run without window.
* Show game tips on scenarios that don't have a special background image
* Remove ProgressWindow as early as possible
* Don't show progress on autosaves and scenario objective saves
* Test website binaries on Travis.

### DIFF
--- a/src/logic/editor_game_base.h
+++ b/src/logic/editor_game_base.h
@@ -136,6 +136,9 @@ public:
     }
 #endif
 
+    // Destroy the loader UI
+    void remove_loader_ui();
+
 	void set_road(const FCoords&, uint8_t direction, RoadSegment roadtype);
 
 	// warping stuff. instantly creating map_objects
--- a/src/logic/editor_game_base.h
+++ b/src/logic/editor_game_base.h
@@ -119,10 +119,11 @@ public:
 	void load_graphics();
 	virtual void cleanup_for_load();
 
-    // Create a new loader UI
-    UI::ProgressWindow& create_loader_ui(const std::vector<std::string>& tipstexts, const std::string& background = std::string());
-    // Set step text for the current loader UI if it's not nullptr.
-    void step_loader_ui(const std::string& text) const;
+	// Create a new loader UI
+	UI::ProgressWindow& create_loader_ui(const std::vector<std::string>& tipstexts,
+	                                     const std::string& background = std::string());
+	// Set step text for the current loader UI if it's not nullptr.
+	void step_loader_ui(const std::string& text) const;
 
 	void set_road(const FCoords&, uint8_t direction, RoadSegment roadtype);
 
@@ -265,7 +266,7 @@ private:
 	std::unique_ptr<InteractiveBase> ibase_;
 	Map map_;
 
-    std::unique_ptr<GameTips> game_tips_;
+	std::unique_ptr<GameTips> game_tips_;
 
 	/// Even after a map is fully loaded, some static data (images, scripts)
 	/// will still be read from a filesystem whenever a map/game is saved.
--- a/src/logic/editor_game_base.h
+++ b/src/logic/editor_game_base.h
@@ -120,14 +120,15 @@ public:
 	virtual void cleanup_for_load();
 
 	/// Create a new loader UI and register which type of gametips to select from.
-    /// If 'show_game_tips' is true, game tips will be shown immediately.
-    /// Optionally sets a background image.
-	UI::ProgressWindow& create_loader_ui(const std::vector<std::string>& tipstexts, bool show_game_tips,
+	/// If 'show_game_tips' is true, game tips will be shown immediately.
+	/// Optionally sets a background image.
+	UI::ProgressWindow& create_loader_ui(const std::vector<std::string>& tipstexts,
+	                                     bool show_game_tips,
 	                                     const std::string& background = std::string());
 
 	/// If 'background' is not empty, change the background image for the loader UI.
-    /// If 'backgound' is empty, show game tips instead.
-    /// There must be a loader ui and no game tips.
+	/// If 'backgound' is empty, show game tips instead.
+	/// There must be a loader ui and no game tips.
 	void change_loader_ui_background(const std::string& background);
 
 	/// Set step text for the current loader UI if it's not nullptr.
@@ -284,7 +285,7 @@ private:
 	// Shown while loading or saving a game/map
 	std::unique_ptr<UI::ProgressWindow> loader_ui_;
 	std::unique_ptr<GameTips> game_tips_;
-    std::vector<std::string> registered_game_tips_;
+	std::vector<std::string> registered_game_tips_;
 
 	/// Even after a map is fully loaded, some static data (images, scripts)
 	/// will still be read from a filesystem whenever a map/game is saved.
--- a/src/logic/editor_game_base.h
+++ b/src/logic/editor_game_base.h
@@ -122,8 +122,19 @@ public:
 	// Create a new loader UI
 	UI::ProgressWindow& create_loader_ui(const std::vector<std::string>& tipstexts,
 	                                     const std::string& background = std::string());
+
+    // Change the background image for the loader UI and remove the game tips
+    void change_loader_ui_background(const std::string& background);
+
 	// Set step text for the current loader UI if it's not nullptr.
 	void step_loader_ui(const std::string& text) const;
+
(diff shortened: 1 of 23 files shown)


Noordfrees (changes requested):
In general these changes look very good to me, this unifies the patchwork of ProgressWindow creations and deletions.

It seems that the loader ui is created once the first time one is needed and then silently replaced whenever a different one is needed. But it is never deleted except when the game/editor is quit so it sticks around in the background which is still not a really "clean" design. So could you please try to add:
- `assert(loader_ui_); loader_ui_.reset(nullptr);` whenever some task that requires a loader ui is completed to keep the background clean
- `assert(!loader_ui_);` in `EditorGameBase::create_loader_ui`, because such silent overwriting should not happen when the first point is fixed
- `assert(loader_ui_);` in `EditorGameBase::step_loader_ui`, because I guess if a task assumes there is a loader ui there just really should be one

gunchleoc:
Thanks for the review! I will look into your points, except for this one:

* `assert(loader_ui_);` in `EditorGameBase::step_loader_ui`, because I guess if a task assumes there is a loader ui there just really should be one

This would break the website binaries, because they don't have a window, and loading the World will step the loader UI. So, adding this in will mean a lot of `if loader_ui != nullptr`, and I'd rather have asserts where we want to ensure that we have one.

Noordfrees:
Ok good point about the binaries. But not asserting a loader ui is present might still mask bugs in real games. So how about having a global variable `g_windowless` which is set to `true` by the web main functions and otherwise always `false`, and implementing it like this:
~~~~
void EditorGameBase::step_loader_ui(const std::string& text) const {
	if (g_windowless) {
		assert(!loader_ui_);
		return;
	}
	assert(loader_ui_);
	loader_ui_->step(text);
}
~~~~

gunchleoc:
I have added all the asserts back in.

I am also removing the loader UI as early as possible rather than waiting for the game/editor to end as in master.

Also, I stopped it from popping up during scenario saves and autosaves. The `g_windowless`idea is not an option because of this. The asserts like we have them in master should be sufficient.

src/logic/editor_game_base.h
```diff
@@ -117,10 +118,26 @@ class EditorGameBase {
...
+	UI::ProgressWindow& create_loader_ui(const std::vector<std::string>& tipstexts,
+	                                     const std::string& background = std::string());
+
+	// Change the background image for the loader UI and remove the game tips
+	void change_loader_ui_background(const std::string& background);
+
+	// Set step text for the current loader UI if it's not nullptr.
+	void step_loader_ui(const std::string& text) const;
+
+#ifndef NDEBUG
```
Noordfrees, line 137: You can remove the `#ifndef` because having this in release builds as well won't hurt even if its unused

Noordfrees (commented):
Code looking good, only one small comment inline.
Will test this now…

src/logic/editor_game_base.h
gunchleoc: Linters like cppcheck or clang tidy will complain about it if I leave it in for the release builds.

Noordfrees (commented):
And while you're already on it, could you replace `uint8_t` with `uint32_t` in the first line of GameTips::update to fix automatic tip exchanging (a feature that never worked because of this typo…)

Noordfrees:
Assert failure when starting widelands with `--editor=maps/Crater.wmf`:
~~~~
widelands: ../src/logic/editor_game_base.cc:353: void Widelands::EditorGameBase::remove_loader_ui(): Assertion `loader_ui_ != nullptr' failed.
~~~~

Noordfrees:
And when loading a game from Main Menu → Singleplayer → Load Game, the tips banner shows up for half a second and then disappears.

gunchleoc:
> Assert failure when starting widelands with `--editor=maps/Crater.wmf`:
> 
> ```
> widelands: ../src/logic/editor_game_base.cc:353: void Widelands::EditorGameBase::remove_loader_ui(): Assertion `loader_ui_ != nullptr' failed.
> ```

The test suite also flagged this up. Should be fixed now :)

> And when loading a game from Main Menu → Singleplayer → Load Game, the tips banner shows up for half a second and then disappears.

The sequence is:

1. Create loader UI with game tips
2. First step messages
3. Preload map
4. Now we can detect special map background -> replace game tips with map background if we have one

Alternatively, we could create the loader UI without game tips, and then either swap the background or create add game tips. What do you think? I have moved them all up as much as possible for now.

Noordfrees:
Can't get it to crash any more :)

> Alternatively, we could create the loader UI without game tips, and then either swap the background or create add game tips. What do you think?

+1. Loading a campaign still causes the tips to flash up briefly and then be replaced with the desired background; your suggestion is the best way to fix this

Noordfrees:
I'm not sure about removing the progresswindow when auto- or scripting-saving. This should IMHO be consistent with the normal saving, and I rather like this screen because saving can take a while during which the game hangs. So I'm in favour of showing the progresswindow in this case as well like in master.

gunchleoc:
> I'm not sure about removing the progresswindow when auto- or scripting-saving. This should IMHO be consistent with the normal saving, and I rather like this screen because saving can take a while during which the game hangs. So I'm in favour of showing the progresswindow in this case as well like in master.

I'm OK with adding it back in for autosaves, but I'd rather not have it when scenario objectives are being saved. I find it very intrusive in tutorial 1, for example. Showing the progress window during saves is a new feature, so people won't miss it. I was away when this was added, so I'm not sure of the reason. I do like it for manual saves though.

Noordfrees:
> I'm OK with adding it back in for autosaves, but I'd rather not have it when scenario objectives are being saved. I find it very intrusive in tutorial 1, for example. Showing the progress window during saves is a new feature, so people won't miss it.

Okay. But please re-add the "Saving the game" string in the lower-left corner just for scenario saves then so the user knows why the game seems to be hanging. ([It was removed because the progresswindow made it obsolete](https://github.com/widelands/widelands/commit/1197e256ce1e476190de0b1f0777f5a306a18746#diff-62340d8164ff3410d24074d8ff7743c1L131))

> I was away when this was added, so I'm not sure of the reason. 

Actually, it was an unintended side effect of d8a789f…

> I do like it for manual saves though.

…and because you're not the only one, we kept it when it was discovered in master ;)

gunchleoc:
Thanks for digging that up! I have added the "Saving game..." string back in.

gunchleoc:
@Noordfrees Can this go in? Travis failure is the Mac build system not installing.

Noordfrees:
Could you try to fix this first?:

> > Alternatively, we could create the loader UI without game tips, and then either swap the background or create add game tips. What do you think?
> 
> +1. Loading a campaign still causes the tips to flash up briefly and then be replaced with the desired background; your suggestion is the best way to fix this

gunchleoc:
> Could you try to fix this first?:
> 
> > > Alternatively, we could create the loader UI without game tips, and then either swap the background or create add game tips. What do you think?
> > 
> > 
> > +1. Loading a campaign still causes the tips to flash up briefly and then be replaced with the desired background; your suggestion is the best way to fix this

What is your suggested fix? I am loading them as fast as I can now (see https://github.com/widelands/widelands/pull/3688/commits/774d1faa5576236cc93f8b28f8d6d48b2cb76d96), and I don't think that it's any slower than on master.

We could create the loader UI later in single player games and only show it after preloading the map ("Preloading map…"). This won't work for multiplayer though, because GameClient is using it as a modal window.

Or do you mean to create the game tips later, after preloading?

Noordfrees:
> Or do you mean to create the game tips later, after preloading?

Yep, that's what I would consider best.
In scenarios, the gametips are initially created, and then removed when the background is set, so they are visible for a frame. IMHO the best solution would be not to show the tips initially, and then unhide them when we know that the map does not define a special background. This could be implemented as a `bool visible_` in GameTips which is initially false, and skipping GameTips::update unless it's set to true.

gunchleoc:
Thanks for sticking with the review- much better now :)

I chose a slightly different design though - I'm registering the wanted game tips strings in EditorGameBase and then only creating the GameTips object as needed.

Noordfrees (approved):
Perfect :)

gunchleoc:
:)

Travis failure is xcode.

PR #3690, opened 2020-02-15
Fixes to empire scenario 4

gunchleoc:
* Fix nil index if Temple of Vesta is trying to donate but was destroyed by enemy
* Give farm plans if player captures the artefact before objective was triggered.


hessenfarmer (approved):
changes looking good although I didn't know this might be a problem

gunchleoc:
1 happened because I wasn't paying attention, and 2 because I knew the scenario and wanted to conquer the field fast.

Thanks for the review!

PR #3702, opened 2020-02-21
Create spritesheets for trees

gunchleoc:
We don't have Blender files for the trees,so I went ahead and created their spritesheets without mipmaps.

Changes in hotspots are there because the tool will crop the images too.


Noordfrees (approved):
Tested it briefly – working as it should, with a much cleaner directory structure :)

gunchleoc:
Thanks!

PR #3710, opened 2020-02-22
Make emergency saves not cause additional crashes

gunchleoc:
Fix problems reported by ASan when emergency saving a single payer game.


Noordfrees:
Could you give instructions how to easily trigger an emergency save so this can be tested?

src/game_io/game_preload_packet.cc
```diff
@@ -111,7 +111,9 @@ void GamePreloadPacket::write(FileSystem& fs, Game& game, MapObjectSaver* const)
 	s.set_string("background", map.get_background());
 	s.set_string("win_condition", game.get_win_condition_displayname());
 	s.set_int("savetimestamp", static_cast<uint32_t>(time(nullptr)));
-	s.set_int("gametype", static_cast<int32_t>(game.game_controller()->get_game_type()));
+	s.set_int("gametype", static_cast<int32_t>(game.game_controller() != nullptr ?
+	                                              game.game_controller()->get_game_type() :
+	                                              GameController::GameType::kReplay));
```
Noordfrees, line 116: Please add a comment explaining why this should default to replay

gunchleoc: Because it needs to default to something and I couldn't think of anything better. Replay is the most general option.

gunchleoc:
> Could you give instructions how to easily trigger an emergency save so this can be tested?

Add `throw wexception("Crash this");` to the code somewhere where it's executed after game start. BuildingWindow constructor should do.

Noordfrees (approved):
I get a lot of memory leaks from MapObject allocations after the crash and emergency save, but the saves can be loaded and continued without any problems. 
Code LGTM

gunchleoc:
The memory leaks are normal, because we can't clean up after an exception that might come from anywhere. Thanks for the review!

Issue #3440, opened 2019-09-09
Building does not complete


bunnybot:
Playing Elven forests I forgot to build a Lime Kiln. Some time later I ran out of Grout so I built a kiln. It quickly reached 85% built and then stopped. Destroying the building and rebuilding does not work (sticks at 85%) but another Lime Kiln built nearby completes normally.
There are sufficient wares in the headquarters to complete the building and wares are being delivered to other buildings along the road.
Savegame attached

------------------------------------
Imported from Launchpad using lp2gh.

 * date created: 2019-05-21T10:59:55Z
 * owner: leightonman
 * the launchpad url was https://bugs.launchpad.net/bugs/1829848

bunnybot:
(by leightonman)

bunnybot:
(by stonerl)
The stone lies next to the barrier close to the lime kiln. Destroying the road reconnecting it will start the delivery.

bunnybot:
(by franku)
We had different bugs with the same issue in the past:

https://bugs.launchpad.net/widelands/+bug/1797213
https://bugs.launchpad.net/widelands/+bug/536367
https://bugs.launchpad.net/widelands/+bug/1656671

Tagged as regression.

Leighton Man, can you remember playing around with the "capacity settings" (e.g. set amount of stones in warehouses, or the construction site)?

bunnybot:
(by leightonman)

bunnybot:
(by leightonman)

bunnybot:
(by leightonman)

bunnybot:
(by leightonman)

bunnybot:
(by leightonman)
I did not alter capacity settings whilst playing the game but hopefully the two replays I have added will give a clue.
Clearly, I saved the game after nearly an hour and resumed later. The first replay shows stones being delivered from the quarry nearby past the barrier towards the headquarters and, at the end, a single stone being delivered towards the building site for the Lime Kiln. If you continue to watch after the replay ends, delivery continues normally and the construction site later completes.
Replay 2 shows a stone stuck at the flag one step on from the barrier but the construction doesn't complete. Unfortunately, I no longer have the saved game as it was overwritten.

bunnybot:
(by gunchleoc)
Sounds like out saveloading bug for the ware transportation, but I can't find the other bug right now.

bunnybot:
(by leightonman)
[...]

PR #3712, opened 2020-02-22
Fix wares stuck on flag after loading savegame

Niektory:
This fixes a bug that causes wares to be stuck on a flag and never picked up after loading a game. (see #1085 and #3440)

This is a partial fix. It should prevent the bug from happening when loading a good savegame, [like this one](https://github.com/widelands/widelands/files/4240001/ware_stuck_at_flag_1.wgf.zip).

However, if the game was saved again after this bug had already occured (like [this one](https://github.com/widelands/widelands/files/4240002/ware_stuck_at_flag_2.wgf.zip)) then the savegame is broken and needs an additional fix. It would probably need to scan all wares at flags after loading and detect those that are marked as acked but have no carrier coming for them.


gunchleoc (approved):
Eek, I always get my min and max crossed. Thanks for finding this one!

gunchleoc:
Appveyor failure is the usual timeout.

I think the patch is good enough with not fixing broken savegames.

PR #3715, opened 2020-02-23
Larger description index

gunchleoc:
- Change `DescriptionIndex` data type to `uint16_t`, because we're running out of IDs
  for Building Descriptions
- Fix usages of `DescriptionIndex`
- Rename function `get_resource` to `resource_index` for consistency.


Noordfrees (approved):
Code LGTM, 
not tested but if travis succeeds it should be fine

PR #3722, opened 2020-02-26
Fix translations formatting

gunchleoc:
- Add command line switches to `fix_formatting.py`
  for formatting selectively
- Add formatting directives to translation update script
  for generated Lua files

### DIFF
--- a/utils/fix_formatting.py
+++ b/utils/fix_formatting.py
@@ -30,49 +30,75 @@ def parse_args():
         description='Fix common whitespace errors in Lua files, run clang-format'
         ' over the code base and pyformat over the utils directory.'
         ' Recurses over all relevant files.')
-    return p.parse_args()
+
+    p.add_argument('-c', '--c++', action='store_true',
+                   help='Format C++ files only')
+    p.add_argument('-l', '--lua', action='store_true',
+                   help='Format Lua files only')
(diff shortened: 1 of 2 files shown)


utils/fix_formatting.py
```diff
@@ -30,49 +30,75 @@ def parse_args():
...
+            sys.stdout.write('.')
+            sys.stdout.flush()
+            call(['clang-format', '-i', filename])
+            call(['git', 'add', '--renormalize', filename])
+        print(' done.')
+
+    if format_lua:
+        directory = args['dir']
+        if not directory:
+            directory = './data'
```
Noordfrees: We also have Lua files in `./test` that should be formatted

Noordfrees (commented):
Changes looking good, one nit inline

Noordfrees (approved):
Looking good :)

utils/fix_formatting.py
gunchleoc: Thanks for catching this one. I am using a set now. Same for Python, because we should format the codecheck rules too in the future. I'd like to review the formatting results though to ensure that they are good, so I'd rather add that directory in a follow-up branch.

gunchleoc:
Merging ow so I can test the translations script on master.

PR #3723, opened 2020-02-26
Let fix_formatting.py format the codecheck directory

gunchleoc:
The only actual code change is in `utils/fix_formatting.py`. The rest is just reformatting + a few fixed typos.

The file deletions/additions are because `git mv` doesn't always work properly. I had to give the rules a file ending so that the formatting script will pick them up.


Noordfrees (approved):
LGTM :)

PR #3724, opened 2020-02-27
Delete unneeded boost headers and replace outdated usages

gunchleoc:
- Replace `boost::function` with `std::function`
- Replace `boost::shared_ptr` with `std::shared_ptr`
- Replace `boost::unordered_map` with `std::unordered_map`
- Add codecheck rule so that these won't regress
- Require including `<memory>` for `shared_ptr` via codecheck
- Delete all boost headers that the compiler won't ask for

The replacements should be safe regarding compiler support, because we're already using them in our codebase.


Notabilis (approved):
Surprising how many unused includes we have... Thanks for cleaning this up!
The code changes are looking good and since it compiles this should be ready for merging.

PR #3727, opened 2020-03-01
Resolve some circular dependencies + shift random library

gunchleoc:


Notabilis (dismissed):
The conflict with trunk has to be resolved, but the changes in this branch can be merged as far as I am concerned.

PR #3736, opened 2020-03-03
Soldier healing

gunchleoc:
- Military sites start healing their soldiers when they're on their way home

Forum discussion: https://www.widelands.org/forum/topic/4590/

### DIFF
--- a/src/logic/map_objects/tribes/militarysite.cc
+++ b/src/logic/map_objects/tribes/militarysite.cc
@@ -751,8 +751,7 @@ void MilitarySite::act(Game& game, uint32_t const data) {
 				            soldier->get_battle()->opponent(*soldier) == nullptr) &&
 				           !get_economy(WareWorker::wwWORKER)->warehouses().empty()) {
 					// Somewhat heal soldiers in the field that are not currently engaged in fighting an
-					// opponent,
-					// but only if there is a warehouse connected.
+					// opponent, but only if there is a warehouse connected.
 					const PlayerNumber field_owner = soldier->get_position().field->get_owned_by();
 					if (owner().player_number() == field_owner) {
 						const unsigned int air_distance =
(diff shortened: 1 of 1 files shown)


stonerl:
I really like the idea. But as far as I understand the code, the healing speed depends on the building a soldier is from? So, a solder from a babarian citadel heals  at 110 health points per second while a soldier from a sentry only heals 40 per second?

Here are the buildings with the lowest healing power per tribe:

* barbarian sentry 80
* empire blockhouse 60
* atlanteans guardhouse 75
* frisians wooden tower 40

The healing rate on the field should be independent of the building but maybe tribe specific? Like:

* barbarians 40
* empire 30
* atlantean 35
* frisians 20

Or we use the same value for every tribe. e.g. 35 which is lower than the lowest healing rate of any tribe.

I, personally prefer the healing rate per tribe. But maybe we should ask in the forum what the suers think.

src/logic/map_objects/tribes/militarysite.cc
```diff
@@ -727,31 +727,47 @@ void MilitarySite::act(Game& game, uint32_t const data) {
...
+						max_total_level = soldier->get_total_level();
+						max_health = soldier->get_current_health() / soldier->get_max_health();
+						soldier_to_heal = soldier;
+					}
+				} else if ((soldier->get_battle() == nullptr ||
+				            soldier->get_battle()->opponent(*soldier) == nullptr) &&
+				           !get_economy(WareWorker::wwWORKER)->warehouses().empty()) {
+					// Somewhat heal soldiers in the field that are not currently engaged in fighting an
+					// opponent,
+					// but only if there is a warehouse connected.
```
stonerl: Could you make these two line one, please? My OCD is killing me :-)

gunchleoc: Done.

stonerl (approved):
Tested it and have no complaints. :-)

hessenfarmer:
we should probably gain more feedback before merging

PR #3742, opened 2020-03-06
add HQ Hunter win condition

asmaps:
As suggested in the [forum](https://www.widelands.org/forum/topic/4787/) I added a win condition where a player is defeated when he looses his headquarter.

Detecting the missing HQ and then destroying all warehouses was the only way I found to achieve this. If there is a simpler possibility I'm open for suggestions of better implementations.

Also if there is a way to get a building type across tribes ("get all warehouses of player xy, no matter which tribe the player is") I'd like to change it to that one.

### DIFF
--- a/data/scripting/win_conditions/defeat_hq.lua
+++ b/data/scripting/win_conditions/defeat_hq.lua
@@ -1,0 +1,60 @@
+-- =======================================================================
+--                         Defeat all Win condition
+-- =======================================================================
+
+include "scripting/coroutine.lua" -- for sleep
+include "scripting/win_conditions/win_condition_functions.lua"
+
+set_textdomain("win_conditions")
+
+include "scripting/win_conditions/win_condition_texts.lua"
+
+local wc_name = "HQ Hunter"
+-- This needs to be exactly like wc_name, but localized, because wc_name
+-- will be used as the key to fetch the translation in C++
+local wc_descname = _("HQ Hunter")
+local wc_version = 1
+local wc_desc = _ "The tribe or team that can destroy all other headquarters wins the game!"
+return {
+    name = wc_name,
+    description = wc_desc,
+    peaceful_mode_allowed = false,
+    func = function()
+        local plrs = wl.Game().players
+
+        -- set the objective with the game type for all players
+        broadcast_objective("win_condition", wc_descname, wc_desc)
+
+        -- Iterate all players, if one is defeated, remove him
+        -- from the list, send him a defeated message and give him full vision
+        repeat
+            sleep(5000)
+
+            for idx,p in ipairs(plrs) do
+                if (#p:get_buildings("barbarians_headquarters") + #p:get_buildings("atlanteans_headquarters") + #p:get_buildings("empire_headquarters") + #p:get_buildings("frisians_headquarters")) == 0 then
+                    for idx,b in ipairs(p:get_buildings("barbarians_warehouse")) do
+                        b:destroy()
+                    end
+                    for idx,b in ipairs(p:get_buildings("atlanteans_warehouse")) do
+                        b:destroy()
+                    end
+                    for idx,b in ipairs(p:get_buildings("empire_warehouse")) do
+                        b:destroy()
+                    end
+                    for idx,b in ipairs(p:get_buildings("frisians_warehouse")) do
+                        b:destroy()
+                    end
+                end
+            end
+            check_player_defeated(plrs, lost_game.title, lost_game.body, wc_descname, wc_version)
+        until count_factions(plrs) <= 1
+
+        -- Send congratulations to all remaining players
+        broadcast_win(plrs,
+        won_game.title,
+        won_game.body,{},
+        wc_descname, wc_version
+        )
+
+    end,
+}
--- a/data/scripting/win_conditions/defeat_hq.lua
+++ b/data/scripting/win_conditions/defeat_hq.lua
@@ -3,6 +3,7 @@
 -- =======================================================================
 
 include "scripting/coroutine.lua" -- for sleep
+include "scripting/table.lua"
 include "scripting/win_conditions/win_condition_functions.lua"
 
 set_textdomain("win_conditions")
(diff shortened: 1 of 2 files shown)


Noordfrees:
Looking good, two coments:
- You also need to destroy all ports since those are also warehouses internally
- What about players who start with a starting condition that gives them a warehouse instead of an HQ initially?

data/scripting/win_conditions/defeat_hq.lua
```diff
@@ -0,0 +1,60 @@
...
+                    end
+                    for idx,b in ipairs(p:get_buildings("atlanteans_warehouse")) do
+                        b:destroy()
+                    end
+                    for idx,b in ipairs(p:get_buildings("empire_warehouse")) do
+                        b:destroy()
+                    end
+                    for idx,b in ipairs(p:get_buildings("frisians_warehouse")) do
+                        b:destroy()
+                    end
```
Noordfrees, line 52: Suggestion: To reduce the amount of typing, you can add `include "scripting/table.lua"`and then use
~~~~
for idx,b in ipairs(array_combine(
   p:get_buildings("barbarians_warehouse"),
   p:get_buildings("empire_warehouse"),
   …etc… )) do
~~~~

gunchleoc:
How about using whichever building is sitting on the starting position as the building that can't be lost? You can access the field by using

`local sf = wl.Game().map.player_slots[plr.number].starting_field`

https://www.widelands.org/documentation/autogen_wl_map/#map

And the building via

`sf.immovable`

https://www.widelands.org/documentation/autogen_wl_map/#field

Noordfrees:
> How about using whichever building is sitting on the starting position as the building that can't be lost? 

This would break with Discovery and New World starting conditions

gunchleoc:
> > How about using whichever building is sitting on the starting position as the building that can't be lost?
> 
> This would break with Discovery and New World starting conditions

And can then be fixed in that branch by picking the first completed port instead. We have a chicken and egg problem with the 2 branches right now.

Noordfrees:
Then this should go directly in this branch. Have a local variable `main_building` for each player, and use a `while not main_building do` loop until an HQ, warehouse, or port is found to assign to it.

hessenfarmer:
if we do it with the loop we should first sleep some time and then do a sorted check. i.e. first check for an HQ at starting field, then for a warehouse, and finally for a port

asmaps:
> How about using whichever building is sitting on the starting position as the building that can't be lost?

How would you know which one is the warehouse/building you have to destroy? Is there a possibility to mark it somehow?

Also is there a possibility to disable a win condition for certain modes as it obviously doesn't really make sense in all conditions.

data/scripting/win_conditions/defeat_hq.lua
asmaps: Thanks. Implemented in 0cbda9d

Noordfrees:
> Also is there a possibility to disable a win condition for certain modes as it obviously doesn't really make sense in all conditions.

That would be best perhaps. We could either disable certain starting conditions (SC) when this win condition (WC) is selected, or disable this WC when certain SCs are used. I believe the latter would be much less difficult to code – perhaps you'd like to take a look at `src/ui_fsmenu/launch_game.cc`, `FullscreenMenuLaunchGame::load_win_conditions` and see whether you could add a check for certain forbidden starting conditions?

gunchleoc:
> Then this should go directly in this branch. Have a local variable `main_building` for each player, and use a `while not main_building do` loop until an HQ, warehouse, or port is found to assign to it.

We could make this generic for all tribes via

`wl.Game():get_building_description(buildingname).type_name == "warehouse"`

asmaps:
> > Also is there a possibility to disable a win condition for certain modes as it obviously doesn't really make sense in all conditions.
> 
> That would be best perhaps. We could either disable certain starting conditions (SC) when this win condition (WC) is selected, or disable this WC when certain SCs are used. I believe the latter would be much less difficult to code – perhaps you'd like to take a look at `src/ui_fsmenu/launch_game.cc`, `FullscreenMenuLaunchGame::load_win_conditions` and see whether you could add a check for certain forbidden starting conditions?

I tried around with this and I think I found the relevant code places, but I don't really know how to implement that :-(

I played some rounds with this WC now and I really like it. I find it most useful for maps with islands, where games otherwise often end as tie because everyone has a whole island and those can't be conquered. Also I noticed that it is still possible to continue playing after being defeated... and even to reestablish everything as warehouses currently being built while getting defeated don't get destroyed obviously. I think one should not be able to build anything after being defeated.

hessenfarmer:
> I played some rounds with this WC now and I really like it. I find it most useful for maps with islands, where games otherwise often end as tie because everyone has a whole island and those can't be conquered. Also I noticed that it is still possible to continue playing after being defeated... and even to reestablish everything as warehouses currently being built while getting defeated don't get destroyed obviously. I think one should not be able to build anything after being defeated.

the thing with islands is only true if the HQ is reachable from the land side. There are a few maps like this, but if your HQ is on an island it does not help.    
The latter is not arising in single player as you get the "you lost" message, there it is only visible in multiplayer. You can reestablish am economy in this case but you can't attack anymore as in the see_all mode attack is disabled. Anyhow having a warehouse in construction with all wares already present in the constructionsite is a rare corner case anyhow. But in this WC it might be easy to fix, as you just need to check for the constructionsites of ports and warehouses as well an destroy them.

gunchleoc:
> > How about using whichever building is sitting on the starting position as the building that can't be lost?
> 
> How would you know which one is the warehouse/building you have to destroy? Is there a possibility to mark it somehow?

Having to find which one it is might actually be fun - a player could build multiple ports to confuse the enemy. Building ports is expensive and will take gold away from solder training.
 
> Also is there a possibility to disable a win condition for certain modes as it obviously doesn't really make sense in all conditions.

We could take care of that in a follow-up branch so you can concentrate on implementing the win condition itself. For which modes would you want to disable it?

Noordfrees:
Superseded by #4648

hessenfarmer:
@asmaps 
As it was your idea and first sketch that now eventually finds its way to trunk, you should be credited for it if you wish. 
In this case please tell us how we should add you to the credits (nickname or real name or both).  
Furthermore I would be happy if you could test the PR superseeding this one

asmaps:
> @asmaps
> As it was your idea and first sketch that now eventually finds its way to trunk, you should be credited for it if you wish.
> In this case please tell us how we should add you to the credits (nickname or real name or both).

That's very nice. Quote with nick and real name, "NativeException" is my more used nick since a while, so please use this :-) Real name "Arne Schauf"
Thanks a lot for implementing this! I had a look into the suggested changes but didn't get very far :-(

> Furthermore I would be happy if you could test the PR superseeding this one

Will do, I don't have much time atm, so I don't know when, but I will give feedback as soon as I can.

PR #3748, opened 2020-03-08
Refactor saveloading ui

jmoerschbach:
Hi,

in #3561 I promised to revisit this code again. I hav done so ;)

(Hopefully) no semantic changes at all, only making the code's intention clearer...

What I did:

- Further separating of concerns: Extracted savegame loading code (SavegameLoader) and savegame deleting code (SavegameDeleter) into own classes
- Introduced small inheritance hierachy for both of these classes to adhere to different loading/deleting of replays/multiplayer/singleplayer

I think there is still room for improvement, for example I do not like the exposure of "internal" objects (via LoadOrSaveGame#table(), #table_box(), #game_details(), #delete_button())

Can someone explain or point me to documentation about how replays are created/saved? What is GameController::GameType::kReplay actually used for? In the old code (and currently in this pr, too) it is not evaluated, is_valid_game always returns true, but I think this might be wrong in the first place?

Feedback very much appreciated :)

Thanks

Jonas


jmoerschbach:
Holy shit, what happened here with the code formatting? Oo
Most of the files I didn't even opened in my editor... How come the format has changed?
Is fixing this somehow automated? Do I need to do something about it?

Noordfrees:
Perhaps you have the deprecated clang-format-githook installed on your machine?

Please go to https://github.com/jmoerschbach/widelands/runs/493304990 and click "Re-run jobs", that should fix the formatting.

jmoerschbach:
Thanks @Noordfrees! Looks better now :)

Will check my git hook tomorrow...

jmoerschbach:
everything is green now.

Review comments gladly appreciated :)

gunchleoc:
> Can someone explain or point me to documentation about how replays are created/saved?

You'll have to dig through the code to find out, because I don't know either.

> What is GameController::GameType::kReplay actually used for? In the old code (and currently in this pr, too) it is not evaluated, is_valid_game always returns true, but I think this might be wrong in the first place?

It is used in the Lua scripting interface to switch off some functions that would crash replays. E.g.

```
int LuaPlayer::message_box(lua_State* L) {
	Game& game = get_game(L);
	// don't show message boxes in replays, cause they crash the game
	if (game.game_controller()->get_game_type() == GameController::GameType::kReplay) {
		return 1;
	}
```

Also used by the save handler - very important!

```
void SaveHandler::think(Widelands::Game& game) {
	if (!allow_saving_ || game.is_replay()) {
		return;
	}
```

src/wui/load_or_save_game.cc
```diff
@@ -135,39 +122,41 @@ bool LoadOrSaveGame::compare_map_name(uint32_t rowa, uint32_t rowb) const {
 std::unique_ptr<SavegameData> LoadOrSaveGame::entry_selected() {
 	std::unique_ptr<SavegameData> result(new SavegameData());
 	size_t selections = table_->selections().size();
-	if (selections == 1) {
+	set_tooltips_of_buttons(selections);
+
+	const std::vector<SavegameData> savegames = get_selected_savegames();
+	game_details_.display(savegames);
+	if (selections > 0 && !selection_contains_directory()) {
```
gunchleoc:
```suggestion
	if (!table_->selections().empty() && !selection_contains_directory()) {
```

src/wui/load_or_save_game.cc
```diff
@@ -135,39 +122,41 @@ bool LoadOrSaveGame::compare_map_name(uint32_t rowa, uint32_t rowb) const {
...
+		delete_->set_enabled(true);
+	} else {
+		delete_->set_enabled(false);
+		delete_->set_tooltip("");
+	}
+	// TODO(GunChleoc): Take care of the OK button too.
+	if (selections > 0) {
```
gunchleoc: You're already checking for emptiness above, so you can restructure this to save a check.

src/wui/load_or_save_game.cc
```diff
@@ -135,39 +122,41 @@ bool LoadOrSaveGame::compare_map_name(uint32_t rowa, uint32_t rowb) const {
...
+		result.reset(new SavegameData(savegames[0]));
+	}
+	return result;
+}
+void LoadOrSaveGame::set_tooltips_of_buttons(size_t nr_of_selected_items) const {
+	if (nr_of_selected_items == 1) {
```
gunchleoc, line 139: You could make this more efficient with a switch statement:

case 0
case 1
default

src/wui/load_or_save_game.cc
```diff
@@ -202,126 +191,28 @@ GameDetails* LoadOrSaveGame::game_details() {
 	return &game_details_;
 }
 
-const std::string LoadOrSaveGame::get_filename(uint32_t index) const {
-	return get_savegame(index).filename;
+void LoadOrSaveGame::select_item_and_scroll_to_it(std::set<uint32_t>& selections) {
+	const uint32_t selectme = *selections.begin();
```
gunchleoc: If you add

```
if (table_->empty()) {
    return;
}
```

on top, you can then have

`table_->select(std::min(selectme, table_->size() - 1));`

src/wui/savegamedeleter.cc
```diff
@@ -0,0 +1,165 @@
...
+	msgBox.run<UI::Panel::Returncodes>();
+}
+
+std::string SavegameDeleter::create_header_for_deletion_failed_window(size_t no_to_be_deleted,
+                                                                      size_t no_failed) const {
+	if (no_to_be_deleted == 1) {
+		return _("The game could not be deleted.");
+	} else {
+		return (boost::format(ngettext(
+		           "%d game could not be deleted.", "%d games could not be deleted.", no_failed)) %
```
gunchleoc, line 96: Add comment:

```
	                        /** TRANSLATORS: Used with multiple games, 1 game has a separate
	                           string. DO NOT omit the placeholder in your translation. */
```

src/wui/savegamedeleter.cc
```diff
@@ -0,0 +1,165 @@
...
+
+	return header;
+}
+
+std::string ReplayDeleter::create_header_for_deletion_failed_window(size_t no_to_be_deleted,
+                                                                    size_t no_failed) const {
+	if (no_to_be_deleted == 1) {
+		return _("The replay could not be deleted.");
+	} else {
+		return (boost::format(ngettext("%d replay could not be deleted.",
```
gunchleoc, line 127: Add translators comment

src/wui/savegameloader.cc
```diff
@@ -0,0 +1,221 @@
...
+	}
+	loaded_games.push_back(SavegameData::create_sub_dir(gamefilename));
+}
+
+ReplayLoader::ReplayLoader(Widelands::Game& game) : SavegameLoader(game) {
+}
+
+bool ReplayLoader::is_valid_gametype(const SavegameData&) const {
+	return true;  // why?? what is the purpose of GameController::GameType::kReplay?
+	              // return gamedata.is_replay(); <-- should be this, right?!
```
gunchleoc: Right :)

gunchleoc (changes requested):
I have added some comments with suggestions. Not tested yet.

src/wui/load_or_save_game.cc
jmoerschbach: ??
selections is of type size_t so it does not have .empty()

src/wui/load_or_save_game.cc
jmoerschbach: Sry, I won't (only if you insist of course)

1. I think it is important for the reader to see the "greater than" to grasp the code's intention quickly. The "default" of your proposed switch would not be the actual default (I suppose it is very rarely that multiple items are selected in this context)
2. it is more code lines

src/wui/savegamedeleter.cc
jmoerschbach: done

src/wui/savegamedeleter.cc
jmoerschbach: done

src/wui/savegameloader.cc
jmoerschbach: yeah... no. When I do this, at least in my tests no games/replays whatsoever are shown anymore. The original code always returns true so "normal" savegames are presented after clicking "Watch replay"-Button. This is why I asked how the system saves a game so it "becomes" a replay...

src/wui/load_or_save_game.cc
jmoerschbach: I don't think so - or I just don't get it.
Above is a "logical and" check. So to avoid checking "selections > 0" twice I would need to make the first if combined statement a nested one, right? So no check saved but nested - which I personally do not like at all...

src/wui/load_or_save_game.cc
jmoerschbach: yeah, nice, thanks :)

src/wui/load_or_save_game.cc
gunchleoc: Oops, I have updated the code suggestion. You can probably get rid of the selections variable then.

src/wui/load_or_save_game.cc
gunchleoc: I prefer the nested. If we have selections, do something, and directories then get special treatment.

Alternatively, you can save some nesting by using an early return:

```
if (table_->selections().empty() {
    return result;
}
// Do something
return result;
```

src/wui/load_or_save_game.cc
gunchleoc: Fine by me, this was just a suggestion :)

src/wui/savegameloader.cc
gunchleoc: I think I remember now that replays were tagged as one of the multiplayer game types. That might be a bug - we never dug into that code. I'm not sure though.

src/wui/load_or_save_game.cc
jmoerschbach: yeah, did that

src/wui/load_or_save_game.cc
jmoerschbach: yeah, I updated a slightly different solution. Looks better now, right? :)

gunchleoc: Yep. I like it :)

src/wui/savegameloader.cc
gunchleoc: Can you add `TODO(jmoerschbach ):` to that comment? I'd say fixing this problem is out of scope for this branch.

ETA: I think I have figures it out - see comment below.

gunchleoc (changes requested):
I have done some testing now, and I'd like to see 2 changes:

1. When saving a game from replay, it has `gametype="4"` and should be accessible from both singleplayer and multiplayer load screen. You can create one by loading a replay and then saving it. I guess that's what `GameController::GameType::kReplay` really is for.
2. When a directory is selected, show the directory name in the game details, just like we do with map details when starting a new game from map

For a follow-up branch, it would be great it the player could create new directories - you can find some code on the editor save map screen for that. The icing on the cake would be to be able to delete empty directories too.

src/wui/savegameloader.cc
jmoerschbach: I added a TODO

jmoerschbach:
Hi @gunchleoc 

sorry for the delayed response...

I implemented 2. Please have look I hope you like it :)

I'm not sure about 1. though. For me it is still quite confusing. For the time being, I would like to have it staying as it is - I am quite sure it is the exact same behaviour as current master. This branch was intended to introduce no different behaviour but only refactoring the code (which is not quite true anymore, since I implemented your wish number 2 ;))

However, I think I found another bug: Loading an unzipped replay crashes the game. Preloading, however, works fine (so it has nothing to do with the code I changed). I belive in current master, unzipped replays are not shown in the table at all, same in b20. Do you have any insights to this?

frankystone:
> However, I think I found another bug: Loading an unzipped replay crashes the game. Preloading, however, works fine (so it has nothing to do with the code I changed). I belive in current master, unzipped replays are not shown in the table at all, same in b20. Do you have any insights to this?

Can't reproduce. Unzipped the file  .wrpl.wgf into a new folder with the same name and watching the replay works fine here. Did you unzipped it into a folder with the same name?

gunchleoc:
> > However, I think I found another bug: Loading an unzipped replay crashes the game. Preloading, however, works fine (so it has nothing to do with the code I changed). I belive in current master, unzipped replays are not shown in the table at all, same in b20. Do you have any insights to this?
> 
> Can't reproduce. Unzipped the file .wrpl.wgf into a new folder with the same name and watching the replay works fine here. Did you unzipped it into a folder with the same name?

It would be good to catch crashes like this in the future. However, we always natively zip the replays even if the option not to compress files has been chosen, so we have never really supported this. No need to fix it in this branch.

src/wui/savegameloader.cc
```diff
@@ -0,0 +1,223 @@
...
+		return;
+	}
+	loaded_games.push_back(SavegameData::create_sub_dir(gamefilename));
+}
+
+ReplayLoader::ReplayLoader(Widelands::Game& game) : SavegameLoader(game) {
+}
+
+bool ReplayLoader::is_valid_gametype(const SavegameData&) const {
+	return true;  // TODO (jmoerschbach): why?? what is the purpose of
```
gunchleoc:
```suggestion
	return true;  // TODO(jmoerschbach): why?? what is the purpose of
```

jmoerschbach:
> 
> 
> Never hard-code word order. Use "Selected %1% directory and %2%" and add a translators comment

I am not quite sure how to do that. Can you show me the complete example?
Actually, I think I want to enforce order here, because in the displayed list, directories are always shown first and then files, for example:
Selected 1 directory and 2 files:
- save/subfolder
- savegameOne
- savegameTwo

Selected 0 directories and 1 file:
- savegameOne

Selected 3 directories and 0 files:
- save/subfolderOne
- save/subfolderTwo
- save/subfolderThree

src/wui/gamedetails.cc
```diff
@@ -172,82 +79,126 @@ void GameDetails::clear() {
...
+			number_of_directories++;
+		} else {
+			number_of_files++;
 		}
-	} else {
+	}
+	std::string name_list = richtext_escape(as_filename_list(gamedata));
+	boost::replace_all(name_list, "\n", "<br> • ");
+
+	std::string combined_header = as_richtext(as_heading_with_content(
```
gunchleoc, line 110: It goes something like this:

```
const std::string header_part_2( (boost::format(ngettext("%d file:", "%d files:", number_of_files)) % number_of_files)
	         .str());	
std::string combined_header = as_richtext(as_heading_with_content(
	   (boost::format(ngettext(
	       "Selected %1% directory and %2%", "Selected  %1% directories and %2%", number_of_directories)) %
	    number_of_directories % header_part_2)
	         .str(),
	   "", style_, true));
```
Add TRANSLATORS comments too so that translators will know how the string will be assembled.

gunchleoc:
> > Never hard-code word order. Use "Selected %1% directory and %2%" and add a translators comment
> 
> I am not quite sure how to do that. Can you show me the complete example?
> Actually, I think I want to enforce order here, because in the displayed list, directories are always shown first and then files, for example:

I am a translator who translates into a language unlike English, so a big -1 on that. The only thing that you will be enforcing here is embarrassing grammar errors and convoluted linguistic contortions to try to make the translations work. It's one of the things that developers get wrong and that wants to make me bang my head against the wall.

jmoerschbach:
hope it is ok now :)

src/wui/gamedetails.cc
```diff
@@ -103,13 +103,15 @@ void GameDetails::show(const std::vector<SavegameData>& gamedata) {
 	std::string name_list = richtext_escape(as_filename_list(gamedata));
 	boost::replace_all(name_list, "\n", "<br> • ");
 
+	const std::string header_second_part(
+	   (boost::format(ngettext("%d file:", "%d files:", number_of_files)) % number_of_files).str());
```
gunchleoc, line 108:
```suggestion
	   /** TRANSLATORS: This is the second part of "Selected %1% directory/directories and %2%" */
	   (boost::format(ngettext("%d file:", "%d files:", number_of_files)) % number_of_files).str());
```

gunchleoc (approved):

PR #3749, opened 2020-03-08
Consistent farmers

### LINKED ISSUE #1672: farmers at home (out of fields)
Wonder if this is another regression. In the save-game you can see some completely or partly inactive farmers although their area overview shows several free spots which could be farmed and they appear to be walkable too. Obviously something is blocking recognition of these spots. (You can find these farmers in the center land where all parties meet for a battle. There are 3 to 4 such defected farmers, one of them completely inactive, one works only one field, etc.) When there is no activity at all then "out of fields" is reported on the building.

Suspicion is that Immovable objects could play a role in this, but the observation is inconsistent. Terrain shouldn't be the problem, but road congestion might be.

 * date created: 2017-03-03T10:41:22Z
 * owner: 7010622-q-deactivatedaccount
 * the launchpad url was https://bugs.launchpad.net/bugs/1669741

hessenfarmer:
made farmers of all tribes consistent. space flag in findspace programm uses now maxcaps for neighboring fields. 
fixes #1672

### DIFF
--- a/src/logic/map_objects/tribes/worker.cc
+++ b/src/logic/map_objects/tribes/worker.cc
@@ -523,7 +523,7 @@ int16_t Worker::findspace_helper_for_forester(const Coords& pos, const Map& map,
 // fields, trees, rocks and such on triangles and keep the nodes
 // passable. See code structure issue #1096824.
 struct FindNodeSpace {
-	explicit FindNodeSpace(BaseImmovable* const ignoreimm) : ignoreimmovable(ignoreimm) {
+	explicit FindNodeSpace () {
 	}
 
 	bool accept(const EditorGameBase& egbase, const FCoords& coords) const {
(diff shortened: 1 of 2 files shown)


src/logic/map_objects/tribes/worker.cc
```diff
@@ -533,7 +533,7 @@ struct FindNodeSpace {
 		for (uint8_t dir = FIRST_DIRECTION; dir <= LAST_DIRECTION; ++dir) {
 			FCoords const neighb = egbase.map().get_neighbour(coords, dir);
 
-			if (!(neighb.field->nodecaps() & MOVECAPS_WALK) &&
+			if (!(neighb.field->maxcaps() & MOVECAPS_WALK) &&
```
Noordfrees: You need to get rid of `ignoreimmovable` now, the change makes it obsolete and it might only cause corner-case bugs.

Noordfrees (approved):
Code LGTM, untested

gunchleoc:
Tested and working - current branch on top, master on bottom.

farms

stonerl:
```
../src/logic/map_objects/tribes/worker.cc:544:17: warning: private field 'ignoreimmovable' is not used [-Wunused-private-field]
        BaseImmovable* ignoreimmovable;
```

hessenfarmer:
upps overlooked should be fixed in #3768

PR #3750, opened 2020-03-09
Detect kVTID_String in static const VariableType kVTID_String(...

### LINKED ISSUE #3747: find_unused_includes.py broken
Travis reports a number of superfluous includes in #3662 . Among others it complains that `editor/scripting/operators.h` includes `editor/scripting/constexpr.h`, which is the only include there. However this include is required for two reasons:
- The classes defined in `operators.h` inherit from `Assignable`, which is included from `scripting.h` via `constexpr.h`
- `constexpr.h` defines some `static const VariableType` constants which are used in `operators.h`

gunchleoc:
Fixes https://github.com/widelands/widelands/issues/3747


Noordfrees (approved):
LGTM, tested and working :)

PR #3762, opened 2020-03-11
New wai files after about 90 rounds of training

tibor95:
Just few rounds of testing after latest changes to AI code (removing 2 "unused" tags)


hessenfarmer (approved):
new wai files should always be ok. Nevertheless I tested them quickly and AI is another step improved I believe.

PR #3764, opened 2020-03-12
Add missing richtext_escape to text_width function and editbox

### LINKED ISSUE #3760: '<' and '>' in player name crash the editor
**Describe the bug**
'<' and '>' characters in player names result in a crash in the editor player menu.

**To Reproduce**
1. Open the editor
2. Press P
3. Enter one of the characters in the field for a player's name

**Expected behavior**
Either we should allow all characters since richtext doesn't make sense there; or chars reserved for richtext should be silently ignored, as happens e.g. when entering 'ä'

**Version:**
 - Ubuntu MATE 19.10
 - a9d183e

**Additional context**
Error message
~~~~
Caught exception (of type 'N2RT15SyntaxErrorImplE') in outermost handler!
The exception said: Syntax error at 1:64: expected an allowed tag, got '</font'. String continues with: '<</font></p></rt>'

This should not happen. Please file a bug report on version r24470[a9d183e@master](Release).
~~~~

gunchleoc:
Fixes https://github.com/widelands/widelands/issues/3760

### DIFF
--- a/src/graphic/text_layout.cc
+++ b/src/graphic/text_layout.cc
@@ -62,7 +62,7 @@ std::string as_richtext_paragraph(const std::string& text, UI::Align align) {
 int text_width(const std::string& text, const UI::FontStyleInfo& style, float scale) {
 	UI::FontStyleInfo info(style);
 	info.set_size(info.size() * scale - UI::g_fh->fontset()->size_offset());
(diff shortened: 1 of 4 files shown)


Noordfrees (approved):
Code LGTM, tested and working :)

PR #3771, opened 2020-03-13
use size_t instead of long unsigned int

### LINKED ISSUE #3738: Compiler warnings in read MapViewPacket
Currently there are a lot of compiler warnings visible on appvehor. (line 2450 ff). Regarding some variables incompatibilities

jmoerschbach:
fixes #3738

I am actually not sure if changing the static_cast to size_t is the best approach. Might be that reverting the use of PRIuS back to %lu the better way?

### DIFF
--- a/src/map_io/map_players_view_packet.cc
+++ b/src/map_io/map_players_view_packet.cc
@@ -127,8 +127,8 @@ namespace {
 	}
 
 // Try to find the file with newest fitting version number
-#define OPEN_INPUT_FILE_NEW_VERSION(                                                               \
-   filetype, file, filename, fileversion, filename_template, version)                              \
+#define OPEN_INPUT_FILE_NEW_VERSION(filetype, file, filename, fileversion, filename_template,      \
+                                    version)                                                       \
 	uint8_t fileversion = version;                                                                  \
 	filetype file;                                                                                  \
 	char(filename)[FILENAME_SIZE];                                                                  \
--- a/src/map_io/map_players_view_packet.cc
+++ b/src/map_io/map_players_view_packet.cc
@@ -161,10 +161,10 @@ namespace {
 
 #define CHECK_TRAILING_BYTES(file, filename)                                                       \
 	if (!(file).end_of_file())                                                                      \
-		throw GameDataError(                                                                         \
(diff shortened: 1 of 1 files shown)


src/map_io/map_players_view_packet.cc
```diff
@@ -777,7 +777,7 @@ void MapPlayersViewPacket::read(FileSystem& fs,
 							   "\"%s\":%" PRIuS ": node (%i, %i) t = R: unexpected end of "
 							   "file while reading time_triangle_last_surveyed",
 							   plnum, survey_times_filename,
```
gunchleoc: Here's an additional fix you can do: the `plnum` variables need casting to `<unsigned int>`. They are `uint8_t`, which is interpreted as a `char` by C++, so we end up with garbage output.

`get_pos()` returns type `size_t`, so your changed casts are the correct approach.

jmoerschbach: did that without completely understanding it ;)

src/map_io/map_players_view_packet.cc
```diff
@@ -161,10 +161,10 @@ namespace {
 
 #define CHECK_TRAILING_BYTES(file, filename)                                                       \
 	if (!(file).end_of_file())                                                                      \
-		throw GameDataError(                                                                         \
-		   "MapPlayersViewPacket::read: player %u:"                                                  \
-		   "Found %" PRIuS " trailing bytes in \"%s\"",                                              \
-		   plnum, static_cast<long unsigned int>((file).get_size() - (file).get_pos()), filename);
+		throw GameDataError("MapPlayersViewPacket::read: player %u:"                                 \
+		                    "Found %" PRIuS " trailing bytes in \"%s\"",                             \
+		                    plnum, static_cast<size_t>((file).get_size() - (file).get_pos()),        \
```
gunchleoc: Can you fix all the plnum and not just the one? Search the file for `MapPlayersViewPacket::read: player`. Thanks :)

jmoerschbach: I have done so.

gunchleoc: Excellent, thanks!

gunchleoc (approved):

PR #3784, opened 2020-03-20
Fix stalling test suite scenarios

### LINKED ISSUE #3728: Test suite sometimes hangs on Clang5 during saveloading
Saveloading fails sometimes with the test suite, which is causing the timeouts on Travis. The game then just keeps running at 1x speed without doing anything.

**To Reproduce**

1. Get an Ubuntu Trusty
2. Install clang version 5.x
3. Open a terminal
4. Delete the `build` directory if you already have one
5. Call
```
export CC=/usr/bin/clang-5.0
export CXX=/usr/bin/clang++-5.0
```
6. Run `./compile.sh -w -a -s -t`
7. Run `./regression_test.py -b ./widelands` until you hit the error. Repeat this step a few times until the bug triggers.

**Version:**
 - OS: Ubuntu Trusty with Clang 5.0
 - Widelands Version: r24393[3c39438@master] (Debug)

gunchleoc:
* `stable_save` now uses last save time and average frame rate for waiting
* pass desired speed explicitly to `stable_save`

Fixes https://github.com/widelands/widelands/issues/3728


src/wui/interactive_base.cc
```diff
@@ -676,6 +676,10 @@ void InteractiveBase::think() {
 	UI::Panel::think();
 }
 
+double InteractiveBase::average_fps() const {
+	return 1000.0 / (avg_usframetime_ / 1000);
```
Noordfrees: please use `1000.0 * 1000.0 / avg_usframetime_` for better readability

test/maps/plain.wmf/scripting/test_inputqueues.lua
```diff
@@ -46,7 +46,7 @@ run(function()
 
    assert_all_queues_are_full(b)
 
-   stable_save(game, "inputqueues")
+   stable_save(game, "inputqueues", 50000)
```
Noordfrees: should be `50 * 1000` like in all the other stable_save calls

test/scripting/stable_save.lua
```diff
@@ -1,13 +1,30 @@
...
    game:save(savename)
    game.desired_speed = 1000
-   sleep(2000)  -- Give the loaded game a chance to catch up
-   game.desired_speed = old_speed
-   sleep(1000)
+
+   -- Wait until save was finished
+   local counter = 0
+   repeat
+      sleep(200)
```
Noordfrees, line 16: Forgotten counter increase.
Also what is the point of breaking after 20 iterations – it looks to me that breaking early would be a bug

test/scripting/stable_save.lua
```diff
@@ -1,13 +1,30 @@
...
+   until game.last_save_time ~= last_save_time or counter > 20
+   print("###### stable_save: new save time " .. game.last_save_time)
+
+   -- Give the loaded game a chance to catch up
+   counter = 0
+   while mapview.average_fps < 20 and counter < 100 do
+      sleep(200)
+      counter = counter + 1
```
Noordfrees, line 24: This pauses execution of the script for up to 20 seconds, but the game continues running. Can we be sure that nothing testcase-critical will happen during this undefined waiting period?

Noordfrees (changes requested):
Code looking good in general, a couple of nits inline though

test/scripting/stable_save.lua
gunchleoc: We already had a sleep here before, just for a fixed, completely arbitrary time. It should be OK as long as we don't launch this function in a coroutine.

I ran the test suite on the buggy setup for about 10 times without any failures, so it should be OK.

gunchleoc:
I have fond the comment regarding the instability of the saving design:
```
void Game::think() {
	assert(ctrl_);

	ctrl_->think();

	if (state_ == gs_running) {
		// TODO(sirver): This is not good. Here, it depends on the speed of the
		// computer and the fps if and when the game is saved - this is very bad
		// for scenarios and even worse for the regression suite (which relies on
		// the timings of savings.
		cmdqueue().run_queue(ctrl_->get_frametime(), get_gametime_pointer());

		// check if autosave is needed
		savehandler_.think(*this);
	}
}
```

Noordfrees (approved):
Okay, so apparently this potential problem already exists in master then. I'd say fixing this is out of scope for this branch and should only be looked into if it becomes a problem sometimes.

gunchleoc:
The best fix would be to run the UI and the game simulation in separate threads where the game thread has no dependencies on the UI thread. That will definitely be a big project.

PR #3792, opened 2020-03-22
Simplifying the Fellowships map, for better playing experience.

tppq:
This modifies the "Fellowships" map.

Expanding to the southern island requires many small military buildings, and one middle sized one. One has to know which location requires the middle sized building. A casual player first makes a small building, dismantles it and build the larger one. This is annoying. Playing should be pleasant not annoying. Requiring the middle sized one was a bad design choice.

This modifies the map slightly so, all buildings can be small while expanding to south island. No other modifications, unless I was careless with mouse. Playtested after the modification, although with one of the slots only.

ref: https://www.widelands.org/forum/topic/4786/


data/maps/Fellowships_v2.wmf/elemental
```diff
@@ -1,12 +1,12 @@
-# Automatically created by Widelands bzr7905[trunk] (Debug)
+# Automatically created by Widelands r24502[97cef98@master] (Release)
 
 [global]
 packet_version="1"
 map_w="96"
 map_h="96"
 nr_players="3"
-name=_"Fellowships"
+name="Fellowships"
```
Noordfrees: Re-add translation markup for these three lines

Noordfrees (commented):
Tested it quickly – looking good :)
One comment inline, then this can go in.

tppq:
Oops. Done.

By the way, translating map names can be sometimes annoying, when agreeing on network game (if both see a localized map name, and fail to understand each others' suggestions). The original could be displayed beside the translated one.

Noordfrees (approved):
> By the way, translating map names can be sometimes annoying, when agreeing on network game (if both see a localized map name, and fail to understand each others' suggestions). The original could be displayed beside the translated one.

We already have an option "show real map names" in the map selection screen which players can use for that.

tppq:
> > By the way, translating map names can be sometimes annoying, when agreeing on network game (if both see a localized map name, and fail to understand each others' suggestions). The original could be displayed beside the translated one.
> 
> We already have an option "show real map names" in the map selection screen which players can use for that.

Oh, yes. I seem to be blind.

PR #3797, opened 2020-03-23
Fix huge soldier healing in the field

Niektory:
Healing was higher than intended because of min() and max() mix-up.


gunchleoc (approved):
Aahrg, I always mix up my min and max. Thanks for fixing!

PR #3800, opened 2020-03-24
Prettier stdout files when regression_test is run with python 3

rikardfalkeborn:
Previously, the output was printed as raw byte strings and the output
started like this:

b'This is Widelands Version r24504[28e2244@reg-test-output] (Debug)\n'b'Adding home directory: ...

i.e., everything on one line, with b' and ' enclosing every output from
widelands. Note that testing was still correct using python 3, but
debugging the output files was not easy.

### DIFF
--- a/regression_test.py
+++ b/regression_test.py
@@ -21,12 +21,16 @@ except AttributeError:
         return iter(d.values())
     def iteritems(d):
         return iter(d.items())
+    def bytes_to_str(bytes):
+        return str(bytes, 'utf-8')
 else:
     # Python 2
     def itervalues(d):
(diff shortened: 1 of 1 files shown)


gunchleoc (approved):
Looks good and tested. Thank you for this fix!

How do you wish to be credited in our developers' list, realname and/or nick?

rikardfalkeborn:
> How do you wish to be credited in our developers' list, realname and/or nick?

Realname is fine.

gunchleoc:
Done https://github.com/widelands/widelands/commit/41167729b1f89e183fe618846f73797ee39274eb

PR #3802, opened 2020-03-25
Add lots of optional braces

gunchleoc:
These have become exhausting during code reviews, so I am adding them in and will enforce them in the future with clang-tidy.

This PR should only have the ones that don't create merge conflicts with other currently open PRs.

Wait for the formatting bot before reviewing.


Noordfrees (commented):
Code LGTM.

While scrolling though the diff I saw some `if (false)` and `if (true)` in AI code. Since you're already cleaning up here, perhaps remove the `if (true)` guards and comment out the `if (false)` blocks?

gunchleoc:
I did see some stuff that would benefit from cleaning up, but I didn't want to touch anything because this diff is so big. We can leave those for another day.

Noordfrees (approved):

Noordfrees:
Okay then :)

PR #3805, opened 2020-03-26
Unify program parsers

gunchleoc:
Refactor program parsers
- Pull out common code into new class MapObjectProgram
- Fix ware demand checks so that they only affect the correct tribes
- Get rid of extraneous calls to EditorGameBase::postload()
- Small performance tweak for trainingsite programs
- Get rid of now empty helper library

The problem with the test suite that kept me from merging this was fixed by https://github.com/widelands/widelands/commit/40cff552840da69c34621ea58613d264769a651e

This branch was reviewed twice already, so I'll merge as soon as the CI goes green.


gunchleoc:
Transient failure on AppVeyor. They are having problems with fetching the git repo sometimes.

PR #3806, opened 2020-03-26
Fix off-by-one in get_road()

rikardfalkeborn:
When the input is int8_t, getRoad() expects it to go from 1 to 6 (inclusive).


src/economy/flag.cc
```diff
@@ -631,15 +631,15 @@ void Flag::propagate_promoted_road(Road* const promoted_road) {
 
 	// Calculate the sum of the involved wallets' adjusted value
 	int32_t sum = 0;
-	for (int8_t i = 0; i < WalkingDir::LAST_DIRECTION; ++i) {
+	for (int8_t i = 1; i <= 6; ++i) {
```
tibor95: Is variable i used somewhere? What was wrong with WalkingDir::LAST_DIRECTION?

Noordfrees (approved):
LGTM, not tested. Good catch :)

src/economy/flag.cc
Noordfrees: The internal road indices range from 0 to 5 but WalkingDirs range from 1 to 6. `get_road(i)` already subtracts 1 from the argument internally, doing it again when calling it was a bug.

gunchleoc: Yes it is the following function is called:

```
Road* Flag::get_road(uint8_t const dir) const {
	if (roads_[dir - 1] && Road::is_road_descr(&roads_[dir - 1]->descr())) {
		return dynamic_cast<Road*>(roads_[dir - 1]);
	}
	return nullptr;
}
```

So, the fix looks correct. Good catch! I'd prefer it if we kept the named upper bound though.

We have `for (int8_t i = 0; i < WalkingDir::LAST_DIRECTION; ++i)` twice, shouldn't the second one be fixed as well?

rikardfalkeborn: > Is variable i used somewhere? What was wrong with WalkingDir::LAST_DIRECTION?

i is only used as index to `get_road()`. I changed `Walkingdir::LAST_DIRECTION` to 6 to match existing loops (see below for an example), but I don't feel very strongly about it.

https://github.com/widelands/widelands/blob/07de47cde1335673d0d7902a769631a3275da038/src/economy/flag.cc#L336-L337

rikardfalkeborn: > So, the fix looks correct. Good catch! I'd prefer it if we kept the named upper bound though.

Sure. Should I change `1` to `WalkIngDir::FIRST_DIRECTION` as well?
> 
> We have `for (int8_t i = 0; i < WalkingDir::LAST_DIRECTION; ++i)` twice, shouldn't the second one be fixed as well?

Hmm, I thought I  did that?

gunchleoc: > > So, the fix looks correct. Good catch! I'd prefer it if we kept the named upper bound though.
> 
> Sure. Should I change `1` to `WalkIngDir::FIRST_DIRECTION` as well?

That would be brilliant :)

I noticed that there are more loops  use 1 and 6 instead of the named flags, it would be great to have those changed as well.

> > We have `for (int8_t i = 0; i < WalkingDir::LAST_DIRECTION; ++i)` twice, shouldn't the second one be fixed as well?
> 
> Hmm, I thought I did that?

I guess I overlooked it.

rikardfalkeborn: > > > So, the fix looks correct. Good catch! I'd prefer it if we kept the named upper bound though.
> > 
> > 
> > Sure. Should I change `1` to `WalkIngDir::FIRST_DIRECTION` as well?
> 
> That would be brilliant :)
> 
> I noticed that there are more loops use 1 and 6 instead of the named flags, it would be great to have those changed as well.

Sure, I fixed the ones found in `flags.cc` and a few in `defaultai.cc`. Also, I removed the hardcoded sizes when operating on `roads_` directly (by using e.g. ranged for loops). Maybe I went to far in a single PR, but there you go. :)

src/economy/flag.cc
```diff
@@ -361,7 +359,7 @@ Waterway* Flag::get_waterway(uint8_t const dir) const {
 /// \return the number of RoadBases connected to the flag
 uint8_t Flag::nr_of_roadbases() const {
 	uint8_t counter = 0;
-	for (uint8_t road_id = 6; road_id; --road_id) {
+	for (uint8_t road_id = WalkingDir::LAST_DIRECTION; road_id != WalkingDir::IDLE; --road_id) {
```
gunchleoc: I think these would look better like this:

```suggestion
	for (uint8_t road_id = WalkingDir::LAST_DIRECTION; road_id >= WalkingDir::FIRST_DIRECTION; --road_id) {
```

rikardfalkeborn: Agreed and fixed.

gunchleoc (approved):

src/economy/flag.cc
gunchleoc: Thanks!

src/economy/flag.cc
gunchleoc: No, that's great. Thanks!

I have fixed the codecheck warning from Travis, so this is ready to go in once the builds come through.

gunchleoc:
We are getting test suite failures:

`widelands: ../src/economy/flag.cc:847: virtual void Widelands::Flag::cleanup(Widelands::EditorGameBase&): Assertion '!road' failed.`

Instructions on how to run the test suite at https://www.widelands.org/wiki/RegressionTests/

rikardfalkeborn:
> We are getting test suite failures:

I'm pretty sure I tested, but only a release build...

gunchleoc:
Your last commit fixed it, so this can go in once the builds are finished :)

PR #3809, opened 2020-03-27
Add optional braces to ui_fsmenu

gunchleoc:


src/ui_fsmenu/internet_lobby.cc
```diff
@@ -255,12 +256,15 @@ void FullscreenMenuInternetLobby::fill_games_list(const std::vector<InternetGame
 }
 
 uint8_t FullscreenMenuInternetLobby::convert_clienttype(const std::string& type) {
-	if (type == INTERNET_CLIENT_REGISTERED)
+	if (type == INTERNET_CLIENT_REGISTERED) {
```
Noordfrees, line 259: `switch (type) {` ?

Noordfrees (approved):
LGTM, one (optional) suggestion inline

src/ui_fsmenu/internet_lobby.cc
Noordfrees: Never mind, let's just merge this now

PR #3810, opened 2020-03-28
Add an option to use SDL mouse cursor

Niektory:
Add an option to use SDL to draw the mouse cursor instead of Widelands drawing it. This results in a more responsive cursor.

It's functional now but the code is slightly hacky. Help welcome.


gunchleoc:
How about making the new objects static? This will get rid of checking for whether it needs to be initialized.

I also think that this shouldn't be an option - either we use this, or we don't. As a player, I wouldn't know what this is about, and I wouldn't care about having to deal with this option. I have tested it and I find the responsiveness of the new way preferable.

Also, there is a codecheck failure:
`/home/travis/build/widelands/widelands/src/ui_basic/CMakeLists.txt:1: ui_basic misses DEPENDS on graphic_image_io, because it includes graphic/image_io.h`

Niektory:
They are static though? This checking whether it needs to be initialized is because it's in Panel::do_run(), which can be run multiple times. I put it there because the previous cursor code was there, but it should probably be moved somewhere else. I'm just not sure where. WLApplication? Graphic? A separate class for the cursor?

Actually according to this comment the whole method should not be there ;)
https://github.com/widelands/widelands/blob/24aac1e3fd7c9ea61f7eba3aa75736490677a586/src/ui_basic/panel.cc#L158

The main reason I made it an option is because this cursor will not show on screenshots/video recordings, which might be undesirable. But enabling it by default might be a good idea, as well as renaming the option to something more descriptive (any suggestions?).

klaus-halfmann:
I get the following Warning when compiling:
sdl_cursor/src/ui_basic/CMakeLists.txt:1: ui_basic misses DEPENDS on graphic_image_io, because it includes graphic/image_io.h

src/ui_basic/panel.cc
```diff
@@ -206,9 +224,15 @@ int Panel::do_run() {
 		if (start_time >= next_draw_time) {
 			RenderTarget& rt = *g_gr->get_render_target();
 			forefather->do_draw(rt);
-			rt.blit((app->get_mouse_position() - Vector2i(3, 7)),
-			        app->is_mouse_pressed() ? default_cursor_click_ : default_cursor_);
-
+			if (get_config_bool("sdl_cursor", false)) {
```
klaus-halfmann: is get_config...() fast enough to check this every time?

klaus-halfmann (commented):
Code looks fine. 
On OSX this uses the size of the system cursor, which is +1 for my new 4k Screen.

src/ui_basic/panel.cc
Niektory: I measured ~6 microseconds on my PC. And the SDL calls take ~30 microseconds. Yeah, probably worth caching.

Niektory:
Okay, I decided to move the cursor drawing code into a separate class. This should be cleaner and also make it easier to add for example more cursors in the future.

get_config_bool() and SDL functions are not called unnecessarily any more so it should be a tiny bit faster now + added cleanup to the destructor.

Renamed the option to "Use system mouse cursor" and enabled it by default.

klaus-halfmann:
Code is fine, I will run the testcases and play a tutorial or such ...

klaus-halfmann:
Hmm, I cannot approve this change, only comment?

Niektory:
I noticed a minor glitch when this feature is enabled. When scrolling the map with the right mouse button, the cursor doesn't stay exactly in one spot but jerks around a bit (it's warped back to the starting position almost instantly, but the movement is still visible). I don't think we can keep the system cursor from moving completely, so we have 3 options:
1. Temporarily switch the feature off when scrolling the map. Probably a bad idea, as the cursor will briefly flicker at the moment it is switched and there might be other glitches.
2. Hide the cursor when scrolling the map.
3. Just leave it as it is.

klaus-halfmann (approved):
- Switching the cursoe when scrolling would have bad efefcts when system cursor is bigge than buitin.
+ Disabling the cursor during scrolling sounds like a good idea
* I checked that curso on the map Editor (OSX) I assume its shortly aligning to the grid coordinates for some reason, now that the cursor is fater this became visible. Can we fix this?

BTW. the 'f' Fullscreen hotkey was lost in R21, do we have a Bug for this or was this done on purpose?

As he the scond Issue is propably not related this can go in.

Niektory:
> Disabling the cursor during scrolling sounds like a good idea

OK, I've hidden the cursor when scrolling. I've also hidden tooltips and selection markers since they looked out of place without the cursor.

> I checked that curso on the map Editor (OSX) I assume its shortly aligning to the grid coordinates for some reason, now that the cursor is fater this became visible. Can we fix this?

Sorry, I'm not sure I understand. If you're talking about the map scrolling issue then I've "fixed" it by hiding the cursor. If you're talking about something else, please elaborate.

> BTW. the 'f' Fullscreen hotkey was lost in R21, do we have a Bug for this or was this done on purpose?

The fullscreen hotkey was changed to Ctrl+F in 0730c7ee72d125e0e9dece5bdaf7854be9b453f2.

klaus-halfmann:
> OK, I've hidden the cursor when scrolling. I've also hidden tooltips and selection markers since they looked out of place without the cursor.

OK for me ingame and in Editor

> > I checked that curso on the map Editor (OSX) I assume its shortly aligning to the grid coordinates for some reason, now that the cursor is fater this became visible. Can we fix this?
> 
> Sorry, I'm not sure I understand. If you're talking about the map scrolling issue then I've "fixed" it by hiding the cursor. If you're talking about something else, please elaborate.

My assumption is that (in the editor) the mouse snaps (locks, latches, clinches) to the grid.
But as you hide the cursor, this will not matter any longer.

Thanks for the hint about the fullscreen key!

Can go in once you have merged master.

Niektory:
No, the problem had nothing to do with the grid. Widelands moves the cursor to the same position every frame, but the OS updates the position in between the Widelands frame and the screen refresh, and that difference in timing is enough that it can move a little.

In other words, Widelands fights the OS for the cursor position and loses because the OS is faster :)

klaus-halfmann:
@gunchleoc I dont see fatal error in the builds so whats the Problem here?
The different cusrsor should not make any difference?

Noordfrees:
The travis failure is a random error on gcc9 which we are getting all the time, not related to this branch. Appveyor fail is the usual timeout. Since you approved I'll merge this now.

Issue #3827, opened 2020-04-03
More detailed soldier preferences for military- and trainingsites


Noordfrees:
https://www.widelands.org/forum/topic/4804
https://www.widelands.org/forum/topic/4803

- [x] Add a third radiobutton to militarysites' soldier preference which cancels soldier exchange for this site
- [ ] Add those three buttons to trainingsites as well so players can choose between training preferably soldiers who have already received training from (other) sites (default, post-#3835 behaviour), or any soldiers (pre-#3835 behaviour), or enforce accepting rookies only
- [ ] Additionally have a checkbox or togglebutton in trainingsite windows to disable kicking out soldiers when training attempts keeps failing (default: kickout enabled)
- [ ] We'll need a third entry in MilitarySite::SoldierPreference and appropriate saveloading support, and everything needs to be exposed to the lua interface and constructionsite settings as well
- [ ] Explain all this in the military tutorial

---

Further suggestions from https://www.widelands.org/forum/topic/5090

- [ ] A fourth radiobutton "Prefer Average" that prefers soldiers whose training levels are close to the average levels of all the player's soldiers
- [ ] UI to manually exchange two specified soldiers between two militarysites
 → This would automatically set the soldier preference to No Exchange unless a modifier like Ctrl is held.
 → Might complicate the user interface too much?
- [ ] Add a modifier (Ctrl) to automatically decrease the capacity by 1 when manually expelling a soldier
- [ ] Add a priority slider for soldier priorities to all military- and trainingsites, so the player can ensure that soldiers are e.g. preferably sent to training before occupying a milsite

tppq:
The build request mentioned by Noordfrees fixes #2940. No radio buttons added, so it does not address this. The commit might make the need for radio buttons go away for trainingsites. Military sites are unaffected.

tppq:
>Add those three buttons to trainingsites as well so players can choose between training preferably soldiers who have already received training from (other) sites (default, post-#3835 behaviour), or any soldiers (pre-#3835 behaviour), or enforce accepting rookies only

Should these be preferences of enforcing? In other words, if already-trained soldiers are not present, should the site remain idle or eventually obtain non-preferred untrained soldiers? If evade-1 soldiers are available, should they be out of luck if enforcing untrained only? (I vote for preferences, not enforcement)

Should the requirement of previous training depend on whether the previous training was give here or not (non-Frisians)? (currently not in code)

[...]

PR #3835, opened 2020-04-08
2940 better trainingsite soldier selection

### LINKED ISSUE #2940: Better selection of soldiers for training sites
I had already posted this in the forum, but it should be placed here:

"In order to find out which soldiers will be sent to a training camp I made the following test:

The HQ contains 20 soldiers with evade 0 and 20 with evade 2. There are enough weapons, helmets and food. Next to the HQ is a training camp.

Each time when the training camp requests a new soldier, the HQ send an untrained soldier.

For the arena I made the following test:

The HQ contains 20 soldiers with hp3/attack5 and 20 without any upgrades. There is enough food and strong beer. Next to the HQ is an arena.

Each time when the arena requests a new soldier, the HQ send an non-upgraded soldier."

The problem is that, effectively, trained soldiers will only be equipped with new weapons AFTER all other soldiers have received the evade training. Similarly, upgraded soldiers will only be sent to the arena AFTER all other soldiers have been upgraded with weapons. This is paramount to a deadlock in the training sequence. It seems that some logic has gone lost here.

It becomes impossible to form a small group of elite soldiers, which I consider to be the core issue of an effective strategy.

 * date created: 2013-11-19T10:52:07Z
 * owner: fredkuijper
 * the launchpad url was https://bugs.launchpad.net/bugs/1252690

tppq:
This branch modifies the soldier selection for trainingsites, in a way which reduces the need of micromanagement. For background: see
https://www.widelands.org/forum/topic/4708/ or issue 2940.

In short: Trainingsite attempts to get soldiers that already are partially trained. If the trainingsite fails to train, it switches strategy and requests soldiers that are below the failing level. If training succeeds, the site starts requesting strong soldiers again.

The "prefer weak" mode makes the site work better in no-gold -like situations.

If the site does not get new soldiers, it will gradually start accepting all trainable soldiers (=old way).

Fixes #2940


tppq:
Fixes #2940

src/map_io/map_buildingdata_packet.cc
```diff
@@ -782,14 +781,16 @@ void MapBuildingdataPacket::read_trainingsite(TrainingSite& trainingsite,
                                               const TribesLegacyLookupTable& tribes_lookup_table) {
 	try {
 		uint16_t const packet_version = fr.unsigned_16();
-		if (packet_version == kCurrentPacketVersionTrainingsite) {
+		// TODO(unknown): remove support for previous packet after release 21, to keep code simple.
+		if (packet_version == kCurrentPacketVersionTrainingsite || packet_version == kCurrentPacketVersionTrainingsite -1) {
```
Noordfrees:
```suggestion
		if (packet_version <= kCurrentPacketVersionTrainingsite && packet_version >= 5) {
```

src/logic/map_objects/tribes/trainingsite.cc
```diff
@@ -371,45 +378,117 @@ void TrainingSite::remove_worker(Worker& w) {
...
-			if (descr().get_train_evade())
-				r.add(RequireAttribute(TrainingAttribute::kEvade,
-				                       descr().get_min_level(TrainingAttribute::kEvade),
-				                       descr().get_max_level(TrainingAttribute::kEvade)));
-			if (descr().get_train_health())
-				r.add(RequireAttribute(TrainingAttribute::kHealth,
-				                       descr().get_min_level(TrainingAttribute::kHealth),
-				                       descr().get_max_level(TrainingAttribute::kHealth)));
+void TrainingSite::update_soldier_request(bool did_incorporate) {
+	Game & game = dynamic_cast<Game&>(get_owner()->egbase());
```
Noordfrees:
```suggestion
	Game& game = dynamic_cast<Game&>(get_owner()->egbase());
```

Noordfrees (changes requested):
Code is looking good to me, only two style suggestions inline. But please go through the diff and add optional `{}` everywhere or our chieftain will have a heart attack when she'll look at it ;)
Will test this now

src/map_io/map_buildingdata_packet.cc
```diff
@@ -841,7 +842,26 @@ void MapBuildingdataPacket::read_trainingsite(TrainingSite& trainingsite,
 				mapsize--;
 				trainingsite.training_failure_count_[std::make_pair(traintype, trainlevel)] =
 				   std::make_pair(trainstall, spresence);
+
 			}
+			// TODO(unknown): Packet version 5 was in build 20. If-statement for savegame compatibility
```
Noordfrees: also label new `TODO`s with your name rather than `unknown` please ;)

Noordfrees:
I played two testgames now. With barbarians everything worked fine, each site preferred soldiers that have been well trained by the other one, with no noticeable delay when no such soldier was available.

But with atlanteans this felt unintuitive. The *Dwaalwegen* (sorry don't remember the original name atm – the one that trains evade,defence,health) preferred soldiers that already have maximum health and evade and only lack defence, even though shields were lacking and the other equipment was plentiful. The *Gefahrelke Gangen* (the one that trains attack) however stubbornly refused to accept any soldiers except perfect rookies – it never accepted a single soldier that already had some promotions.

tppq:
> But with atlanteans this felt unintuitive.

This is surprising. Are you saying that labyrinth let soldiers go without defense training because of no shields, and then similar soldiers (only defense missing) came as replacements? Or did you manually eject the "wrong" soldiers or something like that?

I have not seen anything like that, and frankly would have trouble believing if I would not know that you have spent a lot of time with the game. 

Did the dungeon similarly train rookies with full attack, then release them only to fetch more rookies or did you constantly micromanage the dungeon?

Now when you mentioned it. It probably would be better if the site would "reset to default" if player manually ejects a soldier, because then micromanaging would work as before. I did not think of that before.

src/map_io/map_buildingdata_packet.cc
tppq: I think that the original is more intuitive when one removes the support for old versions, but both work similarly well. No strong feelings.

Noordfrees:
Just played another test game with the same findings. I built the trainingsites and the supporting building chains, waited until the sites were ± filled, then allowed them to start training soldiers. I did no micromanaging and did not manually expel soldiers. Result: 
- The labyrinth always preferred highly trained soldiers. But when I stopped my *Panzersmee* so no more shields were produced, it still took *only* highly trained soldiers. To the effect that it took mostly those already fully trained in evade and health, so it failed to train any soldiers at all even though there were still lots of rookies left in the HQ.
- The dungeon trained 1 hero soldier when I started it the first time, failed to give it the last attack promotion because the weapon did not arrive in time, and then it only ever accepted completely untrained rookies and trained them from attack 0 to attack 4.

Edit: screenshot
wl-3835

src/map_io/map_buildingdata_packet.cc
Noordfrees: This is just for consistency because we hardcode the legacy version numbers everywhere else. And when the version is increased next, using relative versions could have a nasty side effect if this is overlooked.

tppq:
> Just played another test game .... did no micromanaging ....

Strange. I played this with Atlanteans, too, and did not see anything alike. I'll try again, probably tomorrow.  Thanks for the finding, the intention is not to inject new bugs.

tppq:
> Just played another test game with the same findings.

I played two games and did not get these results. I must be doing something differently.  Are you stopping and starting the labyrinth and/or dungeon constantly?

I just realized that the training site becomes more prone to letting soldiers out when stopped. I think this is counter-intuitive: Folks usually stop site to gather resources, and when gathering resources the goal is to get good soldiers instead of a swarm of slightly trained militia. Therefore, increasing eagerness to release soldiers early is not good in this situation. I took the liberty to change that in recent commit.

Anyway. Would it be troublesome for you share a savegame where things run in unexpected way?

Noordfrees:
I played another game with atl and at first thought, typical, the moment I'm asked to demonstrate an issue it's magically gone. The sites started by training heroes only, as they should. 
But when you load the savegame (just let it run for a while without doing anything and watch) you'll find that both sites now train rookies but not those soldiers who have already received the other site's training. (Related to the soldiers being kicked out due to the last training steps failing?)

The turning point was around 2:08, when gold started to run out. From then on, the requirements seem to have been relaxed when they shouldn't have been.

Edit: For example, the second soldier exchange after loading the savegame. You have seen that there is now a rookie with attack 2 in the HQ. So this one should be sent to the labyrinth. But it instead gets a level 0 soldier.

[Replay and savegame](https://github.com/widelands/widelands/files/4460965/3835.zip)

tppq:
>But when you load the savegame (just let it run for a while without doing anything and watch) you'll find that both sites now train rookies but not those soldiers who have already received the other site's training. (Related to the soldiers being kicked out due to the last training steps failing?)

Yes, that is expected: No full training causes the site to prefer non-trained soldiers.

That is intentional. If this is wrong, then it would, of course, be possible to change the code so that the preference does not change in this situation. Will take some time, but nothing too difficult. Should I change this like that?

I understood your yesterday comment so that site would fetch untrained soldiers, even if it succeeds in its own training and still does not prefer already trained but trainable soldiers. That I have not been able to see.

EDIT: Thanks for the savegame. I did not see any obvious problems there, though.

tppq:
> The "prefer weak" mode makes the site work better in no-gold -like situations.

I guess that better is subjective.

Noordfrees:
Ah ok. I would expect that trainingsites always prefer to upgrade soldiers who already have some (other) upgrades. Reason is, an attack-2 soldier plus an evade-2-health-1-defence-1 soldier together are not nearly as strong as one soldier with all these upgrades. So in my opinion the better strategy would be that the trainingsite always prefers heroes if heroes that can be trained are available.

tppq:
I did something. Poorly tested at the moment, obviously.

tppq:
I changed the branch to essentially match with above suggestion. I am not sure if this is now better. The game can enter a state where the same folks pendle back and forth, even if untrained soldiers would be available.

Noordfrees:
Yes, this is better :) Now when the dungeon gets no gold weapons, it is preferring soldiers with high evade/health/defence levels and low attack, as I would expect.   
Edit: And when all soldiers with high ev/he/de already have level 2 attack, *then* the dungeon gets rookies, which is perfectly fine.

I'll play some more games with all tribes now…

tppq:
> Yes, this is better :)

Oh, dear. Thanks for test-playing.

I have been thinking of a way to keep you happy while bringing it a bit closer to what I like.

Plan:
- If any soldier gets any promotions, work as in latest commit.
- If no promotions of any kind and trainees are laid off in a row, start demanding less promoted folks like yesterday.

How would that sound? Do you mind if I break savegame compatibility in this branch while doing that?

tppq:
>I  have been thinking of a way to keep you happy while bringing it a bit closer to what I like.

Done. I even kept savegames compatible with previous commits, although in a slightly hack-y way.

Noordfrees:
Tested again with all tribes on the latest commit and liked what I saw very much, great work!

Did you allow the formatting action to run on your repo? If not, [please enable it](https://github.com/widelands/widelands/#code) and then push a dummy commit to resolve some formatting issues in the diff. Once that is done, this can go in from my side :)

tppq:
> Did you allow the formatting action to run on your repo? 

I suppose so, but I might have failed to do it right.

In my view, the review is still not accepted and changes are requested. I suppose I addressed all your point, even if I commented only one of them.

tppq:
> I suppose so, but I might have failed to do it right.

If you think that this is the case, then please give a hint where to look. Thanks in advance for the advice.

src/map_io/map_buildingdata_packet.cc
tppq: done

Noordfrees (approved):
Please go to https://github.com/tppq/widelands/settings/actions and select `Enable local and third party Actions for this repository` to enable automatic formatting

Noordfrees (changes requested):
Oh and travis is reporting an error in the regression test suite (in test/maps/lua_testsuite.wmf/scripting/editor_test_lua_in_editor.lua) which still needs fixing

tppq:
> Oh and travis is reporting an error in the regression test suite (in test/maps/lua_testsuite.wmf/scripting/editor_test_lua_in_editor.lua) which still needs fixing

Travis no longer does random failures? If so, good.

Noordfrees:
We are getting random failures (currently on GCC9 and clang5) all the time. But when [all builds fail](https://travis-ci.org/github/widelands/widelands/builds/673520106) this is most likely related to the branch…

tppq:
>We are getting random failures (currently on GCC9 and clang5) all the time. But when all builds fail this is most likely related to the branch…

That sound as if I had done something wrong. Yet the errors seem unrelated. Example:

>FAIL: test_conquerable_fields_does_not_crash: [string "scripting/cfield.lua"]:336: expected '404' but was '134'!
>
>stack traceback:
>	[C]: in global 'error'
>	[string "scripting/lunit.lua"]:284: in global 'lunit_do_assert'
>	[string "scripting/lunit.lua"]:102: in global 'assert_equal'
>	[string "scripting/cfield.lua"]:336: in upvalue 'func'
>	[string "scripting/lunit.lua"]:425: in function <[string "scripting/lunit.lua"]:425>
>	[C]: in global 'xpcall'
>	[string "scripting/lunit.lua"]:425: in upvalue 'call'
>	[string "scripting/lunit.lua"]:450: in local 'run'
>	[string "scripting/lunit.lua"]:476: in global 'lunit_run_testcase'
>	[string "scripting/lunit.lua"]:382: in method 'run'
>	[string "scripting/init.lua"]:70: in main chunk
>	[C]: in global 'include'
>	[string "scripting/editor_init.lua"]:3: in main chunk

I did not touch any lua files. If that translates into something wrong in any of the three files modified in this branch: I do not understand what is wrong.

or

>FAIL: roads_and_buildings_test: Couldn't place building!
>stack traceback:
>	[string "scripting/flag.lua"]:151: in upvalue 'func'
>	[string "scripting/lunit.lua"]:425: in function <[string "scripting/lunit.lua"]:425>
>	[C]: in global 'xpcall'
>	[string "scripting/lunit.lua"]:425: in upvalue 'call'
>	[string "scripting/lunit.lua"]:450: in local 'run'
>	[string "scripting/lunit.lua"]:476: in global 'lunit_run_testcase'
>	[string "scripting/lunit.lua"]:382: in method 'run'
>	[string "scripting/init.lua"]:70: in main chunk
>	[C]: in global 'include'
>	[string "scripting/editor_init.lua"]:3: in main chunk
>stack traceback:
>	[C]: in method 'place_building'
>	[string "scripting/flag.lua"]:151: in upvalue 'func'
>	[string "scripting/lunit.lua"]:425: in function <[string "scripting/lunit.lua"]:425>
>	[C]: in global 'xpcall'
>	[string "scripting/lunit.lua"]:425: in upvalue 'call'
>	[string "scripting/lunit.lua"]:450: in local 'run'
>	[string "scripting/lunit.lua"]:476: in global 'lunit_run_testcase'
>	[string "scripting/lunit.lua"]:382: in method 'run'
>	[string "scripting/init.lua"]:70: in main chunk
>	[C]: in global 'include'
>	[string "scripting/editor_init.lua"]:3: in main chunk

How do these changes affect building placement? I see no connection between these errors and the changes done here.  I am confused. What should I fix?

tppq:
>How do these changes affect building placement? I see no connection between these errors and the changes done here. I am confused. What should I fix?

Is there a tutorial how to run these tests locally, to see in more detail what is going on?

Niektory:
> Is there a tutorial how to run these tests locally, to see in more detail what is going on?

https://www.widelands.org/wiki/RegressionTests/#test-scenarios

hessenfarmer:
did a short test to see how this could improve hero making. Well it made one sigle issue a bit easier. it would have cost me probably 20 secs to micromanage this. 
However after setting the atlantean labyrinth to zero soldiers and then to full after having trained a soldier fully with all levels provided there, no soldiers were sent. they were sent only after the full trained soldier was sent to the dungeon.
[replays.zip](https://github.com/widelands/widelands/files/4465433/replays.zip)

tppq:
>Well it made one sigle issue a bit easier. it would have cost me probably 20 secs to micromanage this.

This is the issue that the problem description of the linked ticket asks for.

>However after setting the atlantean labyrinth to zero soldiers and then to full after having trained a soldier fully with all levels provided there, no soldiers were sent.

Not really. If only unpromoted soldiers are available, the site waits for a dozen seconds or so before accepting unpromoted soldiers. You did not wait, that is why.

> they were sent only after the full trained soldier was sent to the dungeon.

That was a coincidence, the full promoted soldier going into dungeon did not affect the behavior of the labyrinth.

tppq:
> We are getting random failures (currently on GCC9 and clang5) all the time.

Now all except two gcc-9 tests passed. Should I worry about the failing gcc-9 tests?

Noordfrees (approved):
Since all other tests are green the GCCs don't matter.

tppq:
>  it would have cost me probably 20 secs to micromanage this.

I have been thinking of a modification where micromanaging (=user interaction) resets this behavior for a small amount of time. I'd be happier to do that in a separate branch and finish this first, though.

Anyway. Do you still see a problem with this merge request?

tppq:
>>it would have cost me probably 20 secs to micromanage this.

>I have been thinking of a modification where micromanaging

Let me explain a bit more. During the time when no soldiers were arriving into labyrinth, the site was requesting for soldiers, that have already been trained, and can still receive more training at the labyrinth (=read: already trained at dungeon). Since there were no such soldiers, nobody came in.

If there would have been such soldiers, they probably would have been vacuumed into military sites preferring heroes. The player could have evicted one during that time, making him go to labyrinth to receive further training. This would play well together with additional military site buttons, like those suggested in #3827 .

As conclusion, I think that this branch works well, but of course in case of games "well" is highly subjective, and probably an even better way, which I have not thought of, does exists. Just please be more verbose. I do not have authorities to press the "merge pull request" button, and thus will not do that.

hessenfarmer:
@tppq 
thanks for the explanation. I think your changes work very well in terms of the original request.   
Anyhow I am not sure whether to merge this as it takes away some control from the player, which he can't do anything about.    
Maybe using your demand code in combination with the requested buttons as in #3827 would be better. I believe this code might get us a lot of bug reports as the waiting time isn't intuitive

tppq:
>  I think your changes work very well in terms of the original request.

Thank you for this statement.

> Anyhow I am not sure whether to merge this as it takes away some control from the player, which he can't do anything about.

Well, I think that this does not change the amount of control the player has.

What this does is it greatly increases the amount of fully promoted soldiers, when no micromanaging happens. I think that the goal would be buildings that do not normally need micromanaging. Training sites are not there currently.

This is not incompatible with adding more state buttons. I would not do it in the same commit, as I'd prefer doing one thing at a time.

> Maybe using your demand code in combination with the requested buttons as in #3827 would be better. I believe this code might get us a lot of bug reports as the waiting time isn't intuitive

Yes, the buttons would fit in.

Your test world was pretty small. In any larger map, the wait time is small compared to walk time (but still beneficial). It is also possible to reduce the amount of time, if user experience is of concern and a bug reports actually happen, down to unnoticeable level. If you feel strongly about this, nothing prevents changing this right away.

Noordfrees:
I did not notice the waiting time to be significant during my tests. Also I don't think that the new behaviour should get buttons to switch it on or off – the player has almost no control over training logic here, but in the old behaviour he did not have any more control than now.
Is there a situation (except perhaps edge cases) where the old behaviour would be preferable over the new? If not, let's merge this, and have a follow-up if (and only if) this feature is reported to have not-yet-discovered side effects.

tppq:
>I did not notice the waiting time to be significant during my tests. Also I don't think that the new behaviour should get buttons to switch it on or off – the player has almost no control over training logic here, but in the old behaviour he did not have any more control than now.

I actually would like to have an explicit button to make the site prefer untrained soldiers, and if somebody wants, one for turning off the selection entirely. Despite this, in my mind this should be merged and see if opinions appear on the forum and do the buttons in a separate task.

Of course, I am very biased here. And still, I am not elder and cannot press the "merge pull request" button.

Noordfrees:
> I actually would like to have an explicit button to make the site prefer untrained soldiers, and if somebody wants, one for turning off the selection entirely

That's just what is suggested in #3827 : Three radio buttons for prefer heroes (new behaviour), accept rookies only, and accept any (old behaviour).

But this is out of scope for this branch and if we want this it should go in a follow-up branch, so I will press the magic red button now :)

PR #3836, opened 2020-04-08
WIP: Net debug windows - Only for windows build

Notabilis:
Only for windows build, don't review.


Noordfrees:
Is this PR still needed?

Notabilis:
Nope, can be removed now.

PR #3839, opened 2020-04-10
Suggested teams truncated

### LINKED ISSUE #3786: Editor: Suggested Teams truncated
The editors map loading screen shows at the bottom right a section "Suggested Teams". This section is truncated at the right and shows only "Suggested T". The rows of teams are also truncated:

suggested_teams
 
**To Reproduce**
1. Start the editor 
2. Open Load Map menu
3. Choose the map 'The Nile' or any other map with suggested teams.

**Expected behavior**
The section "Suggested Teams" should be as wide as the map description, that is the whole space at the right of the table of maps. 

**Version:**
 - OS: Archlinunx
 - Widelands Version: r75[97cef98@master] (Release)

**Additional context**
This applies only to the editor, not when starting a new game over "New Game/Choose a map"

jmoerschbach:
fixes #3786

The width of the box was set statically to 100 and never changed.
The width of the parent box is recalculated during runtime which must also be propagated to the box.

I removed the initialiation (set_size()) in the constructor because it is overriden anyways later on and I was confused why my first tries did not have the desired effect.

### DIFF
--- a/src/wui/mapdetails.cc
+++ b/src/wui/mapdetails.cc
@@ -48,8 +48,7 @@ MapDetails::MapDetails(
                  UI::Align::kLeft,
                  UI::MultilineTextarea::ScrollMode::kNoScrolling),
      descr_(&main_box_, 0, 0, UI::Scrollbar::kSize, 0, style, ""),
-     suggested_teams_box_(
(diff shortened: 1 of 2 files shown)


Noordfrees (approved):
LGTM, tested and working :)

Travis failure is the usual GCC

PR #3841, opened 2020-04-11
Fix road buttons added instead of waterway buttons

Niektory:
Fix a bug causing road buttons instead of waterway buttons to be added in waterway building mode when clicking on a field where the waterway can't be built.
The bug was introduced in a8c236de51e837df467a6f0ea55444b498c2752e


Noordfrees (approved):
LGTM, tested and working. Thanks for cleaning up my errors :)

Noordfrees:
The usual GCC, and a transient appveyor error, merging now

PR #3842, opened 2020-04-11
Restrict waterway length limit

Noordfrees:
Restrict the editor spinbox for the waterway length limit to a reasonable range

https://www.widelands.org/forum/topic/4293/?page=13#post-32428


src/editor/ui_menus/main_menu_map_options.cc
```diff
@@ -204,8 +206,10 @@ void MainMenuMapOptions::update() {
 	size_.set_text((boost::format(_("Size: %1% x %2%")) % map.get_width() % map.get_height()).str());
 	descr_->set_text(map.get_description());
 	hint_->set_text(map.get_hint());
-	// map.get_waterway_max_length() defaults to 0 for older maps
-	waterway_length_box_->set_value(std::max<uint32_t>(1, map.get_waterway_max_length()));
+	// map.get_waterway_max_length() defaults to 0 for older maps, and can be
+	// set to arbitrarily high values by manually editing the map files
+	waterway_length_box_->set_value(std::max<uint32_t>(
+	   1, std::min<uint32_t>(kMaxRecommendedWaterwayLengthLimit, map.get_waterway_max_length())));
```
Niektory: How about
```
#include "base/math.h"
waterway_length_box_->set_value(math::clamp<uint32_t>(
	map.get_waterway_max_length(), 1, kMaxRecommendedWaterwayLengthLimit));
```
EDIT:
Actually the implementation of `SpinBox::set_value()` already does the clamping so this is completely redundant and even `waterway_length_box_->set_value(map.get_waterway_max_length());` would work the same...

Niektory:
If you load a map with a waterway length limit greater than 24, it will display as 24 in the editor but it'll still be greater than 24 if you save the map without changing it.

Noordfrees:
> If you load a map with a waterway length limit greater than 24, it will display as 24 in the editor but it'll still be greater than 24 if you save the map without changing it.

Hmm, what would be the best solution here? I want to keep allowing arbitrarily high limits which players may set in the map files manually, so I don't want to clamp the limit on map loading. Clicking OK in the options window without having changed anything will set the ww-limit to 24 anyway.  
A problem only occurs when players open the map options, see "the limit is 24 which is exactly what I want" and close by right-click without any changes… IMHO a cornercase.

Niektory:
So it sounds like you want to allow it but discourage it, right?
Maybe don't clamp the value and just display a warning "Setting the limit to more than 24 is not recommended!" below the spinbox when it's over 24?

Niektory (approved):
Code looks and works fine. 👍

Noordfrees:
Thanks for the review :)
Will merge when travis is green.

PR #3845, opened 2020-04-13
Removing signal from warelist.

tppq:
I am curious what Travis says of this. Removed some lines of code which I did not fully understand.


Noordfrees:
Travis is refusing to test this branch because of a transient mac failure that is cancelling all our builds currently.
Looking at the diff it seems you are just deleting some dead code, which would probably effect no behaviour change at all except a 1 second decreased compile time.

tppq:
The removal of dead code made tests pass (read: failed before, passed after) with gcc 9 on a local computer.

tppq:
I could have messed something accidentally, hence this branch.

tppq:
I assume too that that was dead code, but since I have not touched those parts, a grain of doubt remains.

tppq:
Now gcc-9 succeeds in Release mode but still fails in Debug mode. There is more to be found.

Noordfrees:
Wow, I wouldn't have expected that deleting unused code can make runtime failures go away :)

tppq:
The signal was called without initialisation. I do not know if that is okay (=did not read enough), but caused a crash in practise.

I have the gcc 9 in a vintage recycled laptop with Vista and Centrino stickers; working on large C++ codebase is slow. I might not have time to look at debug build in the next few (=today, tomorrow) days.

I still think that it would be nice for all tests to succeed. Either make them succeed or remove the uninteresting tests.

hessenfarmer:
Yesterday I had a look into the failing GCC9 builds as well. I tried to run the lua scripts manually on my windows appveyor builds by using the debug console. On windows they exited correctly (appveyor uses GCC9 since last year currently GCC9.3). 
In fact it is always the same one lua that fails (test_rip_second_port_with_worker_in_portdock.lua) while (test_rip_second_port_with_ware_in_portdock.lua) succeeds. The only functional differrence between them is that one time a worker is transported and one time a ware is transported. Furthermore there are some different waiting times. 
So either it is just that we need to add some waiting time before printing All tests passed (as this is what the python script looks for)
or it has some deeper cause related to the economy split for workers and wares.

Noordfrees:
I also ran the failing test locally (gcc 9.2.1) but it worked fine.
Strange about these failures is that the "# All Tests passed" message appears in the log and the test is still considered a failure.

tppq:
>So either it is just that we need to add some waiting time before printing All tests passed (as this is what the python script looks for)

I have no opinions here, since I am not familiar with these tests. Last week I even had not noticed their existence.

I'll probably try to make the debug test succeed later this week, unless you fix this earlier. If you do, I appreciate a message here saying so to avoid unnecessary re-discoveries.

hessenfarmer:
from the lua code what should happen at the end of the test is:
1. pick up the worker
2. print"All tests passed.
3. close the game window

from the log it seems as if 3. is not happening in this special configuration. but this is still a transient failure as it works sometimes. There are still some recent builds being green. So maybe the game doesn't get the close signal due to some timing problems that is why I am suggesting some small sleep times between the last steps, to see what happens.

tppq:
The test case test_rip_second_port_with_worker_in_portdock with gcc9 debug build segfaults near end of execution in Economy::remove_wares_or_workers() line assert(owner().egbase().tribes().worker_exists(id))

The actual problem is that the instance of Economy being called is already deleted; so are at least some of the objects mentioned in above call. Something has been overwritten (corrupted data) after deletion, making the pointers (and references etc) point to illegal address which causes the segfault.

tppq:
Minor improvements. I guess merging this would do no harm and reduces memory consumption and compile time by negligible amounts.

The problem with gcc-9 debug build is a bit tough. I guess it could be studied in a separate branch.

tppq:
> The problem with gcc-9 debug build is a bit tough. I guess it could be studied in a separate branch.

This does not mean that I stopped trying, but it could take a long while.

Noordfrees:
> Minor improvements. I guess merging this would do no harm and reduces memory consumption and compile time by negligible amounts.

Okay. Code LGTM, and this branch already fixes one of these two failures, so I'll merge now. Thanks for your efforts :)

tppq:
> Okay. Code LGTM

Thanks for the encouraging comments.

Do you happen to know, if there is any text written on who should own what ? It is normal that many things get deleted near the end of game. Finding fault is difficult, without knowing what is right. If such discussion exists, it would simplify the difficult part a lot.

tppq:
> Do you happen to know, if there is any text written on who should own what ?

Here, the error is triggered by while loop in map_object.cc/ObjectManager::cleanup()

The economy gets deleted when there are no longer objects on flags. Later, a ship (supposedly not on a flag) deletes associated workers in Ship::cleanup()

I suppose that items & workers on ship also belong to economy. Is this the double free?

If it is, then the unfortunate reservation of memory that invalidates pointers is the assertion check in Economy::remove_wares_or_workers(), which calls EditorGameBase::mutable_tribes() which in turn makes world (which does not exist as it is already gone).

Since the assert is not executed in release mode, no crash.

Noordfrees:
> Do you happen to know, if there is any text written on who should own what ?

We have no documentation for ownership relations at all unfortunately. We try to use smart pointers whereever possible to make this a bit less tricky, but most of the time we just rely on ASan throwing errors about double-frees and memory leaks. ASan doesn't work properly with regression_test.py though.

> Here, the error is triggered by while loop in map_object.cc/ObjectManager::cleanup()
> 
> The economy gets deleted when there are no longer objects on flags. Later, a ship (supposedly not on a flag) deletes associated workers in Ship::cleanup()
> 
> I suppose that items & workers on ship also belong to economy. Is this the double free?
> 
> If it is, then the unfortunate reservation of memory that invalidates pointers is the assertion check in Economy::remove_wares_or_workers(), which calls EditorGameBase::mutable_tribes() which in turn makes world (which does not exist as it is already gone).
> 
> Since the assert is not executed in release mode, no crash.

Did you test this or is this is an assumption? I would be surprised, because `world_` and `tribes_` are `unique_ptr` and should therefore be deleted last of all.
But workers being removed from an already deleted economy by `Ship::cleanup` sounds reasonable to me, at least I don't see any code at first glance that would reliably inform the ships that their economy vanished…

tppq:
>Did you test this or is this is an assumption? I would be surprised, because world_ and tribes_ are unique_ptr and should therefore be deleted last of all.

The "world (or actually tribes_) does not exist" is an assumption. The code, however, behaves as it would not just before crash. It could also be that the instance of economy is corrupted.

The practical problem is that I can only see this with gcc-9, and currently I only have it on a recycled old laptop which is completely unsuitable for this.. .. slow progress. That is why I partially guess. Maybe I need to de-dust some virtual machines.

>But workers being removed from an already deleted economy by Ship::cleanup sounds reasonable to me, 

I would not have written the above post without noticing that.. What is the right solution to try? Normally, it is sensible that the ship removes its contents when it sinks. In this particular case it should not do that.

tppq:
Economy::remove_flag() kills the economy, if no flags. However, workers belong to an economy, and are not notified if their economy goes away. Is this OK?

Noordfrees:
The normal destruction order *during* a game is: Last portdock removed → economy of ships and shipped items is unset → flag removed → economy removed. 
But during end-of-game cleanup these steps may be performed in arbitrary order, so there's no guarantee that the portdock is deleted before its flag. `Economy` sends a `Notification` on destruction, perhaps ships should Subscribe to it and unset their own and their items' economy when the `Note` is received?

tppq:
>Economy sends a Notification on destruction, perhaps ships should Subscribe to it and unset their own and their items' economy when the Note is received?

Sounds reasonable -- The number of ships is small. If economy would start tracking all its workers, the game would become even more heavyweight than what it is now.

Does this also cover the case where a worker needs shipping but is not assigned to a ship yet?

tppq:
Another would be to make the kill order at end of game to follow the same ordering. End of game cleanup does not need to be efficient.

Noordfrees:
> Does this also cover the case where a worker needs shipping but is not assigned to a ship yet?

I don't think so – this will be handled by the PortDock which is typically (?) destroyed by the associated Warehouse. There might be problems is warehouse and portdock are destroyed after the last flag, but I *think* the flag does some magic then to prevent this.

> Another would be to make the kill order at end of game to follow the same ordering. End of game cleanup does not need to be efficient.

That would make sense, but reworking it like this could be a pretty big task…

tppq:
I am now rather sure that the reason why this crashes with gcc-9 and not with earlier ones is that the order in which the destruction happens is different.

PR #3846, opened 2020-04-13
Let Barbarians produce cloth on maps without seafaring

### LINKED ISSUE #3699: Barbarian Weaving Mill needs to work with waterways
The Barbarian Weaving Mill won't produce anything on maps that have waterways but no seafaring.

**To Reproduce**
1. Start a game on the "Golden Peninsula" map
2. Build a Reed Yard, a Weaving Mill and a Ferry Yard
3. The Ferry Yard won't get any cloth

The bug is somewhere in this code:

```
ProductionProgram::ActCheckMap::ActCheckMap(const std::vector<std::string>& arguments) {
	if (arguments.size() != 1 || arguments.front() != "seafaring") {
		throw GameDataError("Usage: checkmap=seafaring");
	}
	feature_ = Feature::kSeafaring;
}

void ProductionProgram::ActCheckMap::execute(Game& game, ProductionSite& ps) const {
	switch (feature_) {
	case Feature::kSeafaring: {
		if (game.map().allows_seafaring()) {
			return ps.program_step(game, 0);
		} else {
			ps.set_production_result(_("No use for ships on this map!"));
			return ps.program_end(game, ProgramResult::kFailed);
		}
	}
	}
	NEVER_HERE();
}
```

Niektory:
Removed redundant check from the Weaving mill, so cloth can be produced on maps without seafaring but with waterways.
The building can't be built on maps without waterways and seafaring anyway so there's no need to check it during production.
Fixes #3699

### DIFF
--- a/data/tribes/buildings/productionsites/atlanteans/shipyard/init.lua
+++ b/data/tribes/buildings/productionsites/atlanteans/shipyard/init.lua
@@ -76,7 +76,6 @@ tribes:new_productionsite_type {
          -- TRANSLATORS: Completed/Skipped/Did not start constructing a ship because ...
          descname = _"constructing a ship",
          actions = {
-            "checkmap=seafaring",
             "construct=atlanteans_shipconstruction buildship 6",
             "sleep=20000",
          }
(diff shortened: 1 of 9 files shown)


Noordfrees:
Code is straightforward, fix tested and working.

The `checkmap` command is then used only by shipyards, but those already have a `map_check` as well. Perhaps we should remove `checkmap` altogether?

Niektory:
It doesn't seem to be needed. Do you want me to nuke it?

Niektory:
[Foo](https://www.monkeyuser.com/2020/levels-of-satisfaction/)

Noordfrees (approved):
LGTM :)
Travis is happy except the usual GCC; transient appveyor failure. I'll merge now

PR #3848, opened 2020-04-13
Fix travis MacOS builds

Noordfrees:
On Travis MacOS, gettext needs to be added to the PATH and some CMake CXX and linker flags have to be set because brew force-linking is no longer allowed for whatever reason


Noordfrees:
This PR seems to fix the travis mac errors. I'd appreciate it if someone could take a quick look at the changes (especially as there are a lot of new (?) compiler warnings in the log), but since it's an urgent travis-only change I'll merge this even without review this evening if all mac checks on travis are green.

Noordfrees:
Travis is green except for the GCC, so this should be merged ASAP.

tppq (approved):
I do not see any problems here.

Noordfrees:
Thanks for the review :)

PR #3853, opened 2020-04-16
Fix wrong textdomain for a string

Noordfrees:
Fix a string not getting translated due to being fetched from the wrong textdomain


tppq (approved):
I am convinced that this modification at very least does no harm. Cannot tell whether it helps, but merging is safe.

Noordfrees:
Thanks for the review :)
Travis failure is a transient error with an apt-get repo.

PR #3855, opened 2020-04-16
Hide economy setting when default target quantity is 0

hessenfarmer:
…e eco setting to be hidden

used for atlantean blackrootflour so far.

### DIFF
--- a/data/tribes/wares/armor/init.lua
+++ b/data/tribes/wares/armor/init.lua
@@ -39,6 +39,10 @@ dirname = path.dirname(__file__)
 --
 --    **default_target_quantity**: A table listing the default target quantity
 --    for each tribe's economy. For example, ``{ atlanteans = 3, empire = 1 }``
+--    This parameter is optional. If not set the economy will always demand this ware.
+--    If set to zero the economy will not demand this ware unless it is required in a
+--    production building. If not set or set to zero the actual target quantity will
+--    not be available in the economy settings window
 --
 --    **preciousness**: How precious this ware is to each tribe. For example,
 --    ``{ atlanteans = 0, empire = 1 }``. We recommend not going higher than ``25``.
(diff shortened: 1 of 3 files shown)


data/tribes/wares/blackroot_flour/init.lua
```diff
@@ -7,7 +7,10 @@ tribes:new_ware_type {
    descname = pgettext("ware", "Blackroot Flour"),
    helptext_script = dirname .. "helptexts.lua",
    icon = dirname .. "menu.png",
-   default_target_quantity = {},
+   default_target_quantity = {
+      atlanteans = 0
+
```
Noordfrees: remove blank line

Noordfrees (commented):
Code LGTM, one style comment.

I tested this and found:
- If I set the target for cornflour to a very high value, then blackroot flour will only be produced if there is demand for it. Otherwise, never.
- If I set the cornflour target to a low value, the mill produces blackroot flour indefinitely.

Only downside I can think of is that it is impossible to stockpile blackroot flour even when cornflour is needed, but that's not really a problem IMHO.

Please add a hint in the documentation that wares with a default target quantity of 0 will be excluded from the economy target settings window.

data/tribes/wares/armor/init.lua
```diff
@@ -39,6 +39,10 @@ dirname = path.dirname(__file__)
 --
 --    **default_target_quantity**: A table listing the default target quantity
 --    for each tribe's economy. For example, ``{ atlanteans = 3, empire = 1 }``
+--    This parameter is optional. If not set the economy will always demand this ware.
```
Noordfrees: That's not correct. The table always has to be specified, but it may be empty.

hessenfarmer: oops you are right

Noordfrees:
Good job :)

PR #3859, opened 2020-04-18
Removing new networking accept code

Notabilis:
The new accept code does not work on Windows, making it impossible to play LAN games with a host running Windows. The reason for this is unknown, but seems to either be in boost or Windows itself. Maybe it can be re-added with some future boost version.

No new code is introduced with this branch, and the "old" accept code should already be reviewed. If someone wants to do so anyway: It is in the #else clause of the removed preprocessor directives.
This branch should be tested under Windows, though, since I don't know whether this ever happened.


klaus-halfmann:
* Pro forma we must test this on OSX/Linux, too
* can we do this for windows only? 
   so we have the new code around until asio has evtentuall fixed this?

klaus-halfmann:
Confirm, works on OSX

Notabilis:
Okay, thanks for testing. It works for Linux on my system, so only Windows is missing when the builds are done.

We can do this for windows only, but personally I am no big fan of having multiple code paths for different boost versions or operating systems. After all, the "new" code isn't objectively better or fixes any bugs, I only consider it to be more elegant. Feel free to convince me otherwise, though.

Noordfrees (approved):
Code LGTM. Since this was tested on all three OSs, this can be merged as soon as appveyor is green. (travis failure is only the gcc)

klaus-halfmann (approved):
Works on Windows and cross  Windows OSX
(we still should file a bug against boost::asio)

klaus-halfmann:
Just on Error with  DEBUG: Ubuntu 18.04 with gcc 9.1

PR #3860, opened 2020-04-19
Allow static linking

TinoM:
WIP:
Another attempt to build and link widelands statically with msys64 mingw64


hessenfarmer:
on my local machine it seems as if MSYS mingw does not handle the -lwinmm and other SDL2 dependencies correct. According to Internet these have to be specified you can get them with sdl2-config --static-libs I think. Maybe we have to use something like -Wl static; -lwinmm .... I gave up trying out as I got frustrated.
However I think the other libraries should work with your changes made.

TinoM:
I had it working for years, but with my own distribution with only static libs available.

Currently i am at a point, where only glbinding is dynamically linked, this seems to be the only package from msys2 which does not include static libraries...

hessenfarmer:
That sounds promising. I'll keep a watch, cause I keep learning a lot about building under windows.
Thanks for your effort.

hessenfarmer:
I got it working with glew (see https://github.com/widelands/widelands/tree/static_link) However I needed to rename the glew  dll.a library to make it work. However I beleive this could be done properly in the FINDGLEW.cmake and could be made working with _DOPTION_GLEW_STATIC. I'll try to include this in the FINDGLEW.cmake and then we could have appveyor deliver static builds

TinoM:
Superseeded. @hessenfarmer has done much more work in his branch.

PR #3864, opened 2020-04-20
Fix defense upgrade descriptions in Training Glade

Niektory:


hessenfarmer (approved):
thanks for spotting and fixing

PR #3865, opened 2020-04-20
Removed `return=no_stats` from productionsite programs

### LINKED ISSUE #3863: Wrong ratio of Granite:Marble produced by Empire Marble Mine
**To Reproduce**
1. Load the savegame: [marble_mine_test.wgf.zip](https://github.com/widelands/widelands/files/4504655/marble_mine_test.wgf.zip)
2. Observe the Marble Mine. It will keep executing the `mine_granite` program, mining in the ratio 3 Granite : 1 Marble until input wares run out. The `mine_marble` program (which would produce 1 Granite and 3 Marble) is never executed. The mine's behavior is the same whether the economy needs only granite, only marble, or both.

In this case the mine is "stuck in the `mine_granite` loop", and this is what usually happens. But I also observed "stuck in `mine_marble` loop" behavior occasionally on other maps. I'm guessing this depends on the exact timing when production starts.

**Expected behavior**
I can think of two behaviors that would make sense:
1. `mine_granite` should only be executed when the economy needs Granite, and `mine_marble` only when the economy needs Marble.
2. Granite and Marble should always be mined in the same ratio (2:2 or whatever is desired). In this case there's probably no need to have two separate programs at all.

**Version:**
 - OS: Manjaro Linux
 - Widelands Version: r24564[5703fc9@master] (Release), the same behavior also observed in Build 20

hessenfarmer:
fixes #3863


Niektory:
Seems to work. Both the Empire Marble Mines and the Atlantean Crystal Mines run their programs in a round-robin order now.

Niektory (approved):
The changes look good.

Looking at the source the problem appears to be that `ProgramResult::kNone` is used for both "program not yet called" and "returned with no_stats" states:
https://github.com/widelands/widelands/blob/9256debc0fe6c8e00dc765d31802237a703bc6e5/src/logic/map_objects/tribes/production_program.cc#L428-L429
https://github.com/widelands/widelands/blob/9256debc0fe6c8e00dc765d31802237a703bc6e5/src/logic/map_objects/tribes/production_program.cc#L607-L609

Should we remove `return=no_stats` completely if it's bugged and not needed?

I still think Empire Marble Mine doesn't need two separate mining programs if the conditions are the same. But it doesn't need to be changed now.

hessenfarmer:
I believe I removed it totally in the lua files. But I would be glad if you would double check. So we shoudl remove it in the c++ part as well as it is dead code now.  will Do this asap.

Niektory:
Yes I checked, it's removed from all lua files.

src/logic/map_objects/tribes/production_program.h
```diff
@@ -110,7 +110,6 @@ struct ProductionProgram : public MapObjectProgram {
 	///    Failed             ::= "failed"
 	///    Completed          ::= "completed"
 	///    Skipped            ::= "skipped"
-	///    None               ::= "no_stats"
```
Niektory, line 113: Remove from line 109 as well?

Niektory (changes requested):
Missed one change on line 109, the rest looks good.

( Some of the code using `ProgramResult::kNone` might be obsolete now, but I'm not sure which parts, so maybe better leave them alone to not break anything :) )

src/logic/map_objects/tribes/production_program.h
hessenfarmer: I left this in line 109 due to this state may still happen if a program doesn't get start.

Niektory: It's in the "Parameter syntax" section of `ActReturn` though. It lists parameters that can be used when returning. So the "not started" state doesn't apply here I think.

hessenfarmer: ok I'll remove it thanks for reviewing.

Niektory (approved):

hessenfarmer:
Niektory has approved this but git doesn't recognize. 
I would be interested why this is the case and how to fix this as he showed great ability codewise.
So he should be allowed approving 

As the branch is approved it should be merged

Noordfrees:
@hessenfarmer you are an admin, right? So you can merge PRs yourself even if automatic merging is blocked.
@Niektory please request to join the team [widelands/elders](https://github.com/orgs/widelands/teams/elders/members) then you'll have push access to origin and your reviews count for automatic merging

hessenfarmer:
@Noordfrees 
Yes I am but it felt wrong to merge my own stuff without a properly recognized review so I hesitated. I would use it for travis or appveyor things and so on

Niektory:
Oops, a small problem slipped through. The changes to training sites' `sleep` programs cause them to increase productivity when there are no soldiers inside. I guess we need `return=failed` or `return=skipped` in place of `return=no_stats` there.

hessenfarmer:
will look into it tonight.

hessenfarmer:
@tppq 
As I assume you understood the code of the trainingsite could you please explain under which circumstances the sleep program is called. Thanks in advance

Niektory:
Hmm... I found 2 places.

This one is called when all upgrade priorities are 0 and all upgrade credits are less than 10; I have no idea when that happens:
https://github.com/widelands/widelands/blob/3dcdf52b09aacc2cf1f687e190b566503715b8a3/src/logic/map_objects/tribes/trainingsite.cc#L799-L801

This is called when there are no soldiers present at the training site with upgradable level for the attempted upgrade type:
https://github.com/widelands/widelands/blob/3dcdf52b09aacc2cf1f687e190b566503715b8a3/src/logic/map_objects/tribes/trainingsite.cc#L830-L831

PR #3870, opened 2020-04-21
Fix economy mismatch in Request loading

### LINKED ISSUE #3866: Worker request not saved in savegame
Bug reported on forum see https://www.widelands.org/forum/post/32641/

to reproduce best use a minimal start condition. Create 1 or 2 buildings without proper workers where no tools are available as well. save the game. close and reload the saved game. build a toolsmithy. No tools are built although the smithy could  work.

Noordfrees:
Ware requests for Warehouses' PlannedWorkers are added to worker economies on saveloading. This was not corrected later under some circumstances.
Fixes #3866


Niektory:
Seems to work, I can't reproduce the bug anymore.

Niektory (commented):
Hmm... So first a wrong request is created in `MapBuildingdataPacket::read_warehouse()`, and then it has to be corrected in `Request::read()`. Quite a roundabout way... Probably worth a comment in both methods.

Is this safe? Is there any danger doing it this way can trigger an exception in `Request::Request()`?
https://github.com/widelands/widelands/blob/0899f4c082af99bdc9bd55c4195d8fc084931674/src/economy/request.cc#L72-L75
EDIT: Oh, `index` is `0` so I guess it's "safe".

Noordfrees:
Added some comments. There's no risk of the exception being triggered with the initialization values because this check is already being performed in master so that has been working all the time.

src/map_io/map_buildingdata_packet.cc
```diff
@@ -458,6 +458,9 @@ void MapBuildingdataPacket::read_warehouse(Warehouse& warehouse,
 
 				uint32_t nr_requests = fr.unsigned_32();
 				while (nr_requests--) {
+					// We habe no information regarding the index or WareWorker type yet.
```
Niektory:
```suggestion
					// We have no information regarding the index or WareWorker type yet.
```

Niektory (approved):
Looks good! Just one typo :)

PR #3871, opened 2020-04-21
Fixes issue #3869 : No training in camp for Empire

### LINKED ISSUE #3869: No training in camp for Empire
A few days ago, I started a game as Empire (red player in Lesser Ring 1.2 map).
Yesterday I was finally able to build a training camp.

As soon as it was finished, 4 soldiers went in there (I had set the capacity to 4). No training happened, since there were no helmets for a while.

The moment helmets arrived, the 4 soldiers left the camp. Now, I no soldiers are going to the camp. Though there are plenty of soldiers sitting in the warehouses.
I tried reducing the capacity to 0, the increasing it again a few times, I tried reloading the savegame. Still, no soldiers go to the training camp.

On the other hand, I see well-trained Atlantean soldiers from AI players, so traning must work for them. Also my Colosseum is working fine.

P.S.: Both at game start and yesterday, I was using widelands from then-current git.

P.P.S.: Savegame: http://www.colecovision.eu/stuff/widelands-no-training/PkK%202.wgf

tppq:
Fixes #3869 probably. I did not test this properly yet, see bug discussions for details. I decided to make this visible anyway.


tppq:
>  I did not test this properly yet.

done by now.

Niektory:
I got a replay desync. I don't know whether it's caused by this but it happens right after a training site is constructed so it's suspicious. Sometimes it happens at 21:20 and sometimes at 23:17.
[3869_desync.zip](https://github.com/widelands/widelands/files/4537114/3869_desync.zip)
Played with cedb5e4

tppq:
> I got a replay desync.

I did horrible amounts of playing last weekend, but replays did not cross my mind..

Great testing! Thanks, I'll take a look (in the evening).

tppq:
>... so it's suspicious. 

You should be. Thanks. This should also have shown in multiplayer, not only replay.

Niektory:
Played a test game using 4fcb936. I still saw some long delays. See these saves:

- [training_delay_test_1.wgf.zip](https://github.com/widelands/widelands/files/4566972/training_delay_test_1.wgf.zip) : A soldier finishes training at the training camp at 6:10:33. His replacement leaves the nearby warehouse at 6:11:23 (50 seconds later).
- [training_delay_test_2.wgf.zip](https://github.com/widelands/widelands/files/4566970/training_delay_test_2.wgf.zip) : A soldier finishes training at the training camp at 6:22:09. His replacement leaves the nearby warehouse at 6:23:22 (73 seconds later).

tppq:
> I still saw some long delays. See these saves

Hello, thanks for playtesting. I looked at the savegames. First casually, and saw nothing wrong. However, looking at the timestamps I realized that you are right: the time to next solder was longer that discussed in the previous ticket. Sorry for that, I am a bit blind in this particular matter, because it worked well in my eyes.

src/logic/map_objects/tribes/trainingsite.cc
```diff
@@ -484,6 +523,14 @@ void TrainingSite::update_soldier_request(bool did_incorporate) {
 				requesting_weak_trainees_ = false;
 				latest_trainee_was_kickout_ = false;
 			}
+			if (kUpperBoundThreshold_ <= repeated_layoff_ctr_ && 0 == soldiers_.size()) {
```
Noordfrees: use `empty()` instead of `size() == 0`

Noordfrees (commented):
Code LGTM, one comment inline.
Has someone tested this further?

hessenfarmer:
will test this tonight if nobody is faster.

src/logic/map_objects/tribes/trainingsite.cc
tppq: No problem.. I usually do not write C++, and can forget details like this.

hessenfarmer:
ok I playtested this a bit. it works if there are superfluous soldiers in warehouses. However i didn't see any soldier in the sites while I was short on soldiers. I am not sure but I believe in the previous version (b20) some soldiers were leaving their military site to get trained even when the milsite requests were not fully fulfilled. Or maybe it were for some new recruits to enter the training rather then going to a milsite first. However training took place while not all milsites manned completely I think. Could you check whether the new algorithm is different to the old one for this case? as by this you would need to micromanage a lot to make the effect happen that a training site chooses the correct soldier, as it doesn't choose any at all although fully supplied.

tppq:
> ok I playtested this a bit. 

Thanks, I appreciate that a lot. I really suck at testing own code: I end up repeating the same play pattern and admiring performance in those cases, leading to limited test coverage, despite number of hours spent.

>  I am not sure but I believe in the previous version (b20) some soldiers were leaving their military site to get trained even when the milsite requests were not fully fulfilled. 

Military sites send out soldiers only if better candidates are available. Sending soldiers away can happen before the new guy has arrived, but not before he is assigned. This feature is unaffected by the bugfix or the new feature being patched.

A side-effect of the recent changes is that militarysite re-builds its request periodically. Because of this, the request is new in terms of gametime. I have not verified this, but it seems that old requests have at least higher odds of being fulfilled when there is a shortage of wares/workers. Therefore, if you have empty military buildings and an empty trainingsite, the military sites are more likely (than previously) to get a new soldier from the barracks. That is not related to this bugfix, though. I encourage you to continue discussing the details after build-21. 

> Could you check whether the new algorithm is different to the old one for this case?

The new  feature, or this fix, does not affect military sites at all. Also, militarysite and trainingsite do not negotiate anything with soldiers, both places just ask for soldiers from their economy, and release soldiers to economy. Could you describe, in other words, what should I check?

> as by this you would need to micromanage a lot

Micromanaging is only needed if there are no unassigned soldiers in the economy. However, in that case the player is making an error in my opinion. Barracks is missing/under-resourced or so. Are you saying that militarysites should realize that trainingsites are empty, and send stationed untrained soldiers away for extra training? That would be much of extra code, and should not be done during pre-release freeze.

Thanks once more for the testing. If I failed to respond to all your concerns, please be patient and raise again.

frankystone:
> ok I playtested this a bit. it works if there are superfluous soldiers in warehouses. However i didn't see any soldier in the sites while I was short on soldiers. I am not sure but I believe in the previous version (b20) some soldiers were leaving their military site to get trained even when the milsite requests were not fully fulfilled.

No, afaik new soldiers went directly 
1. to a warehouse
2. to not full filled military sites

Empty training sites is/was normal if you have militray sites which are not full of soldiers. One has to reduce the amount of soldiers in military sites, to get them into a trainingsite. In my opinion this behavior shouldn't be changed a lot.

hessenfarmer:
Ok as said I was not sure about this. I combined testing this branch with playtesting the campaigns and in empire 4 I had massive lacks of soldiers as the barracks couldn't bring them out as quickly as they were needed at the front. The new Ai makes  difference in this scenario. 

So as Kaputtnik can't confirm my suspicion I think we could have this for b21. Let'S see what users think of this in the final call for testing.

Noordfrees (approved):

PR #3873, opened 2020-04-22
Improve AI's attacking decisions

hessenfarmer:
Bugfix for AI Attack Management


tibor95 (commented):
Buildable_field has these variables (and others):
       field.ally_military_presence = 0;
	field.area_military_capacity = 0;
	field.enemy_military_presence = 0;
	field.enemy_military_sites = 0;
	field.enemy_wh_nearby = false;
	field.military_in_constr_nearby = 0;
	field.military_loneliness = 1000;
	field.military_stationed = 0;
	field.military_unstationed = 0;
	field.own_military_presence = 0;
	field.own_non_military_nearby = 0;

But these are calculated for our site (the site coordinates are center of radius) - just be aware of these, though they are not the same as what are you introducing. You can look for update_buildable_field in default_warfare.cc to see its use

tibor95 (commented):
Also performance consideration - you scan vicinity always, is it not possible to condition it by having some attackers available or so? I mean do not gather this information if attacking can not take place anyway.

hessenfarmer:
> 
> 
> Also performance consideration - you scan vicinity always, is it not possible to condition it by having some attackers available or so? I mean do not gather this information if attacking can not take place anyway.

Good idea will try to implement.

hessenfarmer:
> 
> 
> Buildable_field has these variables (and others):
> field.ally_military_presence = 0;
> field.area_military_capacity = 0;
> field.enemy_military_presence = 0;
> field.enemy_military_sites = 0;
> field.enemy_wh_nearby = false;
> field.military_in_constr_nearby = 0;
> field.military_loneliness = 1000;
> field.military_stationed = 0;
> field.military_unstationed = 0;
> field.own_military_presence = 0;
> field.own_non_military_nearby = 0;
> 
> But these are calculated for our site (the site coordinates are center of radius) - just be aware of these, though they are not the same as what are you introducing. You can look for update_buildable_field in default_warfare.cc to see its use

I knew. Guess where I copied the code from ;-) but it can't be used directly, we need to use the same mechanism on enemy sites

tibor95:
OK, all looks OK to me, I am just waiting if you make that change in regard to performance.

hessenfarmer:
I improved the soldier selection (we now choose the full number of planned attackers if available even if we have to skip some due to not healthy.) and added some comments.   
I moved the vicinity check just in front of the inputs definition. this is only reached if the conditions in line 258 are met. So it is only checked if we have soldiers and our power ratio is above limit.

hessenfarmer:
BTW the wai changes re for testing purposes only they should get overridden by training after the merge.    
After rereading your comment I am now unsure whether you perhaps meant the vicinity check in line 52 if so I can't help as this has always been like this even pre genetic algorithm. And it never caused performance issues. So this should stay as it is, cause I do not want to break anything there.

tibor95:
I looked at changes - everything is OK, but so far I did not compile&run it. Is here somebody else who would do it?

hessenfarmer:
I'll ask in the forum.

Noordfrees:
I get this compiler warning:
```
[4/11] Building CXX object src/ai/CMakeFiles/ai.dir/defaultai_warfare.cc.o
../src/ai/defaultai_warfare.cc: In member function ‘bool DefaultAI::check_enemy_sites(uint32_t)’:
../src/ai/defaultai_warfare.cc:596:28: warning: comparison of integer expressions of different signedness: ‘int’ and ‘std::vector<Widelands::Soldier*>::size_type’ {aka ‘long unsigned int’} [-Wsign-compare]
  596 |  while (a < attackers && b < soldiers.size()) {  // choose soldiers until enough or all evaluated
      |                          ~~^~~~~~~~~~~~~~~~~
```

Tested this branch a bit, watched a game of 8 AIs on Dolomites. Some of them showed the old behaviour (attack with 1 soldier at a time – though such useless attacks were directed only against equal, almost never against stronger forces) – and were promptly vanquished by their more clever neighbours. After 8½ hours, four players were defeated, one had retreated and was struggling hard, and the other three AIs had grown very powerful and were making very good strategic decisions, attacking only when they had stronger soldiers or sufficiently large masses of rookies available, every attack wave a success. If such AIs are used for training, we will have reason to fear the computer players soon :)

hessenfarmer:
> 
> 
> I get this compiler warning:
> 
snipped   

Ok I gues I have to cast here to avoid compiler warning. Sorry my abilities in reading code are better then doing it so betterr be careful with my c++, but I am happy to learn. will be done tonight.        

> Tested this branch a bit, watched a game of 8 AIs on Dolomites. Some of them showed the old behaviour (attack with 1 soldier at a time – though such useless attacks were directed only against equal, almost never against stronger forces)    
  
this is exactly what this change aims at: attack only when you have a chance.    
  
> – and were promptly vanquished by their more clever neighbours. After 8½ hours, four players were defeated, one had retreated and was struggling hard, and the other three AIs had grown very powerful and were making very good strategic decisions, attacking only when they had stronger soldiers or sufficiently large masses of rookies available, every attack wave a success. If such AIs are used for training, we will have reason to fear the computer players soon :)    
   
Hm Fear might be to much of expectation. However thanks for testing. Dolomites is a good idea though.

tppq:
I observed one game in "enemy in sight", no teams.

The attacks I noted made sense. What did not that the AI attacked areas it could not reach (because of stone and berries in between), captured several buildings, then dismantled them and lost those soldiers because of long distance and no roads. I guess this is out of scope of this branch though.

tibor95:
@hessenfarmer  we have user reviews and code looks go to me as far as I can tell, so I approve

tibor95:
Well, I messed up something because I did not see any "approve" button, and dont see even now...

hessenfarmer:
@tibor95 
the apporve button is well hidden in github  :-):
1. click on the files changed in the headline of the PR
2. when the diff is displayed click on review changes in the top right corner. ther you get a comment filed with the approve button.

tibor95 (approved):
all looks good to me, also considering players feedback...

Niektory:
I know I'm late, but there is a problem. Now the AI never attacks an opponent that is globally stronger. Even if all the opponent has locally is a blockhouse with one rookie, the AI will not touch it. You can see it in this savegame - nobody attacks red (set the red player to Human or No AI). [ai_test_3.wgf.zip](https://github.com/widelands/widelands/files/4559793/ai_test_3.wgf.zip)

tibor95:
This is still lesser problem than "always attack". Let have some training and see if this does not improve....

Niektory:
I looked at the code and it seems there's a hard threshold `treshold_ratio`. If the ratio of `my_power` to `enemys_power` (both mean global power) is below `treshold_ratio`, we will never attack. Currently `treshold_ratio` is 87% for Normal AI. It's to some degree controlled by the magic number 5. Depending on that number, the possible range of `treshold_ratio` for Normal AI is 55%~95%.

I don't think there should be a hard threshold for global power at all (because local power is what counts, enemy soldiers on the other side of the map don't really matter), and if there is, it should be possible for it to go lower than 55%.

tibor95:
Threshold ratio in range 55%~95% sounds reasonable to me. I wonder if after training this number will be changed. If we had enough training time and you are right, this number would go down with training.

hessenfarmer:
this value had been defined in this range and for this purpose in the old AI code which was often cited in critics about the current AI, so I reinstated it and made it subject to genetics the range was chosen due to it makes no sense to attack an opponent that is twice as powerful as you. You might get a building or 2 after that he will shift his power to you and mop you up. Better strategy is to grow and hope another player will weaken him imho. So I'd wait for training and even if this corner case might happen the AI in general should be better then before but we should have this in  ind for the future.

Niektory:
> the range was chosen due to it makes no sense to attack an opponent that is twice as powerful as you. You might get a building or 2 after that he will shift his power to you and mop you up. Better strategy is to grow and hope another player will weaken him imho.

If there is another player strong enough to weaken him, then he might be tied up fighting that other player and not be able to easily shift much power to you. You should strike at weak points when you have an opportunity. The current game mechanics reward aggression more than turtling.

The threshold also makes the AI more predictable = easier to exploit and less interesting to play against.

So I disagree and advocate widening the range. Isn't it the point of the genetic algorithm to avoid hardcoding values? Why not let the training decide? If I'm right and lower values work well, we'll have a better AI that works well in more scenarios. If you're right and they make no sense then they'll be rejected in training.

hessenfarmer:
> If there is another player strong enough to weaken him, then he might be tied up fighting that other player and not be able to easily shift much power to you. You should strike at weak points when you have an opportunity. The current game mechanics reward aggression more than turtling.

Correct but only up to a certain point. There is a difference between braveness and suicide. AI of build20 was "super aggressive" as always attacking until no men left and this was not the best strategy.    

> The threshold also makes the AI more predictable = easier to exploit and less interesting to play against.

if this threshold wouldn't be there it would be the same as in b20 always attacking same with having a very low threshhold. what could be done is adding some weight of local power into this decision. But this will have to wait for b22 I think.  Anyhow after training is finished the values are in the .wai files and they only mutate from this same basis with only 0.5% chance. So the exact values of all genetics are always hardcoded. I still believe even with this threshold AI is less predictable then in b20. BTW power is not only military it is tried to make this a value for the sum of all statistics.    

> So I disagree and advocate widening the range. Isn't it the point of the genetic algorithm to avoid hardcoding values? Why not let the training decide? If I'm right and lower values work well, we'll have a better AI that works well in more scenarios. If you're right and they make no sense then they'll be rejected in training.   
  
Well I think we might have a look into this again after b22 which will come with new values and we can see how this works. If you want to ensure this not to be forgotten you might open an issue, But I can assure you there are more things to fix in the Ai then this so we will come back anyway.

PR #3877, opened 2020-04-24
Don't show buildings on minimap that are not visible in main view

### LINKED ISSUE #3778: Minimap reveals unseen enemy buildings
In this screenshot, the minimap shows that the blue player has a building near the screen center. No such building is visible.

The debug console reveals that the building in question is a port which is apparently located partly in seen and partly in unexplored terrain.
Either this building should be shown in the main view as well as on the minimap, or the minimap should not show buildings invisible in the main view.

[Savegame.wgf.zip](https://github.com/widelands/widelands/files/4339510/Savegame.wgf.zip)

Niektory:
Fixes #3778 

Works but is a little ugly and may be buggy - I copy-pasted some code without fully understanding it.

### DIFF
--- a/src/graphic/minimap_renderer.cc
+++ b/src/graphic/minimap_renderer.cc
@@ -46,8 +46,8 @@ int round_up_to_nearest_even(int number) {
 }
 
 uint16_t check_vision(const Widelands::Coords& coords,
-					  const Widelands::Player* player,
-					  const Widelands::EditorGameBase& egbase) {
+                      const Widelands::Player* player,
+                      const Widelands::EditorGameBase& egbase) {
 	const Widelands::Map& map = egbase.map();
 	const int32_t mapwidth = map.get_width();
 	Widelands::FCoords f = map.get_fcoords(coords);
@@ -91,12 +91,13 @@ inline RGBColor calc_minimap_color(const Widelands::EditorGameBase& egbase,
 				color = blend_color(color, kWhite);
 			}
 
-			if (((layers & MiniMapLayer::Flag) && immovable->descr().type() == Widelands::MapObjectType::FLAG)) ||
+			if (((layers & MiniMapLayer::Flag) &&
+			     immovable->descr().type() == Widelands::MapObjectType::FLAG)) ||
 			    ((layers & MiniMapLayer::Building) &&
 			     immovable->descr().type() >= Widelands::MapObjectType::BUILDING &&
 			     check_vision(immovable->get_positions(egbase).front(), player, egbase) > 1)) {
-				color = kWhite;
-			}
+					color = kWhite;
+				}
 		}
 	}
 
@@ -206,7 +207,8 @@ void do_draw_minimap(Texture* texture,
 			}
 
 			if (vision > 0) {
-				texture->set_pixel(x, y, calc_minimap_color(egbase, f, layers, owner, player, vision > 1));
+				texture->set_pixel(
+				   x, y, calc_minimap_color(egbase, f, layers, owner, player, vision > 1));
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


src/graphic/minimap_renderer.cc
```diff
@@ -45,11 +45,30 @@ int round_up_to_nearest_even(int number) {
...
+	const int32_t mapwidth = map.get_width();
+	Widelands::FCoords f = map.get_fcoords(coords);
+	Widelands::MapIndex i = Widelands::Map::get_index(f, mapwidth);
+	move_r(mapwidth, f, i);
+
+	if (player == nullptr || player->see_all()) {
+		return 2;
+	} else if (player != nullptr) {
+		const auto& field = player->fields()[i];
+		return field.vision;
```
Noordfrees:
```suggestion
		return player->fields()[i].vision;
```

src/graphic/minimap_renderer.cc
```diff
@@ -45,11 +45,30 @@ int round_up_to_nearest_even(int number) {
...
 }
 
+uint16_t check_vision(const Widelands::Coords& coords,
+					  const Widelands::Player* player,
+					  const Widelands::EditorGameBase& egbase) {
+	const Widelands::Map& map = egbase.map();
+	const int32_t mapwidth = map.get_width();
+	Widelands::FCoords f = map.get_fcoords(coords);
+	Widelands::MapIndex i = Widelands::Map::get_index(f, mapwidth);
+	move_r(mapwidth, f, i);
```
Noordfrees: Please add a comment explaining what this does

src/graphic/minimap_renderer.cc
```diff
@@ -75,7 +94,8 @@ inline RGBColor calc_minimap_color(const Widelands::EditorGameBase& egbase,
 
 			if (((layers & MiniMapLayer::Flag) && dynamic_cast<Flag const*>(immovable)) ||
 			    ((layers & MiniMapLayer::Building) &&
```
Noordfrees:
```suggestion
			if (((layers & MiniMapLayer::Flag) && immovable->descr().type() == Widelands::MapObjectType::FLAG)) ||
			    ((layers & MiniMapLayer::Building) &&
```

src/graphic/minimap_renderer.cc
```diff
@@ -75,7 +94,8 @@ inline RGBColor calc_minimap_color(const Widelands::EditorGameBase& egbase,
...
-			     dynamic_cast<Widelands::Building const*>(immovable))) {
+			     dynamic_cast<Widelands::Building const*>(immovable) &&
```
Noordfrees:
```suggestion
			     immovable->descr().type() >= Widelands::MapObjectType::BUILDING &&
```

Noordfrees (commented):
Tested and working, some performance suggestions inline.

This is for b22.

src/graphic/minimap_renderer.cc
Niektory: I copy-pasted this line from `do_draw_minimap()`, and testing shows it's wrong here. Removing...

Niektory:
I also tried doing it the other way around - showing the buildings in the main view - since that would be the better way I believe. The "currently partially visible" case is not too bad, but the "formerly partially visible" case is tricky. I see two ways to handle it:

- Store all positions in `Player::Field`. That would also require storing something to distinguish them. Currently it stores only what type of immovable was seen on a node, but not which part of it, so if we store all the positions seen we can't tell what's the primary position.
- Always store the primary position in `Player::Field` - even in unseen nodes - when any of the positions is seen. Don't store secondary positions.

Noordfrees:
Lots of instances of `Player::Field` are allocated in big games so it needs to use as little memory as possible. So storing the `MapIndex` of the MapObject's main Coords in addition to the `Descr*` would be best imho.

gunchleoc:
When a building ranges over multiple fields, I think it would be precise enough for the minimap to draw it on any one of these fields and it doesn't matter which one. So, the minimap could remember the serials of the buildings already drawn.

Niektory:
There will be conflicts with #4046 so I'll wait until that's merged.

Niektory:
There will be conflicts with #4046 so I'll wait until that's merged.

PR #3879, opened 2020-04-26
Add cross-compile-unit optimization with Linux GCC.

tppq:
This modifies the CMakefile:

Add -flto optimization flag with gcc, when building without debugging (=Release). Over Linux only, mostly because I do not have other targets and thus cannot test. Since this is definitely small and non-essential improvement, it is better to stay on safe side.

Binary size shrunk by >15% ;  having lots of calls over compile units shows.

I did not notice anything bad happening after this change.


klaus-halfmann:
WOW, if you aref familia with cross compiling we could try to build this for ARM.
There is a (small) chance the Widelands will run on a Raspi 4 again.
(They are working on OpenGL support for this one)

tppq:
>WOW, if you aref familia with cross compiling we could try to build this for ARM.
There is a (small) chance the Widelands will run on a Raspi 4 again.
(They are working on OpenGL support for this one)

These are different things. In this pull request, an optimization step is performed where the compler looks are methods being called across compilation unit boundaries. Example: carrier calls the push_task of bob.o, which calls schedule_act from map_object.o which calls game.cmdqueue().enqueue and so on. Each of these routines is rather simple. The optimizer notices that the callstack can be simplified; omitting redundant instructions speeds up execution, as there are less instructions to carry out (and indirectly also via enhancing cache hit ratio). Normally, C++ source files are compiled one at a time, and this cannot be done, regardless of -O[n] level

Cross-compiling (=compiling for foreign processor architecture) is a completely different topic. I can take a look at that if that is beneficial, but please make a separate ticket. This pull request is about a cross-{compile unit, not processor platform} issue.

Long time ago, Debian required that all processors architectures were supported and I solved some problems with MIPS and SPARC that were supported at that time ;-) Compared to those, working on raspberry pi should be easy.

====

I now realize that I only tested this modification on Linux and gcc 8 & 9 running on x86_64 processor, not all the others.

Raspberry pi 4 is an ARM A53 (ARMv8) afaik, but Raspbian user ARMv7 instruction set. Which one did you have in mind? If not Raspbian, then which OS?

tppq:
> Compared to those, working on raspberry pi should be easy.

Hmm, I read your comment again. If the problem is with Raspberry's OpenGL, it could be beyond my skill and time limits. At least a real HW would be needed.

Niektory:
Hmm... Checking with GCC 9.3 I couldn't measure any difference in in-game FPS. And it greatly increases linking time. Could you print a message when this is enabled? And maybe add an option to disable it?

tppq:
>  I couldn't measure any difference in in-game FPS.

Even if the binary size shrinks by 15-20%, innermost loops are typically not affected so the expected speed difference is smaller. Differences in the percent-or-two range are difficult to notice visually. This could make a difference in really large (in the sense of economy size) games, but even then the difference is small.

> And it greatly increases linking time. 

This is expected, as optimizer is essentially reading and analyzing all the compile units simultaneously.

> Could you print a message when this is enabled?

From the game, or during compile time?

> And maybe add an option to disable it?

I did not think that anybody wants that, but it is true that the slow down could be annoying. Can do, but not right away; could concentrate on build-21 issues.

Niektory:
> > Could you print a message when this is enabled?
> 
> From the game, or during compile time?

During compile time. Just a simple `message (STATUS "Enabling link-time optimizations")`

> > And maybe add an option to disable it?
> 
> I did not think that anybody wants that, but it is true that the slow down could be annoying. Can do, but not right away; could concentrate on build-21 issues.

That's fine.

TinoM:
Just to comment:
It does not work with mingw on windows. If fails to link the executables. Example widelands.exe:
`[1/1] Linking CXX executable src\widelands.exe
FAILED: src/widelands.exe
cmd.exe /C "cd . && C:\bin\msys64\mingw64\bin\c++.exe -O3 -DNDEBUG  -flto  -mwindows @CMakeFiles\widelands.rsp  -o src\widelands.exe -Wl,--out-implib,src\libwidelands.dll.a -Wl,--major-image-version,0,--minor-image-version,0  && cd ."
lto1.exe: internal compiler error: in add_symbol_to_partition_1, at lto/lto-partition.c:153
libbacktrace could not find executable to open
Please submit a full bug report,
with preprocessed source if appropriate.
See <https://sourceforge.net/projects/msys2> for instructions.
lto-wrapper.exe: fatal error: C:\bin\msys64\mingw64\bin\c++.exe returned 1 exit status
compilation terminated.
C:/bin/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/9.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: error: lto-wrapper failed
collect2.exe: error: ld returned 1 exit status
ninja: build stopped: subcommand failed.`

tppq:
> It does not work with mingw on windows.

I guess that linking is significantly different in Windows world. Visual studio does cross compile unit optimization fine, but I have not tried compiling Widelands with it after WL switched to C++11 (because the version available to Windows 7 did not support it back then). Actually, Windows was my prime suspect of problems while adding the "if (linux)" check.

One could of course try Visual Studio again, but I suppose it is not worth all the trouble.

gunchleoc:
I think reducing the binary size is already a bonus, because there are still people out there who have to pay for their data consumption.

Let's do this for release builds only though, because linking time is already too long for debug builds.

tppq:
> Let's do this for release builds only though, because linking time is already too long for debug builds.

gcc manual advises against doing that anyway: "Link-time optimization does not work well with generation of debugging information.  Combining -flto with -g is currently experimental and expected to produce wrong results."

tppq:
> I think reducing the binary size is already a bonus, because

Outside the scope of this, but I noticed that somebody recently ran png optimizations in Master. However, the images could be shrunk further with advdef tool. This was a bit surprising, since  tools/optimize_pngs mentions advpng which should implement the same thing. Maybe advdef/advpng was missing while optimizing the png files? The reduction in reach is about 2%.

gunchleoc:
> Outside the scope of this, but I noticed that somebody recently ran png optimizations in Master. However, the images could be shrunk further with advdef tool. This was a bit surprising, since tools/optimize_pngs mentions advpng which should implement the same thing. Maybe advdef/advpng was missing while optimizing the png files? The reduction in reach is about 2%.

No, it was present. I have opened a new issue: https://github.com/widelands/widelands/issues/3971

gunchleoc:
I'm getting `CMake Error: The source directory "/home/cuideigin/sources/widelands/github/widelands-codereview/build/USE_FLTO_IF_AVAILABLE=yes" does not exist.`

Compiler: GNU 7.5.0
cmake version 3.10.2

I also think that `n` is not a good choise for the switch, because all our switches switch thins off and could be called `n`. How about `o` as in "optimization" or `c` as in "cross-compile-unit"?

tppq:
> I'm getting `CMake Error: The source directory "/home/cuideigin/sources/widelands/github/widelands-codereview/build/USE_FLTO_IF_AVAILABLE=yes" does not exist.`

Are you using compile.sh or something else? If something else: Could you be more verbose?
If using compile.sh: I modified the whitespace, in case that that caused problems with your setup. Could you please retry?

> I also think that `n` is not a good choise for the switch, because all our switches switch thins off and could be called `n`. How about `o` as in "optimization" or `c` as in "cross-compile-unit"?

No problem.

gunchleoc:
That fixed my problem, thanks!

Now we still need to get the Mac builds to go green.

tppq:
> That fixed my problem, thanks!

Thanks for finding the problem in time.

> Now we still need to get the Mac builds to go green.

The build complains about missing SDL2_IMAGE. I guess that that is unrelated.

Noordfrees:
There seemed to be a transient problem on travis with all our mac builds. It seems to be fixed now.

tppq:
This time Appveyor failed.. error: "Could not resolve host: github.com". I guess this is equally unrelated.

gunchleoc:
Looks like it. I have triggered the failed jobs again.

tppq:
> Looks like it. I have triggered the failed jobs again.

3/4 passed, one seems to have "No artifacts found matching 'Widelands-Debug-x64-1.0~git24712[3ec4df6@HEAD].exe' path" error.

gunchleoc:
> > Looks like it. I have triggered the failed jobs again.
> 
> 3/4 passed, one seems to have "No artifacts found matching 'Widelands-Debug-x64-1.0~git24712[[3ec4df6](https://github.com/widelands/widelands/commit/3ec4df6c2fab6af94c16ebf26ccf08619c8cd7aa)@HEAD].exe' path" error.

That one's OK.

gunchleoc (approved):

Noordfrees:
Just a question, this new feature causes the compile script to print lots of warnings like
```
[2/572] Linking CXX static library src/libbuild_info.a
/usr/bin/ar: src/CMakeFiles/build_info.dir/build_info.cc.o: plugin needed to handle lto object
/usr/bin/ranlib: src/libbuild_info.a(build_info.cc.o): plugin needed to handle lto object
```
How can I silence these warnings? (using gcc10 on Ubuntu MATE 20.04)

gunchleoc:
There might be some clues on:

https://www.reddit.com/r/Gentoo/comments/6b1dlc/plugin_needed_to_handle_lto_object_and_correct/

https://stackoverflow.com/questions/32221221/mingw-x64-windows-plugin-needed-to-handle-lto-object

tppq:
I do not remember installing any plugins for this while testing, and therefore did not pay attention.

I am off-office now. However, if it would be safe to add the -fwhole-program flag, then the plugin is not needed (and even more aggressive optimization is done).

PR #3881, opened 2020-04-27
Add missing check for helmet in Frisian training camp

Niektory:
When helmet was missing the camp was still attempting upgrade and wasting 30 seconds.


hessenfarmer (approved):
good catch. 
Seems we have got a little bit blinded

Noordfrees:
Yes, good catch :)
No need to wait for travis here

PR #3888, opened 2020-05-01
barbarian Poor hamlet fix

### LINKED ISSUE #3867: barbarian poor hamlet does not work as intended
Barbarian poor hamlet can't be solved without dismantling a lumberjack which is not intended.

Proposed solution change the workorder of the metal workshop to first build firetongs and not felling ax. This should not have effect on normal gameplay

hessenfarmer:
fixes #3867 
changed order of tools to produce


Niektory (approved):
The felling ax is still made. At least it's made after the fire tongs, *if* the player has built the smelting works (or increased demand for fire tongs) *before* starting the metal workshop.

It doesn't fix the root cause, but it does make a deadlock slightly less likely.

frankystone:
A comment would be useful. Otherwise in future some one else change the order again, not knowing why this specific order matters.

hessenfarmer:
@Niektory at least it is now consistent with the empire poor hamlet.    m
@frankystone good idea will add a comment

Noordfrees:
Regarding the root cause, worker plans not being deleted when the request is gone – I would consider this not a bug but an implementation detail (in Economy::create_requested_worker). This behaviour could probably be changed, but if we want to attempt it, this should definitely wait for b22.

PR #3889, opened 2020-05-01
fixed ware help

### LINKED ISSUE #3882: Doubled ware descriptions in Tribal Encyclopedia
Several wares (mead, honey bread, iron, Atlantean smoked fish) in the Tribal Encyclopedia have doubled "wares consumed/produced" sections. This is caused by the duplicate programs added in 726d3d3c1dae4a2dc90cc61d690b5c7066c892dc.

double_ware_descriptions

 - Widelands Version: 01ae086ebe550d7365176b8cd1c0024009fb74e7

hessenfarmer:
fixes #3882


data/tribes/scripting/help/ware_help.lua
```diff
@@ -76,44 +76,54 @@ function ware_help_producers_string(tribe, ware_description)
...
+                  .. help_ware_amount_line(ware_descr, amount)
+            end
+            produced_wares_strings[program_name] = produced_wares_string
+         end
+         -- check for doubled entries
+         for j, program_name in ipairs(producing_programs) do
+            for i, prog_name in ipairs(producing_programs) do
+               if (produced_wares_strings[program_name] == produced_wares_strings[prog_name] and prog_name ~= program_name) then
+                  if (help_consumed_wares_workers(tribe, building, program_name) == help_consumed_wares_workers(tribe, building, prog_name)) then
+                     table.remove(producing_programs, i)
```
Niektory: Removing items from a table we're iterating on is dangerous. It can lead to skipping some items. It happens to work now because we only need at most one removal. If there were more it could break.

Niektory (changes requested):

data/tribes/scripting/help/ware_help.lua
hessenfarmer: the idea behind this double loop is to check all programs in the table against the each other starting from the beginning. If any entry in the table is similar to a previous one it does not need to be checked again and can be savely removed this is why we delete the index in the inner loop as it is in advance of the outer loop index. By this all (if multiple) doubled entries after the first one will be deleted before increasing the outr loop index. An entry is only deleted if it is a real doubled entry in terms of input and output. Even a changed order of input wares prevents its deletion. So I beleive this is loop is pretty save. As I can't think of any working alternative and I tried a couple of other, any proposal would be appreciated.

Niektory: The inner loop *can* skip entries after a deletion. They *might* be picked up again in a subsequent outer loop iteration, but I'd rather not rely on this...

I would do it like this:
```
         -- check for doubled entries (identical consumed and produced wares)
         local deduplicated_programs = {}
         for j, prog1_name in ipairs(producing_programs) do
            local duplicate = false
            for i, prog2_name in ipairs(deduplicated_programs) do
               if produced_wares_strings[prog1_name] == produced_wares_strings[prog2_name] and
                     help_consumed_wares_workers(tribe, building, prog1_name) ==
                     help_consumed_wares_workers(tribe, building, prog2_name) then
                  duplicate = true
                  break
               end
            end
            if not duplicate then
               table.insert(deduplicated_programs, prog1_name)
            end
         end
         producing_programs = deduplicated_programs
```

hessenfarmer: ok, this is cleaner. Will change it accordingly. thanks for the tip

Niektory (approved):

Noordfrees (approved):

PR #3890, opened 2020-05-02
Fix wrong colors on Big Endian systems

### LINKED ISSUE #3887: Some texture colors are off on powerpc
Hi,

**Describe the bug**
Some texture colors are off on my powerpc machine, as you can see in the screenshot below.

If you don't have the hardware but know how to fix it, feel free to me send me patches :)

**Screenshots**
widelands

**Version:**
- OS: OpenBSD/macppc 6.7-beta
- Compiler: clang 8.0.1
- Widelands: 20
- Machine: PowerBook G4
- GL infos:
```
OpenGL vendor string: X.Org R300 Project
OpenGL renderer string: ATI RV350
OpenGL version string: 2.1 Mesa 19.2.8
OpenGL shading language version string: 1.20
OpenGL extensions:
OpenGL ES profile version string: OpenGL ES 2.0 Mesa 19.2.8
OpenGL ES profile shading language version string: OpenGL ES GLSL ES 1.0.16
OpenGL ES profile extensions:
```

Niektory:
Reverse the order of bytes when creating SDL surface on big endian architectures.

Fixes #3887 ... hopefully. I have no PowerPC to test.

### DIFF
--- a/src/graphic/sdl_utils.cc
+++ b/src/graphic/sdl_utils.cc
@@ -23,6 +23,10 @@
 
 SDL_Surface* empty_sdl_surface(int16_t w, int16_t h) {
 	SDL_Surface* const surface =
+#if SDL_BYTEORDER == SDL_LIL_ENDIAN
 	   SDL_CreateRGBSurface(SDL_SWSURFACE, w, h, 32, 0x000000ff, 0x0000ff00, 0x00ff0000, 0xff000000);
+#else
+	   SDL_CreateRGBSurface(SDL_SWSURFACE, w, h, 32, 0xff000000, 0x00ff0000, 0x0000ff00, 0x000000ff);
+#endif
 	return surface;
 }


julianaito:
It works fine, great!

Noordfrees (commented):
Code LGTM. Is there some way to print a compile-time info message which one is being used?

Niektory:
> Is there some way to print a compile-time info message which one is being used?

I could add `#pragma message "foo"`, but that looks ugly (similar to compilation warnings). Other ways are more complicated. Probably not worth it.

How about a run-time message?

Noordfrees (approved):
This function is called all the time so it would spam the logs probably. Let's leave it as it is then. Good fix :)

Niektory:
> This function is called all the time so it would spam the logs probably.

It could be added not in this function, but somewhere in the startup code.

Noordfrees:
Good point, didn't think of this. This could go into the lengthy graphics info log displayed while starting.

Niektory:
I added it to the branch. (do I need a new PR?)

Noordfrees:
Pushed your commit directly to master. Thanks :)

PR #3897, opened 2020-05-05
Decrease training site productivity when no soldier to train

Niektory:
It was erroneously being increased before as a consequence of the changes in #3865

### DIFF
--- a/data/tribes/buildings/trainingsites/atlanteans/dungeon/init.lua
+++ b/data/tribes/buildings/trainingsites/atlanteans/dungeon/init.lua
@@ -156,6 +156,7 @@ tribes:new_trainingsite_type {
          descname = _"sleeping",
          actions = {
             "sleep=5000",
+            "return=failed",
          }
       },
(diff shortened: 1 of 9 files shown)


hessenfarmer:
Thanks for taking care I almost forgot this. Although the result would be the same could you change it to return skipped in this cycle, as I believe (still not sure as the code is complicated) this is called in case it is skipping due to no soldier present. If it fails due to no ressources it should return fail already.

Niektory:
OK, changed to `return=skipped` now. I wasn't sure which one is appropriate.

hessenfarmer (approved):
Thanks for fixing

hessenfarmer:
@Noordfrees 
I think we should have this as it fixes a bug introduced in a recent commit from me. Mea Culpa.

Noordfrees:
Okay you can merge as soon as travis is green

hessenfarmer:
Merged. Thanks for cleaning up my faults

PR #3906, opened 2020-05-09
new WAI files - result of about 120 rounds of training

tibor95:
This is training after latest changes related to AI player attacking. I did not tested them and I wonder if the result of training is visible :)


hessenfarmer (approved):
tested them quickly 
looks good for me

PR #3913, opened 2020-05-20
Widelands appveyor project settings updated to base Image Visual Stud…

TinoM:
…io 2019, remove workaround


TinoM:
Ok, this was a dead end...

Issue #3372, opened 2019-09-09
Boost 1.69 breaks compiling


bunnybot:
Latest Boost breaks widelands at least on windows, see recent Appveyor builds.

=> https://github.com/boostorg/logic/pull/5

It seems that also the FindBoost.cmake by Cmake needs an update...

------------------------------------
Imported from Launchpad using lp2gh.

 * date created: 2018-12-14T07:18:42Z
 * owner: tino79
 * the launchpad url was https://bugs.launchpad.net/bugs/1808481

bunnybot:
(by franku)
Hm, the link you postet is about bool/tribool, but the the error on appveyor is about std::snprintf?

Can you explain the context?

Since i will be soon affected by this due to my linux distribution, i would be glad if the underlying bug will be fixed.

bunnybot:
(by franku)
Luckily i have no problems with boost 1.69. Used packages:

Name                     : boost
Version                  : 1.69.0-1
Architektur              : x86_64

Name                     : boost-libs
Version                  : 1.69.0-1
Architektur              : x86_64


Test results:

    Built target wl_tests
Test project /home/kaputtnik/widelands-repo/trunk/build
    Start 1: test_economy
1/4 Test #1: test_economy .....................   Passed    0.19 sec
    Start 2: test_io_filesystem
2/4 Test #2: test_io_filesystem ...............   Passed    0.09 sec
    Start 3: notifications_test
3/4 Test #3: notifications_test ...............   Passed    0.01 sec
    Start 4: test_scripting
4/4 Test #4: test_scripting ...................   Passed    0.02 sec

100% tests passed, 0 tests failed out of 4


I get a deprecation warning though:

   Building CXX object src/random/CMakeFiles/random.dir/random.cc.o
In file included from /usr/include/boost/random/detail/integer_log2.hpp:19,
from /usr/include/boost/random/detail/large_arithmetic.hpp:19,
from /usr/include/boost/random/detail/const_mod.hpp:23,
from /usr/include/boost/random/detail/seed_impl.hpp:26,
from /usr/include/boost/random/mersenne_twister.hpp:30,
from /usr/include/boost/uuid/random_generator.hpp:17,
from /usr/include/boost/uuid/uuid_generators.hpp:17,
from /home/kaputtnik/widelands-repo/trunk/src/random/random.cc:26:
[...]

PR #3916, opened 2020-05-21
Fix snprintf error for Windows build

Adriankhl:
Related to https://github.com/widelands/widelands/issues/3372

Previously, the workaround is to patch the `libintl.h`. While it works, I think this could be dangerous to my local development environment since it could break other softwares.

Here, I propose that undefining `snprintf` is a better fix for the issue since it limits the modification to widelands only.


hessenfarmer:
Did you check whether this does no harm to the usage of snprintf in other parts of the code? 
As the patch does not completely undefine snprintf but prefents redefinition to libintl_snprintf.

Adriankhl:
@hessenfarmer As far as I know, without the `undef`, all `snprintf` in `src`  become `libintl_snprintf`, so basically the changes here are identical to the previous workaround.

hessenfarmer:
the quetion was if the undef without redefinition of the normal snprintf would do any harm? But as it compiles properly and runs suficiently (although I did not test that much) this might be good to go. However it is the decision of Nordfrees whether this will be part of b21 or later

Noordfrees:
Hi @Adriankhl and thank you for this fix :)

Since we can't be 100% sure that this has no side effects we should merge this after b21.

@hessenfarmer how much did you test this? `snprintf` is used by our code only in saving the MapPlayersViewPacket, can you please check whether this is working correctly?

hessenfarmer:
ok played a while now and saved and reloaded as well. nothing happened so seems to work well

hessenfarmer (approved):

gunchleoc:
>  `snprintf` is used by our code only in saving the MapPlayersViewPacket, can you please check whether this is working correctly?

I also found occurrences in

`std::string Md5Checksum::str()`

and in

`third_party/eris/luaconf.h`

PR #3918, opened 2020-05-22
Make emp03 title translatable

Noordfrees:
We have to break string freeze again :(


hessenfarmer (approved):
oops. did not see this. good catch

Noordfrees:
Merged, and pushed the new string to Transifex straight

PR #3922, opened 2020-05-23
Fix recycling center hotspot

### LINKED ISSUE #3921: frisian recycling center jumps back and forth between working and idle
Graphical glitch. probably different hotspots

Noordfrees:
Fixes #3921 
This typo must have slipped in when re-exporting the mipmaps. This is a small anim-hotspot-only change so we can have it for b21.

### DIFF
--- a/data/tribes/buildings/productionsites/frisians/recycling_center/init.lua
+++ b/data/tribes/buildings/productionsites/frisians/recycling_center/init.lua
@@ -35,7 +35,7 @@ tribes:new_productionsite_type {
       working_metal = {
          directory = dirname,
          basename = "working_metal",
-         hotspot = {50, 61}, -- the whole animation is one pixel lower
+         hotspot = {50, 71}, -- the whole animation is one pixel lower
          frames = 10,
          columns = 5,
          rows = 2,


hessenfarmer:
It still jumps by one pixel. However from the comment in the file I don't know whether this is intentional

Noordfrees:
Hm, I can't see any jumping with this hotspot. But if I increase or decrease it by 1 *then* it jumps. The reason this anim has a different hotspot is that the total height of the working_metal spritesheet is 182 whereas the others have 180 due to a blank row having being introduced at the top of each frame for whatever reason.
Do you see the jumping now at 1× zoom or only when zoomed in? If the latter, this can happen with any anim because we would need to specify quarter-pixel steps in hotspots (or an individual hotspot for every zoom level) to avoid that which is not currently possible.

hessenfarmer:
ok it was only visible when zoomend in, and it was still very small. Fits to your explanation above so this is good to go I think. Thanks for fixing

hessenfarmer (approved):

Noordfrees:
Thanks for the review. I opened a new issue (#3925) for the zoomed-in hotspots.

PR #3932, opened 2020-06-01
count_owned_valuable_fields: check for nil

### LINKED ISSUE #3931: lua error in woodgnome
**Describe the bug**
`[../src/scripting/lua_errors.cc:22] [string "scripting/win_conditions/wood_gnome.lua"]:60: bad argument #3 to 'ngettext' (number expected, got nil)`

This happens when a user has 0 trees.

**Version:**
 - Widelands Version: https://github.com/widelands/widelands/commit/5acb402b71c06177278cd3626c4b2cd2e89a0173

stonerl:
...and assign 0 in such cases.

fixes #3931

### DIFF
--- a/data/scripting/win_conditions/win_condition_functions.lua
+++ b/data/scripting/win_conditions/win_condition_functions.lua
@@ -178,6 +178,9 @@ function count_owned_valuable_fields_for_all_players(players, attribute)
          owned_fields[plr.number] = 0
       else
          owned_fields[plr.number] = all_plrpoints[plr.number]
+         if owned_fields[plr.number] == nil or owned_fields[plr.number] =='' then
+           owned_fields[plr.number] = 0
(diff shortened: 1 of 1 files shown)


hessenfarmer:
Although this should work. Wouldn't it be better to initialize the return Variable in the related c++ function with 0 so we always have a result here. 
Anyhow I think this should be fiixed for b21

stonerl:
@hessenfarmer I made the check simpler.

The c++ function just checks if a player has valuable fields and adds all players and their respective fields to a container. If a player has no valuable fields, he doesn't get added and `all_plrpoints[plr.number]` returns `nil`. Testing for `nil` is the easiest way IMHO.

hessenfarmer (approved):
LGTM. This is best I admit.

PR #3933, opened 2020-06-04
Productivity no ressources fix

### LINKED ISSUE #3929: Out of resource messages are triggered too late
The "out of…" messages e.g. of quarries are triggered only when the productionsite reaches 0% productivity.

hessenfarmer:
fixes #3929


Noordfrees (approved):
LGTM. Still takes about two minutes until the message appears but this is an acceptable delay. Thanks for fixing :)

PR #3936, opened 2020-06-06
Fix textdomain for »New Objective(s)« headers

### LINKED ISSUE #3935: Missing  localization of 'New Objective(s)'
The strings 'New Objective' respectively the plural form 'New Objectives' are not translated.

* Change language to non english
* Start a tutorial or a scenario and wait until  a window showing objectives pops up.

**Version:**
 - OS: archlinux
 - Widelands Version: [3fa1b7f@master] (Debug)

**Additional context**
The translation makro is applied in [messages.lua](https://github.com/widelands/widelands/blob/master/data/scripting/messages.lua#L186). So the cause is something different.

Noordfrees:
Fixes #3935

### DIFF
--- a/data/scripting/messages.lua
+++ b/data/scripting/messages.lua
@@ -183,10 +183,13 @@ function new_objectives(...)
       sum = sum + obj.number
    end
 
+   local td = __TEXTDOMAIN
+   set_textdomain("widelands")
    local objectives_header = _"New Objective"
    if (sum > 1) then
       objectives_header = _"New Objectives"
    end
+   set_textdomain(td)
 
    return
       div("width=100%",
--- a/data/scripting/messages.lua
+++ b/data/scripting/messages.lua
@@ -46,6 +46,8 @@ end
 --    :type heading: :class:`string`
 --
 function send_to_all(text, heading)
+   local td = __TEXTDOMAIN
+   set_textdomain("widelands")
    for idx,plr in ipairs(game.players) do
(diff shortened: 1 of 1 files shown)


frankystone (approved):
It works... :-)

Is the intermediate storage of the textdomain needed?

There is also another _-makro in function send_to_all() which is used for some scenarios. Must this also be changed?

Noordfrees:
Well spotted, fixed now :)
The macro `__TEXTDOMAIN` always reflects the current textdomain. `set_textdomain(x)` is only a synonym for `__TEXTDOMAIN = x`. So yes this needs to be cached to reset it to the calling scenario's textdomain.

frankystone:
Thanks.

Is there somewhere a documentation for all these textdomains? There should be one, imho.

Since we had some of such issues in the past, can we have a test, or a script, for testing this?

Noordfrees:
As far as I know the `__TEXTDOMAIN` variable is documented nowhere because there is normally no need to use it. Also the internal TD handling will get more complicated in the addons branch. (btw I only knew how to fix this because I partly rewrote the backend there just a week ago ;) ). 
Which textdomain covers which stuff is documented [in the wiki.](https://www.widelands.org/wiki/TranslatingWidelands/#matching-the-resource-to-the-po-files-in-the-source-code)
It would be nice to have a testcase that covers some i18n-related potential problems, but that would be fragile because translations can be changed at any time…

PR #3944, opened 2020-06-14
use detect_revision.py on appveyor

stonerl:
The version reported by appveyor builds is now the same as the one from 
macOS or Linux. This avoids confusing version mismatches in 
multiplayer games.


Noordfrees:
LGTM :) Can someone with a Windows system please check whether the version is displayed correctly in the main menu?

hessenfarmer:
In game the following string is displayed r24638[1fac1a3@HEAD] ( same as in the installer artifact) I don't know whether this is the same as in linux versions

Noordfrees (approved):
Thanks for testing :)
Since AppVeyor checks out a merge commit rather than the PR branch the `HEAD` string is what I'd expect

PR #3946, opened 2020-06-17
use ubuntu 20.04 for code formatting

stonerl:


.github/workflows/format.yaml
```diff
@@ -15,7 +15,7 @@ jobs:
   run_formatting:
     name: Run fix_formatting.py
     # clang-format has broken dependencies on ubuntu-latest
```
Noordfrees: This comment can be removed now

Noordfrees (commented):
LGTM :)
Although there are no semantics changes I'd rather wait with merging this until just after the release though because of the large diff.

Noordfrees (approved):

PR #3948, opened 2020-06-19
Load replay savegames in multiplayer screens

### LINKED ISSUE #3945: Savegame from replay not shown in loading menu
A savegame that was created from a replay is not shown in the list od loadable savegames (singleplayer or multiplayer). However it is loadable with doubleclick directly from the filesystem. 

see also https://www.widelands.org/forum/post/33535/

to reproduce: choose a replay , save , open e.g sngleplayer-load game , Go to .widelands/save and see the file created , doubleclick 

Version was amazons-7461
75189d8c2182065599fa0f72ee56737f4f644ed6

jmoerschbach:
fixes #3945

It is a workaround though in my opinion.

### DIFF
--- a/src/wui/savegameloader.cc
+++ b/src/wui/savegameloader.cc
@@ -205,6 +205,7 @@ MultiPlayerLoader::MultiPlayerLoader(Widelands::Game& game) : SavegameLoader(gam
 }
 
 bool MultiPlayerLoader::is_valid_gametype(const SavegameData& gamedata) const {
+	// TODO(jmoerschbach): workaround to be able to load replays in multiplayer loading screen
 	return gamedata.is_multiplayer() || gamedata.is_replay();
 }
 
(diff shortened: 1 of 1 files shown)


Noordfrees:
This works fine in multiplayer. In singleplayer however one always has to play as player 1 which is undesirable if one was playing with a different player number before. I'd suggest removing the fix for SP and offering `kReplay` savegames only in MP; that's what it used to be like. Also please add a comment that this is a workaround.

frankystone:
> This works fine in multiplayer. In singleplayer however one always has to play as player 1 which is undesirable if one was playing with a different player number before. 

Yes, this was the same behavior before. I thought there was a bug report for fixing this, but i can't find it.

> I'd suggest removing the fix for SP and offering `kReplay` savegames only in MP; that's what it used to be like. Also please add a comment that this is a workaround.

I would call this suggestion a workaround. If we don't want the player to create save games from replays, the correct approach would be to remove the corresponding button.

In my opinion we should leave it as is (with this branch), and add a bug report to fix the issue in single player mode.

Noordfrees (approved):

PR #3958, opened 2020-07-01
Sort directories & files when running optimize_pngs.py

gunchleoc:
This will make it easier to continue if a file fails.

### DIFF
--- a/utils/optimize_pngs.py
+++ b/utils/optimize_pngs.py
@@ -92,8 +92,9 @@ def collect_pngs(d, prefix):
     """
     pngs = []
     for root, dirs, files in os.walk(d):
+        dirs.sort()
         pngs.extend(os.path.join(root, f)
-                    for f in files
+                    for f in sorted(files)
                     if os.path.splitext(f)[-1].lower() == '.png'
                     and (not prefix or f.startswith(prefix)))
 


TinoM (approved):
This is a very minimal change and in my opinion can be merged.

utils/optimize_pngs.py
```diff
@@ -92,8 +92,9 @@ def collect_pngs(d, prefix):
     """
     pngs = []
     for root, dirs, files in os.walk(d):
+        dirs.sort()
```
TinoM, line 95: we do not use "dirs", remove this line and replace dir in line 94 with  _

TinoM (approved):
Fine with me.

utils/optimize_pngs.py
gunchleoc: I don't think that `_.sort()` is easier for humans to understand than `dirs.sort()`.

The reason for this PR is that it's easier for me to manage running this for release. I run this in multiple terminal tabs using the `-d` option, because it takes a long time for the whol case and I have multiple CPU threads I can use.

TinoM: no, sorry, i did not explain correctly:
the walk functions returns the tuple (root, dirs, files). But the list of directories = dirs is never used! So no need for sorting it.

```
for root,_,files in os.walk(d):
     for f in sorted(files)
     if os......
```

This is what i meant.

gunchleoc: This would remove the sorting by directory that I wanted to have too.

TinoM: Hm, but it does make a difference :(. Sometimes i do not understand Python.

Just ignore my comment ;).

gunchleoc: No problem. Thanks for the review!

Niektory: Beep bop reference bot: https://docs.python.org/3/library/os.html#os.walk

(...) the caller can modify the _dirnames_ list in-place (perhaps using `del` or slice assignment), and `walk()` will only recurse into the subdirectories whose names remain in _dirnames_; this can be used to prune the search, impose a specific order of visiting, or even to inform `walk()` about directories the caller creates or renames before it resumes `walk()` again.

PR #3961, opened 2020-07-03
Replace unknown programs with "work" while saveloading

### LINKED ISSUE #3671: make savegame loading less succeptible for changes in lua
Currently savegame loading easily breaks if lua working programs are changed. 

This should be fixed in a way that if a program in the savegame is not found in the productionsite definition anymore just reinitialize the program working state of the productionsite.    

Perhaps some other parts of saveloading could be hardened this way

gunchleoc:
Alternative proposal to https://github.com/widelands/widelands/pull/3960/

Untested.


Noordfrees (approved):
Tested and working, code LGTM

gunchleoc:
Thanks for testing and for giving me the right idea with your branch :)

PR #3963, opened 2020-07-04
Add some scenario savegame compabitility

gunchleoc:
Savegames for testing are in https://www.widelands.org/forum/topic/4826/?page=3#post-33758


src/logic/map_objects/map_object_program.cc
```diff
@@ -119,16 +119,28 @@ MapObjectProgram::AnimationParameters MapObjectProgram::parse_act_animate(
...
+	// TODO(GunChleoc): Savegame compabitility. Remove after Build 21.
+	if (arguments.size() == 3) {
+		filepath = arguments.at(0) + "/" + arguments.at(1);
+		priority = read_positive(arguments.at(2));
+	} else {
+		if (arguments.size() < 1 || arguments.size() > 2) {
+			throw GameDataError("Usage: playsound=<sound_dir/sound_name> [priority]");
+		}
+		filepath = arguments.at(0);
+		priority = read_positive(arguments.at(1));
```
Noordfrees: This will fail if `priority` is omitted, use the `arguments.size() == 2 ? …` check here and remove that check below

gunchleoc: Correct. Should be fixed now.

Noordfrees (approved):
Code LGTM, tested with the savegames from the forum

gunchleoc:
Failure is a linker timeout. Will merge.

PR #3967, opened 2020-07-07
fixed the helptext_note for frisian beekeeper

### LINKED ISSUE #3965: Untranslated string in beekeeper's helptext
This one's a little easier: in data/tribes/buildings/productionsites/frisians/beekeepers_house/helptexts.lua, there's no pgettext around the helptext note.

hessenfarmer:
fixes #3965

### DIFF
--- a/data/tribes/buildings/productionsites/frisians/beekeepers_house/helptexts.lua
+++ b/data/tribes/buildings/productionsites/frisians/beekeepers_house/helptexts.lua
@@ -15,7 +15,7 @@ end
 
 function building_helptext_note()
    -- TRANSLATORS: Note helptext for a building
-   return "Needs medium-sized fields (barley, wheat, reed, corn or blackroot) or bushes (berry bushes or grapevines) nearby."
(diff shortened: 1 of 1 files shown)


Noordfrees (approved):
LGTM. However I'd wait with merging until after the release because delaying that to wait for one helptext note to be translated would be ridiculous; and if we merge now, all complete translations will be shown as only 99% complete, eek.

hessenfarmer:
Well Gun Chleoc marked it urgent so I provided a fix. I believe we could fix most of the languages by making a final announcement.
Translators were so active they should respond quickly. So I think the 100% will quickly be achieved again. This might even help gaining attention of danish and dutch for the last string fix

at the end it is your decision

gunchleoc:
We can pull 100% matches out of the Transifex translation memory for this string, so I'm merging. Thanks for fixing!

gunchleoc:
P.S. I see, no translation markup at all.

Shall we give translators an extension of a week? The only thing that can crash Widelands are printf placeholders, and we can validate those.

Issue #3968, opened 2020-07-08
shipconstruction_frisians texture broken in Tribal Encyclopedia


gunchleoc:
When looking at the Frisian "Ship Under Construction" entry in the Tribal Encyclopedia, the texture looks broken (see screenshot).

I tried the following fixes that did not work:

- Revert the texture to a state prior to running `optimize_pngs.py`
- Resave the texture with GIMP.

@Noordfrees Would you like to take a look? I'd like this fixed for Build 21 if possible.

![shipconstruction_frisians](https://user-images.githubusercontent.com/4095570/86898062-1d107500-c100-11ea-94af-391dd07709cb.png)

Noordfrees:
??? 
This is how it looks for me:
![wl-shipcfri](https://user-images.githubusercontent.com/48293446/86967765-5cc07680-c16b-11ea-8a54-ca5224c098c4.png)
No idea what might be the cause…

frankystone:
This is what it looks like for me as well. I guess this is caused by frisians shipconstruction images are  spritemaps, but the shipconstruction sites images of other tribes are not.

Niektory:
I linked a fix. If you want a workaround for build 21 without C++ changes, set `representative_frame = 0` in `data/tribes/immovables/shipconstruction_frisians/init.lua`.

PR #3970, opened 2020-07-08
Fix UI bugs

gunchleoc:
I went through the UI with a fine-tooth comb to hunt for bugs

* Fix diverse gettext markup bugs where translations weren't being displayed
* Fix textures for stones resources
* Remove pngrewrite from png optimization tools, because it broke the textures
* Deactivate translation catalog updates
* 1 new translatable string in loader UI

Notes:
* I accidentally committed the current gd translation, which I used for testing. Does no harm though.
* Follow-up issue: https://github.com/widelands/widelands/issues/3968
* Once this is merged, I'll use the scenarios in the tribes_filtering branch to see if pngrewrite broke any other textures
* Shall we push the new string and then give translators another week? We can use the validation project to check for any true positive printf and ngettext errors once the deadline has passed. This won't slow down our release schedule.


Noordfrees:
> Shall we push the new string and then give translators another week? We can use the validation project to check for any true positive printf and ngettext errors once the deadline has passed. This won't slow down our release schedule.

Since you insist we could give translations another week or so, or more if it fits for you. You'll have to take over all the release-related stuff I was intending to do though – I planned to have the release done by the 18th, and I will have not much time from the 23rd on.

gunchleoc:
As I said, this will not delay the release. We can safely pull translations right until we package the release candidate, because we can run the validator to ensure that Widelands won't crash due to translations.

Noordfrees (commented):
The RC is scheduled for the day after tomorrow so it is a bit less than a week… but OK fine :)
Code LGTM, not tested yet

Noordfrees (approved):
Tested. Thanks for fixing :)

PR #3972, opened 2020-07-09
Fix spritesheet representative image display

### LINKED ISSUE #3968: shipconstruction_frisians texture broken in Tribal Encyclopedia

Niektory:
Fixes #3968 

The bug was caused by source and destination rects being switched.


Noordfrees (approved):
LGTM, tested and working. Thanks for fixing :)

gunchleoc:
Thanks! I always get the source and destination rectangles confused.

PR #3976, opened 2020-07-12
Fixes two desyncs in New World & Discovery

### LINKED ISSUE #3975: New World and Discovery desync
**To Reproduce**
1. Start a game with either New World or Discovery starting condition (on the Fellowships map for example).
2. Watch the replay of that game.

or

1. Start a multiplayer game with 2+ players with either New World or Discovery starting condition.

In both cases it desyncs immediately or after a few seconds.

**Version:**
 - OS: Manjaro Linux
 - Widelands Version: r24671[577b9e8@master] (Debug); Build21-RC1 (Debug) (9f487550abd1c66b99e9f53f0136cda150b0a52c)

Noordfrees:
Fixes #3975 
IMHO we should backport this fix into the release branch as this could get ugly…

### DIFF
--- a/data/scripting/starting_conditions.lua
+++ b/data/scripting/starting_conditions.lua
@@ -14,21 +14,22 @@ include "scripting/ui.lua"
 -- =======================================================================
 
 -- RST
--- .. function:: sleep_then_goto(sleeptime, field)
+-- .. function:: sleep_then_goto(player, sleeptime, field)
 --
 --    Wait for the given number of milliseconds, then center the
 --    interactive player's map view on the given Field.
 --
 --    This function returns immediately after calling.
 --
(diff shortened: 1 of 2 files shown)


gunchleoc (approved):
Agreed.

There's a nice git command for porting the changes to the other branch: https://git-scm.com/docs/git-cherry-pick

Noordfrees:
Personally I prefer to add `.patch` to the PR's URL, copypasting to a text file and using `git apply ../patch`. That's a nice thing about git, there are often many ways of doing something so it suits everyone's preference. Applied the patch to the branch like this.

gunchleoc:
Good.

I'll merge this now so we'll have it in master too.

PR #3982, opened 2020-07-13
Build a static linked executable with msys2 (Appveyor&Github Workflow)

TinoM:
This adds a new OPTION_BUILD_WINSTATIC to our CMake Script to allow a standalone widelands.exe.
This is currently only tested with the MSYS2 environment and targeted for our Appveyor builds.
Pro: We do not have to include all the dll files in the installer script (changes in dependecies always broke this).
Con: Currently only Glew possible, no GLBindings (Bug in the GLBinding-Package for MSYS2, no static libs possible).


hessenfarmer:
Converted to draft as it seems to be hard to get the OSX part of travis to work

TinoM:
Ah, while the windows binaries are now ok, travis still fails due to issues with ICU...

hessenfarmer:
so we are back at OSX trying to do the impossible (linking crt0 statically, which is not possible on OSX), this seems to be a dependency to one of our changes in identifying packages, but I don't know which one.

TinoM:
But why? We do only enable static linking on the windows platform...

TinoM:
Currently trying to simplify our CMAKe files in another branch without any static linking.
Maybe this will help...

hessenfarmer:
> 
> 
> But why? We do only enable static linking on the windows platform...

If I only knew that ;-)     
I am a bit lost about this as I can't do the trial an error locally.     
I already pinged @stonerl via the forum but he has very limited capacity to look into this.   
Only other person to ask about the specifics for OSX and cmake for this case might be @SirVer

hessenfarmer:
I went through the logs and the following line of the failed build indicates that sdl2main is linked statically causing probabaly the atttempt to link crt0 statically as well.    
-- Found SDL2main: /usr/local/lib/libSDL2main.a (found version "2.0.12")

hessenfarmer:
Hooray, it works on OSX
it was a stupid missing If statement that led to link boost always statically

hessenfarmer:
ready for review again, let's see how it does on travis.
However this would need to be checked for compatibility with the out of box trusty installation.    
maybe we might not have compatibility there

gunchleoc:
The required CMake version is too high for my Trusty

`CMake 3.7 or higher is required.  You are running version 3.5.1`

If I lower it to 3.5, it won't find ICU or SDL2.

gunchleoc:
`preload.lua` will become obsolete with https://github.com/widelands/widelands/pull/4134, but in current master it is still needed.

TinoM:
Hmm, do we really want to hardcode the libs in the cmake file? I liked the parameter setting via the cmake  call in the Travis/Appveyor/GitHub workflow scripts better.
E.g. I have installed msys2 in a slightly different path and have my own build.ps1 script. Now I have to also maintain my own cmakelist.txt? Or do manual -DVar=value parameters override the cmake default values?

TinoM:
Sorry, closing this was of course a mistake!

hessenfarmer:
Regarding the variables I found it difficult to always type the whole string into the commandline of my msys so I think it to be a good idea to control everything with the buildstatic option. As they are currently declared as internal this should according tom the cmake doc imply that they are forced. maybe you could just try this in your setup. However if we make them of type STRING rather then INTERNAL they should not overwrite -D decalrations from the commandline.  So if you could try this if it works for you this should be the way to go.

gunchleoc:
@TinoM All green now. Are you OK with the libs coding, or do you wish to have further changes made before merging?

TinoM:
No, let's get this in now...

TinoM:
Hmm, first i got an server error on merging, then it wanted to sync master, and now i am not sure what out best practice is: Merge commit, Squash or rebase?
@gunchleoc or @hessenfarmer could you please merge? Thx!

PS:
There seems to be an internal variable for Cmake which defines the file suffix the library search will prefer. On windows it is ".a" or ".dll.a" for static/dynamic. Maybe this is a way to force somehow that cmake defaults to the static libs. I'll try that in another branch at some time...

gunchleoc:
I always squash and ensure all the important information is in the commit message. This will give us 1 commit per pull request, just like we used to have with bazaar. It also makes it easier to write the changelog if we have less commits.

I am keeping the `static_link` branch in the AppVeyor definition, so if you want to do a follow-up, just reuse the branch name.

PR #3983, opened 2020-07-13
Parse production outputs

gunchleoc:
* The building "outputs" table was originally added for the Encyclopedia, and is now being used by the AI as well. We can now easily get that information from the productionsite programs, uncluttering the Lua configuration.
* 2 separate commits for C++ code and Lua configuration.

@tibor95 There's a difference for the AI here - Trainingsites are no longer classified as having `BuildingAttribute::kBarracks`. Does this fix an AI bug or add one?

### DIFF
--- a/src/website/map_object_info.cc
+++ b/src/website/map_object_info.cc
@@ -101,10 +101,15 @@ void write_buildings(const TribeDescr& tribe, EditorGameBase& egbase, FileSystem
 				}
 			}
 			// Produces workers
-			if (productionsite->output_worker_types().size() > 0) {
+			if (productionsite->output_worker_types().size() > 0
+				|| productionsite->type() == Widelands::MapObjectType::TRAININGSITE) {
 				JSON::Array* json_workers_array = json_building->add_array("produced_workers");
 				for (DescriptionIndex worker_index : productionsite->output_worker_types()) {
 					json_workers_array->add_empty(tribe.get_worker_descr(worker_index)->name());
+				}
+				// Trainingsites
+				if (productionsite->type() == Widelands::MapObjectType::TRAININGSITE) {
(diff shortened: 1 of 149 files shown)


tibor95:
Trainingsites as Barracks??? This was never my intention, and it worked that way? I need to look into code

EDIT:
Trainingsites are not considered as barracks, because only productionsite can be, also we have this check in the AI code:
if (count_buildings_with_attribute(BuildingAttribute::kBarracks) != 1) {

hessenfarmer:
As far as I can remember TRAININGSITES never had that Attribute

gunchleoc:
The attribute is set then the building has the tribe's soldier as an outpout. It was configured as such in the Lua files for use in the Encyclopedia.

ETA: I have now seen that there is a check for productionsite type higher up, so never mind. No impact on the AI.

hessenfarmer:
I did some testing. Ai seems to behave "normal", which is expectable as the help seems to be unchanged as well.

src/website/map_object_info.cc
```diff
@@ -101,11 +101,16 @@ void write_buildings(const TribeDescr& tribe, EditorGameBase& egbase, FileSystem
 				}
 			}
 			// Produces workers
-			if (productionsite->output_worker_types().size() > 0) {
+			if (productionsite->output_worker_types().size() > 0
+				|| productionsite->type() == Widelands::MapObjectType::TRAININGSITE) {
```
hessenfarmer, line 105: What is the purpose of this line above and the addition below?

gunchleoc: The training sites used to have "soldier" in their output table, but they don't technically produce soldiers. So, I'm adding a manual remedy for the website encyclopedia here.

Noordfrees (approved):
Code LGTM, tested briefly and working

gunchleoc:
Thanks for the review!

PR #3988, opened 2020-07-16
Immediately-upgraded constructionsites should not burn

### LINKED ISSUE #3816: Immediately-upgraded construction sites should not burn when destroyed
When creating a new building and immediately upgrading it, the site should not burn down then turn to ashes if it is destroyed. First-level buildings under construction simply disappear when destroyed, which helps save time if a site was placed mistakenly.

Presumably this is a left-over from a time before immediate upgrades were added, as upgrade construction sites were only available once a building was completed; in this case the burning cycle seems to be appropriate.

1. Start a game
1. Place a new multi-level building and then immediately upgrade it 
1. Immediately destroy the construction site
1. Observe that the site burns down and leaves ashes behind, before eventually clearing

**Expected behavior**
Site is immediately removed

**Version:**
Ubuntu 19.10
Widelands r24525[b2e6914@master]

Noordfrees:
Fixes #3816

### DIFF
--- a/src/logic/map_objects/tribes/constructionsite.cc
+++ b/src/logic/map_objects/tribes/constructionsite.cc
@@ -510,10 +510,10 @@ Construction sites only burn if some of the work has been completed.
 ===============
 */
 bool ConstructionSite::burn_on_destroy() {
-	if (work_completed_ >= work_steps_)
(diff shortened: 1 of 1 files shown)


gunchleoc (approved):
Thanks for the fix :)

Issue #3989, opened 2020-07-17
Extract information from MapObjectProgam


gunchleoc:
Extract available information from Worker, Immovable and Productionsite programs to

* Simplify the Lua API
* Automate some AI hints
* Enhance the Encyclopedia

Branches:

- [x]  Deduce building outputs from worker and productionsite programs and remove `outputs` table https://github.com/gunchleoc/widelands/tree/parse-production-outputs PR: https://github.com/widelands/widelands/pull/3983

- [x]  Deduce whether a building collects a ware from map https://github.com/widelands/widelands/pull/4618
- [x]  Deduce AI mining hints https://github.com/widelands/widelands/pull/4489
- [x]  Deduce which resources/bobs/immovables are collected from/placed on the map https://github.com/widelands/widelands/pull/4618

frankystone:
Maybe related: #3956

hessenfarmer:
closing this as all Tasks mentioned above have been implemented.
resetted the version impelemented to v1.0

PR #3999, opened 2020-07-19
Deduce whether a building collects a ware from map

gunchleoc:
Re. https://github.com/widelands/widelands/issues/3989

`collected_map_resources` is a set now, because the Frisian hunter also creates fur. This might affect the AI scoring for that building a bit. @tibor95

### DIFF
--- a/data/tribes/buildings/productionsites/atlanteans/fishers_house/init.lua
+++ b/data/tribes/buildings/productionsites/atlanteans/fishers_house/init.lua
@@ -25,7 +25,6 @@ tribes:new_productionsite_type {
    },
 
    aihints = {
-      collects_ware_from_map = "fish",
       needs_water = true,
       basic_amount = 1,
       prohibited_till = 540,
--- a/data/tribes/buildings/productionsites/atlanteans/hunters_house/init.lua
+++ b/data/tribes/buildings/productionsites/atlanteans/hunters_house/init.lua
@@ -26,7 +26,6 @@ tribes:new_productionsite_type {
    },
 
    aihints = {
-      collects_ware_from_map = "meat",
       prohibited_till = 380
    },
 
--- a/data/tribes/buildings/productionsites/atlanteans/quarry/init.lua
+++ b/data/tribes/buildings/productionsites/atlanteans/quarry/init.lua
@@ -24,9 +24,7 @@ tribes:new_productionsite_type {
       },
    },
 
-   aihints = {
-      collects_ware_from_map = "granite"
-   },
+   aihints = {},
 
    working_positions = {
       atlanteans_stonecutter = 1
--- a/data/tribes/buildings/productionsites/atlanteans/well/init.lua
+++ b/data/tribes/buildings/productionsites/atlanteans/well/init.lua
@@ -32,7 +32,6 @@ tribes:new_productionsite_type {
 
    aihints = {
       basic_amount = 1,
-      collects_ware_from_map = "water"
    },
 
    working_positions = {
--- a/data/tribes/buildings/productionsites/atlanteans/woodcutters_house/init.lua
+++ b/data/tribes/buildings/productionsites/atlanteans/woodcutters_house/init.lua
@@ -25,9 +25,7 @@ tribes:new_productionsite_type {
       },
    },
 
-   aihints = {
-      collects_ware_from_map = "log",
-   },
+   aihints = {},
 
    working_positions = {
       atlanteans_woodcutter = 1
--- a/data/tribes/buildings/productionsites/barbarians/fishers_hut/init.lua
+++ b/data/tribes/buildings/productionsites/barbarians/fishers_hut/init.lua
@@ -32,7 +32,6 @@ tribes:new_productionsite_type {
    },
 
    aihints = {
-      collects_ware_from_map = "fish",
       needs_water = true,
       prohibited_till = 490,
       very_weak_ai_limit = 1,
--- a/data/tribes/buildings/productionsites/barbarians/hunters_hut/init.lua
+++ b/data/tribes/buildings/productionsites/barbarians/hunters_hut/init.lua
@@ -34,7 +34,6 @@ tribes:new_productionsite_type {
    },
 
    aihints = {
-      collects_ware_from_map = "meat",
       prohibited_till = 480,
       basic_amount = 1,
       very_weak_ai_limit = 1,
--- a/data/tribes/buildings/productionsites/barbarians/lumberjacks_hut/init.lua
+++ b/data/tribes/buildings/productionsites/barbarians/lumberjacks_hut/init.lua
@@ -32,7 +32,6 @@ tribes:new_productionsite_type {
    },
 
    aihints = {
-      collects_ware_from_map = "log",
       basic_amount = 1,
    },
 
--- a/data/tribes/buildings/productionsites/barbarians/quarry/init.lua
+++ b/data/tribes/buildings/productionsites/barbarians/quarry/init.lua
@@ -31,9 +31,7 @@ tribes:new_productionsite_type {
       },
    },
 
-   aihints = {
-      collects_ware_from_map = "granite"
-   },
+   aihints = {},
 
    working_positions = {
       barbarians_stonemason = 1
--- a/data/tribes/buildings/productionsites/barbarians/well/init.lua
+++ b/data/tribes/buildings/productionsites/barbarians/well/init.lua
@@ -32,7 +32,6 @@ tribes:new_productionsite_type {
    },
 
    aihints = {
-      collects_ware_from_map = "water",
       basic_amount = 1
    },
 
--- a/data/tribes/buildings/productionsites/empire/fishers_house/init.lua
+++ b/data/tribes/buildings/productionsites/empire/fishers_house/init.lua
@@ -26,7 +26,6 @@ tribes:new_productionsite_type {
    },
 
    aihints = {
-      collects_ware_from_map = "fish",
       needs_water = true,
       prohibited_till = 410
    },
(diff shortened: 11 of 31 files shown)


data/tribes/buildings/productionsites/atlanteans/fishers_house/init.lua
```diff
@@ -25,7 +25,6 @@ tribes:new_productionsite_type {
    },
 
    aihints = {
-      collects_ware_from_map = "fish",
```
tibor95, line 28: I am lost now, how will fishers and other special buildings identified without this hint?

tibor95: Ou, I got it, this indeed looks like very good change...

tibor95:
I would not worry about frisian hunter, the impact on scoring is desired and ok

gunchleoc:
Thanks for having a look. Can this branch go in?

tibor95:
Yes, the code looks OK, for me, did somebody run it in DEBUG build to see if all asserts are satisfied? And for all tribes....

gunchleoc:
No assert failures. Thanks for the review!

hessenfarmer:
Well, 
this might not work for amazons this way, as there we have a building that collects fish and meat which might fail some assertions.   
Furthermore this adds a collect wares from map hint to every farm and the empire vineyard as well which hadn't been the case before. need to check the code whether this will lead to wrong decisions.

gunchleoc:
> this might not work for amazons this way, as there we have a building that collects fish and meat which might fail some assertions.

I think the AI needs fixing for that anyway. We won't want too many tribe-specific exceptions - it's not maintainable.

One could theoretically want to build a tribe that has no fisher or no hunter, and the AI can't handle it because of the assertions.

> Furthermore this adds a collect wares from map hint to every farm and the empire vineyard as well which hadn't been the case before. need to check the code whether this will lead to wrong decisions.

Why were those hints left out? Was there a specific reason, or were they simply overlooked.

tibor95:
no hunter or no fisher should be doable, after changing the assertion this might work fine

as for farms as collectors - one of differences here is that fishes or trees are or are not in vicinity (yes, trees are 'renewable'), but farms are creating own fields and can do it everywhere where some free space is.

gunchleoc:
> as for farms as collectors - one of differences here is that fishes or trees are or are not in vicinity (yes, trees are 'renewable'), but farms are creating own fields and can do it everywhere where some free space is.

We will be able to do something smarter when I finish https://github.com/gunchleoc/widelands/tree/collection-help because then we will know which immovables a building creates and collects, as well as who is supporting whom without need for extra AI hints.

I'll set this to draft now and let's hold off on it until the other branch is more advanced - this branch might become obsolete then.

hessenfarmer:
Well, 
I had another look into the code. and this branch definitly would lead to problems in the AI.
Reason is the evolution of the collects_ware_from map hint. first it was invented to combine some bools which were used to identify well, lumberjack and quarry. With the frisians it was enhanced to identify other special buildings by this rather then by their output cause the identification by output was ambigous (this cured a bug in empire (piggery) as well).
By this it evolved to form relations between buildings together with the "supports production of" hint.

We need to keep this relations as well as the identification of the "basic" special buildings. In the current version this relation would be broken for frisians_aqua_farm and frisians_claypit as well as for frisians_claypit and frisians_charcoal_burners_house (although it was not done for the latter correctly.     

for me there are 2 solutions: 
1. change the supports_production_of hint to reflect the building name(s) rather then taking the detour with the collects ware hint. This would give a tribe designer good control of the buildings that need to be placed to each other.    
2. We could probably deduce the same information from the declarations of working area overlaps. "true" means a relation between the 2 buildings is needed.

tibor95:
were bo.production_hints affected? Because I see a line here:

} else if (!bo.production_hints.empty() && !bo.is(BuildingAttribute::kSupportingProducer)) {

I hope it is evaluated the same...

hessenfarmer:
> 
> 
> were bo.production_hints affected? Because I see a line here:
> 
> } else if (!bo.production_hints.empty() && !bo.is(BuildingAttribute::kSupportingProducer)) {
> 
> I hope it is evaluated the same...

They are not affected in this branch, as they result from the "supports production of" AI hint. However their counterpart to identify a relation was the "collects_ware_from_map" hint which is changed by this PR so we need a way to reinstate the relation for the AI if we fix these issues

tibor95:
I went through check_building_necessity() where is a lot of logic and GA magic to to say if a building should be built. Bottom line is that buildings are split into these categories:

- trainingsites
- mines
- recruitment sites
- Hunter/fisher/well
- lumberjack
- ranger
- quarry
- supporters (no input / no output ?)
- barracks
- mines
- producers (max_needed_preciousness is calculated)
- shipyard

As long as you create a building that match a category no problem. But we have a lot of buildings (in frisian tribe) that combine producer and supporter. For now such buildings are treated as producers but it it not optimal. Obviously new category is needed and new code in check_building_necessity() as well.

And clear definition of supporting producers:
- have inputs
- have outputs
- produce something on map that is collected by somebody else
- question: do they need some input to produce this "something on map"
- question: this something is then stored in warehouse or is transparently used by some nearby producer?

This check_building_necessity() is too overgrown :(

Also if we have kSupportingProducer we might have also kPureSupporter and these too be exclusive...

tibor95:
Also is a gamekeeper a supporter? Because animals roam so the mutual positions of gamekeeper and hunter does not matter. Maybe we should note down and specify what categories of productionsites we have

hessenfarmer:
well, I believe we had a clear definition of supporter. this is defined as having the AI hint "supports production of"    
for most of the buildings this is being a "pure" supporter. For this it doesn't matter if there are inputs necessary or not.
Necessity for these buildings is then only dependent by the neededness of the supported ware and the productivity of the producers which are supported. While determining the location of such a building it needs to be adjacent to the supported building. There is no direct attribute for them as they are identified by !bo.production_hints(empty)     

If the supporter produces something as well necessity is also dependent of the neededness of the ware he produces that is why we identify these buildings separatly. 

regasrding their necessity they are different but regarding their location they are similar. so defini8ng them exclusivly might add redundancy in some parts of the code. although defining them with a kAtrribute might increase readability and understandability of the code.   

I don't think this checks to be overgrown, in fact the more circumstances and necessities are evaluated the better the decision of the AI would be.

for the hunter vs gamekeeper part their relation is not that strong but is there, cause they are more effective if put in vicinity so for me yes the gamekeeper is a supporter of the hunter, although the hunter doesn't require strictly a supporter

tibor95:
There is max_needed_preciousness - calculated for outputs of a production site - this defines if building is needed
Supporting sites does not have this - other logic is applied.

Supporting productionsites can have 0 max_needed_preciousness but still be needed due to necessity of supported ware. Try to review the AI code to see how would AI cope with supporting productionsites whose outputs are not needed, but whose "supporting" is needed...

hessenfarmer:
> 
> 
> There is max_needed_preciousness - calculated for outputs of a production site - this defines if building is needed
> Supporting sites does not have this - other logic is applied.

Just checked lines 5538 ff for this logic. This could use some improvement probably. Next time I do a "watch the AI" test game I'll keep a look on the AI building such sites.    

> 
> Supporting productionsites can have 0 max_needed_preciousness but still be needed due to necessity of supported ware. Try to review the AI code to see how would AI cope with supporting productionsites whose outputs are not needed, but whose "supporting" is needed...       
    
Should be no problem as the get_stocklevel function calculates the lowest value of outputs and supported wares to determine whether the ware is needed.

tibor95:
Check for line (yes, typo is there):
// Calulate preciousness
and below - only outputs are used for max_needed_preciousness, are not they?

tibor95:
Theoretically, supporting productionsites should be evaluated two times: as a producer and as a supporter, and higher result would be used. Or have AI somehow combine results...

hessenfarmer:
> 
> 
> Check for line (yes, typo is there):
> // Calulate preciousness
> and below - only outputs are used for max_needed_preciousness, are not they?

yes, but it is faked to a positive value in line 5200 and therefore the neededness is calculated by lines 5703ff. and there I added some Weights to take stocklevel into account while determining neededness. 
Anyhow I need to check in an AI only game whehter there are shortfalls of these buildings to improve this part of the code probably.  
But this discussion went fr from the topic of this branch I believe (although I enjoy everytime to discuss AI code with you)

tibor95:
OK, then. 
This function is indeed full of ad-hoc hacks and hard to follow.
And is almost 1000 lines long, far beyond any "good practice" rule. I am thinking about splitting it into multiple functions, basically GA stuff for each building category into separate function.

gunchleoc:
> for me there are 2 solutions:
> 
> 1. change the supports_production_of hint to reflect the building name(s) rather then taking the detour with the collects ware hint. This would give a tribe designer good control of the buildings that need to be placed to each other.
> 
> 2. We could probably deduce the same information from the declarations of working area overlaps. "true" means a relation between the 2 buildings is needed.

This is exactly what I'm working on in https://github.com/gunchleoc/widelands/tree/collection-help

Thank you for your analysis; I'll use that information for designing it over there.

tibor95:
AI should know for such buildings what product they are supporting (the output of supported building) and subsequently be able to figure out if we have enough of this product on stocks.

hessenfarmer:
> 
> 
> AI should know for such buildings what product they are supporting (the output of supported building) and subsequently be able to figure out if we have enough of this product on stocks.

Yes for sure, but in my proposal this could be obtained from the supported buildings output wares.    
Refactoring defaultai.cc is definitly a good idea, however might be a huge task which needs proper testing.

hessenfarmer:
as this is explicitly superseeded by the linked PR I'd vote to close this

PR #4000, opened 2020-07-19
Replace buildferry with createbob

gunchleoc:
One less program for modders to learn ;)

### DIFF
--- a/src/logic/map_objects/tribes/worker_program.cc
+++ b/src/logic/map_objects/tribes/worker_program.cc
@@ -695,8 +695,8 @@ createbob
    :arg string bob_name: The bob type to add to the selection. Specify as many bob
       types as you want.
 
-   Adds a bob (usually an animal) to the map at the worker's current location.
-   Randomly select from the list of ``bob_name``. Example::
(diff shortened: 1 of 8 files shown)


src/logic/map_objects/tribes/worker_program.cc
```diff
@@ -722,7 +730,8 @@ buildferry
 ^^^^^^^^^^
 .. function:: buildferry
 
-   Adds a new instance of this tribe's ferry to the map at the worker's current location. Example::
+   **DEPRECATED** Adds a new instance of this tribe's ferry to the map at the worker's current
+location. Example::
```
Noordfrees: Just replace the whole comment and usage example with a line "DEPRECATED, use `createbob=TRIBENAME_ferry`" instead

gunchleoc: Done

Noordfrees (approved):
LGTM, not tested

gunchleoc:
I have tested all ferries and the gamekeeper.

PR #4001, opened 2020-07-20
Switch to new version scheme

### LINKED ISSUE #3979: Ambiguous version number
I noticed you made some changes to the version naming.
https://github.com/widelands/widelands/blob/build21-rc1/WL_RELEASE#L1

```
$ grep -H '' */WL_RELEASE
widelands-build15-src/WL_RELEASE:build-15
widelands-build16-src/WL_RELEASE:build-16
widelands-build17-src/WL_RELEASE:build-17
widelands-build18-src/WL_RELEASE:build-18
widelands-build19-repack-src/WL_RELEASE:build-19
widelands-build19-src-gcc7/WL_RELEASE:build-19
widelands-build19-src/WL_RELEASE:build-19
widelands-build20/WL_RELEASE:build-20
widelands-build21-rc1/WL_RELEASE:Build21-RC1
```

First of all, lowercase is newer than uppercase.
```
$ rpmdev-vercmp "b" "B"
b > B
```

Secondly, the lack of a dash means that we have a newer version.
```
$ rpmdev-vercmp "build-21" "build21"
build-21 < build21
```

So, in the end, we have this:
```
$ rpmdev-vercmp "build-21" "Build21"
build-21 > Build21
```

In addition, you used `-` to specify release candidate, but according to **Linux Standard Base** and **Freedesktop** you should use `~`. Both [LSB](https://refspecs.linuxfoundation.org/lsb.shtml) and [XDG](https://www.freedesktop.org/wiki/) ([AppSteam](https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html]) rely on the [RPM](http://rpm.org/) standard.

http://refspecs.linux-foundation.org/LSB_5.0.0/LSB-Core-generic/LSB-Core-generic/swinstall.html

> Applications shall either be packaged in the RPM packaging format as defined in this specification, or supply an installer which is LSB conforming (for example, calls LSB commands and utilities).

> Note: Supplying an RPM format package is encouraged because it makes systems easier to manage. This specification does not require the implementation to use RPM as the package manager; it only specifies the format of the package file and requires that implementations must have some method of installing conforming packages.

https://www.freedesktop.org/software/appstream/docs/api/appstream-as-utils.html#as-utils-compare-versions

> The version compare algorithm is also used by RPM.

https://github.com/rpm-software-management/rpm/blob/rpm-4.15.1-release/lib/rpmvercmp.c#L12-L137

wrong:
```
$ rpmdev-vercmp "Build21-RC1" "Build21"
Build21-RC1 > Build21
```

valid:
```
$ rpmdev-vercmp "Build21~RC1" "Build21"
Build21~RC1 < Build21
```

Since you are changing the versioning scheme, you might consider removing the entire `Build` part.

```
$ rpmdev-vercmp "Build21" "21"
Build21 < 21
```

The version starting with the number will always be newer than any old one.

For development versions, you can always use `~`.

```
$ rpmdev-vercmp "21~rc1" "21"
21~rc1 < 21
```

```
$ rpmdev-vercmp "21~r24630" "21~rc1"
21~r24630 < 21~rc1
```

```
$ rpmdev-vercmp "21~rc1" "21~rc2"
21~rc1 < 21~rc2
```

And if you really need to:
```
$ rpmdev-vercmp "21~rc1+r24668" "21~rc1"
21~rc1+r24668 > 21~rc1
```

This should fix the whole mess with package versioning in various distributions.

For example, in Debian, the version string must (or at least should) start with a number.

https://www.debian.org/doc/debian-policy/ch-controlfields.html#version

> The upstream_version may contain only alphanumerics 6 and the characters . + - ~ (full stop, plus, hyphen, tilde) and should start with a digit.

Please keep in mind that nowadays different package formats can coexist in a one system. It is also possible that one type of package replaces another. If I good remember, this was the case with Chromium in Ubuntu 20.04 LTS, where `Snap` replaced `DEB` (during the upgrade process).

This is how versioning of Widelands packages in various distributions looks:
https://pkgs.org/search/?q=widelands

Distribution: Fedora (rpm)
Version string: `0`
Release string: `0.77.build20.fc32`
Link: https://fedora.pkgs.org/32/fedora-x86_64/widelands-0-0.77.build20.fc32.x86_64.rpm.html

Distribution: openSUSE (rpm)
Version string: `build20`
Release string: `lp152.2.2`
Link: https://opensuse.pkgs.org/15.2/opensuse-oss-x86_64/widelands-build20-lp152.2.2.x86_64.rpm.html

Distribution: OpenMandriva (rpm)
Version string: `b20`
Release string: `1`
Link: https://openmandriva.pkgs.org/4.1/openmandriva-unsupported-release-x86_64/widelands-b20-1-omv4001.x86_64.rpm.html

Distribution: Mageia (rpm)
Version string: `b20`
Release string: `1.mga7`
Link: https://mageia.pkgs.org/7.1/mageia-core-release-x86_64/widelands-b20-1.mga7.x86_64.rpm.html

Distribution: PCLinuxOS (rpm)
Version string: `b20`
Release string: `3pclos2020`
Link: https://pclinuxos.pkgs.org/rolling/pclinuxos-x86_64/widelands-b20-3pclos2020.x86_64.rpm.html

Distribution: ALT Linux (rpm)
Version string: `0`
Release string: `alt7_0.69.build19`
Link: https://altlinux.pkgs.org/p9/classic-x86_64/widelands-0-alt7_0.69.build19.x86_64.rpm.html

Distribution: Debian (deb)
Version string: `20`
Release string: `2+b1`
Link: https://debian.pkgs.org/11/debian-main-amd64/widelands_20-2+b1_amd64.deb.html

Distribution: Ubuntu (deb)
Version string: `20`
Release string: `2build2`
Link: https://ubuntu.pkgs.org/20.04/ubuntu-universe-amd64/widelands_20-2build2_amd64.deb.html

Distribution: Solus (eopkg)
Version string: `19`
Release string: `15-1`
Link: https://solus.pkgs.org/rolling/solus-shannon-x86_64/widelands-19-15-1-x86_64.eopkg.html

Distribution: Gentoo (ebuild)
Version string: `0.20`
Release string: `r1`
Link: https://packages.gentoo.org/packages/games-strategy/widelands

Distribution: Slackware (tarball)
Version string: `build19`
Release string: `1_slonly`
Link: https://slackware.pkgs.org/14.2/slackonly-i586/widelands-build19-i586-1_slonly.txz.html

Distribution: Arch (tarball)
Version string: `20`
Release string: `9`
Link: https://archlinux.pkgs.org/rolling/archlinux-community-x86_64/widelands-20-9-x86_64.pkg.tar.zst.html

Distribution: FreeBSD (tarball)
Version string: `b20`
Release string: `6`
Link: https://freebsd.pkgs.org/13/freebsd-amd64/widelands-b20_6.txz.html

Distribution: NetBSD (tarball)
Version string: `20nb3`
Release string: -
Link: https://netbsd.pkgs.org/9.0/netbsd-amd64/widelands-20nb3.tgz.html

As you can see, the same game version can have a different package version: `b20`, `build20`, `0`, `0.20`, `20`, `20nb3`, etc. I wouldn't be surprised if someone called it `20.0.0` (or `20.0` in short), `0.20.0` or `0.0.20`.
Please keep in mind that `20` != `20.0` and `20.0` != `20.0.0`.

```
$ rpmdev-vercmp "20" "20.0"
20 < 20.0
```

```
$ rpmdev-vercmp "20.0" "20.0.0"
20.0 < 20.0.0
```

Anyway, I suggest either sticking to the old format or (if you want to change it anyway) switch to a more clear and legible scheme.
So, in my opinion it should be either `build-21~rc1` or `21.0.0~rc1` (or `21~rc1` if you really want to). Personally, I prefer the second one.
As for the GitHub version, the `21.0.0-rc1` is an equivalent of `21.0.0~rc1` in RPM. If you want to add the revision number on GitHub, you can use something like this: `21.0.0-rc1+20200711`. It is an equivalent of `21.0.0~rc1-20200711` in RPM. On GitHub, don't forget about the `v` at the beginning to produce the correct [tag version](https://semver.org/#is-v123-a-semantic-version): `v21.0.0-rc1+20200711`.
See [Semantic Versioning](https://semver.org/) for details.

Noordfrees:
Fixes #3979 
When this is merged I'll update the github milestone and the daily PPA, and I'll also write a summary of the versioning rules on the ReleasingWidelands wiki page.

Note that the string in the lower-right corner of the main menu will look like `1.0~git24681[2efff03@master](Release)` – the *actual version* is only the part `1.0~git24681`! The `[2efff03@master]` is additional branch info and `(Release)` is the build configuration.


utils/detect_revision.py
```diff
@@ -11,6 +11,12 @@
 import re
 import string
 
+# The name of the next stable release. We use semantic version strings in the form
+# "1.0~..." where "1.0" is the next release. Update this string after every release,
+# and also if we have to rename the next target during the release cycle.
+# See https://github.com/widelands/widelands/issues/3979 for a full discussion.
+next_stable_version = '1.0'
```
gunchleoc: How about we put this in a configuration file, just like we do with `WL_RELEASE`?

Noordfrees: Done

gunchleoc (approved):

utils/win32/widelands.rc.cmake
```diff
@@ -21,9 +21,10 @@
 
 A ICON MOVEABLE PURE LOADONCALL DISCARDABLE "@CMAKE_CURRENT_SOURCE_DIR@/../utils/win32/WL.ico"
 
-//
-// TO CHANGE VERSION INFORMATION, EDIT  OPTIONS BELOW...
-//
+// To change version information, edit FILEVERSION and PRODUCTVERSION. Keep these two identical.
+// Version "A,B,C,D" translates to version "0x000A000B.0x000C000D",
+// e.g. for version "1.2" we need "0,1,0,2".
```
hessenfarmer: I beleive this is wrong,
the current value of this attribute in the windows file properties is 0.21.0.1,
so the 4 16 bit Integers are all separated by a point in the file properties.
this leads to 1.2 translates to 1,2,0,0 Which in fact will be 1.2.0.0 in the properties I believe

hessenfarmer (changes requested):

utils/win32/widelands.rc.cmake
Noordfrees: Okay changed the explanation. I have no windows system available now to check this so I had to rely on microsoft's poor documentation

Noordfrees:
@hessenfarmer can this go in now?

hessenfarmer (approved):

frankystone:
Shouldn't the new version scheme be announced on the website?

Packagers should also be informed, but i think this should be done on the next release.

Noordfrees:
No need IMHO. We can announce it on the website at the next feature freeze, and inform the packagers in the 1.0 release announcement. As for now this would be only of theoretical interest.

gunchleoc:
Yep, let's not inform packagers right now. It would only confuse them as they are currently still packaging Build 21.

PR #4003, opened 2020-07-20
Fix ware/worker economy mismatch in set_target_quantity

gunchleoc:


Noordfrees (approved):
Code LGTM, not tested

PR #4006, opened 2020-07-20
Stack lua textdomains in C++

### LINKED ISSUE #3964: Lua scripts cross-manipulate textdomains
In tutorial 3 (seafaring) there's a window where the translation is not shown, alerting the player that the second ship is ready and shipbuilding should be stopped. It corresponds to the function 'shipyard_production', which is the only function in tutorial03_seafaring.wmf/scripting/texts.lua.

Noordfrees:
Fixes #3964 
Fixes partly #4020


gunchleoc:
Code LGTM, but we should test saveloading. The new functions probably need to be added to the bottom of

`static const char* kPersistentGlobals[]`.

Not tested yet.

gunchleoc (approved):
Done a bit of testing now. Let's have it :)

Issue #3691, opened 2020-02-16
Merge "transform" and "grow" immovable programs


gunchleoc:
The "transform" and "grow" immovable programs both do the same, except for some differences in parameters.

1. [ ] Replace "grow" with "transform", making sure that tha parameters work
2. [ ] Add saveloading compatibility code for games

hessenfarmer:
I believe the difference is that transform is not taking terrain affinity into account while grow does. So this is a big difference in result.

gunchleoc:
I did indeed miss that. I'm wondering whether we could make this difference a parameter. I'd be fine with not changing anything though.

hessenfarmer:
I would vote for keeping things as they are. Mainly for keepin the telling program names which make reading the immovable programs easier.

Noordfrees:
Currently trees have programs like `"remove=21","grow=…"`. IMHO it would be better if this was unified into one program that accepts an optional param for a remove chance and autodetects whether the immovable has terrain affinity. The remove chance should then be weighted with the terrain goodness, and instead of increasing the growth time for bad terrains remove the immovable with a chance that is the higher the worse the terrain matches the preference.

gunchleoc:
The tribe immovables also use `remove` (ashes, harvested fields). Shipconstruction uses `transform` without a separate `remove`. So, if we want to combine `grow` with `remove`, how about combining `transform` and `remove` too? If `transform` is called without an object, simply remove the object.

gunchleoc:
The merging of the programs has been rejected to give Lua coders more control.

PR #4007, opened 2020-07-21
Add "success:" prefix to number parameters in "grow", "remove" and "transform" immovable programs

gunchleoc:
Re. https://github.com/widelands/widelands/issues/3691

@Noordfrees @hessenfarmer How do you like the new syntax?

I think in the long run, it would be good to give all numbers a `parameter_name:` prefix for readability. If you agree, I can collect them and open a new issue.


src/logic/map_objects/immovable_program.h
```diff
@@ -66,39 +66,18 @@ struct ImmovableProgram : public MapObjectProgram {
...
-	/// Like ActTransform but the probability is determined by the suitability.
-	struct ActGrow : public Action {
-		ActGrow(std::vector<std::string>& arguments, const ImmovableDescr&);
-		void execute(Game&, Immovable&) const override;
-
-	private:
-		std::string type_name;
+		bool removal_wanted;
+		uint8_t transform_probability;
+		uint8_t removal_probability;
```
Noordfrees: Apply `_` to the variable names for consistency

src/logic/map_objects/immovable_program.cc
```diff
@@ -110,94 +201,152 @@ void ImmovableProgram::ActPlaySound::execute(Game& game, Immovable& immovable) c
...
-		   f, type_name, owner_type, owner, nullptr /* former_building_descr */);
+	const bool will_transform =
+	   descr.has_terrain_affinity() && transform_probability == 0 ?
+	      (game.logic_rand() % TerrainAffinity::kPrecisionFactor) <
+	         probability_to_grow(descr.terrain_affinity(), f, map, game.world().terrains()) :
+	      transform_probability == 0 || game.logic_rand() % 256 < transform_probability;
+
+	if (will_transform) {
+		if (bob) {
+			game.create_ship(f, type_name, player);
```
Noordfrees: No need to delete the immovable afterwards?

src/logic/map_objects/immovable_program.cc
```diff
@@ -110,94 +201,152 @@ void ImmovableProgram::ActPlaySound::execute(Game& game, Immovable& immovable) c
...
+** DEPRECATED** Use ``transform=<immovable_name>`` instead.
+*/
+
+/* RST
+
+remove
+------
+** DEPRECATED** Use ``transform=remove:[<chance>]`` or ``transform=`` instead.
+*/
+// TODO(GunChleoc): Savegame compatibility, remove after Build 22
```
Noordfrees: There will never be a version called Build22 ;)
Use `v1.0` instead.

Noordfrees (commented):
Syntaxwise this looks good to me. A few comments inline.

Noordfrees:
Is there a reason why this is not marked as Fixes the linked issue?

hessenfarmer:
Well in any case +1 for the "parameter:number" part as it increases readability and understandability.   
   
For the other change to combine grow, transform, remove into one I am not that convinced as it decreases readability and is less intuitive for me, but this might be due to habitual patterns on my side

gunchleoc:
> Is there a reason why this is not marked as Fixes the linked issue?

There is a proposal in the issue to determine `remove` from the terrain affinity. I have not addressed that here, since probability calculus is not my strong suit.

src/logic/map_objects/immovable_program.cc
gunchleoc: Thanks for catching this bug

src/logic/map_objects/immovable_program.cc
```diff
@@ -110,94 +201,152 @@ void ImmovableProgram::ActPlaySound::execute(Game& game, Immovable& immovable) c
...
 
-	if ((game.logic_rand() % TerrainAffinity::kPrecisionFactor) <
-	    probability_to_grow(descr.terrain_affinity(), f, map, game.world().terrains())) {
-		MapObjectDescr::OwnerType owner_type = descr.owner_type();
-		Player* owner = immovable.get_owner();
+	const bool will_transform =
+	   descr.has_terrain_affinity() && transform_probability_ == 0 ?
+	      (game.logic_rand() % TerrainAffinity::kPrecisionFactor) <
+	         probability_to_grow(descr.terrain_affinity(), f, map, game.world().terrains()) :
+	      transform_probability_ == 0 || game.logic_rand() % 256 < transform_probability_;
```
Noordfrees: > There is a proposal in the issue to determine remove from the terrain affinity. I have not addressed that here, since probability calculus is not my strong suit.

Then what about something like:
```
bool will_transform = true;
if (descr.has_terrain_affinity()) {
	if ((game.logic_rand() % TerrainAffinity::kPrecisionFactor) >= probability_to_grow(descr.terrain_affinity(), f, map, game.world().terrains()) {
		if (transform_probability_) { will_transform = false; }
		else { immovable.remove(game); return; }
	}
} else {
	will_transform = transform_probability_ == 0 || game.logic_rand() % 256 < transform_probability_;
}
```

gunchleoc: What I mean is that currently, the removal probability is 100% hard-coded and always the same for a tree species, independent of terrain type. If we change this, it will need rebalancing. So, I'd rather not touch it in this branch.

gunchleoc:
@hessenfarmer Would it be more readable to you if we added a blank line above the terrain affinity assignment? Like this:

```
world:new_immovable_type{
   name = "cirrus_wasteland_sapling",
   -- TRANSLATORS: This is a fictitious tree. Be creative if you want.
   descname = _ "Cirrus Tree (Sapling)",
   editor_category = "trees_wasteland",
   size = "small",
   attributes = { "tree_sapling" },

   terrain_affinity = terrain_affinity,
   programs = {
      program = {
         "animate=idle 61000",
         "transform=cirrus_wasteland_pole remove:44",
      },
   },
   spritesheets = {
      idle = {
         directory = dirname,
         basename = "sapling",
         fps = 8,
         frames = 5,
         rows = 3,
         columns = 2,
         hotspot = { 6, 13 }
      }
   },
}
```

hessenfarmer:
With regard to readability I meant it was clear what each program did in the past due to their names. Now we have one program name with a lot of options and it is not really clear at first glance what the line of code does. 
Especially the transform= as replacement for remove is hard to understand for beginners I believe. Furthermore I don't see the gain in this change, as we almost have the same amount of lines and letters.

gunchleoc:
How about keeping "remove" as it is and just merging the "grow" and "transform" programs?

This is how the programs are being used in master right now:

- When "grow" is called, the immovable always has terrain affinity
- When "transform" is called, the immovable only has terrain affinity when it's a deadtree, where an explicit chance is given.

So, we're basically defining twice that an immovable has terrain affinity. This makes master is easier to read than this branch, but harder to write.

hessenfarmer:
> How about keeping "remove" as it is and just merging the "grow" and "transform" programs?    
     
that would make it even more complicated, I believe.

> 
> This is how the programs are being used in master right now:
> 
>     * When "grow" is called, the immovable always has terrain affinity     
    
From my understanding it is the opposite. in the grow program Terrain affinity is taken into account and needs to be defined with the immovable.     

> 
>     * When "transform" is called, the immovable only has terrain affinity when it's a deadtree, where an explicit chance is given.

Transform doesn't care about terrain affinity may it be defined or not. It is only dependent on probability given or will work always. e.g. wheat fields or similar.
so the tree programs are defined as follows:
first each tree has a chance to die due to undefined reason (like with acid rain in realtiy f.e.) this chance is bigger with young trees as older trees are more robust. 
then it might grow to the next step if the terrain is allowing to do so, else this cycle is repeated in the next season.
an old tree doesn't grow anymore it just has a chance to die eventually and become a deadtree else it seeds new trees in the vicinity (terrain affinity is needed for that.    
    
> 
> So, we're basically defining twice that an immovable has terrain affinity. This makes master is easier to read than this branch, but harder to write.     
   
I disagree we are defining it only once with the immovable, but we decide to take it into account or not in the programs.
So how would your proposal deal with the old tree that needs to be transformed without terrain affinity although he has it defined.

gunchleoc:
I wasn't talking bout how the C++ code is doing things, I was talking about how the 2 programs are being used in practice in the Lua files:

* Whenever an immovable has terrain affinity, the person defining the immovable used the grow program, except for when transforming to deadtree, where an explicit probability is given
* Whenever an immovable does not have terrain affinity, the person defining the immovable used the transform program

So, which one of the 2 is being used can be deduced from that, which is what I did in this branch.

hessenfarmer:
Well, 
I understand what this branch does. And for the current state of the programs you are probably right in your deductions. However when I started to look into the lua codes I appreciated when there were telling names and I appreciated when the chain of events was clear, so I would prefer the old style of the programs. Furthermore it would give the Lua authors more control and more flexibility in what they can do with immovables. And they got a good error message if they were attempting something not possible (like using grow without having terrain_affinity defined).
to give an example you might look into the definition of the amazon special trees. I decided to keep the grow mechanism and therefore I needed to define the 3 trees for all types of world to make them growable on each map. It would have been more easy to just make them grow equally every where with using the transform command rather then grow, and just use terrain affinity for the placement decisions.  For me this was an active decision to go this way, after this branch this decision would not be possible anymore.   
Anyhow I just wanted to deliver my opinion and I am not requesting changes, so feel free to do what you deem suitable.

gunchleoc:
So, let's reinstate the "grow" program but keep the new "remove" option and get rid of the "remove" program?

You are obviously feeling strongly about this and since you work a lot more with the Lua files than I do, I am willing to go with your preference on this.

hessenfarmer:
> 
> 
> So, let's reinstate the "grow" program but keep the new "remove" option and get rid of the "remove" program?

that would result in having 2 different syntaxes for remove:
transform=   and
transform=remove:30

I would prefer
remove=success:30  and 
remove=

> 
> You are obviously feeling strongly about this and since you work a lot more with the Lua files than I do, I am willing to go with your preference on this.

Well, I have a strong opinion on this format changes but not strong feelings. I am just expressing my strong opinion what would be best to understand for a lua scripter (assuming me being representative for this of course ;-))
Where I do have strong feelings is removing options from the scripters as this limits creativity in scenario / tribe design. (in fact we managed to implement really cool features due to this freedom in scripting,although it needed misuse of some of them and there are rooms for improvement)

gunchleoc:
I'd be interested in what these "misuses" are and if the syntax can be changed to something more fitting.

gunchleoc:
@hessenfarmer @Noordfrees This is ready for review again

Noordfrees (approved):

hessenfarmer:
> 
> 
> I'd be interested in what these "misuses" are and if the syntax can be changed to something more fitting.

Well, I can't remember all of them, but mainly we used some AI hints in a way not expected from the description to make the Ai handle new tribes correctly (frisians, amazons).   
But I believe we will come to them when you continue your work.

hessenfarmer (approved):
Much more happy with this now.     
On small thing that could be changed in a follow on branch perhaps:
the probabilities are not given in Percent but in x/254 % may be we should make them read like real percentages with having a notation ,like 0.5 , 1.0, 1.5, as this would be more readable from the lua persepctive and should not change the code much.

gunchleoc:
@hessenfarmer great minds think alike; I opened an issue today to discuss this https://github.com/widelands/widelands/issues/4047

hessenfarmer:
> 
> 
> > I'd be interested in what these "misuses" are and if the syntax can be changed to something more fitting.
> 
> Well, I can't remember all of them, but mainly we used some AI hints in a way not expected from the description to make the Ai handle new tribes correctly (frisians, amazons).
> But I believe we will come to them when you continue your work.

Ok Now I found one while trying to implement the removal of outputs for amazons. look at the rare tree plantation code of them where I documented an ugly hack. 
Edit: I tested it now. The change in the outputs branch definitly broke AI handling of amazons
Edit2: but this doesn't affect this branch

gunchleoc:
I have run the test manually and compiling was green before the master merge, so merging this now.

PR #4028, opened 2020-07-28
Fix map compatibility for MapObject Loaders

### LINKED ISSUE #4026: Incompatible Scenarios
Several of our scenarios are no longer compatible since #3980: 

```
 Reading Map Objects ... WidelandsMapLoader::load_map_complete() for 'A Place to Call Home' took 46ms
Fatal exception: map objects: 

UnhandledVersionError: Deeses Speel is mit eener ollen Versioon vun Widelands sekert worden un kann nich mehr laden, of dat is eene neje Versioon, wat Widelands noch nich mit klaarword.

Packet Name: MapObjectPacket
Saved Version: 1
Current Version: 2
FATAL ERROR - game crashed. Attempting emergency save.
lastserial: 0
```

We need to either reinstate the MapObjectPacket compatibility code, or go through **all** maps and resave them with build21 if necessary.

@gunchleoc

gunchleoc:
Fixes https://github.com/widelands/widelands/issues/4026

The repetition of the packet version constant is a bit ugly, but we'd get circular dependencies otherwise.

### DIFF
--- a/src/logic/map_objects/tribes/building_settings.cc
+++ b/src/logic/map_objects/tribes/building_settings.cc
@@ -145,7 +145,7 @@ constexpr uint8_t kCurrentPacketVersion = 2;
 constexpr uint8_t kCurrentPacketVersionMilitarysite = 1;
 constexpr uint8_t kCurrentPacketVersionProductionsite = 2;
 constexpr uint8_t kCurrentPacketVersionTrainingsite = 1;
-constexpr uint8_t kCurrentPacketVersionWarehouse = 2;
+constexpr uint8_t kCurrentPacketVersionWarehouseAndExpedition = 2;
 
 // static
 BuildingSettings* BuildingSettings::load(const Game& game,
@@ -330,7 +330,7 @@ void WarehouseSettings::read(const Game& game,
 	BuildingSettings::read(game, fr, tribes_lookup_table);
 	try {
 		const uint8_t packet_version = fr.unsigned_8();
-		if (packet_version == kCurrentPacketVersionWarehouse) {
+		if (packet_version == kCurrentPacketVersionWarehouseAndExpedition) {
 			launch_expedition = fr.unsigned_8();
 			const uint32_t nr_wares = fr.unsigned_32();
 			const uint32_t nr_workers = fr.unsigned_32();
@@ -356,7 +356,7 @@ void WarehouseSettings::read(const Game& game,
 			}
 		} else {
 			throw UnhandledVersionError(
-			   "WarehouseSettings", packet_version, kCurrentPacketVersionWarehouse);
+			   "WarehouseSettings", packet_version, kCurrentPacketVersionWarehouseAndExpedition);
 		}
 	} catch (const WException& e) {
 		throw GameDataError("WarehouseSettings: %s", e.what());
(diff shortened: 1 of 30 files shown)


Noordfrees:
Having the packet version repeated in three places doesn't look good to me. How about refactoring the code to have one central file for ALL packet version constants which is included by every file that checks for a version?

gunchleoc:
I'd prefer to keep it local - this is the only use case for needing it centrally. How about 1 separate header file just for kMapObjectPacketVersion, and leave the rest as is?

Noordfrees:
`kCurrentPacketVersionWarehouse` is also being used in multiple places (MapBuildingdataPacket and ExpeditionBootstrap saveloading code). So +1 for moving only these two version constants into their own header.

src/logic/map_objects/tribes/building_settings.cc
```diff
@@ -145,7 +145,7 @@ constexpr uint8_t kCurrentPacketVersion = 2;
 constexpr uint8_t kCurrentPacketVersionMilitarysite = 1;
 constexpr uint8_t kCurrentPacketVersionProductionsite = 2;
 constexpr uint8_t kCurrentPacketVersionTrainingsite = 1;
-constexpr uint8_t kCurrentPacketVersionWarehouse = 2;
+constexpr uint8_t kCurrentPacketVersionWarehouseAndExpedition = 2;
```
Noordfrees: Why did you rename this? This is only for building _settings_, no expedition-related stuff here.

src/logic/map_objects/tribes/ship.cc
```diff
@@ -1143,51 +1144,57 @@ const Bob::Task* Ship::Loader::get_task(const std::string& name) {
...
-		expedition_->seen_port_buildspaces.clear();
-		uint8_t numofports = fr.unsigned_8();
-		for (uint8_t i = 0; i < numofports; ++i) {
-			expedition_->seen_port_buildspaces.push_back(read_coords_32(&fr));
-		}
-		// Swimability of the directions
-		for (uint8_t i = 0; i < LAST_DIRECTION; ++i) {
-			expedition_->swimmable[i] = (fr.unsigned_8() == 1);
+void Ship::Loader::load(FileRead& fr, uint8_t packet_version) {
+	if (packet_version == kCurrentPacketVersionMapObject) {
```
Noordfrees: `packet_version` here refers to the _ship_ packet version. There is no need to compare `kCurrentPacketVersionMapObject` with anything here as far as I can see.

src/logic/map_objects/tribes/building_settings.cc
gunchleoc: Because of 
```
	// Keep this synchronized with kCurrentPacketVersionWarehouse in MapBuildingDataPacket!!
	static const uint16_t kCurrentPacketVersion = 8;
```

So it gives people a hint about where it is being used - when loading a warehouse and when loading an expedition.

src/logic/map_objects/tribes/ship.cc
gunchleoc: Yep, that's a bug, and it broke the test suite too. Fixed.

src/logic/map_objects/tribes/building_settings.cc
Noordfrees: That hint refers to the relation between a Warehouse and an ExpeditionBootstrap. But this file is for BuildingSettings which are used only by ConstructionSites. These are neither synchronized with the warehouse/expedition packets nor meant to be.

gunchleoc: Duh, now I get it. My brain was in the new header file.

Noordfrees (approved):

PR #4033, opened 2020-07-30
Fix nodecaps calculation error

### LINKED ISSUE #1482: Impassable terrain enables buildings
See the attached screenshot.
In the first situation, no mine can be built on the mountain, and no building on the mountain meadow.
If you add some impassable terrain (as in the second part), buildings suddenly become buildable.
This is highliy unlogical (if I cannot walk on some terrain, there is no space to build a building) and works with snow and lava (both where you cannot build flags on the border), but not with water or swamp (where this is possible).

 * date created: 2013-11-04T09:11:12Z
 * owner: wl-zocker
 * the launchpad url was https://bugs.launchpad.net/bugs/1247760

Noordfrees:
Fixes #1482


gunchleoc:
This will undo the work done in https://github.com/widelands/widelands/commit/9673135b723090945522708eee7ea5a024a8d5d6 so I'm not sure that it's the proper fix.

Noordfrees:
Had a look through the bzr history and found that this commit was introduced to fix [this bug](https://bugs.launchpad.net/widelands/+bug/566970). Placing a milsite close to unreachable terrains with a passage only to the west, northwest and/or northeast (but no direct connection from there to the building's flag) may make it impossible for attacking soldiers to reach the milsite, which is why that commit imposed some restrictions for that case. These have the linked issue as a side-effect. My solution is to make those restrictions even stricter – with my fix, it will not be possible to place a building in any location where it was until now buildable. So I believe that the old bug will not regress.

src/logic/map.cc
```diff
@@ -1733,31 +1733,25 @@ int Map::calc_buildsize(const EditorGameBase& egbase,
...
  */
 bool Map::is_cycle_connected(const FCoords& start, uint32_t length, const WalkingDir* dirs) const {
+	if (!(start.field->get_caps() & MOVECAPS_WALK)) {
+		return false;
+	}
 	FCoords f = start;
-	bool prev_walkable = start.field->get_caps() & MOVECAPS_WALK;
-	uint32_t alternations = 0;
-
 	for (uint32_t i = 0; i < length; ++i) {
```
gunchleoc: Since we're touching this function anyway, we could modernize the loop and get rid of the "length" function argument.

```
for (WalkingDir* dir : dirs) {
    f = get_neighbour(f, dir);
```

Noordfrees: My compiler doesn't like it since the array is passed as a pointer function argument, so I converted it to vector instead.

gunchleoc:
I have tried to recreate the conditions of the original savegame.

This change means we lose building plots, and it might make maps with tight building spaces like Fjords unplayable.

Comparison screenshot - we lose 3 big building plots here, and the passage can no longer be guarded by a Fortress:

nodecaps

Multiplayer savegame for testing: [nodecaps.wgf.zip](https://github.com/widelands/widelands/files/5005693/nodecaps.wgf.zip)

Noordfrees:
Unfortunately we have only three options here:
- Always enable the building spot → regression of the old bug where players can seal themselves off completely
- Always disable it → some maps may need to be rechecked
- Keep it as it is → the bug that *less* walkable terrain results in *more* building spaces.

I would consider the second option the lesser evil

gunchleoc:
Your argument has convinced me. We will need to audit all official maps though to ensure that they are still playable, e.g. we should have some big building plots on those screenshots so that players can use a big militarysite strategically. The forum community could help.

gunchleoc:
Another idea: Would it be worth the effort to add a switch somehow, in order not to make unofficial maps unplayable? This couldn't be just the usual packet version, because then saveloading a game would always switch to the new algorithm.

Noordfrees:
We could use the "official" map tag to reinstate the current behaviour. I think e.g. Long Long Way might have some trouble after this change. The question is when we want to get rid of this "legacy compatibility" code though.

gunchleoc:
Using the "official" tag would mean that unofficial maps can never benefit from the new feature. It would need to be a separate switch if we want to support an option. If we don't want the extra code, the official maps should be curated before we merge this branch, or shortly thereafter.

frankystone:
The bug report is from year 2013 and i can't remember other people mentioned this issue in the past. At least i can't find any forum comments when searching the (launchpad) bug number.

There might be some maps on the homepage which would be affected by this solution and may be unplayable after this solution has been applied.

In my opinion we should ask in the forum what solution the players prefer. WorldSavior has knowledge of all maps, imho, and other players as well. In this difficult decision we should take their meaning into account.

tppq:
I think that this is a good change. Since build 22 is far in the horizon, making a modification which might introduce a need to rework maps is not that bad.

The ancient sun of fire is one of my favorites, I vote for adjusting it instead of dropping.

> Comparison screenshot - we lose 3 big building plots here, and the passage can no longer be guarded by a Fortress

I think that one should *not* be able to build fortresses to these corridors. The corridors are close to each other. Nearby player can accidentally block entry altogether. This is bad, if the nearby player is in same team and controller by AI (=it is not even possible to ask it to replace the fortress with something smaller).

> There might be some maps on the homepage which would be affected by this solution

Authors should be able to upload new revisions of maps. Re-uploading with version number in name is a bit silly. Unrelated, though.

gunchleoc:
> Authors should be able to upload new revisions of maps. Re-uploading with version number in name is a bit silly. Unrelated, though.

Yep, that's https://github.com/widelands/widelands-website/issues/196

Noordfrees:
I went through all the official maps with the editor and fixed the two that needed fixing. So can this be merged now?

gunchleoc:
Yes, please :)

gunchleoc (approved):

PR #4036, opened 2020-07-30
Use player fields for terrain rendering

### LINKED ISSUE #1470: Player field structure does not remember terrain
When the water in the Atlantean campaign begins to raise, the land under trees etc. vanishes. However, the trees are still visible. I am not sure if this is only a matter of visibility: In areas once discovered and now covered by the fog of war, the terrain changes, but the immovables seem not to be updated (similar to normal games). I think the most logic way would be not to update the terrain as well.
The issue affects immovables (trees, stones, the ruins), but also flags from the road.

 * date created: 2013-09-20T11:45:51Z
 * owner: wl-zocker
 * the launchpad url was https://bugs.launchpad.net/bugs/1228096

Noordfrees:
Fixes #1470


gunchleoc:
This doesn't fix the issue - see the tree at the mouse cursor that should be gone, and the grass next to it.

persistent-tree

Noordfrees:
Um, why should this tree become invisible? The field is not supposed to change at all once it is no longer seen

gunchleoc:
The point of the screenshot is that we have an immovable surrounded by water. Whether the terrain was changed to water while it was still visible or afterwards doesn't change that fact - either both should be gone, or none.

Noordfrees:
This is the correct behaviour: First, the node with the bush becomes unseen. So the bush remains "visible" forever. The terrains to the north and/or northwest of the node are still land at that time. Only later, these terrains also become water (which is revealed because the node NW of the bush is seen). That's why all terrains are water but the bush is not removed from sight when it is truly deleted.

gunchleoc (approved):
Duh, now I get it. Fix confirmed :)

PR #4041, opened 2020-07-31
Fix some cert warnings with clang-tidy

gunchleoc:
* [cert-err34-c]: Replace `atoi` with `boost::lexical_cast` and  `const string` literals with `constexpr const char* const`
* [cert-msc30-c], [cert-msc50-cpp]: Our uses of `std::rand` are not security-related, so marked them as `NOLINT`


src/graphic/style_manager.cc
```diff
@@ -93,7 +93,8 @@ void StyleManager::init() {
 	scrollbarstyles_.clear();
 
 	LuaInterface lua;
-	std::unique_ptr<LuaTable> table(lua.run_script(kTemplateDir + "init.lua"));
+	std::unique_ptr<LuaTable> table(
+	   lua.run_script((boost::format("%s%s") % kTemplateDir % "init.lua").str()));
```
Noordfrees:
```suggestion
	   lua.run_script((boost::format("%1%init.lua") % kTemplateDir).str()));
```

src/network/network_player_settings_backend.cc
```diff
@@ -39,7 +39,8 @@ void NetworkPlayerSettingsBackend::set_player_ai(PlayerSlot id,
 		ComputerPlayer::ImplementationVector::const_iterator it = impls.begin();
 		if (impls.size() > 1) {
 			do {
-				size_t random = (std::rand() % impls.size());  // Choose a random AI
+				// Choose a random AI
+				size_t random = (std::rand() % impls.size());  // NOLINT
```
Noordfrees:
```suggestion
				const size_t random = (std::rand() % impls.size());  // NOLINT
```

src/ui_basic/fullscreen_window.cc
```diff
@@ -35,7 +37,7 @@ namespace UI {
 
 FullscreenWindow::FullscreenWindow()
    : UI::Panel(nullptr, 0, 0, g_gr->get_xres(), g_gr->get_yres()),
-     background_image_(kTemplateDir + "fsmenu/background.png") {
+     background_image_((boost::format("%s%s") % kTemplateDir % "fsmenu/background.png").str()) {
```
Noordfrees:
```suggestion
     background_image_((boost::format("%1%fsmenu/background.png") % kTemplateDir).str()) {
```

src/ui_basic/window.cc
```diff
@@ -106,11 +106,15 @@ Window::Window(Panel* const parent,
...
-     pic_background_(g_gr->images().get(kTemplateDir + "wui/background.png")),
+     pic_lborder_(
+        g_gr->images().get((boost::format("%s%s") % kTemplateDir % "wui/left.png").str())),
+     pic_rborder_(
+        g_gr->images().get((boost::format("%s%s") % kTemplateDir % "wui/right.png").str())),
+     pic_top_(g_gr->images().get((boost::format("%s%s") % kTemplateDir % "wui/top.png").str())),
+     pic_bottom_(
+        g_gr->images().get((boost::format("%s%s") % kTemplateDir % "wui/bottom.png").str())),
+     pic_background_(
+        g_gr->images().get((boost::format("%s%s") % kTemplateDir % "wui/background.png").str())),
```
Noordfrees: Use `window_image_path("...")` for those as well

src/ui_basic/window.cc
```diff
@@ -64,12 +70,6 @@ constexpr int16_t kVerticalBorderThingyHeight = 20;
 constexpr int16_t kVerticalBorderMiddleLength =
    (kVerticalBorderTotalLength - 2 * kVerticalBorderThingyHeight);
 
-static const std::string kWindowImageClose = kTemplateDir + "wui/window_close.png";
-static const std::string kWindowImagePinned = kTemplateDir + "wui/window_unpin.png";
-static const std::string kWindowImageUnpinned = kTemplateDir + "wui/window_pin.png";
-static const std::string kWindowImageMinimize = kTemplateDir + "wui/window_minimize.png";
-static const std::string kWindowImageMaximize = kTemplateDir + "wui/window_maximize.png";
```
Noordfrees, line 71: Reintroduce them as `constexpr const char* const`. Some of these constants are used in multiple places so they should be typed only once.

gunchleoc: The only way to to that is to not use `kTemplateDir`, because I have never found a way to directly concatenate 2 `const char* const` variables. So, we'd type `kTemplateDir` 4 times instead. If we don't come up with anything else, we could turn them into macros. In the long run, I want to make the template dir configurable anyway, so I'd be fine with a macro here if you agree.

ETA: Nevermind, We can just make `"wui/window_unpin.png"` etc constexpr and use `window_image_path`

Noordfrees (approved):

PR #4044, opened 2020-08-01
Animation duration units

gunchleoc:
Implement more human-readable duration units, e.g. `1m20s500ms`.

Notes:

* I had also considered using `1:20.500` as a notation, but then it would not be immediately clear from `1:20` whether that's hours-minutes or minutes-seconds
* The function `read_duration` is protected rather than private because there will me more uses, e.g. for the decay parameter in `construct=`
* I have only changed the trees so far to keep the diff smaller

TODOs for follow-up branches:

* Update tribes' `init.lua` files
* Update documentation
* Add deprecation warning

@hessenfarmer @Noordfrees


Noordfrees (approved):
Very nice :)
Code LGTM; tested briefly and no problems observed

PR #4050, opened 2020-08-02
Add new option: Allow diagonal scrolling with numeric keypad

### LINKED ISSUE #3986: Keyboard speed control is too difficult on some Laptop keyboards
## Describe the bug
Speed control is some kind of broken. `Page Up` and `Page Down` doesn't work as expected when you use numeric keypad.

## To Reproduce

1. Run any scenario
2. Use `Page Up`/`Page Down` on your numeric keypad when `Num Lock` is off.

## Expected behavior

According to UI, it should speed up/down the game.

## Actual behavior

`Home`, `End`, `Page Up` and `Page Down` on numeric keypad are used to navigate the map.

## Version:

 - OS: `EL7`
 - Widelands Version: `Build 21~rc1`, `flatpak`

## Additional context

It is hard to say if this is a bug or not. The problem is the ambiguity of UI description.
Anyway, this can be problematic on notebooks that don't have normal navigation keys (e.g. Lenovo Legion, ASUS ROG, etc). They provide a numeric keypad instead.

asus-rog

scx:
Page Up/Down on the numeric keypad don't work as expected. They should
speed up/down the game. Unfortunately, they are assigned for diagonal
scrolling.

By this patch we want to revert the old behavior, and at the same time
allow users to use numpad for diagonal scrolling when they explicitly
enable this feature.

Closes widelands/widelands#3986

### DIFF
--- a/src/wlapplication.cc
+++ b/src/wlapplication.cc
@@ -825,6 +825,7 @@ bool WLApplication::init_settings() {
 	get_config_bool("single_watchwin", false);
 	get_config_bool("ctrl_zoom", false);
 	get_config_bool("game_clock", true);
+	get_config_bool("numpad_diagonalscrolling", false);
 	get_config_bool("inputgrab", false);
 	get_config_bool("transparent_chat", false);
 	// Undocumented. Unique ID used to allow the metaserver to recognize players
--- a/src/wui/interactive_gamebase.cc
+++ b/src/wui/interactive_gamebase.cc
@@ -32,6 +32,7 @@
 #include "logic/map_objects/tribes/ship.h"
 #include "logic/player.h"
 #include "network/gamehost.h"
+#include "wlapplication_options.h"
 #include "wui/constructionsitewindow.h"
 #include "wui/dismantlesitewindow.h"
 #include "wui/game_chat_menu.h"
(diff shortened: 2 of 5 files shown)


scx:
@stonerl @gunchleoc @Noordfrees
Can you take care of this?

src/wui/interactive_gamebase.cc
```diff
@@ -375,10 +376,15 @@ bool InteractiveGameBase::handle_key(bool down, SDL_Keysym code) {
...
 	}
+	const bool numpad_diagonalscrolling = get_config_bool("numpad_diagonalscrolling", false);
 
 	if (down) {
 		switch (code.sym) {
+		case SDLK_KP_9:
 		case SDLK_PAGEUP:
+			if (code.sym == SDLK_KP_9 && ((code.mod & KMOD_NUM) || numpad_diagonalscrolling)) {
+				break;
+			}
```
scx, line 387: Don't speed up game when the key is `Page Up`/`9` on NumPad and:
- NumPad is on

or:

- diagonal scrolling is enabled

src/wui/mapview.cc
```diff
@@ -619,16 +620,16 @@ bool MapView::scroll_map() {
 	int32_t distance_to_scroll_y = 0;
 
 	// check the directions
-	if (kUP || kNP7 || kNP8 || kNP9) {
+	if (kUP || kNP8 || (numpad_diagonalscrolling && (kNP7 || kNP9))) {
```
scx, line 623: Perform diagonal scrolling only if this feature is enabled.

src/wlapplication.cc
```diff
@@ -825,6 +825,7 @@ bool WLApplication::init_settings() {
 	get_config_bool("single_watchwin", false);
 	get_config_bool("ctrl_zoom", false);
 	get_config_bool("game_clock", true);
+	get_config_bool("numpad_diagonalscrolling", false);
```
scx, line 828: Diagonal scrolling should be disabled by default, because it conflicts with time control.

scx (commented):
Explanation of submitted changes.

Noordfrees (approved):
Code LGTM, tested and working :)

PR #4051, opened 2020-08-02
Clear some clang-analyzer checks

gunchleoc:
* `clang-analyzer-core.CallAndMessage`
* `clang-analyzer-core.DivideZero`


src/economy/ferry_fleet.cc
```diff
@@ -354,22 +354,22 @@ void FerryFleet::act(Game& game, uint32_t /* data */) {
 	}
 	while (!pending_ferry_requests_.empty() && !idle_ferries.empty()) {
 		// The map is sorted by ascending gametime
-		Waterway* ww = pending_ferry_requests_.begin()->second;
+		Waterway& ww = *pending_ferry_requests_.begin()->second;
 
-		Ferry* ferry = nullptr;
+		Ferry* ferry = idle_ferries.front();
```
Noordfrees: Bug! See 17 lines below

Noordfrees (changes requested):
Some of those safety checks look just stupid to me. What's the point of using `std::max(1, infspace_count)` directly after `assert(infspace_count > 0);`? Or using `closest ? closest->serial() : 0` after `assert(closest);`? That is ugly.

src/economy/ferry_fleet.cc
gunchleoc: Should be fixed now

gunchleoc:
> Some of those safety checks look just stupid to me. What's the point of using `std::max(1, infspace_count)` directly after `assert(infspace_count > 0);`? Or using `closest ? closest->serial() : 0` after `assert(closest);`? That is ugly.

This is safer for release builds, where the assert isn't being checked. I have already found at least one instance in the code where asserts are dependent on input from Lua files, which is a bug, because we can't assert that.

If you still think these are too ugly, I can revert them though and tag them as `// NOLINT`

src/economy/flag.cc
```diff
@@ -101,7 +101,7 @@ void Flag::load_finish(EditorGameBase& egbase) {
 			log("WARNING: worker %u is in the capacity wait queue of flag %u but "
 			    "its waitforcapacity task is for map object %u! Removing from "
 			    "queue.\n",
-			    worker.serial(), serial(), state->objvar1.serial());
+			    worker.serial(), serial(), state ? state->objvar1.serial() : 0);
```
Noordfrees: Will never be null because we have a `if (state == nullptr) { return true; }` a few line above

Noordfrees (commented):
Okay then. One comment inline where I would really prefer a NOLINT.

Noordfrees (approved):

PR #4053, opened 2020-08-03
Use new animation duration syntax in buildings

gunchleoc:


hessenfarmer (commented):
changes are ok 
However we should change the sleep commands to the duration notation as well.
The tribe specific buildings in the scenarios need to be adopted as well

gunchleoc:
The workers are also still missing. I'll definitely do sleep next when these are through. There will be more, e.g. ship construction decay.

hessenfarmer (approved):

PR #4054, opened 2020-08-03
Fix scenario select screen at low resolution

### LINKED ISSUE #3985: Campaign scenario selection screen text overlaps at 800x600 resolution
## Describe the bug

UI is barely tested when it comes to window mode and low resolution (800x600). E.g. text does not fit in the box.
It applies mainly to the scenario description. Difficulty selector is affected as well.

### Bug 1

1. Run **Widelands** in window mode (800x600)
2. Select `Single Player`
3. Select `Campaigns`
4. Select `Empire`

### Bug 2

1. Run **Widelands** in window mode (800x600)
2. Select `Single Player`
3. Select `Campaigns`
4. Select `Frisians`
5. Select `Difficulty`

## Actual behavior

widelands-21-rc1-window-scenario

widelands-21-rc1-window-scenario-difficulty

## Expected behavior
There should be a scroll bar, like in the `campaign` screen.

widelands-21-rc1-window-campaign

## Version:
 - OS: EL7
 - Widelands Version: `Build 21~rc1`, flatpak

## Additional context

`Build 20` is not affected, but only because descriptions are shorter there. There is also no difficulty selector.

scx:
Fixes widelands/widelands#3985

### DIFF
--- a/src/ui_fsmenu/load_map_or_game.cc
+++ b/src/ui_fsmenu/load_map_or_game.cc
@@ -37,11 +37,11 @@ FullscreenMenuLoadMapOrGame::FullscreenMenuLoadMapOrGame()
 void FullscreenMenuLoadMapOrGame::layout() {
 	// UI coordinates and spacers
 	tablex_ = get_w() * 47 / 2500;
-	tabley_ = get_h() * 17 / 50;
+	tabley_ = get_h() * 14 / 50;
 	tablew_ = get_w() * 711 / 1250;
 	tableh_ = get_h() * 6083 / 10000;
 	right_column_x_ = tablex_ + tablew_ + right_column_margin_;
-	buty_ = get_h() * 9 / 10;
+	buty_ = get_h() * 84 / 100;
 	butw_ = (get_w() - right_column_x_ - right_column_margin_) / 2 - padding_;
 	buth_ = get_h() * 9 / 200;
 	right_column_tab_ = get_w() - right_column_margin_ - butw_;
--- a/src/ui_fsmenu/scenario_select.cc
+++ b/src/ui_fsmenu/scenario_select.cc
@@ -154,7 +154,9 @@ void FullscreenMenuScenarioSelect::layout() {
 	header_box_.set_size(get_w(), tabley_);
 	table_.set_size(tablew_, tableh_);
 	table_.set_pos(Vector2i(tablex_, tabley_));
-	scenario_details_.set_size(get_right_column_w(right_column_x_), tableh_ - buth_ - 4 * padding_);
+	scenario_details_.set_size(get_right_column_w(right_column_x_), tableh_ - buth_ - 6 * padding_ - scenario_difficulty_.get_h() - scenario_difficulty_header_.get_h());
+	scenario_details_.set_desired_size(
+	   get_right_column_w(right_column_x_), tableh_ - buth_ - 6 * padding_ - scenario_difficulty_.get_h() - scenario_difficulty_header_.get_h());
 	scenario_details_.set_pos(Vector2i(right_column_x_, tabley_));
 	scenario_difficulty_.set_size(get_right_column_w(right_column_x_), scenario_difficulty_.get_h());
 	scenario_difficulty_.set_pos(
--- a/src/ui_fsmenu/scenario_select.cc
+++ b/src/ui_fsmenu/scenario_select.cc
@@ -154,9 +154,13 @@ void FullscreenMenuScenarioSelect::layout() {
 	header_box_.set_size(get_w(), tabley_);
 	table_.set_size(tablew_, tableh_);
 	table_.set_pos(Vector2i(tablex_, tabley_));
-	scenario_details_.set_size(get_right_column_w(right_column_x_), tableh_ - buth_ - 6 * padding_ - scenario_difficulty_.get_h() - scenario_difficulty_header_.get_h());
+	scenario_details_.set_size(get_right_column_w(right_column_x_),
+	                           tableh_ - buth_ - 6 * padding_ - scenario_difficulty_.get_h() -
+	                              scenario_difficulty_header_.get_h());
 	scenario_details_.set_desired_size(
-	   get_right_column_w(right_column_x_), tableh_ - buth_ - 6 * padding_ - scenario_difficulty_.get_h() - scenario_difficulty_header_.get_h());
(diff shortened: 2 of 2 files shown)


src/ui_fsmenu/load_map_or_game.cc
```diff
@@ -37,11 +37,11 @@ FullscreenMenuLoadMapOrGame::FullscreenMenuLoadMapOrGame()
...
 	tablex_ = get_w() * 47 / 2500;
-	tabley_ = get_h() * 17 / 50;
+	tabley_ = get_h() * 14 / 50;
 	tablew_ = get_w() * 711 / 1250;
 	tableh_ = get_h() * 6083 / 10000;
 	right_column_x_ = tablex_ + tablew_ + right_column_margin_;
-	buty_ = get_h() * 9 / 10;
+	buty_ = get_h() * 84 / 100;
 	butw_ = (get_w() - right_column_x_ - right_column_margin_) / 2 - padding_;
 	buth_ = get_h() * 9 / 200;
```
scx, line 46: **tabley_factor**: table relative position (top)
**tableh_factor**: table relative height
**buty_factor**: buttons relative position (top)
**buth_factor**: buttons relative height

For **tabley_factor**, _14/50_ is probably the highest possible value that would allow 3 items to be displayed in a combo box at low resolution.
**tabley_factor** + **tableh_factor** - **buth_factor** = _17/50_ + _6063/10000_ - _9/200_ = _0.9013_ ~= _9/10_ = **buty_factor**
If we change **tabley_factor**, we also need to modify **buty_factor**.
_14/50_ + _6063/10000_ - _9/200_ = _0.8413_ ~= _84/100_

src/ui_fsmenu/scenario_select.cc
```diff
@@ -154,7 +154,13 @@ void FullscreenMenuScenarioSelect::layout() {
 	header_box_.set_size(get_w(), tabley_);
 	table_.set_size(tablew_, tableh_);
 	table_.set_pos(Vector2i(tablex_, tabley_));
-	scenario_details_.set_size(get_right_column_w(right_column_x_), tableh_ - buth_ - 4 * padding_);
+	scenario_details_.set_size(get_right_column_w(right_column_x_),
+	                           tableh_ - buth_ - 6 * padding_ - scenario_difficulty_.get_h() -
+	                              scenario_difficulty_header_.get_h());
```
scx, line 159: We need to add the difficulty label height + padding + difficulty combobox height + padding.

src/ui_fsmenu/scenario_select.cc
```diff
@@ -154,7 +154,13 @@ void FullscreenMenuScenarioSelect::layout() {
...
+	scenario_details_.set_desired_size(
+	   get_right_column_w(right_column_x_), tableh_ - buth_ - 6 * padding_ -
+	                                           scenario_difficulty_.get_h() -
+	                                           scenario_difficulty_header_.get_h());
```
scx, line 163: Idea borrowed from [campaign_select.cc](https://github.com/widelands/widelands/blob/build21/src/ui_fsmenu/campaign_select.cc#L78-L79).

scx (commented):
Explanation of submitted changes.

Noordfrees:
Hm, after this change there is a noticeable strip of unused space at the bottom of the campaign/scenario select screens which doesn't look good IMHO. I think the problem with the dropdown dropping partly out of the window is a bug in dropdown code, so it should be fixed there rather than using this workaround. 

Description-textarea height fix confirmed and working.

gunchleoc:
Fiddling with the values is fragile. We should use a `UI:Box` instead, just like everything else in the `ScenarioDetails` does.

scx:
@Noordfrees 

> Hm, after this change there is a noticeable strip of unused space at the bottom of the campaign/scenario select screens which doesn't look good IMHO

This is the whole point - make some empty space at the bottom. Anyway, I think it's just a matter of taste. Previously, the lower graphics were partially covered by buttons, which also didn't look good.

widelands_build21_scenario_select_fix1

widelands_build21_scenario_select_fix2

Since we already have empty space at the top, why can't we have it at the bottom?

>  I think the problem with the dropdown dropping partly out of the window is a bug in dropdown code, so it should be fixed there rather than using this workaround.

To tell the truth, I really don't know how it could be solved. We can't put it outside the window. Even if it were possible (but it isn't), we also need something that would work in fullscreen mode.

I see a few options:
- the list could expand in opposite direction when there is not enough space at the bottom
- move the combobox a little up over the buttons, but without touching the table and buttons
- we could use a scrollbar inside the combobox (IMHO, this is the best idea)

However, even with a scrollbar, we need at least 3 visible items to make the list clear and readable.

widelands_build21_scenario_select_after_expanded_scrollbar

Alternatively, we could put this combobox on top, but that's just a workaround.

scx:
@gunchleoc 

> Fiddling with the values is fragile. We should use a UI:Box instead, just like everything else in the ScenarioDetails does.

I fail to see how it could help. Can you provide some mockups?

Noordfrees:
Dropdowns usually drop open towards the top of the screen when there isn't enough space below them. That they don't do that here is a bug.

> Since we already have empty space at the top, why can't we have it at the bottom?

For consistency with the other screens, and because masses of unused space don't look good IMHO

> > Fiddling with the values is fragile. We should use a UI:Box instead, just like everything else in the ScenarioDetails does.
>
> I fail to see how it could help.

Boxes intelligently layout their child elements without the need to specify masses of fragile magic numbers

scx:
> Dropdowns usually drop open towards the top of the screen when there isn't enough space below them. That they don't do that here is a bug.

So it is already implemented, just doesn't work properly, right?

> For consistency with the other screens

Any examples?

> Boxes intelligently layout their child elements without the need to specify masses of fragile magic numbers

But the combox itself fits very well. The problem appears only when it is expanded. Can `UI:Box` handle this? Or is there at least any chance the list will start opening towards the top of the screen?

gunchleoc:
The dropdowns have only been used within UI::Box so far, so migrating this outdated code to UI::Box might already fix the problem. We won't know if the bug goes away or not until we try.

Niektory:
> > Dropdowns usually drop open towards the top of the screen when there isn't enough space below them. That they don't do that here is a bug.
> 
> So it is already implemented, just doesn't work properly, right?

Yes. Fixed in #4123

Noordfrees:
Superseded by #4411.

PR #4055, opened 2020-08-03
Convert worker animate= to new duration syntax

gunchleoc:
Documentation, deprecation warnings and scenario units will follow in a separate branch.

### DIFF
--- a/data/tribes/workers/frisians/reed_farmer/init.lua
+++ b/data/tribes/workers/frisians/reed_farmer/init.lua
@@ -26,9 +26,9 @@ tribes:new_worker_type {
       harvestreed = {
          "findobject=attrib:ripe_reed radius:1",
          "walk=object",
-         "animate=harvesting duration:12s",
+         "animate=harvesting duration:6s",
          "callobject=harvest",
-         "animate=harvesting duration:1ms",
+         "animate=harvesting duration:6s",
          "createware=reed",
          "return"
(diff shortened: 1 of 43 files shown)


data/tribes/workers/frisians/reed_farmer/init.lua
```diff
@@ -18,17 +18,17 @@ tribes:new_worker_type {
...
          "return"
       },
       harvestreed = {
          "findobject=attrib:ripe_reed radius:1",
          "walk=object",
-         "animate=harvesting 12000",
+         "animate=harvesting duration:12s",
          "callobject=harvest",
-         "animate=harvesting 1",
+         "animate=harvesting duration:1ms",
```
hessenfarmer: that might be a bug we should consult @Noordfrees about this. (1ms would be barely visible)

hessenfarmer (approved):
Ok if the line marked is intended

data/tribes/workers/frisians/reed_farmer/init.lua
Noordfrees: Indeed this should best be twice `6s` rather than `12s` + `1ms` 😁

gunchleoc: Done.

I have run the test suite locally and will merge now to give our CI a break.

PR #4061, opened 2020-08-04
New testsuite stage in Travis

gunchleoc:


Noordfrees:
In times when there are fewer PRs around, having 4 jobs in this stage would be more efficient. So perhaps we could additionally test the two GCC-4.8's in the 3rd stage? (as this is the one that fails most frequently with some shadowed declaration)

gunchleoc:
Good idea - done

Noordfrees (approved):
LGTM :)

gunchleoc:
Travis has reached the last stage and jobs are going green, so let's have it now.

PR #4063, opened 2020-08-05
Add codecheck rule for image files

gunchleoc:
This new rule checks whether image files loaded via `g_gr->images().get()` exist. It can only handle string literals though.


gunchleoc:
I managed to hack the base path for Travis, so this is ready for review now.

cmake/codecheck/rules/images_exist.py
```diff
@@ -0,0 +1,93 @@
...
+"""
+
+import os.path
+import re
+
+
+FIND_LITERALS = re.compile(r"""g_image_cache->get\("(.*?)"\)""")
+# We skip pointers and function calls here
+FIND_VARIABLE_NAMES = re.compile(
+    r"""g_image_cache->get\(([0-9a-zA-Z_-]*?)\)""")
```
Niektory:
```suggestion
    r"""g_image_cache->get\(([0-9a-zA-Z_]+?)\)""")
```
Just a small tweak.

cmake/codecheck/rules/images_exist.py
```diff
@@ -0,0 +1,93 @@
...
+                    variables.append(match)
+
+    # Try to identify images that are fetched by variable name.
+    # We won't be able to check everything, e.g. when the filename is passed via
+    # a function parameter
+    for variable in variables:
+        found = False
+        no_matches = 0
+        found_on_line = 1
+        FIND_VARIABLES = re.compile(variable + r"\w+=\w+\"(.*?)\"")
```
Niektory:
```suggestion
        FIND_VARIABLES = re.compile(r"\b" + variable + r"""\s*=\s*"(.+?)";""")
```
This was not matching anything so I changed it.

Niektory (changes requested):

cmake/codecheck/rules/images_exist.py
gunchleoc: Any particular reason you are removing the `-` character here? Changing `*` to `+` looks good.

Niektory: Variable names can't have `-`, so I don't know what it's doing there.

gunchleoc: Oops, let's get rid of it then.

Niektory (approved):

PR #4065, opened 2020-08-05
Move MS-Windows packaging and integration to a GitHub action

### LINKED ISSUE #3667: Frequent AppVeyor Timeouts
The 64-bit debug builds on AppVeyor frequently fail due to a timeout. The other three builds take about one hour each, this one often timeouts after 2h.

Noordfrees:
This action compiles PRs and master on Windows, runs the basic tests (the regression test suite does not work on GitHub's runners), and uploads the installer binaries (which will be stored for 90 days). 
The purpose of this change is to use AppVeyor for 32-bit master builds only – the action is a bit faster, and many jobs can run in parallel. Offering Release (~ 30 minutes) and Debug (~ 110 minutes) builds, 64-bit only.
I very much hope travis and appveyor will be agreeable with the changes introduced here.

Closes https://github.com/widelands/widelands/issues/3667


Noordfrees:
It is working :)
Still waiting for travis and appveyor.
This is ready for review now

TinoM:
This is very cool, I started to have a look into the windows integration of Travis but his is in the very early stage and no longer necessary then...

Is it possible to create and store artifacts? Then we could deactivate Appveyor completely

Noordfrees:
Uploading artifacts is also possible, but GitHub gives us only 500 MB storage for free which is enough only for a single installer (edit: at least I *think* that this also goes for public repos, they're not very clear about that). It would be possible to upload the binaries to some server if we have a suitable place of course.

Edit: by getting away from appveyor we would have to drop 32-bit support, do we want that?

gunchleoc:
For now, let's keep AppVeyor for the 32 bit, and for building 64 bit on master. Dropping it later is easier than reinstating it.

If people want to test a branch on Windows, 32 bit builds will also run on 64 bit, so that should be sufficient.

Noordfrees:
I read the documentation again and it does sound like artifacts storage would be free and unlimited for us. So how about using this action to package 64-bit installers on master as well (by uploading the installer after the successful run), and keeping AppVeyor around only for 32-bit master builds?
(artifacts will be stored for 90 days)

gunchleoc:
I have translated InnoSetup into Scottish Gaelic, and it's been available officially for years. Why doesn't GitHub have it? *headdesk*

/rant

Noordfrees:
Yay! The installers can be found here: https://github.com/widelands/widelands/actions/runs/200171078
Would be nice if someone with access to a windows machine can check that they work

gunchleoc:
I tested the debug installer and when I start Widelands after finishing the installation, it' missing `libboost_regex-mt.dll`.

Also, GitHub forces people to log in before they can download the installers. So, I think we should continue to use AppVeyor for all master and release branches builds.

Noordfrees:
Hm, that's strange… perhaps this DLL needs to be added to the innosetup script? (@hessenfarmer ?)

TinoM:
The CMake File was explicitely changed to not link Boost statically. I'll try to fix this in the branch directly, if this is ok (tried a sub-merge-request, but this does not trigger the workflows....)

Noordfrees:
Thanks :) 
The workflow is currently set up to build only master and PRs to master, I'll change this to build all PRs

gunchleoc:
Both installers are working now :)

Since you reinstated the AppVeyor x64 builds, it's building them for this PR too. They should only be reinstated for the master branch, not for pull requests. This looks to be non-trivial, so I'd be fine with deactivating x64 builds on AppVeyor for now. We can always figure that one out in a follow-up branch.

gunchleoc (approved):

hessenfarmer:
just a short question: Why are we not using this for 32 bit as well.

Noordfrees:
Because GitHub provides only 64-bit runner systems, they don't let us choose an architecture

TinoM:
> Because GitHub provides only 64-bit runner systems, they don't let us choose an architecture

But this does not matter, it does depend on the compiler. And as far as i can see, the msys2 install on the image does download the 32bit repository, too.
So it should be possible to also install the 32bit-toolchain and 32bit dependencies and compile a 32bit executable.
(Some more if/then cases in the yaml like with appveyor...)

But let's implement this in another branch....

gunchleoc:
> But let's implement this in another branch....

+1

hessenfarmer:
Tino was faster but I wanted to seuggest the same thing.
Let's have this and try to have 32 bit in separate branch.
Good work @Noordfrees. Now travis might be our bottleneck

Noordfrees:
32-bit follow-up branch is up: https://github.com/widelands/widelands/actions/runs/202829222
Would be great if someone with a real 32-bit windows system could try whether the installers work (they should all be up within two hours)

hessenfarmer:
one small nit the builds should be stripped before putting them into innosetup. Same as on appveyor. this would save us about 80 MB per debug build in the installer

Noordfrees:
Ok will add that. Just out of interest what does this command do exactly?

Noordfrees:
Before I open the PR for the follow-up branch. We have two options: 
- Either we use AppVeyor only for master and disable it for PRs, and use the new action for PRs only.
- Or we get rid of AppVeyor completely and use the new action for everything. This is what I'd prefer because one build platform is easier to maintain than two.

Which one do we want? @gunchleoc @hessenfarmer @TinoM

Noordfrees:
and now innosetup fails to find `libspeex-1.dll` both on appveyor and the action

gunchleoc:
GitHub requires everybody to get an account at GitHub before they can download the binaries. I think this is not acceptable for master.

TinoM:
Yes, let's keep Appveyor for now, it is no required check, so it does not hurt.
And both use msys2, so keeping both in sync should be not much work.

hessenfarmer:
I can take care about the changed dll file tonight. I just need to look into my local msys environment to get the correct dll. This is due to an msys update.     
Regarding appveyor, I'd like to keep it for master at least. First for the reason GunChleaoc provided. Second to see whether we really have unlimited space on github.

@Noordfrees 
the strip command removes unneeded debug information from the exe and by this makes it smaller.
see https://man7.org/linux/man-pages/man1/strip.1.html

PR #4068, opened 2020-08-06
Only show MultilineTextarea scrollbar when really needed

### LINKED ISSUE #4060: Scenario details textarea uses scrollbar inconsistently
Admittedly, a rare cornercase. Found this while testing #4054; this only seems to happen with very specific MultilineTextarea dimensions and row counts.
The description of emp01 in low german needs a scrollbar:
shot0001
And so does emp02, but only barely:
shot0002
emp03 doesn't need one:
shot0003
And selecting emp02 *again* – it suddenly doesn't need one:
shot0004

**Expected behaviour:**
The same text in the same bounds should either always or never use a scrollbar.

Niektory:
Change the code that detects whether the scrollbar is needed to always try without the scrollbar first.
Previously the first try was decided by the previous presence/absence of the scrollbar.

Fixes #4060

### DIFF
--- a/src/ui_basic/multilinetextarea.cc
+++ b/src/ui_basic/multilinetextarea.cc
@@ -83,7 +83,9 @@ void MultilineTextarea::set_text(const std::string& text) {
  */
 void MultilineTextarea::recompute() {
 	// We wrap the text twice. We need to do this to account for the presence/absence of the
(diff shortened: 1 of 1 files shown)


Noordfrees (approved):
LGTM; tested and working :)

Issue #4047, opened 2020-08-02
Refactor probability ranges


gunchleoc:
At the moment, probability ranges are defined according to the range of `uint8_t`, i.e. we have 256 values at our disposal.

* This is not intuitive for humans.
* We also want to increase the range for finer control of sound effects (https://github.com/widelands/widelands/issues/3417). The proposal there is for a factor of 5, which would give us 1280 values. Also not intuitive for humans.

I propose to change the range to 1000 values (promille). This could be done in 2 different ways:

1. `"playsound=sound/farm/horse promille:750"` (Keep the current syntax)
2. `"playsound=sound/farm/horse probability:0.75"` (Use floats in Lua and convert to int in the engine)

@Noordfrees @hessenfarmer Do you have any preferences here?

Noordfrees:
+1
I'd prefer to have chances in the range `]0.0, 1.0[` so the second one would look best to me. But please call the parameter `chance` to save poor lua scripters some typing ;)

gunchleoc:
+1 for calling it `chance`

hessenfarmer:
I would like to have it written in percent which would be 99.9, 99.8, 99.7 and so on
Don't mind whther it will be chance or probability both are quite understandable

gunchleoc:
Shall we call it `percent` then to make the unit obvious? I'm still OK with `chance` though. Or we could do:

`"playsound=sound/farm/horse chance:7.5%"`

Noordfrees:
How about allowing both `chance:75%` and `chance:0.75`? More coding effort but also more scripter-friendly

gunchleoc:
This would make it easier to write according to personal preference, but harder to read. I'd rather not have a mix of orthography options.

The percent notation will be more accessible to normal people, and the probability notation to people who are used to doing math-related stuff.

hessenfarmer:
my preferred choice would be percent:87.4

Noordfrees:
`percent` wouldn't make it clear what the percentage refers to. In this example it might also refer to e.g. the relative volume of the sound effect.

hessenfarmer:
chance would be ok as well for me

gunchleoc:
Let's go with `chance:75%` then, this will make the unit type clear. For sounds, we can then have `priority:75%`

hessenfarmer:
so having a deciumal chance would mean chance:75.1% correct?
I would love that

gunchleoc:
[...]

PR #4070, opened 2020-08-06
Transform and remove by percent

gunchleoc:
Change probability Lua specification from [0, 255] to percentage for `transform` and `remove` programs.

Re. #4047


Noordfrees (approved):
One sphinx warning, otherwise LGTM.
```
/home/benedikt/wl/debug/doc/sphinx/source/autogen_immovable_programs.rst:129: WARNING: Explicit markup ends without a blank line; unexpected unindent.
```

Issue #10, opened 2019-09-09
Wrong variable used as array index caused out of bound acces


bunnybot:
Sigsegv while loading savegame (attached) in r1821.

Env:
Debian Testing, widelands compiled using build-widelands.sh with gcc 4.1.2 20061028

More detailed description including gdb backtrace in attachment.

------------------------------------
Imported from Launchpad using lp2gh.

 * date created: 2006-11-24T23:37:00Z
 * owner: marsik
 * the launchpad url was https://bugs.launchpad.net/bugs/535810

bunnybot:
(by marsik)
Logged In: YES 
user_id=1653433
Originator: YES

And the savegame..

bunnybot:
(by sigra)
Logged In: YES 
user_id=31104
Originator: NO

First the loading failed because the savegame referred to the building pressworks, which was deleted in revision 1153 (http://widelands.svn.sourceforge.net/viewvc/widelands?view=rev&revision=1153). Of course Widelands should not terminate after such an error (it should show an error message and return to the load game menu), but that is another bug.

After removing all lines containing "presworks=true" from ~/.widelands/ssave/hrajemesi.wgf/map/allowed_buildings, and loading the savegame again while running Widelands in Valgrind, I got this:

==19089== Invalid read of size 4
==19089==    at 0x8138E82: Player::set_area_seen(Coords, unsigned, bool) (stl_bvector.h:86)
==19089==    by 0x8090DEE: Editor_Game_Base::conquer_area(unsigned char, Coords, Building_Descr const*) (editor_game_base.cc:220)
==19089==    by 0x81A76F6: Warehouse::init(Editor_Game_Base*) (warehouse.cc:422)
==19089==    by 0x805C0C4: Building_Descr::create(Editor_Game_Base*, Player*, Coords, bool, Building_Descr*) (building.cc:113)
==19089==    by 0x8091F60: Editor_Game_Base::warp_building(Coords, char, int) (editor_game_base.cc:597)
==19089==    by 0x81B3F94: Widelands_Map_Building_Data_Packet::Read(FileSystem&, Editor_Game_Base*, bool, Widelands_Map_Map_Object_Loader*) (widelands_map_building_data_packet.cc:96)
==19089==    by 0x81C0A2B: Widelands_Map_Loader::load_map_complete(Editor_Game_Base*, bool) (widelands_map_loader.cc:246)
[...]

PR #4074, opened 2020-08-07
Add copy paste support

### LINKED ISSUE #3943: Enhance gamechatpanel
As of now, gamechatpanel does not support copy'n'paste as one is used to from literally every text edit component in modern systems

**Implementation steps**
- [ ] Enable text selection (via shift+arrow keys)
- [ ] Enable text copy (via strg+c) 
- [ ] Enable text insertion (via strg+v)

jmoerschbach:
fixes #3943

In fact, enables copy&paste not only to gamechat panel but to editbox of course.

- [x] highlight selected text
- [x] also implement for MultilineEditBox
- [x] fix scroll offset issue (not caused by this branch, but anyways...)
- [ ] ~implement selection and copy also in MultilineTextarea~ postponed due to more effort (selection via mouseclick must be implemented)


Noordfrees:
Tested and working, nice feature :) 

In my opinion it should also be enabled for MultilineEditboxes. 
Highlighting the selection is probably not trivial, you may have to wait for https://github.com/gunchleoc/widelands/tree/refactor-wordwrap. Then we should also allow other uses of selection IMHO, such as deleting the selected test when pressing a key as in text editors.

gunchleoc:
Without highlighting is already a good start :+1: 

It would also be good to have clipboard support for error messages in the future - that would be in the MultilineTextarea.

frankystone:
Can this also be added to the debug console (reachable in debug builds if pressing F6)

Noordfrees:
It already is, that's the best part of this feature ;)
(this change applies to *every* single-line edit box)

frankystone:
Yippieh :-)

gunchleoc:
I just pasted something long to the end of the string without doing anything else first, and then hit the arrow buttons + backspace a couple of times in random order. The cursor was plotted with an offset once the whole string started fitting into the edit box.

I used the map options for testing.

jmoerschbach:
> 
> 
> Tested and working, nice feature :)
> 
> In my opinion it should also be enabled for MultilineEditboxes.

I will have a look. Currently, there are only 2 instances of MultilineEditboxes used (in Editor), so lower prio imho
>
> Highlighting the selection is probably not trivial, you may have to wait for https://github.com/gunchleoc/widelands/tree/refactor-wordwrap. 

I implemented first try, looks good imho. Please have a look :)
>
>Then we should also allow other uses of selection IMHO, such as deleting the selected test when pressing a key as in text editors.

implemented, please have a look

I also added some advanced selection actions (ctrl+a, shift+end, shift+pos1)

>
>I just pasted something long to the end of the string without doing anything else first, and then hit the arrow buttons + backspace a couple of times in random order. The cursor was plotted with an offset once the whole string started fitting into the edit box.
>
>I used the map options for testing.

This is not caused by this branch, since same behaviour is also visible on current master (just type random charactes instead of pasting them).
However I will have a look how to fix this ;)

jmoerschbach:
Ready for review.

There is room for improvement: duplicate code (especially between editbox and multilineeditbox), some hard to understand code fragments, for example 
`while ((m_->text[++m_->caret] & 0xc0) == 0x80) {
				}`

I will tackle these on a separate refactoring branch after my holidays.

Also, selection via mouse press might be a interesting feature, especially for read only widgets like MultilineTextarea.

Niektory:
Screenshot_2020-08-15_06-43-11
Selecting like this and pressing [Delete] causes a crash.
```
terminate called after throwing an instance of 'std::out_of_range'
  what():  basic_string::basic_string: __pos (which is 8) > this->size() (which is 1)
(...)
#8  std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::basic_string(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, unsigned long, unsigned long) (this=<optimized out>, __str=..., __pos=<optimized out>, __n=<optimized out>)
    at /build/gcc/src/gcc-build/x86_64-pc-linux-gnu/libstdc++-v3/include/bits/basic_string.h:481
#9  0x0000555556a951ea in UI::EditBox::check_caret() (this=0x7fffffff9638) at ../src/ui_basic/editbox.cc:665
#10 0x0000555556a8d8ea in UI::EditBox::handle_key(bool, SDL_Keysym) (this=0x7fffffff9638, down=true, code=...) at ../src/ui_basic/editbox.cc:282
(...)
```
Can you also implement 'Cut' using [Ctrl+X]?

jmoerschbach:
> 
> 
> Screenshot_2020-08-15_06-43-11
> Selecting like this and pressing [Delete] causes a crash.
> 
> ```
>   what():  basic_string::basic_string: __pos (which is 8) > this->size() (which is 1)
> (...)
> #8  std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::basic_string(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, unsigned long, unsigned long) (this=<optimized out>, __str=..., __pos=<optimized out>, __n=<optimized out>)
>     at /build/gcc/src/gcc-build/x86_64-pc-linux-gnu/libstdc++-v3/include/bits/basic_string.h:481
> #9  0x0000555556a951ea in UI::EditBox::check_caret() (this=0x7fffffff9638) at ../src/ui_basic/editbox.cc:665
> #10 0x0000555556a8d8ea in UI::EditBox::handle_key(bool, SDL_Keysym) (this=0x7fffffff9638, down=true, code=...) at ../src/ui_basic/editbox.cc:282
> (...)
> ```
> 
> Can you also implement 'Cut' using [Ctrl+X]?

crash fixed and [Ctrl+X] implemented :)

jmoerschbach:
does anybody know why check complains about shadowing a member? I do not see any member `end` or `start` in the type hierarchy... Maybe a gcc 4.8 false positive?

Niektory:
> does anybody know why check complains about shadowing a member? I do not see any member end or start in the type hierarchy... Maybe a gcc 4.8 false positive?

They are `UI::Panel` methods.

https://github.com/jmoerschbach/widelands/blob/add_copy_paste_support/src/ui_basic/panel.h#L114-L115

gunchleoc:
GCC 4.8 can't distinguish between function and variable names for this check. We'll drop support for it soon (draft PRs are already up), so you can just wait it out if you want.

jmoerschbach:
perfect, thx for the explanation. I'll wait then instead of renaming the variables :)

klaus-halfmann:
Works fine on macOS, but it must be CMD-X/C/V not CTRL. Steven Jobs will ot tolerate that ;-)
I found no example usage of the CMD-key by scaaing the code with fgrep. Where could I find such an example.

klaus-halfmann (changes requested):
@jmoerschbach can you replace `SDL_GetModState() & KMOD_CTRL` wit some inline function? 
And then insert a single #ifdef APPLE . as this is your private repo I cannot commit to it.
If not should I attach a diff?

gunchleoc:
This is what SDL has to offer:

```
#define KMOD_CTRL   (KMOD_LCTRL|KMOD_RCTRL)
#define KMOD_SHIFT  (KMOD_LSHIFT|KMOD_RSHIFT)
#define KMOD_ALT    (KMOD_LALT|KMOD_RALT)
#define KMOD_GUI    (KMOD_LGUI|KMOD_RGUI)
```

Maybe `KMOD_GUI` is the one?

klaus-halfmann:
Lets try KMOD_GUI, ill try to insert it, but I dont know If I have time

gunchleoc:
You could replace it just so we'll know whether it will work and then leave the #ifdef stuff to the branch author :)

jmoerschbach:
GUI is the one according to https://wiki.libsdl.org/SDL_Keycode
I'll integrate it today

jmoerschbach:
@klaus-halfmann please test :)

src/ui_basic/editbox.cc
```diff
@@ -39,6 +41,12 @@ namespace {
 
 constexpr int kMarginX = 4;
 constexpr int kLineMargin = 1;
+bool inline ctrl() {
```
klaus-halfmann: Well maybe soemthing like copy_paste_modifier() but well ...

klaus-halfmann (approved):
Tested on macOS works for me

klaus-halfmann:
What has failed here? Trusty failed but AFAIK this is some other issue?
Shall I Update and try again?

jmoerschbach:
@klaus-halfmann gcc 4.8 reports a false postive about shadowing a member variable, that is all, see https://github.com/widelands/widelands/pull/4074#issuecomment-675968668

I hope gcc 4.8 support is dropped soon as @gunchleoc promised since I do not want to be forced to rename variables because of a bug of an outdated compiler ;)

gunchleoc:
The Travis branch is ready for review https://github.com/widelands/widelands/pull/4130

hessenfarmer:
Travis branch got merged. So this should be rechecked after merging master

gunchleoc (changes requested):
When marking some text and hitting CTRL-C, CTRL-V to replace the text with itself, crash.

```
terminate called after throwing an instance of 'std::out_of_range'
  what():  basic_string::replace: __pos (which is 7) > this->size() (which is 0)

Thread 1 "widelands" received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
51	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) backtrace
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007ffff434b8b1 in __GI_abort () at abort.c:79
#2  0x00007ffff4d3e957 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x00007ffff4d44ae6 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00007ffff4d44b21 in std::terminate() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007ffff4d44d54 in __cxa_throw () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00007ffff4d4085d in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#7  0x00007ffff4dd7c23 in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::insert(unsigned long, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#8  0x0000555556014445 in UI::EditBox::handle_textinput (this=0x6230000db980, input_text="No Name") at ../src/ui_basic/editbox.cc:454
#9  0x00005555560121f5 in UI::EditBox::handle_key (this=0x6230000db980, down=true, code=...) at ../src/ui_basic/editbox.cc:231
```

While nobody would need to replace text with itself, it can happen by accident, so we shouldn't crash here.

jmoerschbach:
fixed. good catch, thx :)

gunchleoc (approved):
No more bugs found. Thanks for this great feature!

PR #4075, opened 2020-08-07
Fix disappearing language stats

### LINKED ISSUE #3874: Language stats disappear in options menu
- Start Widelands with the language set to Try System Language. (on a system with non-english locale).
- Open the options menu. You will see a stats hint like "De Översetten na Plattdüütsch is to 85% kumpleet." next to the language dropdown.
- Select any other entry in the language dropdown.
- Select Try System Language again. The language stats hint is gone.

Observed with cbfb8f8; OS language `nds`

Noordfrees:
Fixes #3874


gunchleoc (commented):
I don't remember why I added that boolean, so code LGTM.

gunchleoc (approved):

PR #4078, opened 2020-08-08
Fix cppcoreguidelines-pro-type-static-cast-downcast

gunchleoc:
Also, pick up some stragglers for already existing checks that my system flagged up.


Noordfrees (approved):
LGTM

PR #4083, opened 2020-08-08
Show buildhelp semi-transparent on PreviouslySeen nodes

### LINKED ISSUE #3055: Building houses sometimes impossible
I play trunk/7556. Sometimes it's not possible to build something where it should be possible. In some cases clicking the location dozens of times resolved the issue.

In the attached game, for example, there are two places with iron ore next to the existing mine. On one of them it should be possible to build a mine, however it's impossible.

 * date created: 2015-10-18T13:16:26Z
 * owner: jugal
 * the launchpad url was https://bugs.launchpad.net/bugs/1507299

Noordfrees:
Fixes #3055 
Also moves some player-specific code from the IBase to the IPlayer


src/wui/interactive_player.cc
```diff
@@ -450,10 +478,10 @@ void InteractivePlayer::draw_map_view(MapView* given_map_view, RenderTarget* dst
...
 			bool show_port_space = has_expedition_port_space(f->fcoords);
 			if (show_port_space || buildhelp()) {
-				const auto* overlay = get_buildhelp_overlay(
-				   show_port_space ? f->fcoords.field->maxcaps() : plr.get_buildcaps(f->fcoords));
-				if (overlay != nullptr) {
-					blit_field_overlay(dst, *f, overlay->pic, overlay->hotspot, scale);
+				if (const auto* overlay = get_buildhelp_overlay(
+				       show_port_space ? f->fcoords.field->maxcaps() : plr.get_buildcaps(f->fcoords))) {
+					blit_field_overlay(dst, *f, overlay->pic, overlay->hotspot, scale,
+					                   f->seeing == Widelands::SeeUnseeNode::kVisible ? 1.f : 0.4f);
```
gunchleoc:
```suggestion
					                   f->seeing == Widelands::SeeUnseeNode::kVisible ? 1.f : 0.3f);
```

Code LGTM, but I think the difference isn't noticeable enough yet. It would pop even better at `0.2f`, but then the port space icon would be too hard to see. SO, I think `0.3f` is our best option.

gunchleoc (approved):

PR #4086, opened 2020-08-09
Rename main immovable program to 'main'

gunchleoc:
The main program for immovables is called `program`, and for productionsites, it's called `work`. This is inconsistent, so I have decided to call them both `main`.

A second branch will follow for the productionsites.


src/logic/map_objects/immovable_program.cc
```diff
@@ -46,8 +46,8 @@ allow workers to interact with an immovable (e.g. a tree will need a "fall" prog
 woodcutters to remove the tree).
 
 It is not mandatory for immovables to define programs. If the immovable defines a program named
-``program``, this program will be started as the main program on creation. Immovables without such a
-program will simply display their main animation indefinitely.
+``main``, this program will be started as the main program on creation. Immovables without such a
+program will simply display their 'idle' animation indefinitely.
```
Noordfrees, line 50: add a note that the name "program" is reserved and deprecated

src/logic/map_objects/immovable.cc
```diff
@@ -249,12 +259,12 @@ const TerrainAffinity& ImmovableDescr::terrain_affinity() const {
 }
 
 void ImmovableDescr::make_sure_default_program_is_there() {
-	if (!programs_.count("program")) {  //  default program
+	if (!programs_.count("main")) {  //  default program
```
Noordfrees: Make that string a constant in `map_object.h`

src/logic/map_objects/immovable.cc
```diff
@@ -193,8 +193,18 @@ ImmovableDescr::ImmovableDescr(const std::string& init_descname,
...
 		}
 		try {
-			programs_[program_name] = new ImmovableProgram(
-			   program_name, programs->get_table(program_name)->array_entries<std::string>(), *this);
+			// TODO(GunChleoc): Compatibility, remove after v1.0
+			if (program_name == "program") {
+				log("WARNING: The main program for the immovable %s should be renamed from 'program' "
+				    "to 'main'\n",
+				    name().c_str());
+				programs_["main"] = new ImmovableProgram(
```
Noordfrees: If there is a program called "main" and another program called "program" then this compatibility name change must be skipped

src/logic/map_objects/immovable_program.cc
gunchleoc: There is a warning message already - should that not be enough?

src/logic/map_objects/immovable.cc
gunchleoc: The chance of that is slim, so I'll just add an additional warning. Keeping it as "program" will be just as wrong.

src/logic/map_objects/immovable_program.cc
Noordfrees: If the documentation says nothing but you get an unexpected warning, people will say that the doc is missing something

gunchleoc: I have added a note.

Noordfrees (approved):

Issue #3706, opened 2020-02-21
Accessibility ideas for eye tracking


gunchleoc:
From the forum: https://www.widelands.org/forum/topic/4642/

- Add buttons for things that need a right-click
- Add navigator buttons for scrolling the map (where - bottom of minimap?)
- Hover actions are difficult - button to toggle all tooltips on?

Noordfrees:
1 – a little button with an X in the corner of every window's titlebar would be nice
2 – let's have an option to enable scrolling the map when the cursor is near the edges
3 – I don't understand, the tooltips are shown the moment the mouse points on an item, and the TT doesn't go away when you move the mouse only a bit

gunchleoc:
I like your suggestions :)

For 3, maybe it's difficult to keep the eyes still long enough? I don't know how the apparatus works though, so I'm just guessing.

Niektory:
I suspect the problem with tooltips is that you need to move your eyes to read them (as they don't show up exactly at the cursor's location, and some are quite large), but when you do, the mouse cursor moves as well causing them to disappear.

PR #4088, opened 2020-08-10
Enhance production site help

gunchleoc:
* Add "creates" and "collects" sections to productionsites in the Encyclopedia
* Use created/collected info to add some purpose icons
* Add menu icons to rocks
* Make tree menu icons smaller
* Add regression tests for supporting/supported productionsites

Re https://github.com/widelands/widelands/issues/3989


Noordfrees:
The Collect section looks a bit broken. For the frisian farm it says:
  Collects: [field] → [farm] Barley Field (Ripe)
  Creates: [farm] → [field] → [beekeepers house] Beekeeper's House
This should always give the name of the created immovable imho, and perhaps additionally the consuming building for created immovables.

gunchleoc:
> The Collect section looks a bit broken. For the frisian farm it says:
> Collects: [field] → [farm] Barley Field (Ripe)
> Creates: [farm] → [field] → [beekeepers house] Beekeeper's House
> This should always give the name of the created immovable imho, and perhaps additionally the consuming building for created immovables.

How would this look for berry bushes though? Something like this OK?

> Long • List • Of • Many • Berry • Bushes from: Berry Farm

BTW We currently have the same pattern for trainingsites where the names of the weapons are not listed.

Noordfrees:
Perhaps draw only the images of the immovables, producers and consumers; and put their names and the "from"/… strings into a tooltip?

gunchleoc:
We can't do tooltips for individual richtext string elements. It would also make the text unreadable with eye trackers https://github.com/widelands/widelands/issues/3706

gunchleoc:
Here are side-by-side screenshots of the 2 discussed variants - this branch vs. adding immovable names.

immovable_names

I think the added immovable names make it hard to see which other building I would need to place close to this one, which is the point of the information. It would look the same in the "Collects" section, I accidentally coded it in the "Creates" section.

hessenfarmer:
Best would be a dirty hack to summarize the relevant property of the list of images. 
in the example above this could look like 
[Berry Farm] -> Berry Bushes (medium)
[images of the medium berry bushes]
-> [Beekepers House] Beekeper's House
No clue however how to do this

gunchleoc:
> Best would be a dirty hack to summarize the relevant property of the list of images.
> in the example above this could look like
> [Berry Farm] -> Berry Bushes (medium)
> [images of the medium berry bushes]
> -> [Beekepers House] Beekeper's House
> No clue however how to do this

We might do this if we use attributes and make them translatable. That would require a bigger code change though that I don't want to go into for this branch.

I am also using the exact same pattern as we have for trainingsite weapons where we don't list the weapon names, so I don't think it's that important right now.

hessenfarmer:
> 
> 
> > Best would be a dirty hack to summarize the relevant property of the list of images.
> > in the example above this could look like
> > [Berry Farm] -> Berry Bushes (medium)
> > [images of the medium berry bushes]
> > -> [Beekepers House] Beekeper's House
> > No clue however how to do this
> 
> We might do this if we use attributes and make them translatable. That would require a bigger code change though that I don't want to go into for this branch.
> 
> I am also using the exact same pattern as we have for trainingsite weapons where we don't list the weapon names, so I don't think it's that important right now.

Agreed. so the right screenshot with only the images would fit better.

gunchleoc:
That's the current state of this branch then.

hessenfarmer (approved):
If @Noordfrees has no further objections this branch looks ok for me

Noordfrees:
Let's have it then :)

PR #4096, opened 2020-08-11
Add libspeex

gunchleoc:
Innosetup in GitHub action is complaining about a missing dll. Let's see if this fixes it.

### DIFF
--- a/.github/workflows/windows.yaml
+++ b/.github/workflows/windows.yaml
@@ -19,7 +19,7 @@ jobs:
         $env:Path += ";C:\msys64\usr\bin"
         pacman.exe --noconfirm -Syyuu
         pacman.exe --noconfirm -Syyuu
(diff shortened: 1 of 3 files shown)


Noordfrees:
AppVeyor also needs this DLL: https://ci.appveyor.com/project/widelands-dev/widelands/builds/34579093

TinoM:
As far as i can see, Speex was removed as a dependency, so we do not need it any more. Just removing the entry from the innosetup file should be enough.
If you have a look at older builds, speex was just pulled in as a dependency automatically.

(Because we are using the pre compiled packages, we have to include more than just the ogg libs we are actually using..)

TinoM (approved):
tested locally: widelands.exe does no longer need the speex dll to start.

TinoM:
With the only change left in the innosetup script, no travis or appveyor run needed.

PR #4098, opened 2020-08-11
Move 32-bit MS-Windows integration for PRs to the action

Noordfrees:
Build PRs on actions, 64 and 32 bit, Release and Debug.

We can then disable AppVeyor for PRs in the Webhooks section of the repo settings.


hessenfarmer (approved):
LGTM

hessenfarmer:
I disabled the pull requests buolds in appveyor now that this got merged

hessenfarmer:
@Noordfrees 
is there any way to cancel a PR build if the branch gets updated otherwise we will have a lot of running actions for no use. making it hard to find the relevant one

Noordfrees:
There's no need really, as far as I can see we do not seem to hit any limit on concurrent jobs (yet). But each job could check for remote changes after compilation and cancel all 4 jobs if there are any, so we would have cancellation checks 30 and 60 minutes after start.
→ Edit: No, this is not possible because we build a merge commit, so the info which branch and which commit the action corresponds to is not easily obtained

Note that pushes by the formatting bot to not currently trigger workflows. I'm working on a follow-up branch to change this.

Issue #3678, opened 2020-02-07
Increase maximum map resources in editor to 50


gunchleoc:
Allow map editors to place higher amounts of resources. See discussion in https://www.widelands.org/forum/topic/4688/

We agreed on an amount of 50.

Noordfrees:
We should then also add more stages for the editor resource icons since only 4 icons per resource would be too inexact

frankystone:
There is also inconsistency in the UI:

One can set the amount to a value of 63, but when setting the resources there will be max. amount of 20 for Gold/Coal/Iron/Stone/Fish and max. amount 50 for Water:

![resource_amount](https://user-images.githubusercontent.com/6730556/77826338-0e80ed80-710f-11ea-8d89-775783f6f674.jpg)

If all resources have a max. amount of 50, it shouldn't be possible to set a bigger value in the ui.

Noordfrees:
Instead of hardcoding the spinbox values we should set the upper border to the selected resource's max amount whenever the selection changes.

PR #4099, opened 2020-08-11
Use named parameters in ActMine

gunchleoc:
The worker experience parameter is now optional.
Also, fix hard-coded max resources.

Ideas for better parameter names are welcome :)

@hessenfarmer @MarkMcWire @Noordfrees 

Re #4047


Noordfrees (commented):
I'd suggest renaming `workarea` to `radius` for consistency with workers. Code LGTM, not tested.

src/logic/map_objects/tribes/production_program.cc
```diff
@@ -1343,16 +1404,16 @@ void ProductionProgram::ActMine::execute(Game& game, ProductionSite& ps) const {
 		} while (mr.advance(*map));
 	}
 
-	//  how much is digged
-	int32_t digged_percentage = 100;
+	//  how much is dug
+	unsigned dug_percentage = MapObjectProgram::kMaxProbability;
```
hessenfarmer, line 1408: I belive this won't work as kMaxProbability is 10000 instead of 100

src/logic/map_objects/tribes/production_program.cc
```diff
@@ -1393,7 +1454,8 @@ void ProductionProgram::ActMine::execute(Game& game, ProductionSite& ps) const {
 		//  there is a sufficiently high chance, that the mine
 		//  will still produce enough.
 		//  e.g. mines have chance=5, wells have 65
-		if (chance_ <= 20) {
+		if (depleted_chance_ <= game.world().get_resource(resource_)->max_amount() *
```
hessenfarmer: Don't undestnad this change we need to have a small value her to show empty mines and notify the player while not notifiying the player about a depleted (normal) well

hessenfarmer (changes requested):
I agree with Nordfrees on radius.
ressources could be yield (shorter and more appropriate)
depleted could be yield_dead or deadyield
experience  might be train_fail

src/logic/map_objects/tribes/production_program.cc
gunchleoc: `10000` corresponds to `100%` now, because we represent floating point as int. E.g. `33.33%` is represented as `3333`.

I have printed some test log output and tested this to verify that my scaling is correct.

src/logic/map_objects/tribes/production_program.cc
gunchleoc: Needs to be scaled up, just the same way as I commented above.

At the moment, `game.world().get_resource(resource_)->max_amount() == 20` for all 4 resources via their Lua configuration. So, no semantic change there either until we start implementing https://github.com/widelands/widelands/issues/3678 where the old code would cause a bug.

gunchleoc:
* `workarea` => `radius` done
* `resources`=> `yield`done
* depleted could be yield_dead or deadyield: Not a fan. How about `when_depleted`?
* experience might be train_fail: Soldiers get trained, workers gain experience. I don't think we should mix up our terminology here. It also makes me think that it's a chance of failure, rather than a change of success in spite of a failure. `experience_on_fail` could work, but is pretty long and not 100% accurate either. What is actually does is `experience_on_fail_when_depleted`.

src/logic/map_objects/tribes/production_program.cc
hessenfarmer: this is not correct as this has nothing to do with the initial amount or the max amount. it is just an arbitrary value defined as threshold to declare an empty state on a building which mines something. the result is every building with a yield_depleted higher then 20 will never be empty --> no message in the news window, no empty animation. So mixing in the ressources amount would do no good here

hessenfarmer:
ok understood the max Probability thing now. Thanks for explanation
Problem with the names is that they are only useful if they are self explaining.
so what about yield_empty as this is what it does.

no glue for the worker param so far

hessenfarmer:
learn_on _fail is the shortest so far I can imagine.

src/logic/map_objects/tribes/production_program.cc
gunchleoc: Thanks for explaining - I misunderstood the intent. Will change it back to 20.

gunchleoc:
How about `experience_on_fail`? "learn" doesn't mean anything in Widelands, but "experience" does.

`yield_empty`makes `yield` harder to understand. How about `when_empty`?

hessenfarmer:
> 
> 
> How about `experience_on_fail`? "learn" doesn't mean anything in Widelands, but "experience" does.
> 
> `yield_empty`makes `yield` harder to understand. How about `when_empty`?

Well I have no strong opinion on this rather then that there is no short form to really make it telling so I am ok with the above. Maybe we could consider yield_if_empty as I believe this to be conditional and not temporal. but in the end it doesn't matter as we need to look in the docs to really understand the params.

hessenfarmer (approved):

gunchleoc:
Let's go with `when_empty` then. This parameter means "the chance that you can still get some while the mine is exhausted", while "yield" means "the percentage of resources you can mine with a chance of 100% until you will exhaust the mine"

PR #4100, opened 2020-08-11
CMake Cleanups

TinoM:
Trying to cleanup the cmake files...


CMakeLists.txt
```diff
@@ -1,48 +1,10 @@
...
 
-# This policy is not known to versions prior 3.1 and would result in errors,
-# if set on such systems. This can be removed when cmake_minimum_required is set
-# to 3.1 or newer by using:
-# cmake_policy(VERSION 3.1)
-if(POLICY CMP0054)
-   cmake_policy(SET CMP0054 NEW)
-endif(POLICY CMP0054)
-
-# This policy is not known to versions prior 3.11 and would result in errors,
```
hessenfarmer, line 16: well from the explanation in the text I would have expected we would need this on Trusty probably as we only require 3.9.
However as it compiles it could be good to go as well

gunchleoc: I get this on my trusty VM
 
```
CMake Error at CMakeLists.txt:3 (cmake_minimum_required):
  CMake 3.9 or higher is required.  You are running version 2.8.12.2
```

TinoM: Yes, i was hoping to use at least 3.12, but the Trusty images on Travis have only 3.9 available.
I see no errors in the logs, so the policy referring to be needed <3.11 seems to be no longer necessary (maybe due to updated cmake package files..)

@gunchleoc : Do you have "cmake3" as command available?

gunchleoc: There is a cmake3 package available for Trusty. When I install it, I get:

```
CMake Error at CMakeLists.txt:3 (cmake_minimum_required):
  CMake 3.9 or higher is required.  You are running version 3.5.1
```

My VM is a default installation + some installed dev packages, so that's the kind of setup we should be aiming for.

TinoM: Ah, interesting. I'll try to figure out why on Travis the Trusty comes with 3.9.
Would lowering our requirement to 3.5 ok? Or do you insist on staying compat to 2.8.x ?

[And with Trusty outlived its Standard Support Live time in April '19 maybe we should move to Xenial as lowest requirement / supported Distro.]

gunchleoc: 3.5 is fine - it's available out of the box from the package manager.

I think we should try to still support trusty for a while, unless it becomes too ugly. Launchpad PPAs still support it.

gunchleoc:
The older version doesn't know ICU yet:

```
CMake Error at CMakeLists.txt:108 (find_package):
  By not providing "FindICU.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "ICU", but
  CMake did not find one.

  Could not find a package configuration file provided by "ICU" with any of
  the following names:

    ICUConfig.cmake
    icu-config.cmake

  Add the installation prefix of "ICU" to CMAKE_PREFIX_PATH or set "ICU_DIR"
  to a directory containing one of the above files.  If "ICU" provides a
  separate development package or SDK, be sure it has been installed.

```

TinoM:
https://gitlab.kitware.com/cmake/cmake/blob/master/Modules/FindICU.cmake :

`FindICU
-------

.. versionadded:: 3.7
`
I do not find any detailled list of package (resources) Travis is using for it's build images. 

So difficult to say, why there is Cmake 3.9 available and your Trusty ist stuck with 3.5.

gunchleoc:
This has nothing to do with Travis. I am testing on my own local VM that's just a default Ubuntu Trusty install + some development packages for building stuff.

hessenfarmer:
Can we copy the FindICU from CMAKe 3.7 or later to our repo without violating any license so this would result the same for newer CMAKE and the trusty build?
Maybe adding a comment that this file can be deleted after dropping trusty

TinoM:
> This has nothing to do with Travis. I am testing on my own local VM that's just a default Ubuntu Trusty install + some development packages for building stuff.
Yeah, i know / didn't mean that. So Travis is using a "non default" Trusty appereantly. 

I have only Windows available, so i am dependent on the Travis builds to check. So if we can figure out if there is maybe a nice backport apt repository for Trusty with some more recent build tools (Travis is using), we can require newer versions to make the maintanance of our build scripts easier.

gunchleoc:
I'll test with my Trusty to see what we need to put back in for it to work.

gunchleoc:
The needed changes are in https://github.com/widelands/widelands/pull/4120

We'd lose the ICU and the GLEW cleanups.

hessenfarmer:
@TinoM 
as we found a solution for Trusty (direct users to where to obtain an actual cmake) could you please have a look into whether we would have this one or the static_link branch. I believe they are identical for the Mac OS and Linux parts but I'd like to have it doublechecked.
As the static_link is now working I would prefer to have the static windows builds

gunchleoc:
They are not identical https://github.com/widelands/widelands/compare/static_link...cmake_housekeeping - this branch has even more cleanups in it.

Shall we merge the static link branch first, then update this one and merge it after?

hessenfarmer:
> 
> 
> They are not identical [static_link...cmake_housekeeping](https://github.com/widelands/widelands/compare/static_link...cmake_housekeeping) - this branch has even more cleanups in it.

This is one thing that doesn't work correctly in github. The shown diff is not what is the real diff as it is somehow influenced by master ( igues it is based on commits rather then really doing a compare, but I don't know.
YYou can check this by manually diffing cmakelists.txt

> 
> Shall we merge the static link branch first, then update this one and merge it after?   
   
If we take them both I prefer to have this cmake file first as this should not result in conflichts while it could be the other way round

gunchleoc:
Shell we revert to before https://github.com/widelands/widelands/pull/4100/commits/4d1f3de1e627992e2be9963ca899371b968f0e21 (lower CMake version for Trusty didn't help) and then merge?

TinoM:
> Shell we revert to before [4d1f3de](https://github.com/widelands/widelands/commit/4d1f3de1e627992e2be9963ca899371b968f0e21) (lower CMake version for Trusty didn't help) and then merge?

No need. We do require the two policy guards because they are for < 3.11 and < 3.12 and we want to use Cmake on Trusty on Travis with 3.9. I'll set this as a minimum.

TinoM:
> If we take them both I prefer to have this cmake file first as this should not result in conflichts while it could be the other way round

Yes, i think this is a bit less work and should require solving less conflicts (if at all).

hessenfarmer:
Ok so let's have this branch first as soon as it get's green

hessenfarmer (approved):

hessenfarmer:
All green so I merge

Noordfrees:
Our daily PPA is failing on Trusty and Xenial since this change. Should we try to fix it or drop support for these two?

gunchleoc:
I have dropped Trusty.

Noordfrees:
ok but Xenial is also failing
`  CMake 3.9 or higher is required.  You are running version 3.5.1`

TinoM:
I've lowered cmake requirements to 3.5

PR #4105, opened 2020-08-12
Show partially seen buildings in main view

### LINKED ISSUE #3778: Minimap reveals unseen enemy buildings

Niektory:
Fixes #3778
Supersedes #3877

[Test savegame](https://github.com/widelands/widelands/files/5060650/vision_test.wgf.zip). I haven't tested this thoroughly yet.


src/logic/player.cc
```diff
@@ -1231,7 +1231,11 @@ void Player::rediscover_node(const Map& map, const FCoords& f) {
 					map_object_descr = nullptr;
 				} else if (upcast(Building const, building, base_immovable)) {
 					if (building->get_position() != f) {
-						// This is not the building's main position so we can not see it.
+						// This is not the building's main position. We don't store the building's
+						// description here but in its main position instead.
+						Field& field_main =
+						   fields_[map.get_fcoords(building->get_position()).field - &first_map_field];
+						field_main.map_object_descr = map_object_descr;
```
Noordfrees, line 1238: You need to update the other field's ConstructionSiteInformation as well

src/wui/interactive_player.cc
```diff
@@ -420,18 +428,22 @@ void InteractivePlayer::draw_map_view(MapView* given_map_view, RenderTarget* dst
...
 	Workareas workareas = get_workarea_overlays(map);
-	auto* fields_to_draw = given_map_view->draw_terrain(gbase, &plr, workareas, false, dst);
+	FieldsToDraw* fields_to_draw = given_map_view->draw_terrain(gbase, &plr, workareas, false, dst);
 	const auto& road_building_s = road_building_steepness_overlays();
 
 	const float scale = 1.f / given_map_view->view().zoom;
 
+	// Store the coords of partially visible buildings
+	// so we can draw them later when we get to their main position.
+	std::vector<Widelands::Coords> deferred_coords;
```
Noordfrees: Since the order does not matter, an `std::set` would make for better readability IMHO because you can then use `count()` rather than that ugly `std::find()`

Noordfrees (commented):
Two comments inline, not tested yet.

What will happen when a player clicks on an enemy big military building that is only partially seen?

Niektory:
> What will happen when a player clicks on an enemy big military building that is only partially seen?

Right now it'll behave as if it was not seen. Not intuitive, I'll see how to change that.

Noordfrees:
Code LGTM, but the case of half-seen enemy military sites still needs fixing. I can now open the attack dialogue but I can select zero soldiers as long as the site is not being seen by a scout.

Niektory:
> Code LGTM, but the case of half-seen enemy military sites still needs fixing. I can now open the attack dialogue but I can select zero soldiers as long as the site is not being seen by a scout.

It should work now.

I use this map for testing. The distance between the blue player and the other players is 24. [vision_test_24.wmf.zip](https://github.com/widelands/widelands/files/5065676/vision_test_24.wmf.zip)

Noordfrees (approved):

PR #4112, opened 2020-08-13
Localedir parameter

### LINKED ISSUE #4097: Add new commandline parameter for the locale dir
Currently only the datadir is a parameter and widelands expects the localedir to be in the datadir.
It does no longer crash with a missing localedir, but it would be nice to be able to specificy the localedir independently.

On my windows development machine, the locale dir with the translation get build in the build dir, while the datadir resides in the git checkout dir.
I've added an extra xcopy to my build script to copy the translations and modified .gitignore, but it would rather not.

@gunchleoc you have been working on this, but i am not sure, if and what branch is valid?

I've also checked: widelands does not accept a windows symbolic link named "locale" in the datadir on windows.

TinoM:
Resolves #4097 :

Allow setting of a localedir indepentendly from the datadir.
Also show default homedir on windows systems.

### DIFF
--- a/src/wlapplication.cc
+++ b/src/wlapplication.cc
@@ -880,7 +880,7 @@ bool WLApplication::init_settings() {
  */
 void WLApplication::init_language() {
 	// Set the locale dir
-	if (localedir_) {
+	if (!localedir_.empty()) {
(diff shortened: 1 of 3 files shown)


gunchleoc (approved):
Code LGTM, not tested

PR #4113, opened 2020-08-13
Speedup player vision packet

### LINKED ISSUE #4109: Player vision packet is too slow
Writing new Europa map with Build 21:

```
Writing Exploration Data ... took 18ms
Writing Players Unseen Data ... took 70ms
```

And on Master:

```
Writing Players View Data ... took 13565ms

```

This is far too slow. I have started working on this https://github.com/gunchleoc/widelands/tree/speedup-player-vision-packet

gunchleoc:
Closes https://github.com/widelands/widelands/issues/4109

Not as fast as it used to be, but workable. I used the same trick as for the player statistics.

Also, nodes that were never seen are skipped.


gunchleoc:
This is ready now.

src/map_io/map_players_view_packet.cc
```diff
@@ -48,8 +56,229 @@ void MapPlayersViewPacket::read(FileSystem& fs, EditorGameBase& egbase) {
...
+				}
+
+				// Read numerical field infos as combined strings to reduce number of hard disk write
+				// operations
+				std::vector<std::string> field_vector;  // Some data for all player fields
+				std::vector<std::string> data_vector;   // Single complex record for a field
+				std::string parseme;
+
+				// Seeing
+				parseme = fr.c_string();
```
klaus-halfmann, line 100: Please add example of expected string like n|m wheren n and m are ...

src/map_io/map_players_view_packet.cc
```diff
@@ -48,8 +56,229 @@ void MapPlayersViewPacket::read(FileSystem& fs, EditorGameBase& egbase) {
...
+				const MapIndex no_of_seen_fields = fr.unsigned_32();
+				if (seen_fields.size() != no_of_seen_fields) {
+					throw wexception("Read %" PRIuS
+					                 " unseen fields but detected %d when the packet was written\n",
+					                 seen_fields.size(), static_cast<unsigned>(no_of_seen_fields));
+				}
+
+				// Skip data for fields that were never seen
+				if (no_of_seen_fields == 0) {
+					continue;
```
klaus-halfmann, line 122: Comment: How can this happen? does not each player see _some_ fie lds?

src/map_io/map_players_view_packet.cc
```diff
@@ -48,8 +56,229 @@ void MapPlayersViewPacket::read(FileSystem& fs, EditorGameBase& egbase) {
...
+				parseme = fr.c_string();
+				boost::split(field_vector, parseme, boost::is_any_of("|"));
+				assert(field_vector.size() == seen_fields.size());
+
+				size_t counter = 0;
+				for (auto& field : seen_fields) {
+					field->owner = stoi(field_vector[counter]);
+					++counter;
+				}
+
```
klaus-halfmann, line 136: assert (counter == no_of_seen_fields) ??

src/map_io/map_players_view_packet.cc
```diff
@@ -48,8 +56,229 @@ void MapPlayersViewPacket::read(FileSystem& fs, EditorGameBase& egbase) {
...
+					                 " unseen fields but detected %d when the packet was written\n",
+					                 seen_fields.size(), static_cast<unsigned>(no_of_seen_fields));
+				}
+
+				// Skip data for fields that were never seen
+				if (no_of_seen_fields == 0) {
+					continue;
+				}
+
+				// Owner
```
klaus-halfmann: Better: parse list of Owner (IDs ?) seperatred by |

src/map_io/map_players_view_packet.cc
```diff
@@ -48,8 +56,229 @@ void MapPlayersViewPacket::read(FileSystem& fs, EditorGameBase& egbase) {
...
+					assert(data_vector.size() == 4);
+
+					field->border = from_unsigned(stoi(data_vector[0]));
+					field->border_r = from_unsigned(stoi(data_vector[1]));
+					field->border_br = from_unsigned(stoi(data_vector[2]));
+					field->border_bl = from_unsigned(stoi(data_vector[3]));
+					++counter;
+				}
+
+				// Map objects
```
klaus-halfmann, line 232: Do .... with Map objects

src/map_io/map_players_view_packet.cc
```diff
@@ -180,83 +409,173 @@ void MapPlayersViewPacket::read(FileSystem& fs, EditorGameBase& egbase) {
 	}
 }
 
+unsigned to_unsigned(bool value) {
```
klaus-halfmann: inline

src/map_io/map_players_view_packet.cc
```diff
@@ -180,83 +409,173 @@ void MapPlayersViewPacket::read(FileSystem& fs, EditorGameBase& egbase) {
...
-
-			fw.unsigned_8(static_cast<uint8_t>(f.seeing));
-
-			if (f.seeing != SeeUnseeNode::kPreviouslySeen) {
-				continue;
+		// Write numerical field infos as combined strings to reduce number of hard disk write
+		// operations
+		{
+			// Seeing
+			std::ostringstream oss("");
```
klaus-halfmann, line 453: Can we presize oss based on some reasomnable estimation?

klaus-halfmann (changes requested):
Please document the strings emitted - will be needed in case we must examine such a file.

Please either refactor this into smaller fucntions,
or - better - invent some iterator/lambda function aproach to avoid the duplicated code.

I wonder why our streams are not buffered so that we gain such a big speedup?

Will testplay europe now anyway ....

src/map_io/map_players_view_packet.cc
gunchleoc: It's constructor doesn't have a pre-sizing argument, so no, I don't think so.

src/map_io/map_players_view_packet.cc
gunchleoc: I don't understand what you mean.

src/map_io/map_players_view_packet.cc
gunchleoc: Isn't that what I am doing with `boost::split(field_vector, parseme, boost::is_any_of("|"));`?

src/map_io/map_players_view_packet.cc
gunchleoc: Will do :)

src/map_io/map_players_view_packet.cc
klaus-halfmann: Its only about the comment
// Owner -> parse list of Owner (IDs ?) seperated by |

src/map_io/map_players_view_packet.cc
klaus-halfmann: // Map objects -> // Parse the Map Objects from a string like 'X|y|z ..."

src/map_io/map_players_view_packet.cc
gunchleoc: The game is saveloaded before it starts in order to create a backup copy, so it happens there.

klaus-halfmann: continue; // happens in Map Editor or on first file of a replay

src/map_io/map_players_view_packet.cc
gunchleoc: Does something like this do?
```
// Some data for all player fields
// While saving, we iterated all fields for a property and separated the fields by '|'.
std::vector<std::string> field_vector;

// Single complex record for a field
// If a property is multidimensional (e.g. each field has 3 roads), the values are separated by '*'.
std::vector<std::string> data_vector;
```

klaus-halfmann: Fine!

gunchleoc:
> Please either refactor this into smaller fucntions,
> or - better - invent some iterator/lambda function aproach to avoid the duplicated code.

We would end up with a separate function for each loop, because the code is not uniform enough. So, lambda functions like I used for the player statistics code won't add any readability.
 
> I wonder why our streams are not buffered so that we gain such a big speedup?

Me too. I had a quick look and decided not to go there.

src/map_io/map_players_view_packet.cc
gunchleoc: I have added some comments now.

src/map_io/map_players_view_packet.cc
gunchleoc: I'm not using the string trick for the map objects.

klaus-halfmann:
I used the new (great) continue game fature and got:
```
Reading Objective Data ... took 23ms
Reading Scripting Data ... PANIC: unprotected error in call to Lua API (bad permanent value (no value))
Abort trap: 6
```

normal multiplayer worked as expected ....

klaus-halfmann (approved):
Played  Europe with all Slots filled with AIs fore some hours. 
Autosave felt not as bad as before.

gunchleoc:
> I used the new (great) continue game fature and got:
> 
> ```
> Reading Objective Data ... took 23ms
> Reading Scripting Data ... PANIC: unprotected error in call to Lua API (bad permanent value (no value))
> Abort trap: 6
> ```
> 
> normal multiplayer worked as expected ....

That's a different package and doesn't reproduce for every game. So, unrelated to this branch.

src/map_io/map_players_view_packet.cc
```diff
@@ -180,83 +424,173 @@ void MapPlayersViewPacket::read(FileSystem& fs, EditorGameBase& egbase) {
...
+			const MapIndex upper_bound = egbase.map().max_index() - 1;
+			for (MapIndex m = 0; m < upper_bound; ++m) {
+				const Player::Field& f = player->fields_[m];
+				oss << static_cast<unsigned>(f.seeing) << "|";
+				if (f.seeing != SeeUnseeNode::kUnexplored) {
+					seen_fields.insert(&f);
```
Noordfrees, line 459: This is a bit late now, but…
why are you also saving fields in `kVisible` state? These will be rediscovered automatically anyway when the MapObject seeing them is being loaded. I have the feeling that this packet just became a lot slower…

Noordfrees: by changing this I can reduce the write time for Europa with Fogless from now 125s to 7s. Will make a PR

PR #4114, opened 2020-08-13
Remember dismantlesite's state when unseeing a field

### LINKED ISSUE #1344: Dismantle sites in the fog of war look like constructionsites
We had a similar bug for constructionsites before. 

When dismantle sites are seen in the fog of war the base graphic is used. It should display the last known progress picture of the dismantle side. The solution that was implemented for constructionsites is too special to accomodate for dismantle sides. A unified solution should be found.

 * date created: 2011-11-05T20:34:43Z
 * owner: sirver
 * the launchpad url was https://bugs.launchpad.net/bugs/886664

### LINKED ISSUE #4109: Player vision packet is too slow

Noordfrees:
Fixes #1344 
Closes #4109 – along with the necessary saveloading changes I also sped up view saveloading (Europa: 1.6 seconds in a debug build)

### DIFF
--- a/src/logic/player.cc
+++ b/src/logic/player.cc
@@ -1226,7 +1226,7 @@ void Player::rediscover_node(const Map& map, const FCoords& f) {
 		field.border_bl = field.border && bl_vision != SeeUnseeNode::kUnexplored &&
 		                  fields_[bl_index].border && fields_[bl_index].owner == field.owner;
 		{
-			const MapObjectDescr* map_object_descr;
+			const MapObjectDescr* map_object_descr = nullptr;
 			if (const BaseImmovable* base_immovable = f.field->get_immovable()) {
 				map_object_descr = &base_immovable->descr();
 
@@ -1237,6 +1237,7 @@ void Player::rediscover_node(const Map& map, const FCoords& f) {
 					if (building->get_position() != f) {
 						// This is not the building's main position. We don't store the building's
 						// description here but in its main position instead.
+						map_object_descr = nullptr;
 						FCoords main_coords = map.get_fcoords(building->get_position());
 						Field& field_main = fields_[main_coords.field - &first_map_field];
 						if (field_main.seeing != SeeUnseeNode::kVisible) {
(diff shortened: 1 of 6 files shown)


gunchleoc:
I have had a quick look at the saveloading code - mine is better, because it will also speed up map loading if all players saw all nodes. So, we should still have both branches.

Noordfrees:
Okay. Since your branch touches only the saveloading code perhaps you could merge it into this one to make the change as a whole easier to review?

src/logic/player.cc
```diff
@@ -92,6 +93,10 @@ void terraform_for_building(Widelands::EditorGameBase& egbase,
 
 namespace Widelands {
 
+Player::Field::PFB::PFB() {
```
gunchleoc: What does PFB mean?

gunchleoc (approved):
Let's get this one in and I'll merge in the other branch. It's easier for me that way.

src/logic/player.cc
Noordfrees: `PartiallyFinishedBuilding`. This abbreviation is used in a few other places as well. I would have preferred a nameless `union` here since it is used nowhere else but the compiler demanded an explicit constructor.

gunchleoc: Well, if I have to ask, the name is not clear. Can you rename all of them to `PartiallyFinishedBuilding`please?

Niektory:
Tested and working, but #4105 will need changes if unexplored fields are not saved.

Niektory:
Something isn't right:

shot0022

Noordfrees:
Good catch, fixed

Niektory:
Works fine now. Can you add this tweak? No need to run this code for unexplored fields anymore. *-snip-* Actually I'll make this a PR.

PR #4116, opened 2020-08-14
Rebuild show/hide menu when toggling via hotkeys

### LINKED ISSUE #4069: Wrong labels in showhide menu
- Start a new game
- Press S, C, U
- Open the showhide menu in the bottom toolbar. The census, stats, visibility entries will be called "show census" / "show stats" / "hide buildings" rather than "hide census" / "hide stats" / "show buildings".

Niektory:
Fixes #4069

### DIFF
--- a/src/editor/editorinteractive.cc
+++ b/src/editor/editorinteractive.cc
@@ -434,6 +434,7 @@ void EditorInteractive::showhide_menu_selected(ShowHideEntry entry) {
 		toggle_resources();
 	} break;
 	}
+	showhidemenu_.toggle();
 }
 
 void EditorInteractive::load(const std::string& filename) {
--- a/src/editor/editorinteractive.cc
+++ b/src/editor/editorinteractive.cc
@@ -434,7 +434,6 @@ void EditorInteractive::showhide_menu_selected(ShowHideEntry entry) {
 		toggle_resources();
 	} break;
 	}
-	rebuild_showhide_menu();
 }
 
 void EditorInteractive::load(const std::string& filename) {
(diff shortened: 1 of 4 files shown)


Noordfrees:
Tested and working, but there's a regression: If I press U/C/S while the showhide menu is open, it closes. In master it stays open.

Niektory:
> Tested and working, but there's a regression: If I press U/C/S while the showhide menu is open, it closes. In master it stays open.

That (closing the menu when hotkey pressed) was already the behavior of "Show/Hide Building Spaces [Space]" and "Show/Hide Minimap [M]". So I made the rest consistent with those. I understand that all of them should remain open after pressing the hotkeys?

What about on-click behavior? Should it remain unchanged? Hard to tell from the code if it's intended or accidental (the menus that remain open after clicking do so because they are toggled _twice_; and not all in the same way).

Noordfrees:
When clicking, the dropdown closes and then toggle() is invoked to reopen it. IMHO they should always stay open, both when clicking and when pressing a key.

gunchleoc:
We need them to stay open when the player is adjusting the game speed, so keeping them open always would make the behavior consistent.

gunchleoc:
> When clicking, the dropdown closes and then toggle() is invoked to reopen it. IMHO they should always stay open, both when clicking and when pressing a key.

Just in the toolbar though. We'll want them to autoclose in the multiplayer setup screen.

Noordfrees:
That's what I meant :)

Niektory:
Alright, now they don't close when hotkeys are pressed. On-click behavior unchanged (they stay open when clicked).

Noordfrees (approved):
LGTM, tested and working well :)
The only annoyance is that the showhide menu's width changes when the longest entry is toggled but that is not related to this branch.

PR #4117, opened 2020-08-14
Stop text input fields consuming some hotkeys

### LINKED ISSUE #3726: Editbox consumes Ctrl-modfied hotkeys
**Describe the bug**
When the keyboard focus is on an input field, Ctrl-… hotkeys don't work.

**To Reproduce**
1. Open the debug console (F6)
2. Press Ctrl-F → Nothing happens

**Expected behavior**
The hotkey should work.
Hotkeys like 'M' that are also characters should be consumed (as they are), but if Ctrl is down this is *not* a character so it should be treated as a hotkey instead.

**Version:**
 - OS: Ubuntu 19.10
 - Widelands Version: 0427609af0c6ad06311bc4ee841b70c7247b4165

Niektory:
Fixes #3726

Stop text input fields consuming Ctrl-modified hotkeys, F1-F12 function keys and the Pause key.

### DIFF
--- a/src/ui_basic/panel.cc
+++ b/src/ui_basic/panel.cc
@@ -950,11 +950,20 @@ bool Panel::do_key(bool const down, SDL_Keysym const code) {
 		return false;
 	}
 
-	// If we handle text, it does not matter if we handled this key
-	// or not, it should not propagate.
-	if (handle_key(down, code) || handles_textinput()) {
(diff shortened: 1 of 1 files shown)


Noordfrees (approved):
LGTM, tested and working :) Perhaps add PageUp and PageDown to the list?

gunchleoc (approved):

PR #4119, opened 2020-08-14
Rename main productionsite program to 'main'

gunchleoc:
Also:

* Shift check for `ActCall` to after all programs are loaded, so we can call them now anything we like.
* Check that productionsites have a `main` program defined.

### DIFF
--- a/src/logic/map_objects/tribes/production_program.cc
+++ b/src/logic/map_objects/tribes/production_program.cc
@@ -1887,8 +1887,7 @@ const Buildcost& ProductionProgram::recruited_workers() const {
 
 void ProductionProgram::validate_calls(const ProductionSiteDescr& descr) const {
 	for (const auto& action : actions_) {
-		const ActCall* act_call = dynamic_cast<const ActCall*>(action.get());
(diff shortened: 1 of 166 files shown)


src/logic/map_objects/tribes/production_program.cc
```diff
@@ -1891,4 +1884,23 @@ const Buildcost& ProductionProgram::produced_wares() const {
 const Buildcost& ProductionProgram::recruited_workers() const {
 	return recruited_workers_;
 }
+
+void ProductionProgram::validate_calls(const ProductionSiteDescr& descr) const {
+	for (const auto& action : actions_) {
+		const ActCall* act_call = dynamic_cast<const ActCall*>(action.get());
+		if (act_call != nullptr) {
```
Noordfrees:
```suggestion
		if (upcast(const ActCall, act_call, action.get())) {
```

Noordfrees (commented):
Code LGTM, one readability nit. Please convert the emp04 custom units as well

Noordfrees (approved):

PR #4120, opened 2020-08-14
Cmake housekeeping

gunchleoc:
This is what we'd need to put back in to make it work on my trusty.

The eris changes were forced on me by git, because it would not allow me to switch away from the master branch. It's just line endings. So, I guess we'll better cherry-pick my commits if we want this.

I'm not adamantly opposed to dropping Trusty and GCC 4.8. If we want this, I'd like to ask the community first though to make sure nobody of our regulars has an old system like that.


hessenfarmer:
did you try to use the newest FindICU.cmake file from Tinos link and leaving the icu cleanup in our Cmake settings?

gunchleoc:
Yes, it did not work.

TinoM:
To be honest: My vote is against keeping these dedicated cmake files and workarounds for Trusty.
`
https://docs.travis-ci.com/user/reference/trusty/

Trusty is EOL by Canonical, try updating to a newer image and see our Trusty to Xenial Migration Guide.
`
And: we can even keep Trusty around in Travis, because it does still compile widelands (most likely some backport repos).

But i don't mind if you want to merge this.

hessenfarmer:
Finally I got the static_link branch to work on OSX (and most probably on linux as well) However I am afraid that it might not work with the things needed for trusty compatibility in this branch.

hessenfarmer:
What about telling the users where to obtain a suitable version of cmake for trusty (by googling I found a Link to a ppa on the knitware homepage)

gunchleoc:
Can you give me the link for testing?

And if we drop it, I'd be fine with dropping GCC 4.8 as well; then we can get rid of boost::regex.

hessenfarmer:
Well I googled for the verions in differnt distros and got there

https://gitlab.kitware.com/cmake/community/-/wikis/CMake-Versions-on-Linux-Distros   

if you click the ppa link for trusty you get there

https://software.opensuse.org/download/package?package=cmake&project=science%3Adlr

hessenfarmer:
if we could keep it by provding a meaningfull errormessage and a hint in the homepage or if we drop it this should enable the static builds for windows as they are working now.

gunchleoc:
This doesn't work because the public GPG keys for the packages are not available.

https://apt.kitware.com/ no longer supports Trusty either.

I'd say let's discard this PR and drop Trusty support. Which branch do we want to merge first, Windows static build or the original CMake housekeeping?

hessenfarmer:
Well, 
Another alternative would be to point them to https://cmake.org/files/ where they can find a Bash script to install a prebuild binary. May be this would work on trusty for at least v3.7. Although it should work for higher versions as well.

hessenfarmer:
I believe from a cmake point of view both branches are almost identical. So we could go directly for the Static_link branch. However I would like if @TinoM could doublecheck they are identical

gunchleoc:
> Well,
> Another alternative would be to point them to https://cmake.org/files/ where they can find a Bash script to install a prebuild binary. May be this would work on trusty for at least v3.7. Although it should work for higher versions as well.

We have a winner! Instructions are up on the wiki https://www.widelands.org/wiki/Building%20Widelands/#ubuntu-debian

So, we can go ahead now with the cleanup :)

hessenfarmer:
Closed due to other solution identified

PR #4121, opened 2020-08-14
Build MacOS with GitHub

gunchleoc:
* New GitHub action for building latest MacOS with homebrew
* Consolidate all actions except code formatting into 1 yaml file

MacOS version is `10.15`. We're currently building `10.13`and `10.14` on Travis. @stonerl @SirVer @klaus-halfmann Can building `10.14` on Travis builds be dropped safely to speed things up?


.github/workflows/build.yaml
```diff
@@ -82,3 +93,41 @@ jobs:
...
+#        DMGPATH="$(pwd)"
+#        DMGFILE="$(ls *.dmg)"
+#        echo "DMG file is: $DMGPATH/$DMGFILE"
+#        echo "::set-env name=dmg::$DMGPATH/$DMGFILE"
+
+#    - name: Uploading DMG
+#      uses: actions/upload-artifact@v2
+#      with:
+#        name: Widelands ${{ matrix.config }} ${{ matrix.compiler }} AppImage
+#        path: ${{ env.dmg }}
```
Noordfrees: The `env` variables are not being evaluated correctly in this context. Did you try moving the DMG to a hardcoded path (e.g. `${{ github.workspace }}/Widelands-${{ github.sha }}-${{ matrix.config }}.dmg`) and passing that as the artifact `path`?

gunchleoc: I'd need to reprogram `buildapp.sh` for that, which I don't want to touch because I don't have a Mac. Also, the contents of `${{ env.dmg }}` are correct, so I don't know if that would help.

For example, in this run: https://github.com/gunchleoc/widelands/actions/runs/208860783

It complains about

> No files were found with the provided path: /Users/runner/work/widelands/widelands/build_wl/widelands_10.9_1.0~git1[302d788@build-macos].dmg. No artifacts will be uploaded.

The path is identical to the one created by the build step:

```
Creating DMG ...
created: /Users/runner/work/widelands/widelands/build_wl/widelands_10.9_1.0~git1[302d788@build-macos].dmg
DMG file is: /Users/runner/work/widelands/widelands/build_wl/widelands_10.9_1.0~git1[302d788@build-macos].dmg
```

Noordfrees: Just do `mv OLD_DANGEROUS_PATHNAME NEW_WELLDEFINED_PATHNAME` directly after buildapp.sh – unless renaming would damage the DMG? The new pathname needs to be passed to the uploader action without accessing any environment variables

gunchleoc: Thanks, I have pushed an attempt

.github/workflows/build.yaml
```diff
@@ -82,3 +93,41 @@ jobs:
...
+        fetch-depth: 1
+    - name: Building
+      run: |
+        mkdir build_wl
+        cd build_wl/
+        ../utils/macos/build_app.sh ${{ matrix.compiler }} ${{ matrix.config }} ../
+        DMGPATH="$(pwd)"
+        DMGFILE="$(ls *.dmg)"
+        echo "DMG file is: $DMGPATH/$DMGFILE"
+        mv $DMGFILE ${{ github.workspace }}\Widelands-${{ github.sha }}-${{ matrix.config }}-${{ matrix.clang }}.dmg
```
Noordfrees: `matrix.clang` will *not* work ;) Neither will backslashes, probably

gunchleoc: Duh. Fixed.

.github/workflows/build.yaml
```diff
@@ -82,3 +93,41 @@ jobs:
...
+  macos:
+    needs: [clang_tidy]
+    strategy:
+      matrix:
+        config:
+          - release
+          - debug
+        compiler:
+          - clang
+    name: MacOs ${{ matrix.config }} ${{ matrix.compiler }}
```
Noordfrees:
```suggestion
    name: MacOS ${{ matrix.config }} ${{ matrix.compiler }}
```

.github/workflows/build.yaml
```diff
@@ -82,3 +93,41 @@ jobs:
...
+
+  macos:
+    needs: [clang_tidy]
+    strategy:
+      matrix:
+        config:
+          - release
+          - debug
+        compiler:
+          - clang
```
Noordfrees, line 105: Are you planning to add more compilers sometime?

Noordfrees (approved):

.github/workflows/build.yaml
gunchleoc: Yes. Simply adding "gcc" to the matrix isn't sufficient though, so I skipped it in this branch.

Issue #1062, opened 2019-09-09
work area preview doesn´t cover unseen fields


bunnybot:
When a building´s work area is displayed, the circles indicating the area aren´t drawn over fields that have never been seen by the player (are black).
I think this was fixed some time ago and is a regression in current trunk.

------------------------------------
Imported from Launchpad using lp2gh.

 * date created: 2017-10-23T06:41:48Z
 * owner: nordfriese
 * the launchpad url was https://bugs.launchpad.net/bugs/1726269

bunnybot:
(by nordfriese)

bunnybot:
(by sirver)
Thanks for reporting. Will fix soonish.

bunnybot:
(by gunchleoc)
Fixed in build20-rc1

PR #4122, opened 2020-08-14
1344 halfseen dismantlesites tweak

Niektory:
No need to run this code for unexplored fields anymore.


gunchleoc:
I think this part can be shifted as well:

```
		// Draw work area markers.
		if (has_workarea_preview(f->fcoords, &map)) {
			blit_field_overlay(dst, *f, grid_marker_pic_,
			                   Vector2i(grid_marker_pic_->width() / 2, grid_marker_pic_->height() / 2),
			                   scale);
		}
```

Noordfrees:
> I think this part can be shifted as well:

That would regress #1062

gunchleoc:
OK then :+1:

PR #3847, opened 2020-04-13
Overhaul of SP and MP launch game screen

jmoerschbach:
**Update**

I am getting closer... ;) There are still some (minor) issues, see todo list (the very first one is a showstopper imho).
Feedback appreciated :)
yes, there are still some logs which will be deleted before merging... also new classes will replace the old ones.

* fixes #3831
* fixes #3886
* fixed #3928 
* fixes #3942 
* fixes partly #947
* fixes partly  #956

- [x] fix automatic relayout of dropdowns (fixed a bug in basedropdown -> needs extra careful review since I still don't get box layout)
- [x] decide what to do with help button in mp screen (hard coded position in layout())
- [ ] decide whether mp screen should also be completely reset when map changes (as sp now does)
- [x] fine tune paddings
- [x] add missing license header
- [x] SP: map description is not localized
- [x] SP: set tribes according to map defaults
- [x] MP: address scrolling problem when many-player-map is swapped by few-player-map (empty scrolling area) 
- [x] use predefined player names from map
- [x] SP: keep player name when switching player position/number

---

* add scrollbar iff necessary to playerbox
* adapt width so scrollbar is aligned nicely
* extract some local variables

personally, I dislike the forced scrollbar of the chat panel, see screenshot
grafik

Is there a specific reason why it is forced to be displayed? It can be easily be fixed by changing
https://github.com/widelands/widelands/blob/8ca731ae521c49992931c0d9152eb3b3ea2b8163/src/wui/game_chat_panel.cc#L45
to 
`UI::MultilineTextarea::ScrollMode::kScrollLog`

This of course would be visible in the internet lobby, too, since the game_chat_panel is also used there.

What do you think? :)

PR #4125, opened 2020-08-15
Make the Widelands window resizable

### LINKED ISSUE #1836: Should be resizable in windowed mode without any fixed resolutions
With a physical screen of 1600x900, I'd like to play in a window that is whatever is left after panels (Gnome), but of the fixed resolutions available, the largest I can get that fits is 1280x800.  I should be able to get 1600x836 (I think).  Also, if resizing is implemented, I should start in the same resolution I last exited the game with.

 * date created: 2010-07-03T15:39:06Z
 * owner: mjshannon3
 * the launchpad url was https://bugs.launchpad.net/bugs/601403

### LINKED ISSUE #3974: Assert failure in Panel::set_desired_size(int w, int h)
`widelands: ../src/ui_basic/panel.cc:312: virtual void UI::Panel::set_desired_size(int, int): Assertion 'h < 3000' failed.`

I don't see anything strange in the UI when running a release build.

**To Reproduce**
1. Load [stormflood_panel_desired_size.wgf.zip](https://github.com/widelands/widelands/files/4906434/stormflood_panel_desired_size.wgf.zip) in a debug build and attack the enemy

**Version:**
 - OS: Linux Mint, Windows
 - Widelands Version: Build 21

Niektory:
Fixes #1836
Fixes #3974

Mostly works, but some screens don't react properly (the loading screen for example).


src/ui_basic/panel.cc
```diff
@@ -797,6 +795,10 @@ void Panel::do_draw_inner(RenderTarget& dst) {
  * \param dst RenderTarget for the parent Panel
  */
 void Panel::do_draw(RenderTarget& dst) {
+	// TODO(Niektory): Are the asserts needed/correct?
```
gunchleoc: This is the point in the code where panel size matters, so shifting it here is good. Can 4k displays become higher than 3000 though?

gunchleoc (approved):
Don't worry about the loading screen, it doesn't accept any events really. So, that's a different problem.

The Encyclopedia and the savegame screen won't resize/shift when the window resizes. We can handle that later if you want.

Everything else is working fine :)

src/ui_basic/panel.cc
Niektory: As in height? Typically no, but if you rotate them or stack two of them vertically then yes. With this feature you can easily stretch the window over multiple monitors :-)

TinoM (approved):
Tested locally on windows.

src/ui_basic/panel.cc
gunchleoc: I meant both dimensions. How about setting it to 10000 or something? The purpose of the upper bound it to make sure there are no really crazy values, e.g. due to erroneous initialization.

klaus-halfmann:
It was unclear if I have time this week, but it looks like I can spend some time for reviews.
I will start here, but if there are other, more importat MRs around, please assign me two or three.

src/ui_basic/panel.cc
Niektory: Can be 10000, sure. I don't really have an opinion on this. I just had to change it in this PR because it was too easy to trigger it by just resizing the window.

Niektory:
This one is definitely not a priority :-) I'm still tweaking it.

klaus-halfmann:
Works fine on OSX, on all screeens I tested it for now.
Does not work during Map loading, I assume there are to many events on the way as of all the progress messages.
I will do more testplay now. A great Feature!

klaus-halfmann:
* Single Player Lauch Screen result in Layouts of the Window

Unrelated?: at some point all my buildings whereh about 20% transparent.

src/ui_basic/panel.cc
```diff
@@ -797,6 +795,9 @@ void Panel::do_draw_inner(RenderTarget& dst) {
  * \param dst RenderTarget for the parent Panel
  */
 void Panel::do_draw(RenderTarget& dst) {
+	assert(desired_w_ <= std::max(10000, g_gr->get_xres()));
```
klaus-halfmann, line 793: Please comment what this assert does and what this 10000 is used for

klaus-halfmann (approved):
Just one small improvment suggestion.
Looks like our layout algorithms are not that bad.
SDL does a good job on resizing.

Niektory:
OK, I'm done. I got the resolution notifications working with resizing so all the screens that work with Ctrl+F should now also work with resizing. Those that don't will have to be fixed separately.

gunchleoc:
> Unrelated?: at some point all my buildings whereh about 20% transparent.

That's a new feature where you can switch building transparency on and off.

@klaus-halfmann When you have some time, I'd be interested whether our Mac builds work. For example, these ones: https://github.com/widelands/widelands/actions/runs/212279481

frankystone:
Works also on a two monitor setup with different resolutions: 1920x1080 and 1680x1050 :-)

klaus-halfmann (approved):
Now I got the Panel assertion

klaus-halfmann:
Now I get an assertion that is familiar:

4   widelands                     	0x000000010a553945 UI::Panel::set_desired_size(int, int) + 1173 (panel.cc:311)
5   widelands                     	0x000000010a3717c3 UI::Box::update_desired_size() + 3123 (box.cc:143)
6   widelands                     	0x000000010a376719 UI::Box::add(UI::Panel*, UI::Box::Resizing, UI::Align) + 1145 (box.cc:317)
7   widelands                     	0x000000010c969593 UI::SuggestedTeamsBox::show(std::__1::vector<std::__1::vector<std::__1::vector<unsigned char, std::__1::allocator<unsigned char> >, std::__1::allocator<std::__1::vector<unsigned char, std::__1::allocator<unsigned char> > > >, std::__1::allocator<std::__1::vector<std::__1::vector<unsigned char, std::__1::allocator<unsigned char> >, std::__1::allocator<std::__1::vector<unsigned char, std::__1::allocator<unsigned char> > > > > > const&) + 5539 (suggested_teams_box.cc:124)
8   widelands                     	0x000000010a73158d FullscreenMenuLaunchMPG::load_map_info() + 6749 (launch_mpg.cc:618)

Assertion failed: (w >= 0), function set_desired_size, file ../src/ui_basic/panel.cc, line 311.

I start widelands on any of my two monitors, window size: 1280 x 800 (from options)
The Map "Four Castels" causes the issue (In debug builds only AFAIK), same on master.
Not sure if we should fix this here.

Niektory:
> Now I get an assertion that is familiar:

I think #3847 fixes it.

gunchleoc:
I think we can remove those asserts too - the ones in do_draw should be sufficient.

Niektory:
> I think we can remove those asserts too - the ones in do_draw should be sufficient.

OK – I removed them.

Niektory:
Oh wait, I didn't actually remove the ones that caused the problem. Correcting.

PR #4133, opened 2020-08-18
Fix text baseline and vertical alignment

### LINKED ISSUE #590: Font renderer uses wrong baseline for some words
Some words are drawn with a 1 pixel lower baseline, causing the text layout to look irregular. Especially noticeable in campaign messages with long text.

Attached is a screenshot of a tooltip where the baselines of a lowered word and its correctly layouted neighbours are marked in red.

 * date created: 2019-05-03T15:55:49Z
 * owner: nordfriese
 * the launchpad url was https://bugs.launchpad.net/bugs/1827544

Niektory:
Fixes #590.

- Determine the real ascent of a rendered string.
- Valign: When aligning text make sure all text nodes in a line are aligned to the same baseline.
- Valign: Change bottom alignment of images to align to the very bottom, not to the text baseline.
- Valign: Top and center alignment were moving some nodes too far up. This was fixed.


Niektory:
**Before:**
shot0031
shot0029

**After:**
shot0026
shot0028

gunchleoc:
You could try the fixes from https://github.com/widelands/widelands/compare/master...gunchleoc:refactor-wordwrap?expand=1#diff-d8edc7eb2cce36691a369a2d95847412 to see if the lineskip function will give better results.

Niektory:
> You could try the fixes from https://github.com/widelands/widelands/compare/master...gunchleoc:refactor-wordwrap?expand=1#diff-d8edc7eb2cce36691a369a2d95847412 to see if the lineskip function will give better results.

No, I don't think it makes any difference.

Niektory:
I'm going to try to do this properly.

Niektory:
OK, so there are two problems here.

1. [Finding out the text baseline in TextNode::hotspot_y()](https://github.com/widelands/widelands/blob/afba019008c6533c5ebf287d82f119374258ca9b/src/graphic/text/rt_render.cc#L638). SDL_ttf does not always give us text with the same height and ascent. Unfortunately, it also provides no obvious way to find out the ascent for a given chunk of rendered text (or rather, the obvious way gives incorrect results). I've spent way too much time trying to figure it out and this is the best method I could manage – at least it works for the cases I tested. [SDL_ttf bug](https://bugzilla.libsdl.org/show_bug.cgi?id=5263).

2. [Vertical alignment in Layout::fit_nodes()](https://github.com/widelands/widelands/blob/afba019008c6533c5ebf287d82f119374258ca9b/src/graphic/text/rt_render.cc#L552-L567): This code does not preserve the text baseline. It only preserves it when the height and ascent of all the text is the same (which is not always true) or in the bottom-align case (that basically does nothing). Most text in the game seems to be top-aligned though.

So I'd like to ask: How is the vertical alignment supposed to work exactly? What are the use cases? Is `hotspot_y` a meaningful value for nodes other than `TextNode`? Should we special case `TextNode`s to always preserve the baseline? Or just make them bottom-aligned by default?

gunchleoc:
The baseline could make sense when we have italic and bold fonts mixed with normal fonts on the same line. Mixing various font sizes on the same line has always been broken.

Niektory:
To clarify: Because of SDL_ttf behavior, the height of TextNodes can differ even when the font face/size/style is the same. So it's important to align to the baseline.

Niektory:
Just one more question... How should "align to bottom" work? It used to work like in the first example.

bottom_alignment

gunchleoc:
I think the second version is better for align bottom.

Also, descenders aren't perfect yet - it cuts across the background highlight in this screenshot:

descender

Niektory:
> Also, descenders aren't perfect yet - it cuts across the background highlight in this screenshot:

Hmm, it's a separate problem. It's because this font renders a rather tall surface:

shot0057

When this language is selected and applied, the dropdown height increases (though I don't see what's the mechanism for this at first glance) and the name fits:

shot0058

Not sure if it's worth fixing.

gunchleoc (approved):
OK, that would need fixing in the listselect object then and is of low priority, since it should only affect the language selection menu.

hessenfarmer:
Travis error is only the transient SDL error 
so shall we merge this?

PR #4135, opened 2020-08-19
Add Frisians to lua_testsuite map

gunchleoc:
The tribes_filtering branch broke the testsuite. This should fix it.


gunchleoc:
Jobs are going through, so let's have this now as a hotfix.

Issue #4020, opened 2020-07-25
Untranslated strings


s-tdh:
I can't imagine this not being known, but I can't find any bug reporting this:
There are a number of strings that are not translated/translatable:
- Ship names
- Custom buildings in campaigns (Vesta's temple in emp04, Imperial warehouse in fri02 - interestingly the broken forester etc. in emp04 are translated).

klaus-halfmann:
Thanks @s-tdh for finding this and all the other "small" things. Fact is that playtesting is a boring task and checking all translations even more dull. So such errors are only notices by +/- new developers/testers using a foreign language. I must admit I did it in once for relase 20 and after 3 days I gave up. Perhaps we should add a hint "Howto report bugs" in every tutorial, so we get more reports like your one.

Keep on poking us with all the "small" things we should fix, so we can improve this game.

hessenfarmer:
That is indeed not a "small" or trivial one. Custom buildings are translated if they have the same name as the original building, cause they use the translated string of the original Building (emp04 foresters, farms, mills, etc. ) if the strings are altered they are not translated because they don't belong to any textdomain.

Noordfrees:
It would be enough to add `'../../data/campaigns/%(name)s/scripting/*/*.lua',` to `buildcat.py:114` and to set the scenario's textdomain in each of the buildings' `init.lua`s

hessenfarmer:
ok I'll try that and provide a PR.

hessenfarmer:
@Noordfrees 
Would it make sense if you could incorporte this in the stack textdomains branch

gunchleoc:
For the shipnames, I have a branch on that that I never proposed for review: https://github.com/gunchleoc/widelands/tree/localizeable-shipnames

Noordfrees:
Is this branch functional, or what's the reason why you didn't propose it?

Localizing shipnames would make #3947 a tiny bit harder to fix. These two should probably be fixed in one branch IMHO.

gunchleoc:
The reason I didn't propose it is that I have too many branches that I can't get through code review, so I have to prioritize.

PR #4140, opened 2020-08-21
Program documentation

gunchleoc:
* Update Production Program documentation
* Modernize tribe code for empire scenario 4

All that's left to do now is to update the animations' sound priority to percent notation, and then the program API overhaul is done :)


data/campaigns/emp04.wmf/scripting/tribes/temple_of_vesta/init.lua
```diff
@@ -8,6 +8,7 @@ tribes:new_warehouse_type {
    -- TRANSLATORS: This is a building name used in lists of buildings
    descname = pgettext("empire_building", "Temple of Vesta"),
    helptext_script = path.dirname(__file__) .. "helptexts.lua",
```
hessenfarmer: this doesn't work on my Windows machine unless I add "map:" .. in front of the path.dirname

hessenfarmer (approved):

hessenfarmer (changes requested):
see my comment 
see also https://www.widelands.org/forum/topic/4926/
which might be related to that

data/campaigns/emp04.wmf/scripting/tribes/temple_of_vesta/init.lua
gunchleoc: The helptexts will get a different handling soon, but we do need to fix the problem with the animations.

hessenfarmer: well as the fix for the helptext in this particular case is easy we might simply add it in.

gunchleoc: I have deleted the line to avoid merge conflicts with https://github.com/gunchleoc/widelands/commit/66509b76460f5267f110b15218995d4e6c6c9007

gunchleoc:
I have undone the animation changes to the scenario now - better safe than sorry.

data/campaigns/emp04.wmf/scripting/tribes/temple_of_vesta/init.lua
hessenfarmer: Is the undelete of the line intentional. if yes why shouldn't we add just the "map:/" .. string to make it work for the moment being . Otherwise we should reopen #4020

hessenfarmer (approved):
Changes themselves are ok.
not tested though

data/campaigns/emp04.wmf/scripting/tribes/temple_of_vesta/init.lua
gunchleoc: I expect to open a PR with a fix for the helptexts tomorrow.

PR #4148, opened 2020-08-23
Load world units on demand

gunchleoc:
* World map object descriptions are now registered via `register.lua`
* World map object descriptions are loaded on map load
* Split terrains into separate files so that they can be loaded separately
* Reorganize files according to editor category
* `world/init.lua` now defines the editor category members
* Map objects no longer know about their editor categories
* Internal map object names now need to be globally unique
  * Rename all resources, e.g. `coal` -> `resource_coal` to fix conflicts with wares
  * Rename `ashes` terrain to `ashes1` to fix conflict with tribe immovable
* Bottom-align icons in editor tools

I will probably do away with the Tribes and World classes in the future and just use the DescriptionManager for everything. I'll need to take a careful look at how this impacts Lua scripting before I do this, but it would simplify the code.


hessenfarmer (approved):
LGTM. Did not find any obvious errors.
Hard to review though
not tested

gunchleoc:
Sorry, I couldn't make this one any smaller. Thanks for the review!

I think I have done enough testing for this one, including hacking the Atlantean scenario to fill with a terrain not originally present on the map.

PR #4153, opened 2020-08-24
Shift tribe helptexts

gunchleoc:
Shift tribe helptexts to central files in order to move the tribe-dependent information out of units.

This is needed for the wares and immovables, and I have done the same for buildings and workers to keep the design orthogonal.

Deletions of helptext files will follow in a separate branch to keep this diff navigable.


data/tribes/initialization/atlanteans/helptexts.lua
```diff
@@ -0,0 +1,927 @@
...
+--
+-- .. code-block:: lua
+--
+--    push_textdomain("tribes_encyclopedia")
+--    local result = {
+--       buildings {
+--          barbarians_ax_workshop = {
+--             lore = {
+--                -- TRANSLATORS: Lore helptext for a barbarian production site: Ax Workshop
+--                pgettext("barbarians_building", "‘A new warrior’s ax brings forth the best in its wielder – or the worst in its maker.’")
```
Noordfrees: This syntax is a bit awkward. `lore = pgettext("foo", "bar"),` would be more readable to me. How about allowing plain strings as values as well as those tables, at the scripter's choice?

hessenfarmer:
In my opinion it would make sense to have this Information in the units file in the tabele of the relevant instance. Because in the end you have to maintain the same tables for each unit in 2 separate files in this setup

data/tribes/initialization/atlanteans/helptexts.lua
gunchleoc: That sounds doable.

gunchleoc:
> In my opinion it would make sense to have this Information in the units file in the tabele of the relevant instance. Because in the end you have to maintain the same tables for each unit in 2 separate files in this setup

We'd have a problem with xgettext, unless we don't mind having the ship names in the encyclopedia textdomain where they don't belong. Alternatively, we could stick the shipnames into a separate file to solve that problem. I think I like that.

hessenfarmer:
> 
> 
> > In my opinion it would make sense to have this Information in the units file in the tabele of the relevant instance. Because in the end you have to maintain the same tables for each unit in 2 separate files in this setup
> 
> We'd have a problem with xgettext, unless we don't mind having the ship names in the encyclopedia textdomain where they don't belong. Alternatively, we could stick the shipnames into a separate file to solve that problem. I think I like that.

Would be ok for me. Alternatively we could think about merging "tribes" textdomain and "tribes encyclopedia" textdomian. For me they belong closely togehter and I always found it difficult to switch between them while translating to keep consistency. If we feel this gets to big as a whole splitting it up per tribe probably would make sense in terms of pluggability

gunchleoc:
I created the separate textdomains, because it makes it easier for new translators to prioritize. The tribes textdomain is very important, the encyclopedia textdomain is not.

Splitting it up per tribe would not be doable, because multiple tribes share the same wares and immovables.

hessenfarmer:
Ok understood the reason.
what about shifting the ship names to the ships Init.lua per tribe, but I am ok with any other file as well.

gunchleoc:
The ships' init.lua is a great idea, thanks!

gunchleoc:
I have shifted the Barbarian helptexts over to `units.lua`. Does this look good to you?

hessenfarmer:
yep like it that way so we have all relevant information for a tribe in one place.

gunchleoc:
Conversion is done. I still need to update the documentation.

For the Frisians, do we want cornfield, blackrootfield, grapevine etc listed in the encyclopedia? If not, I can remove them from the list and the tribe will still work.

Translations for wares and workers are broken now. Hopefully, https://github.com/widelands/widelands/pull/4170 will fix that.

gunchleoc:
Merging master fixed the wares and worker translations, so this branch is ready for final review now.

data/campaigns/emp04.wmf/scripting/tribes/init.lua
```diff
@@ -12,17 +12,39 @@
...
 --       empire = {
 --          buildings = {
---             "empire_brewery1",
---             "empire_brewery2",
---             "empire_well1",
---             "empire_temple_of_vesta"
+--             {
+--                name = "empire_brewery1",
+--                helptexts = {
+--                   -- TRANSLATORS: Purpose helptext for an empire production site: Brewery
```
hessenfarmer: as this is a documentation comment will this confuse gettext as well?

data/campaigns/emp04.wmf/scripting/tribes/init.lua
```diff
@@ -61,24 +83,99 @@
...
+-- * You can also add ``.png`` files to your subdirectories.
+-- * Adding helptexts is optional, but will units without a helptext will print a warning on the console.
 --
--- See :ref:`defining_tribe_units` for further details on these files.
+-- See :ref:`defining_tribe_units` for further details on these files, and :ref:`lua_tribes_tribes_helptexts` on the helptext format.
+
+-- For formatting time strings
+include "tribes/scripting/help/global_helptexts.lua"
+
+local buildingctxt = "empire_building"
```
hessenfarmer: these variables have been dropped in the tribes units.lua files in the  commit fixing gettetxt errors. Are they safe here?

data/tribes/initialization/atlanteans/units.lua
```diff
@@ -229,228 +326,466 @@ tribes:new_tribe {
...
+            preciousness = 7,
+            helptexts = {
+               -- TRANSLATORS: Helptext for an atlantean ware: Spidercloth
+               purpose = pgettext("atlanteans_ware", "Spidercloth is made out of spider silk in a weaving mill. It is used in the toolsmithy and the shipyard. Also some higher developed buildings need spidercloth for their construction.")
+            }
          },
       },
       {
-         -- Food
+         -- [[ Food ]]
```
hessenfarmer: what is the reason for having these brackets here and not having them for the other categories?

src/logic/map_objects/map_object.cc
```diff
@@ -253,21 +253,21 @@ MapObjectDescr IMPLEMENTATION
...
                                const LuaTable& table)
-   : MapObjectDescr(init_type,
-                    init_name,
-                    init_descname,
-                    table.has_key("helptext_script") ? table.get_string("helptext_script") : "") {
+   : MapObjectDescr(init_type, init_name, init_descname) {
+	if (table.has_key("helptext_script")) {
+		// TODO(GunChleoc): Compatibility - remove after v1.0
+		log_warn("Helptexts script for %s is obsolete - please move strings to "
+		         "tribes/initializations/<tribename>/helptexts.lua",
```
hessenfarmer: needs to be units.lua in the log warning

hessenfarmer (approved):
some small nits
rest looks good
not tested

hessenfarmer:
should be good to go as soon as it is green

PR #4161, opened 2020-08-25
Update CMakeLists.txt

TinoM:
=> https://gitlab.kitware.com/cmake/cmake/blob/master/Modules/FindICU.cmake

This was only added with version 3.7.

Still leaves us with a failing daily ppa build failure for Xenial.


TinoM:
Ok, lets re-include the FindICU module for cmake 3.5 on xenial.

CMakeLists.txt
```diff
@@ -3,6 +3,8 @@ project (widelands)
 cmake_minimum_required (VERSION 3.5)
 cmake_policy(VERSION 3.5)
 
+message(STATUS "Using CMake version $(CMAKE_VERSION)")
```
hessenfarmer: I believe this needs to have the curly brackets to work

hessenfarmer (approved):
LGTM.
Should probably fix the xenial builds. Maybe worth a try on the trusty ppa

Noordfrees:
no point trying Trusty it comes with CMake 2.8.12.2 (besides it's no longer supported anyway)

frankystone:
If an OS version xy isn't officially supported anymore, it shouldn't be supported by widelands. Regardless if it is called windows xy,  linux distribution xy or MacOs xy.

hessenfarmer:
> 
> 
> no point trying Trusty it comes with CMake 2.8.12.2 (besides it's no longer supported anyway)

Well, Trusty was supported officially with a cmake update to 3.5 as far as I know that is why I was thinking about it

gunchleoc (approved):
> Well, Trusty was supported officially with a cmake update to 3.5 as far as I know that is why I was thinking about it

Only if you download and compile CMake manually, which can't be done by the PPA AFAIK, because it's not available from the official Ubuntu package repositories. So, if you have a Trusty you'll have to compile Widelands yourself. At least that was the last state of my testing as I remember it.

TinoM:
Ah, again we are drifting off-topic:
- Trusty (14.04) is in Extended Security Maintanance, so no more public  support
-- Trusty by default comes with cmake 2.8.x, but e.g. the Trusty Images on Travis have a 3.x installed.
-- So if anybody wants to, technically it is possible to build widelands on Trusty
- Xenial (16.04) recieves updates until April 2021, so we should support it 
- So we should not drop support and get the Daily Launchpad PPA for this on track (which is the aim of this merge requests)

Again, only because we do not build widelands (automatically) on any integration platform (Github, Travis, Appveyor,...) it does not mean it is not possible.
It just requires extra effort to get all the tools (cmake, compiler..) on older platforms to the necessary standard and we cannot support (meaning: giving help, adapting buildfiles...) for every combination...


https://ubuntu.com/about/release-cycle

hessenfarmer:
all green so I merged

Noordfrees:
Still failing :(
```
CMake Error at cmake/WlFunctions.cmake:183 (add_library):
  Target "widelands_options" links to target "OpenGL::GL" but the target was
  not found.  Perhaps a find_package() call is missing for an IMPORTED
  target, or an ALIAS target is missing?
  src/CMakeLists.txt:132 (wl_library)

[… lots of more blocks like that …]
```
https://launchpadlibrarian.net/495090761/buildlog_ubuntu-xenial-amd64.widelands_2%3A1.0~bzr9478-202008261246~ubuntu16.04.1_BUILDING.txt.gz

PR #4163, opened 2020-08-25
Load about screen tab contents on demand

### LINKED ISSUE #1695: Do not preload tabs in About screen
At the moment, all tabs in the About screen are preloaded, which means that the user has to wait for all the Lua to run and all texts to be rendered.

It would be better to load each tab on demand.

Investigate if we can do the same for the Encyclopedia (fill the Listselects only on demand) - performance isn't an issue there, but it would make debugging the Lua scripts easier.

 * date created: 2017-12-28T14:09:15Z
 * owner: gunchleoc
 * the launchpad url was https://bugs.launchpad.net/bugs/1740397

gunchleoc:
Closes #1695


Noordfrees (approved):
LGTM, tested and working :)

PR #4166, opened 2020-08-26
Fix an #ifdef in wlapplication messages

gunchleoc:
* Fix compiler warning about `#elif _WIN32`
* Show game data path on MacOS


hessenfarmer (approved):
LGTM

PR #4171, opened 2020-08-27
Get rid of duplicate translation fetching for world

gunchleoc:


src/logic/map_objects/world/world.cc
```diff
@@ -24,6 +24,7 @@
 #include <SDL_surface.h>
 
 #include "base/i18n.h"
```
hessenfarmer: is this include still needed here?

hessenfarmer (approved):
one small nit rest looks good

PR #4182, opened 2020-08-29
Fix description loading in LuaBases

gunchleoc:


Noordfrees (approved):
LGTM. Not tested as I don't see how.

gunchleoc:
For testing, you'd have to write Lua scripts that fetch them for tribes that haven't been loaded. Too much work.

Issue #4180, opened 2020-08-29
Tooltip layouting issues


Noordfrees:
The following screenshots illustrate two issues with very broad (> half the screen width) tooltips:
- When the mouse pointer is near the screen centre, the tooltip does not fit onto the screen when it is displayed to the lefr of the mouse pointer, nor when it is displayed to the right. In this situation it is cut off. **Expected:** it should be displayed with an offset so that it is completely on the screen.
- Even though the tooltip is very broad, the win condition name is wrapped over two lines. It should use only one line here.

![tooltipbug1](https://user-images.githubusercontent.com/48293446/91634999-12ce7100-e9f5-11ea-8378-1a65ce85980d.png)
![tooltipbug2](https://user-images.githubusercontent.com/48293446/91635001-1530cb00-e9f5-11ea-8660-bf8798423f03.png)
![tooltipbug3](https://user-images.githubusercontent.com/48293446/91635006-1661f800-e9f5-11ea-993c-a7c64d2c6acd.png)

PR #4185, opened 2020-08-30
Add terrain affinity to editor tooltips

### LINKED ISSUE #1967: Improve the help system for the editor
It does not seem to be possible to remove immovables in the map editor. This is kind-of frustrating, if one has accidentally put something into a wrong place.

Unless I have read the the editor tutorial poorly, the only solution is to restart from scratch, in which case this could be a documentation bug.

Forum discussion for trees/terrains help: https://wl.widelands.org/forum/topic/1870/

 * date created: 2013-07-24T12:00:33Z
 * owner: kxq
 * the launchpad url was https://bugs.launchpad.net/bugs/1204482

gunchleoc:
* When placing terrains, show preferred trees
* When placing trees, show preferred terrains
* Get rid of obsolete "likes trees" workaround

Closes #1967

Notes:
* I left out the names, because there isn't enough space and I can't control the line breaks to make them clean
* For the trees, the tooltips are sometimes rendered upwards instead of downwards. That's https://github.com/widelands/widelands/issues/4180 and fixing it is out of scope for this branch.


frankystone:
Nice :-)

One nit: Terrain images are not shown in a column. See red marking. This applies also to the tree-tooltip, but here it is not that noticeable.
editor_terrain_tooltips

frankystone:
Maybe add a threshold of trees/terrains to which they are shown. E.g. Show only trees that have at least a chance to grow of 20%? Same for terrains.

I think it is not of interest if a tree on summer water has chance of 1,9%.

gunchleoc:
> Terrain images are not shown in a column.

This is a limitation of our font renderer and can't be fixed in this branch. Showing only 1 column would make the tooltips too tall though.

> Maybe add a threshold of trees/terrains to which they are shown

Good idea, will do.

frankystone:
Looks good now :-)

PR #4188, opened 2020-08-30
Improve UI::Box

### LINKED ISSUE #4062: UI::Box can't handle deletion of child members
See discussion in https://github.com/widelands/widelands/pull/3847 and https://github.com/widelands/widelands/pull/4023

jmoerschbach:
Fixes #4062 

Add support for deletion of panels for UI::Box and ignoring invisible panels while layouting.

I also have a solution according to @gunchleoc proposal https://github.com/widelands/widelands/issues/4062#issuecomment-669097595 with publishing notifications, but this of course notifies ALL existing boxes which in turn need to iterate over their `items_` to find the (one) panel which was deleted and then relayout themselves. To me, this looked like much overhead... Opinions?

I provided a boolean switch to enable/disable ignoring of invisible items while layouting. Imho, this should be removed (set to always ignore invisible items) but as I am not sure if there are any spacing hacks (add a invisible panel for enforcing padding or whatever) I included it for now...

### DIFF
--- a/src/editor/ui_menus/main_menu_map_options.cc
+++ b/src/editor/ui_menus/main_menu_map_options.cc
@@ -241,8 +241,13 @@ MainMenuMapOptions::MainMenuMapOptions(EditorInteractive& parent, Registry& regi
      main_box_(&tabs_, padding_, padding_, UI::Box::Vertical, max_w_, get_inner_h(), 0),
      tags_box_(&tabs_, padding_, padding_, UI::Box::Vertical, max_w_, get_inner_h(), 0),
      teams_box_(&tabs_, padding_, padding_, UI::Box::Vertical, max_w_, get_inner_h(), 0),
-     inner_teams_box_(
-        &teams_box_, padding_, padding_, UI::Box::Vertical, max_w_, get_inner_h() / 2),
+     inner_teams_box_(&teams_box_,
+                      padding_,
+                      padding_,
+                      UI::Box::Vertical,
+                      max_w_,
+                      get_inner_h() / 2,
+                      kSuggestedTeamsUnitSize),
 
      name_(&main_box_, 0, 0, max_w_, UI::PanelStyle::kWui),
      author_(&main_box_, 0, 0, max_w_, UI::PanelStyle::kWui),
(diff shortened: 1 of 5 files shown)


gunchleoc:
> I also have a solution according to @gunchleoc proposal #4062 (comment) with publishing notifications, but this of course notifies ALL existing boxes which in turn need to iterate over their items_ to find the (one) panel which was deleted and then relayout themselves. To me, this looked like much overhead... Opinions?

This is inefficient, so if you have a better solution, you don't have to use my idea at all :)

src/ui_basic/box.cc
```diff
@@ -33,7 +33,8 @@ Box::Box(Panel* const parent,
          uint32_t const orientation,
          int32_t const max_x,
          int32_t const max_y,
-         uint32_t const inner_spacing)
+         uint32_t const inner_spacing,
+         bool const ignore_invisible_items)
```
gunchleoc: Is there a situation where we would not want to ignore them?

jmoerschbach: No, I in general there should be no such situations. However, there are > 100 calls to Panel::setVisible() and for sure I didn't catch all of these during my limited testing. I found it safer to keep the Box' behaviour as before. But maybe it is more straight forward to always ignore invisible items and if some layouting bugs occur because of that, we will fix them...? 
(I'd prefer not to have the boolean flag)

jmoerschbach: After some more testing, I removed the boolean flag. I did not encounter any problems so far :)

jmoerschbach:
Hi @Noordfrees 

as UI::Box can now handle removing panels, I simplyfied the deletion code in the suggested teams box in editor. Please have a look if you are ok with that ;)

src/editor/ui_menus/main_menu_map_options.cc
```diff
@@ -349,16 +354,18 @@ MainMenuMapOptions::MainMenuMapOptions(EditorInteractive& parent, Registry& regi
...
 	teams_box_.add(&inner_teams_box_, UI::Box::Resizing::kFullSize);
+	teams_box_.add_space(padding_);
 	teams_box_.add(&new_suggested_team_, UI::Box::Resizing::kFullSize);
 	new_suggested_team_.sigclicked.connect([this]() {
 		SuggestedTeamsEntry* ste =
 		   new SuggestedTeamsEntry(this, &inner_teams_box_, eia().egbase().map(),
 		                           max_w_ - UI::Scrollbar::kSize, Widelands::SuggestedTeamLineup());
 		inner_teams_box_.add(ste);
-		inner_teams_box_.add_space(kSuggestedTeamsUnitSize);
+		//		inner_teams_box_.add_space(kSuggestedTeamsUnitSize);
```
gunchleoc: Remove this line, then the branch is good to go :)

gunchleoc (approved):

src/editor/ui_menus/main_menu_map_options.cc
jmoerschbach: ?
I already removed it yesterday... ;)

gunchleoc: I was looking at the wrong view then.

matthiakl:
The player configuration in multiplayer selection screen does not properly. After clicking on the player type selection, the row vanishes. By removing the changes in box.cc of this PR everything works fine.

PR #4195, opened 2020-08-31
fixed usage of static glew

### LINKED ISSUE #4186: Windows build won't run
I downloaded a development build from a GitHub action (x64, release) and got this:

glew-dll

Looks to me like the static linking doesn't work.

hessenfarmer:
fixes #4186 
now using target GLEW::glew_s when static build required.


gunchleoc (approved):
Tested and working :)

PR #4199, opened 2020-09-01
Replace boost::regex with std::regex

gunchleoc:


Noordfrees (commented):
Code LGTM. Could you deprecate boost::regex via a codecheck rule?

gunchleoc:
> Code LGTM. Could you deprecate boost::regex via a codecheck rule?

No need, as it won't compile without adding the regex module to CMake again.

Noordfrees (approved):

hessenfarmer:
this fails with clang 5

gunchleoc:
It does. I have done some testing and we need to shift the minimum compiler version to Clang 6. I'll take care of the Travis setup in another branch.

PR #4205, opened 2020-09-02
Reset dropdown list selection when toggled off

### LINKED ISSUE #4204: Dropdowns select entry without click
- Open the options menu
- Open the resolution or language dropdown by clicking on it
- Mouse over any entry in the list (other than the initially selected one) so it is highlighted
- Move the mouse far away (without mousing over any other list entries) so the dropdown auto-closes. The dropdown still shows the previous selection, as expected.
- Click OK. The resolution/language will be set to the value that you last moused over rather than the displayed one.

Niektory:
If a dropdown is toggled off without the user confirming the selection by clicking or pressing space/enter, reset the list selection back to the previously selected value.
Fix #4204

### DIFF
--- a/src/ui_basic/dropdown.cc
+++ b/src/ui_basic/dropdown.cc
@@ -416,6 +416,9 @@ void BaseDropdown::set_list_visibility(bool open) {
 }
 
 void BaseDropdown::toggle_list() {
+	if (list_->is_visible()) {
+		list_->select(current_selection_);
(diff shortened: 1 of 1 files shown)


Noordfrees (approved):
LGTM, tested and working :)

gunchleoc:
Failures are instabilities in Travis, so I'll merge now.

PR #4213, opened 2020-09-03
Fix BaseListselect key handling when no entry selected

Noordfrees:
Fixes the issue mentioned at https://github.com/widelands/widelands/pull/4197#issuecomment-686342278

To reproduce the bug, open the Showhide dropdown in the editor, mouse over any entry, press G (now no entry is highlighted), and then press any key.


gunchleoc (commented):
Code LGTM, not tested yet

Niektory:
Tested, works fine.

gunchleoc (approved):

PR #4214, opened 2020-09-03
Add a Japanese font

Niektory:
I've been experimenting with adding a separate font for Japanese, as well as making it larger.

For rationale see [Unicode FAQ – Chinese and Japanese](https://www.unicode.org/faq/han_cjk.html#2).
Currently Widelands uses a font with Simplified Chinese shapes.
For illustration, some characters that share the same code point in Unicode but look different in Japanese and Simplified Chinese: 曜骨直言練禍突謁
Simplified Chinese and Japanese comparison: 曜骨直言練禍突謁

There are some limitations. It can't be reliably detected what language a given string is in, so it's determined by the locale setting. This causes trouble when using Chinese when Japanese locale is set. Latin text is also rendered inconsistently, but this is solvable.

So whether it's worth the trouble is an open question.


src/graphic/text/bidi.cc
```diff
@@ -725,8 +742,16 @@ UI::FontSet const* find_fontset(const char* word, const UI::FontSets& fontsets)
 	UI::FontSets::Selector selector;
 	if (has_script_character(word, UI::FontSets::Selector::kArabic)) {
 		selector = UI::FontSets::Selector::kArabic;
+	} else if (strcmp(word, "简体中文") == 0 || strcmp(word, "繁體中文") == 0) {
```
gunchleoc, line 745: Please comment that these are the locale names. Not sure if the extra cost is worth the performance hit just for the language selection list.

src/graphic/text/bidi.cc
```diff
@@ -725,8 +742,16 @@ UI::FontSet const* find_fontset(const char* word, const UI::FontSets& fontsets)
...
+		selector = UI::FontSets::Selector::kCJK;
+	} else if (strcmp(word, "日本語") == 0) {
+		selector = UI::FontSets::Selector::kJapanese;
+	} else if (get_locale().rfind("ja", 0) == 0 && has_script_character(word, UI::FontSets::Selector::kJapanese)) {
```
gunchleoc, line 749: `has_script_character` is expensive, so nest this differently so we won't have to run it twice. Something like this:

```
if (has_script_character(word, UI::FontSets::Selector::kCJK)) {
    if (get_locale().rfind("ja", 0) == 0) // Japanese
    else // CJK
} else if (has_script_character(word, UI::FontSets::Selector::kJapanese)) ...
```

gunchleoc (commented):
We only expect mixed locales between Japanese/Chinese in the language selection menu, so that should be fine of the font is a tiny bit off there.

src/graphic/text/bidi.cc
gunchleoc: P.S. or store the result of  `(has_script_character(word, UI::FontSets::Selector::kJapanese))` in a variable to avoid calling it twice.

hessenfarmer:
@Niektory do you have plans to continue working on this?

hessenfarmer:
@widelands/elders 
As @Niektory hasn't responded yet and had no contributions on github for the past 10 months I think we should close his/her PR's unless somebody else volunteers to work on this. Any ideas?

PR #4218, opened 2020-09-05
Include <cstdlib> in src/map_io/coords_profile.cc

ryandesign:
Hi, widelands build21 fails on OS X 10.9 with Apple LLVM version 6.0

```
widelands-build21-source/src/map_io/coords_profile.cc:32:21: error: use of undeclared identifier 'strtol'
        const long int x = strtol(endp, &endp, 0);
                           ^
widelands-build21-source/src/map_io/coords_profile.cc:33:21: error: use of undeclared identifier 'strtol'
        const long int y = strtol(endp, &endp, 0);
                           ^
1 warning and 2 errors generated.
```

This was [reported to MacPorts](https://trac.macports.org/ticket/61028).

According to `man 3 strtol`, one must include <stdlib.h> to access this function. (Perhaps on later OS/LLVM versions that happens indirectly.)

### DIFF
--- a/src/map_io/coords_profile.cc
+++ b/src/map_io/coords_profile.cc
@@ -19,6 +19,8 @@
 
 #include "map_io/coords_profile.h"
 
+#include <cstdlib>
+
 #include "base/wexception.h"
 #include "io/profile.h"
 #include "logic/widelands_geometry.h"


ryandesign:
`strtol` is used in coords_profile.cc but I put the `#include <stdlib.h>` into coords_profile.h because I didn't see any existing system includes in coords_profile.cc but there is one in coords_profiles.h so I assumed that's where you want them. If you'd rather have the include directly in coords_profile.cc let me know.

ryandesign:
Apologies, your codecheck suite didn't like my change; it said:

> You have included a deprecated C header. Remove it!

I am not very familiar with C++. I believe perhaps I should have included `<cstdlib>` instead; I will try that.

gunchleoc:
Yes, put ` <cstdlib>` into the cc file.

We have separate include blocks depending on include type:

1. First, if it's a cc file, its header file
2. C++ libs
3. Other libs
4. Our own code

There muse be a blank line between those blocks, or our codecheck will complain.

ryandesign:
> If you'd rather have the include directly in coords_profile.cc let me know.

I did move it to coords_profile.cc since that's where it's used.

gunchleoc (approved):

PR #4223, opened 2020-09-05
Add Codecheck rule for <cstdlib> to make libc++ happy

gunchleoc:
@ryandesign This should prevent regressions of this one.


ryandesign:
Thanks.

I would expect this issue to affect any operating system using older versions of clang, not just Macs.

With the three previous PRs I submitted, I was able to compile widelands on OS X 10.9 without needing the additional `#include`s added here, but if those functions are used in those files, then I agree it's probably good form to include the headers in which those functions are defined.

gunchleoc:
`master` is building fine with clang 5 and 6 on Linux, so I don't know why you had problems with clang 6.

ryandesign:
Using libc++ or libstdc++ on Linux? This was with libc++ on OS X.

gunchleoc:
No idea, whichever CMake picks on a default Ubuntu installation. We only ask for `-static-libstdc++` explicitly for Windows.

ryandesign:
Ok. Well, if Ubuntu uses libstdc++, then that could be a relevant difference that might explain why you don't see the issue there.

cmake/codecheck/rules/include_cstdlib.py
```diff
@@ -0,0 +1,78 @@
...
+"""Checks that cstdlib is included for some of its functions that we use
+commonly.
+
+Older Macs that build with libc++ instead of libstc++ need this.
+"""
+
+import re
+
+FUNCTION_REGEX = re.compile(
+    r"""(\s|std::)(abs|ato[fil]|atoll|strto[dfl]|strtol[ld]|strtoul[l]{0,1}|std::[s]{0,1}rand|malloc|getenv)\(""")
```
Niektory:
```suggestion
    r"""((\s|std::|\(|\[)(abs|ato[fil]|atoll|strto[dfl]|strtol[ld]|strtoul[l]{0,1}|malloc|getenv)|(\bstd::[s]{0,1}rand))\(""")
```
This will catch a little more, for example `(strtol(foo) + bar)`

cmake/codecheck/rules/include_cstdlib.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def does_include_cstdlib(lines, fn):
+    includes_cstdlib = False
+    for lineno, line in enumerate(lines, 1):
+        if 'include <cstdlib>' in line:
```
Niektory:
```suggestion
        if line.startswith('#include <cstdlib>'):
```
No commented out includes for you :P

Niektory (approved):
LGTM, suggested minor improvements.

Noordfrees:
Is the error message meant to look like this?
```
/home/benedikt/wl/splashscreen/src/ui_basic/progresswindow.cc:106: This file uses  but does not include <cstdlib>.
```

Niektory:
Well it does seem a little bit off :-P
Fix: #4264

Issue #3223, opened 2019-09-09
Use clang-tidy to clean up the codebase


bunnybot:
I found another cleanup tool: http://clang.llvm.org/extra/clang-tidy/

I'll attach a script with instructions.

This is generating a lot or warnings, it's probably best if we run this one check at a time. Some of the checks also have a -fix option to fix stuff automatically.

------------------------------------
Imported from Launchpad using lp2gh.

 * date created: 2017-11-08T19:55:54Z
 * owner: gunchleoc
 * the launchpad url was https://bugs.launchpad.net/bugs/1731034

bunnybot:
(by sirver)
That tool is written by some of the people in my office :).

bunnybot:
(by gunchleoc)
Thank them for a great tool then!

I found it via this post: https://forum.freegamedev.net/viewtopic.php?p=74965#p74965 which links to the following video about common bugs: https://www.youtube.com/watch?v=lkgszkPnV8g

Here are the stats on first run:

   1 [cert-dcl03-c]
  11  [cert-dcl50-cpp]
   1  [cert-err52-cpp]
  51  [cert-err58-cpp]
   2  [cert-err61-cpp]
  16  [clang-analyzer-alpha.core.CastToStruct]
   1  [clang-analyzer-alpha.core.IdenticalExpr]
  12  [clang-analyzer-alpha.core.SizeofPtr]
 211  [clang-analyzer-alpha.deadcode.UnreachableCode]
  10  [clang-analyzer-alpha.security.ArrayBound]
   1  [clang-analyzer-alpha.security.ArrayBoundV2]
   4  [clang-analyzer-alpha.unix.cstring.OutOfBounds]
  10  [clang-analyzer-alpha.unix.Stream]
   4  [clang-analyzer-core.CallAndMessage]
   2  [clang-analyzer-core.DivideZero]
   3  [clang-analyzer-core.NonNullParamChecker]
   2  [clang-analyzer-core.NullDereference]
   1  [clang-analyzer-core.UndefinedBinaryOperatorResult]
   1  [clang-analyzer-core.uninitialized.Assign]
   2  [clang-analyzer-cplusplus.NewDelete]
   7  [clang-analyzer-cplusplus.NewDeleteLeaks]
   8  [clang-analyzer-deadcode.DeadStores]
   2  [clang-analyzer-security.insecureAPI.strcpy]
   1  [clang-diagnostic-unused-const-variable]
[...]

PR #4224, opened 2020-09-06
Get rid of using namespace in scripting directory

gunchleoc:
Re. https://github.com/widelands/widelands/issues/3223


Noordfrees (approved):

Noordfrees:
Travis failure is the usual transient SDL error

PR #4183, opened 2020-08-30
Fix window resizing problems

Niektory:
- Do not save window size on every resize event to avoid saving a maximized size. Instead save it only on application shutdown and in the options screen, and only if not maximized or fullscreen. (Fix #4151)
- Track the maximized state separately so we can save and restore it.
- Temporarily set the window as not resizable whenever we resize it ourselves. SDL can lose track of the real window size if we don't do this. See SDL bugs [2452](https://bugzilla.libsdl.org/show_bug.cgi?id=2452), [4574](https://bugzilla.libsdl.org/show_bug.cgi?id=4574). (Fix #4160)
- Remove some unnecessary window resizes.
- Add *Maximized* and *Fullscreen* to the *Window Size* dropdown; remove the *Fullscreen* checkbox.

PR #4226, opened 2020-09-06
Set SDL minimum requirement to v2.0.5

gunchleoc:
c.f. discussion in https://github.com/widelands/widelands/pull/4183


gunchleoc:
We need to support SDL 2.0.4. for Xenial, which has been addressed in #4183. Closing this one as obsolete.

Issue #1092, opened 2019-09-09
Toolbar images need replacing


bunnybot:
The toolbar frame images are just placeholders and need replacing.

For details, see https://www.widelands.org/forum/topic/4602/

------------------------------------
Imported from Launchpad using lp2gh.

 * date created: 2019-08-10T17:50:55Z
 * owner: gunchleoc
 * the launchpad url was https://bugs.launchpad.net/bugs/1839724

frankystone:
I think stonerl has created some images. They can be improved, but this bug isn't valid anymore i think.

PR #4231, opened 2020-09-06
Frisian toolbar icons

Noordfrees:
Re #1092
shot0000


gunchleoc:
I think it could do with some desaturation - it is a bit too distracting and draws the eye away from the button icons.

gunchleoc:
Now the flag is transparent and the logo is not - the logo needs fixing now, because it is too noisy.

gunchleoc (approved):
Can you remove the toolbar section from the Atlanteans before you merge? It's only there to ensure that the code path is executed, which the Frisians will now do for us.

PR #4242, opened 2020-09-09
Optimize ObjectManager::object_still_available()

Niektory:
This significantly speeds up huge games like the one from this thread: https://www.widelands.org/forum/topic/4938/


gunchleoc (approved):
Thanks for finding this 13 year old gem :)

Issue #1558, opened 2019-09-09
Code cleanup: change struct to class where appropriate


bunnybot:
We currently have a bit of a mix between structs and classes, where classes would be more appropriate.

http://google-styleguide.googlecode.com/svn/trunk/cppguide.html#Structs_vs._Classes

Use a struct only for passive objects that carry data; everything else is a class.

The struct and class keywords behave almost identically in C++. We add our own semantic meanings to each keyword, so you should use the appropriate keyword for the data-type you're defining.

structs should be used for passive objects that carry data, and may have associated constants, but lack any functionality other than access/setting the data members. The accessing/setting of fields is done by directly accessing the fields rather than through method invocations. Methods should not provide behavior but should only be used to set up the data members, e.g., constructor, destructor, Initialize(), Reset(), Validate().

If more functionality is required, a class is more appropriate. If in doubt, make it a class.

For consistency with STL, you can use struct instead of class for functors and traits.

Note that member variables in structs and classes have different naming rules.

------------------------------------
Imported from Launchpad using lp2gh.

 * date created: 2014-09-30T07:45:18Z
 * owner: gunchleoc
 * the launchpad url was https://bugs.launchpad.net/bugs/1375624

gunchleoc:
See also https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c2-use-class-if-the-class-has-an-invariant-use-struct-if-the-data-members-can-vary-independently

PR #4244, opened 2020-09-09
Improve debug window

### LINKED ISSUE #4167: Debug window won't scroll
The debug window aggressively scrolls to the top of the window.

To reproduce: Use a debug build, click on a field, choose the Watch tab, click the DBG button, try to scroll down in the field-debug window.

matthiakl:
* Fix scrolling (#4167): update text only when changed
* Open bobs with double click: with selection every bob is opened when scrolling through using the mouse wheel
* Change structs to classes (relates to #1558)

### DIFF
--- a/src/wui/game_debug_ui.cc
+++ b/src/wui/game_debug_ui.cc
@@ -37,19 +37,6 @@
 #include "ui_basic/window.h"
 #include "wui/interactive_base.h"
 
-struct MapObjectDebugPanel : public UI::Panel, public Widelands::MapObject::LogSink {
-	MapObjectDebugPanel(UI::Panel& parent, const Widelands::EditorGameBase&, Widelands::MapObject&);
-	~MapObjectDebugPanel() override;
-
-	void log(const std::string& str) override;
-
-private:
-	const Widelands::EditorGameBase& egbase_;
-	Widelands::ObjectPointer object_;
-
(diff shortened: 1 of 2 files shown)


src/wui/game_debug_ui.cc
```diff
@@ -339,7 +281,10 @@ void FieldDebugWindow::think() {
 		}
 	}
 
-	ui_field_.set_text(str.c_str());
+	if (str.compare(text_) != 0) {
```
gunchleoc:
```suggestion
	if (!text_.empty()) {
```

gunchleoc (commented):
1 code style nit, rest LGTM :)

src/wui/game_debug_ui.cc
matthiakl: Then the text would only be set during the first call, but will not update if the field properties change, while the window is open.

gunchleoc: I can never remember what 0 means in the string compare... `if (text_ == str)`? Strings have a direct equality operator these days, so there's no need for using the compare function.

matthiakl: 0 means equal, therefor `(str.compare(text_) != 0)` is the same as `(str != text_)`

gunchleoc (approved):

Issue #3585, opened 2019-11-20
CIs instable


jmoerschbach:
Please fill out the relevant sections below and delete the rest.

**Describe the bug**
CIs (especially Travis) oftentimes fail where they should not. Usually it is only one or two of the many build configurations that fail while all the others succeed.

**To Reproduce**
1. create a pull request to trigger CI

**Expected behavior**
If no real problem exists (compile error, formatting problem, failing test due to codechange), all build configurations should succeed.

**Version:**
Travis CI environment

**Additional context**
See for example

- https://travis-ci.org/widelands/widelands/jobs/614169005?utm_medium=notification&utm_source=github_status

- https://travis-ci.org/widelands/widelands/jobs/614278136?utm_medium=notification&utm_source=github_status

It is always saying
Caught exception (of type '10WException') in outermost handler!

**The exception said: [/home/travis/build/widelands/widelands/src/wlapplication.cc:371] Failed to initialize SDL, no valid video driver: No available video device**

This should not happen. Please file a bug report on version r52[9c09168@HEAD](Release).

and remember to specify your operating system.

stonerl:
I'm not sure whether we can change any of this or not. Most of the time (90%) it is the xfvb.service that is, for some reason, not running anymore and therefor letting the test fail.

jmoerschbach:
is there really nothing we can do? Almost every time a CI build is triggered, at least one of the build configuration fails. quite annoying...

gunchleoc:
Since we can't fix the instabilities in the CI environment, I have changed the code to log and exit(0) rather than throw an exception.

stonerl:
Well, it kind of worked, I guess. But I assume we can't do anything when xvfb dies during the test suite run?

[https://travis-ci.org/widelands/widelands/jobs/629354906](https://travis-ci.org/widelands/widelands/jobs/629354906)

gunchleoc:
We could use a different return code and check for that return code when running the test suite to make it ignore the problem.

PR #4245, opened 2020-09-09
Handle video problems of helper applications in CI environment

matthiakl:
relates to #3585

The CI instability was only fixed for the main application. In recent builds the same issue occurs for the website tools.
So they should exit the same way.
This might also require an adaption in the build script to process the exit  code.

Examples:
* https://travis-ci.org/github/widelands/widelands/jobs/725022315
* https://travis-ci.org/github/widelands/widelands/jobs/724618982

```
[100%] Built target wl_map_object_info
1856 +++'[' linux = linux ']'
1857 +++cd ..
1858 +++'[' ON = ON ']'
1859 +++mkdir temp_web
1860 +++build/src/website/wl_map_object_info temp_web
1861 [00:00:00.000 real] INFO: selected language: en
1862 [00:00:00.002 real] INFO: using locale en_US.utf-8
1863 [00:00:00.054 real] ERROR: Exception: [/home/travis/build/widelands/widelands/src/website/website_common.cc:35] Unable to initialize SDL: No available video device.
1864 The command "./.travis.sh build" exited with 1.
```

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -1,4 +1,17 @@
-set -ex
+set -x
+trap 'err_handler' ERR
+
+# Called if a command returns a non-zero value
+err_handler() {
+   result=$?
+   if [ $result -eq 2 ] ; then
+      # Catch instabilities with SDL in CI environment
+      echo "SDL initialization failed. TEST SKIPPED.";
+   else
+      # Exit with this error
+      exit $result
+   fi
+}
 
 # Create build folder.
 mkdir build
--- a/src/website/CMakeLists.txt
+++ b/src/website/CMakeLists.txt
@@ -8,6 +8,7 @@ wl_library(website_common
   DEPENDS
     base_exceptions
     base_i18n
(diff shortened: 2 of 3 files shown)


gunchleoc:
Yes, the return code needs handling in .travis.sh. You can read it by comparing `$?` directly after the test call. Here's an example from `utils/merge_and_push_translations.sh`:

```
# Update authors file
utils/update_authors.py
if [ $? -eq 0 ] ; then
  echo "Updated authors";
else
  echo "Failed updating authors";
  exit 1;
fi
```

matthiakl:
Didn't find the build-script in the tree at first. It should work now.
See https://stackoverflow.com/a/19622569 as reason for err_handler.

gunchleoc (approved):

gunchleoc:
Mac error is an upload timeout, so let's have it. Thanks for fixing :)

PR #4247, opened 2020-09-10
Fix heap use after free in ShipFleet

Niektory:
I can only reproduce the issue by loading this old savegame saved using the Amazons branch and then quitting: [406.wgf.zip](https://github.com/widelands/widelands/files/5198430/406.wgf.zip) It still loads in master though.

The change to `object_still_available()` in 0de9148ec8c1c2598c25c10dff4b6fd9d083797f made the old workaround not work anymore. I investigated what was causing the problem and found out that two instances of the same ship were in one ShipFleet. When cleaning up only one was being removed, the other becoming a dangling pointer.


gunchleoc (approved):
I's not obvious why the ship is being added twice in the first place, so code LGTM

Issue #3686, opened 2020-02-12
Add option to the main menu to change speed


stonerl:
Followup for #3458 

Something like this had been present already in [build 14](https://www.widelands.org/wlmedia/wlscreens/screens/Current%20Development/Advanced%20options%20menu.jpeg)

![Advanced options menu](https://user-images.githubusercontent.com/2091312/74352963-87bfbd80-4db9-11ea-8e97-025a70f91917.jpeg)

Noordfrees:
The poll's purpose sounded to me as if the default speed increase should apply to both SP and MP. So perhaps each player's desired speed should be initially set to this value in MP.

IMHO the setting should have steps of 0.25×, with a default of 1×, a minimum of 0.25× (0 means the game starts with a black screen, which would look like a bug), and a reasonable maximum like 5× or 10×. 
This step size might require some tweaking to spinbox code though.

stonerl:
Changed the title 😀

PR #4262, opened 2020-09-12
Default show hide options

### LINKED ISSUE #3265: Show Buildhelp should be a display flag
Buildhelp is currently implemented as a boolean. It should be a display flag instead, just like dfShowCensus etc.

Same for draw_resources_, draw_immovables_, draw_bobs_ in the editor.

This will give us a consistent implementation and then make it easier to define preferences, e.g. implementing the buildhelp state into singleplayer saveloading could be done without incrementing the packet version.

 * date created: 2018-04-18T05:27:09Z
 * owner: gunchleoc
 * the launchpad url was https://bugs.launchpad.net/bugs/1764942

### LINKED ISSUE #4073: Add starting options for show/hide menu
At the moment, many players hit space, c, s on every game start. It would be nice to have options to set defaults for all entries in the Show/Hide menu.

matthiakl:
Closes #4073
This adds checkboxes to the game options tab to select the default show/hide states when starting a new game.
The current states are saved in displayflags and restored when loading a savegame (this was only partially implemented).


Niektory:
The defaults set in the options work fine when starting a game. Saving the states in a savegame works in single player but not in multiplayer.

gunchleoc:
That's because the `GameInteractivePlayerPacket` isn't loaded in multiplayer games.

src/ui_fsmenu/options.cc
```diff
@@ -222,6 +223,14 @@ FullscreenMenuOptions::FullscreenMenuOptions(FullscreenMenuMain& fsmm,
      game_clock_(&box_game_, Vector2i::zero(), _("Display game time in the top left corner")),
      numpad_diagonalscrolling_(
         &box_game_, Vector2i::zero(), _("Allow diagonal scrolling with the numeric keypad")),
+     showhide_info_(
+        &box_game_, 0, 0, 100, 0, UI::PanelStyle::kFsMenu, _("Show / Hide default values")),
```
gunchleoc:
```suggestion
        &box_game_, 0, 0, 100, 0, UI::PanelStyle::kFsMenu, _("Show / Hide options for new games")),
```

src/ui_fsmenu/options.cc
```diff
@@ -426,6 +452,14 @@ void FullscreenMenuOptions::layout() {
 		game_clock_.set_desired_size(tab_panel_width, game_clock_.get_h());
 		numpad_diagonalscrolling_.set_desired_size(
 		   tab_panel_width, numpad_diagonalscrolling_.get_h());
+
+		showhide_info_.set_desired_size(tab_panel_width, show_buildhelp_.get_h());
```
gunchleoc: Can you try adding these and all other similar elements with a different resize option to the box (kFullsize of kFillspace, I can never remember which is which)? We should then be able to get rid of all these set_desired_size calls.

gunchleoc (commented):
2 nits, rest LGTM

src/ui_fsmenu/options.cc
matthiakl: It does not work for all elements, but for checkboxes, which we mostly have, it does.

src/game_io/game_interactive_player_packet.cc
```diff
@@ -74,11 +74,7 @@ void GameInteractivePlayerPacket::read(FileSystem& fs, Game& game, MapObjectLoad
 			if (InteractiveBase* const ibase = game.get_ibase()) {
 				ibase->map_view()->scroll_to_map_pixel(center_map_pixel, MapView::Transition::Jump);
 
-				uint32_t const loaded_df =
-				   InteractiveBase::dfShowCensus | InteractiveBase::dfShowStatistics;
-				uint32_t const olddf = ibase->get_display_flags();
-				uint32_t const realdf = (olddf & ~loaded_df) | (display_flags & loaded_df);
-				ibase->set_display_flags(realdf);
+				ibase->set_display_flags(display_flags);
```
Noordfrees, line 81: Shouldn't the `dfDebug` flag get special treatment here?

matthiakl: Good point. Saving a game in debug build and opening it in release shows the debug item.
I fixed it by ignoring the saved value and set it dependent on the build type.

Noordfrees:
Tested and working, very good :) But the options screen is looking a bit too cluttered now IMHO – move the new checkboxes into a new tab? 
And add a comment in interactive_base.h to remind coders who add new display flags to add appropriate checkboxes in the options please.

matthiakl:
I don't find it too cluttered. Maybe it helps adding some space above the show/hide options? With a reasonable resolution there is enough free space in the tabs that can be filled (in small resolution everything is squeezed together anyway).

Any suggestion how to name the new tab? Maybe "View"?

Noordfrees:
Several of the devs use Widelands at small resolutions, so is needs to look good at all window sizes. The default game speed spinner for #3686 should also go in that tab when it gets implemented, so how about calling this tab "New Games" (and renaming the "Game" tab to "In-Game" to avoid confusion)?

Noordfrees (approved):
Great :)

Niektory:
Hmm, now when I load Build 21 savegames the buildings are always hidden.

matthiakl:
That's because the ShowBuildings flag wasn't present in build 21, so when loading the game its value is 0. To keep the buildings visible the packet version needs to be incremented and a default flag value of 1 applied to older versions.

frankystone:
I get this also for new games. See #4291

Noordfrees:
I guess `kDefaultDisplayFlags` should be added as default value to `get_config_int` in `interactive_base.cc:181`

Issue #4253, opened 2020-09-11
Player vision code is very slow


Niektory:
And I mean **very** slow. Profiling shows it is the biggest CPU eater right now by far.

Here's the problem. Every time a bob moves, its entire view area from its old position is recalculated:
https://github.com/widelands/widelands/blob/ff36ede854ab98cab34e20d6b2c2a3002c38a27b/src/logic/map_objects/bob.cc#L929-L934
That means every field separately:
https://github.com/widelands/widelands/blob/ff36ede854ab98cab34e20d6b2c2a3002c38a27b/src/logic/player.cc#L1486-L1489
And for each of these fields, iterate through all of our+allied buildings+bobs, and calculate if this field is in their field of vision:
https://github.com/widelands/widelands/blob/ff36ede854ab98cab34e20d6b2c2a3002c38a27b/src/logic/player.cc#L1419-L1445

That gives us up to `(moving bobs) * (π * radius^2) * (bobs+buildings)` vision checks. Oops!

**Potential solutions:**

1. Cache the number of object that see each field so there's no need to calculate it all the time. Basically what was done before 02947d690dae1f87131a6db4e686c047939cbdab (#4046).
2. Reduce the number of checked fields by not checking the fields that are seen from the moving bob's new position.
3. When checking a group of nearby fields and finding that a seer sees one of them, check that seer first in subsequent iterations.

**ADDED:**

4. When evaluating an area, do an initial iteration calculating the distance of all seers to the center field and put the ones that are close into a list. Then for the other fields only check seers from that list.
5. Divide the map into sectors. When checking visibility, only consider seers from the same and adjacent sectors.
6. Iterate seers belonging to the field's owner first.

I think option 1 is the only real solution. The other options could give good gains in an average case but would still leave big lag spikes in some scenarios.

**Version:** ff36ede854

Noordfrees:
It used to count only the number of map objects that see each field. This was very fragile so I removed it. I favour option 2. Option 3 also sounds good but a lot of effort. We could additionally tweak the iteration of player teams a bit so the seers belonging to the field's owner (if he's in the player's team) are iterated first.

Niektory:
[...]

PR #4270, opened 2020-09-15
Optimize player vision, part 1

Niektory:
See #4253

I decided to go for this optimization first, because... well, mostly because it's relatively easy ;-) and I was curious how much it would speed things up.

It works by building a list of nearby objects when updating an area, so all objects are only evaluated once per area, instead of once per field. When updating individual fields only this list of nearby objects is evaluated, which greatly reduces the number of iterations on large maps.

**Benchmark of this branch**

See https://github.com/widelands/widelands/issues/4253#issuecomment-692189232 for comparison. It's roughly 20 times faster than master on average (or 500 times faster when dismantling towers), but still 170 times slower than build-21 on [the tested savegame](https://github.com/widelands/widelands/files/5219868/406.wgf.zip).

Time spent in player vision code per 1 second gametime:

- Release
gametime 51,163,247 - 51,164,252 took 3,081,653 µs
gametime 51,164,252 - 51,165,253 took 571,586 µs
gametime 51,165,253 - 51,166,255 took 598,782 µs
gametime 51,166,255 - 51,167,255 took 619,366 µs
gametime 51,167,255 - 51,168,256 took 586,841 µs

- Debug
gametime 51,163,247 - 51,164,252 took 14,686,967 µs
gametime 51,164,252 - 51,165,253 took 2,579,212 µs
gametime 51,165,253 - 51,166,255 took 2,723,073 µs
gametime 51,166,255 - 51,167,255 took 2,924,197 µs
gametime 51,167,255 - 51,168,256 took 2,641,972 µs

Tower dismantling test:

- Release
remove_seer() at (37,235) took 6,019 µs
remove_seer() at (4,248) took 3,158 µs

- Debug
remove_seer() at (37,235) took 17,339 µs
remove_seer() at (4,248) took 8,877 µs


Niektory:
Using static casts here gives a 25% speed boost but clang_tidy complains. Should I not use them? Ideally we wouldn't need to cast here but changing it is out of scope of this branch.

src/logic/player.cc
```diff
@@ -1479,14 +1484,52 @@ void Player::update_vision(const Area<FCoords>& area, bool force_visible) {
 	if (egbase().objects().is_cleaning_up()) {
 		return;
 	}
+
+	// Build a list of nearby objects.
+	// This way we only evaluate all objects (which is very expensive) once.
+	// When updating individual fields, only the objects from this list are evaluated.
+	std::list<std::pair<int, const MapObject*>> nearby_objects;
```
Noordfrees, line 1529: Pull the code to build the list out into a new function `seers_for(FCoords)`

src/logic/player.cc
```diff
@@ -1479,14 +1484,52 @@ void Player::update_vision(const Area<FCoords>& area, bool force_visible) {
...
+				const Bob* b = static_cast<const Bob*>(seer);
+				int dist =
+				   egbase().map().calc_distance(area, b->get_position()) - b->descr().vision_range();
+				if (dist <= area.radius) {
+					nearby_objects.push_back(std::make_pair(dist, b));
+				}
+			}
+		}
+	}
+	// Sort the list, so the objects that are more likely to be a hit are evaluated first.
```
Noordfrees: How much is the performance gain by sorting them, vs the additional complexity introduced by performing the sorting?

src/logic/player.cc
```diff
@@ -1418,39 +1417,45 @@ void Player::remove_seer(const MapObject& m, const Area<FCoords>& a) {
...
-				upcast(const Bob, b, mo);
-				assert(b);
-				if (egbase().map().calc_distance(f, b->get_position()) <= b->descr().vision_range()) {
-					return true;
-				}
+	}
+	for (const MapObject* mo : nearby_objects) {
+		if (mo->descr().type() >= MapObjectType::BUILDING) {
+			assert(is_a(Building, mo));
+			const Building* b = static_cast<const Building*>(mo);
```
Noordfrees: How about keeping two lists of seers, one for `Building*` and one for `Bob*`, so we don't need to upcast all the time?

Niektory: For some reason I didn't want to do it but on second thought that should be worth it.

src/logic/player.cc
Niektory: I didn't keep the numbers but AFAIR the gain was negligible in the average case, but substantial in the dismantling tower test. Basically it benefits big area seers, should also help with scouts (the tested savegame didn't have them) etc.

src/logic/player.cc
Niektory: Ah, I remember the reason. It would be tricky to have two lists without duplicating the code in multiple places. I thought a better solution would be to make a new type for seers. And I decided against it because I wanted to keep this diff (and the work involved) small as it's just a temporary solution that will be replaced later.

src/logic/player.cc
Niektory: Good idea. Great idea even because it helped me find a bug :-)

src/logic/player.cc
Niektory: Dismantling is 3-5 times slower without sorting.

Niektory:
There was a bug, the list was being built even when `force_visible=true`. After fixing it It's 1.8 times faster compared to the first post's numbers.

src/logic/player.cc
```diff
@@ -1418,39 +1417,47 @@ void Player::remove_seer(const MapObject& m, const Area<FCoords>& a) {
...
+			if (egbase().map().calc_distance(f, b->get_position()) <= b->descr().vision_range()) {
+				return true;
 			}
 		}
 	}
 	return false;
 }
 
 void Player::update_vision(const FCoords& f, bool force_visible) {
+	update_vision(f, force_visible, seers_);
```
Noordfrees: What about the seers of team partners?

Noordfrees (commented):
Looking good, one more comment inline.
Regarding performance are you aware that `is_a(` works by performing a `dynamic_cast` and checking whether the result is non-null

src/logic/player.cc
Niektory: Good catch. Fixed.

Niektory:
> Regarding performance are you aware that `is_a(` works by performing a `dynamic_cast` and checking whether the result is non-null

Yes, I figured it's fine to keep the slower safety check in debug builds.

Noordfrees (commented):
Code LGTM now, tested and no problems observed.
Just one suggestion: Add something like 
```
using SeersList = std::list<const MapObject*>;
``` 
for readability and pass a `const SeersList&` to `should_see` and `update_vision`

Noordfrees (approved):

PR #4272, opened 2020-09-16
Fix windows moving on their own when docking is enabled

### LINKED ISSUE #4255: Trainingscamp window moves down by click on start/stop
Clicking on the start/stop button of an empire trainingscamp let the window move down.

**To Reproduce**
1. Load attached save game
2. Make sure the trainingscamp is in upper third of the screen
3. Open the trainibgscamp window
4. Click on start/stop

On each click the window moves a bit down. This happens until the window bottom reaches the bottom of the screen.

[test.wgf.zip](https://github.com/widelands/widelands/files/5208050/test.wgf.zip)

**Expected behavior**
The window should stay at the same position.

**Version:**
 - OS: archlinux
 - Widelands Version: [7b55967@splashscreen] (Debug) but this happens also in master

Niektory:
Fix #4255

The Window::docked_*_ member variables are removed to simplify the logic.
This also subtly changes how bottom-docked window borders look:

- Before:
screenshot before the change
- After:
screenshot after the change

I don't think it's worse but if you disagree it can be changed back.


Noordfrees (approved):
LGTM, tested and working :)

PR #4277, opened 2020-09-17
Toolbar fix

### LINKED ISSUE #4233: Wrong toolbar tabbing order
The tabbing order in the InteractivePlayer's, the InteractiveSpectator's, and the Editor toolbar is messed up because the respective buttons/dropdowns are being created in the wrong order.

The tabbing order for a simple line of menu icons should always be left-to-right.

### LINKED ISSUE #4276: Mixing mouse and keyboard control causes game crash
**Describe the bug**
The game crashes when hitting enter (or space in dev-version) after opening a submenu from the toolbar with a mouse click.

**To Reproduce**
1. Click on one of the toolbar buttons with a submenu (e.g. show/hide)
2. Don't select anything
3. Press 'Enter' (or 'Space' in dev-version) 

**Expected behavior**
The game should handle the keypress without crashing

**Version:**
 - OS: Arch Linux
 - Widelands Version: "Build 21" (also tested in current dev-version)

**Additional context**
Log output:
```
widelands: /mnt/storage/Programme/widelands/src/ui_basic/dropdown.h:291: 
    const Entry& UI::Dropdown<Entry>::get_selected() const [with Entry = InteractiveBase::MapviewMenuEntry]: 
    Zusicherung »BaseDropdown::has_selection()« nicht erfüllt.
```

matthiakl:
Fixes #4233
Fixes #4276 by always selecting an entry in dropdown lists


matthiakl:
Looks like Travis got stuck while installing packages. Is there a way to restart the build script?
 
**Log:**
Updating Homebrew
$ rvm $brew_ruby do brew update 1>/dev/null
==> Downloading https://homebrew.bintray.com/bottles-portable-ruby/portable-ruby-2.6.3_2.yosemite.bottle.tar.gz
######################################################################## 100.0%
==> Pouring portable-ruby-2.6.3_2.yosemite.bottle.tar.gz
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

The build has been terminated

Noordfrees (commented):
LGTM, but one bug: When using Enter to toggle an entry in the showhide dropdown menu, the dropdown closes. It should stay open like it does when clicking the item instead.

matthiakl:
> LGTM, but one bug: When using Enter to toggle an entry in the showhide dropdown menu, the dropdown closes. It should stay open like it does when clicking the item instead.

This bug has been present before and wasn't intended to be fixed at first. However after looking into the code again I found other cases where nothing is selected and Enter could cause a crash (Use a shortcut while the menu is open). Now, after rebuilding the menus, the last selected item is selected again, and the menu stays open.

Noordfrees:
The last-selection-memorizing works fine, nice feature :)
The dropdown still doesn't stay open though when using Tab until the showhide menu is selected, then Space to open it, then Space again to toggle the selected entry

matthiakl:
You need to use enter, to keep the menu open. Because of the `toggle_list() `in the `set_value()` / `selected()` call, the Enter and Space behavior is switched. I currently don't see a way to implement this without changing the general dropdown behavior.
https://github.com/widelands/widelands/blob/cc6e4fd9d89c3e12c016855d117e5d125e331e64/src/ui_basic/dropdown.cc#L460-L467

Noordfrees:
The main toolbar and a submenu in the suggested teams edit box are the only places that use a `kPictorialMenu` dropdown. So how about changing the dropdown behaviour just for this type?

matthiakl:
> The main toolbar and a submenu in the suggested teams edit box are the only places that use a `kPictorialMenu` dropdown. So how about changing the dropdown behaviour just for this type?

Should be fixed now. On a second look, it wasn't too difficult. Just check the visibility again after set_value to determine the correct space behavior.

src/editor/editorinteractive.cc
```diff
@@ -384,6 +387,8 @@ void EditorInteractive::rebuild_showhide_menu() {
 	                  g_image_cache->get("images/wui/menus/toggle_buildhelp.png"), false, "",
 	                  pgettext("hotkey", "Space"));
 
+	showhidemenu_.select(ShowHideEntry::kBuildingSpaces);
```
Noordfrees: This line shouldn't be needed as an entry is selected anyway down in line 412

Noordfrees (commented):
Now the dropdowns close when using Enter and stay open when using Space. That's the same as for all other dropdowns so it's fine.

Code LGTM, one comment inline.

P.S. please merge master into this branch

Noordfrees (approved):

PR #4240, opened 2020-09-08
Unify tribes world

gunchleoc:
- Replace `Tribes` and `World` with `Descriptions`
- Consistent function naming in `Descriptions`
- No longer distinguish tribe from world immovables

Fixes https://github.com/widelands/widelands/issues/4258

Things that still need doing in follow-up branches:

- Unify `wl.Tribes()` and `wl.World()` for scripting
- Update the Sphinx documentation
- Unify legacy lookup tables

PR #4281, opened 2020-09-22
made dither layer unique to prevent hard edges

### LINKED ISSUE #4280: Dither Layer for terrains should be unique over all 4 worlds
The Terrains "Summer_swamp" and "wasteland_hard_ground_2" have the same dither layer resulting in sharp edges of triangles 

See Forum Thread https://www.widelands.org/forum/topic/4964/

Other terrains to be checked

hessenfarmer:
fixes #4280

### DIFF
--- a/src/logic/map_objects/world/terrain_description.cc
+++ b/src/logic/map_objects/world/terrain_description.cc
@@ -112,6 +112,13 @@ TerrainDescription::TerrainDescription(const LuaTable& table, Widelands::World& 
 		log_warn("Terrain '%s' contains obsolete 'tooltips' table", name().c_str());
 	}
 
+	for (DescriptionIndex di = world.get_nr_terrains(); di; --di) {
+		const TerrainDescription& t = world.terrain_descr(di - 1);
+		if (t.dither_layer_ == dither_layer_) {
+			throw GameDataError("Terrain %s has the same dither layer %i as %s", name_.c_str(), dither_layer_, t.name_.c_str());
+		}
+	}
+
 	if (table.has_key("enhancement")) {
 		enhancement_ = table.get_string("enhancement");
 		if (enhancement_ == name_) {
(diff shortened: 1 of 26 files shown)


Noordfrees:
LGTM, but IMHO we should code a check in C++ to ensure that no two terrains have the same layer as this would regress easily

hessenfarmer:
> 
> 
> LGTM, but IMHO we should code a check in C++ to ensure that no two terrains have the same layer as this would regress easily

+1 for this however I am not good in c++ coding so I would appreciate if somebody else with more experience could do this.

Noordfrees:
Okay I can code such a check, but it might be better to wait with this until #4240 is in to avoid merge conflicts

Noordfrees:
Easier than I thought, implemented this now :)

src/logic/map_objects/world/terrain_description.cc
```diff
@@ -115,7 +115,8 @@ TerrainDescription::TerrainDescription(const LuaTable& table, Widelands::World&
 	for (DescriptionIndex di = world.get_nr_terrains(); di; --di) {
 		const TerrainDescription& t = world.terrain_descr(di - 1);
 		if (t.dither_layer_ == dither_layer_) {
-			throw GameDataError("Terrain %s has the same dither layer %i as %s", name_.c_str(), dither_layer_, t.name_.c_str());
+			throw GameDataError("Terrain %s has the same dither layer %i as %s", name_.c_str(),
+			                    dither_layer_, t.name_.c_str());
```
SuperSandro2000, line 119: mix of tabs and spaces

hessenfarmer: the spaces are inserted by our fix_formatting script and are ok from my side. 
So for me code looks good.

hessenfarmer:
@Noordfrees how shall we proceed from my side I can approve your changes. But I shouldn't approve mine

Noordfrees:
Then we have hereby approved each others' changes :) So I'll merge now

Issue #4286, opened 2020-09-25
Can't load build21 save: flagdata: 22: ware #0 (1570): not found


Niektory:
Loading this save from the mission Neptune's Revenge results in an exception.
[Neptunetest2.wgf.zip](https://github.com/widelands/widelands/files/5280187/Neptunetest2.wgf.zip)

```
[00:00:16.261 real] INFO: Reading Map Objects ... 
[00:00:16.582 real] WARNING: Unknown animation 'empire_outpost' for immovable 'destroyed_building', using main animation instead.
[00:00:16.583 real] WARNING: Immovable 'destroyed_building', size of program 'main' seems to have changed.
[00:00:16.583 real] ERROR: ImmovableActionData::load: type ^
                                                             not known
(...)
[00:00:16.703 real] INFO: Reading Flagdata Data ... 
[00:00:16.707 real] INFO: WidelandsMapLoader::load_map_complete() for 'Neptune’s Revenge' took 556ms
[00:00:16.709 real] INFO: GameLoader::load() took 8055ms
[00:00:16.709 real] ERROR: Fatal exception: flagdata: 22: ware #0 (1570): not found
```

**Version:**
 - OS: Manjaro Linux
 - Widelands Version: 1.0~git24895[d288d1a@master] (Debug)

Noordfrees:
The game loads fine with build 21. There seems to be a difference in the loaded map objects:

b21: 
```
(gdb) p mol.objects_
$4 = std::map with 1582 elements
```
master:

```
(gdb) p mol.objects_
$6 = std::map with 1558 elements
```

Noordfrees:
The problem is that immovables have two different packet versions. Loading immovables which use the `kCurrentPacketVersionImmovableNoFormerBuildings` version load fine, immovables with the `kCurrentPacketVersionImmovable` version fail.

PR #4287, opened 2020-09-26
Optimize player vision, part 2

### LINKED ISSUE #4253: Player vision code is very slow

Niektory:
A.k.a. The Real Deal.

Counting the number of seers for each field is back, with some twists. The vision values are:
0 = unexplored (never seen)
1 = explored/previously seen (not seen now)
\>1 = currently seen
2 = not seen by us, but seen by a team member
\>2 = seen by us; each seer adds +2, the revealed state (via Lua `reveal_fields()`) adds +1; so:
3 = revealed, no seers
4 = not revealed, 1 seer
5 = revealed, 1 seer
6 = not revealed, 2 seers
and so on.

Each team member counts only their own seers, making team splitting possible. We don't care about team vision if we can see the field ourselves. Encoding the revealed state in the counter saves memory when a lot of fields are revealed. Performance is comparable to Build 21. In other words, fast :-)

Fix #4253


Niektory:
OK, this is mostly ready and testable. Saveloading might still need changes (it saves more than strictly necessary), but it should work. Savegame compatibility is not done though (how many versions do I need to support?).

Noordfrees:
The packet was newly introduced after b21, so breaking compatibility affects only development versions. If it is possible without too much effort to provide compatibility, please do so, but if not it's not the end of the world.
After a quick look at the diff – please don't reinstate the numeric vision comparisons all over the code. If possible *all* comparisons should take place in dedicated functions in player.cc. Will test this soon.

Niektory:
Should there be two vision types – numeric and enum – with conversions when appropriate, or just the numeric type with more helper functions?

Noordfrees:
Perhaps an enum with types like `kUnexplored, kPreviouslySeen, kVisible` like now (and other constants if you need them) and conversion functions from the raw numeric vision to the enum types would be best. If the use of the vision number doesn't reasonably fit into such an enum, just more helper functions are fine.

Niektory:
I ended up doing both, so `Vision` got some helper functions and is also convertible to the enum. `vision.is_visible()` and `vision == VisibleState::kVisible` are equivalent, I tend to use whatever is shorter. And in some places the enum type is used directly because only the 3 states are relevant.

Niektory:
Savegame compatibility is in. Packet versions 4 and 3 work. 2 and 1 I couldn't test because of #4286.

"See all" mode is currently not working properly. Anything I should keep in mind when fixing it?

src/logic/hide_or_reveal_field_mode.h
```diff
@@ -17,15 +17,10 @@
...
+#define WL_LOGIC_HIDE_OR_REVEAL_FIELD_MODE_H
 
 namespace Widelands {
-// Do not change the order! It is stored in savegames.
-enum class SeeUnseeNode {
-	kUnexplored = 0,      // Never seen yet
-	kPreviouslySeen = 1,  // Previously seen
-	kVisible = 2          // Currently visible
-};
+enum class HideOrRevealFieldMode { kHide, kHideAndForget, kReveal };
```
Noordfrees: Move the enum to `vision.h` and delete this file

src/logic/vision.h
```diff
@@ -0,0 +1,102 @@
...
+enum class VisibleState {
+	kUnexplored = 0,      // Never seen yet
+	kPreviouslySeen = 1,  // Previously seen
+	kVisible = 2          // Currently visible
+};
+
+// Extended vision information encoded as a number with some helper functions.
+// See Player::Field for explanation of values.
+struct Vision {
+	uint16_t value;
```
Noordfrees: Should this value be accessed directly? Better make it `private`, add a getter function, and either some functions or a `friend` for modifications.

Noordfrees (commented):
Found no problems while testing so far. Code LGTM, a few comments inline.

Noordfrees (approved):
LGTM, no problems observed

gunchleoc:
We have merge conflicts now

PR #4292, opened 2020-09-28
Draw checkbox borders on top of focus overlay

### LINKED ISSUE #4216: Editor categorized selection needs different focus highlighting
In the attached screenshot, it's impossible to tell whether the Wisent has been selected or not without looking at the text on the bottom.

animals_focus

For animals and immovables, we could add a fill rect, but that wouldn't work for the terrains, because they're square.

Noordfrees:
Fixes #4216


gunchleoc:
Thanks for the fix :)

PR #4293, opened 2020-09-28
Fix poor hamlet warning message localization

Noordfrees:
Fixes an untranslated message in the Poor Hamlet starting conditions


hessenfarmer:
Well, I don't understand why we have to push the textdomain a second time

Noordfrees:
The first `push_textdomain()` affects only the starting condition's definition table. The message is sent from an inner function which is only created here and then stored until use, so the `_()` markup for the message is being evaluated only at a later time when `func()` gets actually called by the engine. By this time the `"tribes"` textdomain has already been popped so we need to push it again.

hessenfarmer (approved):
tested and working

PR #4301, opened 2020-09-29
Refactor string assembly in MapDetailsBox

### LINKED ISSUE #4309: Map description in MP/SP setup screen does not scroll
**Describe the bug**
The small map details textbox in both SP and MP does not scroll with the mouse wheel

**To Reproduce**
1a. Select a singleplayer map
or
1b. Start a multiplayer game and select a map
2. Move the cursor over the map description
3. Try scrolling using the mouse wheel

**Expected behavior**
The description should follow the mouse wheel.

**Version:**
 - OS: Arch Linux
 - Widelands Version: [`472f9102`@master] (Debug)

Noordfrees:


jmoerschbach:
Scrolling works. Not quite sure if RTL works, too (not much experience and I do not speak any rtl language). But I would expect that it would say sth like "64x64 :Size" (so not only right-bound but also swapped?)

Don't know...
grafik

Noordfrees:
I don't know any RTL languages either, but I expect they'd translate `"Size: %s"` as `"%s :eziS"` to get the correct layout (`"64×64 :eziS"`). We have a number of other places in the code where strings with colons are marked for translation in the same way.

src/ui_fsmenu/mapdetailsbox.cc
```diff
@@ -90,48 +90,62 @@ void MapDetailsBox::update_from_savegame(GameSettingsProvider* settings) {
...
-				infotext += "):\n    ";
-				break;
+			if (tribeinfo.name == p.tribe) {
+				i18n::Textdomain td("tribes");
+				return std::string(_(tribeinfo.descname));
 			}
 		}
+		return g_style_manager->font_style(UI::FontStyle::kDisabled)
+		   .as_font_tag((boost::format(_("invalid tribe ‘%s’")) % p.tribe).str());
+	};
```
jmoerschbach: move this locale lambda further down where it is actually used

jmoerschbach (commented):
well then, let's hope translaters create the correct layout :)

Code LGTM (although I don't have much experience with boost::format)

maybe extract the noisy string assembly in another method so we have equal abstraction level (both `show_map_description `and `show_map_description_savegame`):

`set_map_description_text(assemble_infotext().str())`

jmoerschbach (approved):
LGTM :)

gunchleoc:
`Size:`etc are left-to-right because they're still in English. This is on purpose and correct.

Issue #3876, opened 2020-04-23
Trainers are impatient


hessenfarmer:
After optimizing the fail cycle of all work programs which now fail before sleep. we fail very quickly in training while not supplied fully and constantly. leading to massive soldier traffic between HQ/warehouse and trainingsite. So trainers should be more patient (allow more fails) then in the past. So I propose to increase trainer patience by factor 2.

tppq:
Build 21 or later?

Noordfrees:
IMHO this is not critical so I'd say b22

tppq:
Trainer patience select the amount of failed training before layoffs begin. They do not affect the rate.

hessenfarmer:
@tppq 
correct. But the rate of how fast the training attempts fail has been increased by the fail before sleep change. In the past we tried a trainings step, we slept 20 or so and then we failed, then the next step. now we fail, try the next and after all failed we wait until 10 secs after the first fail and repeat. Trainer patience should take this into account imho. But I am ok with having it in b22.

tppq:
>  But the rate of how fast the training attempts fail has been increased by the fail before sleep change.

That explains it. Thanks for telling.

tppq:
@hessenfarmer , would you happen to have a pointer to the issue where the fail cycle optimization was done?

hessenfarmer:
I think it was with 
https://github.com/widelands/widelands/commit/b2e84e87410a273e145039819a23012e02d2bf1c

tppq:
Thanks. I should also check #1663 , might be related.

PR #4305, opened 2020-09-30
Atlantean dungeon balancing tweaks

Noordfrees:
Proposed by the-x via PM:
>  Dungeons Bug fixed:
> - Trainingssite now skips when there is no fitting soldier inside
> - It can now train every soldier to its individual maximum without needing to close
all Trainingssites slot and open only 1 slot
> - Working time adjusted for the individual level of the soldier
> - Sorts soldiers a lot faster by accelerating the skipping process

### DIFF
--- a/data/tribes/buildings/trainingsites/atlanteans/dungeon/init.lua
+++ b/data/tribes/buildings/trainingsites/atlanteans/dungeon/init.lua
@@ -259,7 +259,7 @@ tribes:new_trainingsite_type {
          -- TRANSLATORS: Completed/Skipped/Did not start sleeping because ...
          descname = _"sleeping",
          actions = {
-            "sleep=duration:5s",
+            "sleep=duration:1s",
             "return=skipped",
          }
       },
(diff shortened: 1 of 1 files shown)


hessenfarmer:
-1 
This is totally unbalanced with the other tribes. I don't think it does what is proposed in the description. 
e.g. trainer patience needs to go up not down to reduce kicking out soldiers constantly.     

So changes like this should be made in congruence for all tribes, and maybe they shouzld be discussed in the forums.

Bjoern2Settler:
+1

Wow, an excellent Idea. Especially "Sorts soldiers a lot faster by accelerating the skipping process" is a very nice working.

Hope this gets done for all four tribes soon as it solves a lot of bugs like Traininingssites doing nothing at all / waiting as well as it can now train every soldier to its individual maximum without needing to close
all Trainingssites slot and open only 1 slot

hessenfarmer:
the proposed changes do not achieve what is described in the first comment. Nor have these changes been discussed by the community. In fact they will worsen an already existing bug. (#3876)
So I propose to close this PR

PR #4307, opened 2020-09-30
Reverse dismantlesite wares dropout order

Noordfrees:
https://www.widelands.org/forum/post/34889/

### DIFF
--- a/src/logic/map_objects/tribes/dismantlesite.cc
+++ b/src/logic/map_objects/tribes/dismantlesite.cc
@@ -138,7 +138,7 @@ bool DismantleSite::init(EditorGameBase& egbase) {
 	for (const auto& ware : count_returned_wares(this)) {
 		WaresQueue* wq = new WaresQueue(*this, ware.first, ware.second);
 		wq->set_filled(ware.second);
-		consume_wares_.push_back(wq);
+		consume_wares_.insert(consume_wares_.begin(), wq);
 		work_steps_ += ware.second;
 	}
 	return true;


frankystone:
Are the materials when building a building also in the order ground to top?

I didn't observed how the order is when a building get build, but from my experience it isn't. E.g. mostly i have to wait for one stone until the building get completed, and i believe a stone isn't used for the roof...

To be consistent maybe it is better to have no order at all, instead they should be freed randomly on dismantling.

Noordfrees:
When constructing a building, the constructionsite will consume the first-best available ware at each work step. So if all wares are present by the time the builder arrives they are consumed top-to-bottom (i.e. first granite later reed). Only if wares arrive one at a time then the csite just picks whatever happens to be available already.

When dismantling, the order doesn't matter in 99% of cases, so this changes an arbitrary behaviour to a different arbitrary behaviour, and the reason is that dismantling the roof before the floor seems more logical. It's a cosmetic change, I don't feel strongly about this. Random order, or round-robin like for preserved wares, would be a bit more effort though.

Bjoern2Settler:
-1 

When dismantling, the order doesn't matter. It's a cosmetic change.

Noordfrees:
… which is precisely why this change does not have a negative effect so it's not an argument against.

hessenfarmer (approved):
Code LGTM.
Not tested.
As this now reversed order is dependent on the definitions in lua. It might be good to check the definition order there as well.

Noordfrees:
The wares are internally sorted by ascending DescriptionIndex which depends on the wares loading order

hessenfarmer:
Ok didn't know that. 
Is this loading order deterministic? if yes how is it derived?

Noordfrees:
It must be deterministic, else replays and network games would desync all the time. This used to be the order in which they were listed in the init.lua, don't know how it's done now – perhaps by alphabetical directory name?

hessenfarmer:
so we should wait and doublecheck with @gunchleoc how this is derived and or how it can be controlled. 
As the bug was reported for b21 we need to check how the current behaviour with the plugable tribes changes is

hessenfarmer:
Did some research 
If I understand this 

https://github.com/widelands/widelands/blob/cba9976e95329a7fe93340de2c605f5245a160d3/src/logic/map_objects/tribes/tribe_descr.cc#L360

correctly. The order is defined by column and row of each ware item. so the change should be ok. needs some testing though I believe

Bjoern2Settler:
I wouldnt change that. One reason is that it must be deterministic, else replays and network games would desync all the time.

Wait arent there more important Pull Requests, like this for example:

The Icons in the attack window can be overdone

enlarge them to a nice size, because you click very often on them in singleplayer

by default all units are attacking, because you click very often on them in singleplayer. If there is only one soldier attacking by default you need to change that order every single time

"random attacks" in game setup. In order to not have always the same boring one hero attacks szenario, randomize it. Of all given soldiers in the area a random number of units will attack. It can be set to a special number like three quarters of the units or all or something completely different.

Noordfrees:
@Bjoern2Settler Throwing out pointless remarks on the forum is one thing. But PRs on GitHub are for *developers* and we will keep them clean. So I will be very quick to press the Delete Comment button if you spam us with more unrelated posts here.

Additionally GitHub has a feature to limit conversations to team members only, which we did not need to use – yet.

Bjoern2Settler:
Looks like an unfriendly community, as there necessary changes arent made whilst cosmetic in every detail. You seem to not want a playable game that is bugfixed.

Dont understand why Nordfrees deletes important bugfixes and fixes that aim at improving service and controls of the gameplay?

hessenfarmer:
hm,
in current master the order of the returned wares is e.g. for attlanteans log planks granite. same as the order (sorting) in the constructionsite. However this is not as it is defined in the columns of the tribe and I have no glue why this is the case.

frankystone:
hm, may i do not understand correctly. Anyway here is an example of an empire tower made with current master:

dismantle_order

[Return on dismantle is in another order](https://github.com/widelands/widelands/blob/master/data/tribes/buildings/militarysites/empire/tower/init.lua#L19): Planks, Granite, Marble column.

Noordfrees:
Why is everyone always thinking that lua tables define some kind of order? ;)
Tables are *not* sorted in any way. The keys are iterated in arbitrary order, and the wares are sorted internally by ascending DescriptionIndex, which in turn is defined by the order in which the wares are loaded during game start. And this order is now defined by… some implementation details of the new loading code that nobody except GunChleoc understands ;)

hessenfarmer:
Well, obviously the sorting of the wares is different from tribe to tribe. So the problem described in the forums is currently only valif for empire and not for atlanteans which I tested. 
So the fix in this PR won't fix it for all tribes. Therefore I recommend to close or postpone it and to open an issue instead, to continue collecting some information about the current behaviour. so we might get able to determine the correct place for fixing

Noordfrees:
Then how about sorting the wares by descending tribe-specific wares order (the one that is used in the stock window, inverted)?

frankystone:
The Tooltip for dismantling shows also another order than the dismantle site window:

dismantle_order_tooltip

If i understand this PR correctly the dismantle order for this Empire Trainingscamp will be: Marble column, Marble, Log, Planks, Granite. This is also not from top to ground. In general i think this will raise the question: Which is the most acceptable order when dismantling?

Is granite used for the base (grounding) or marble? When will gold be returned? Are planks used for the roof or for the interior?

If the underlying issue is _really_ disturbing, we should go for random or round robin, imho.

Noordfrees:
So many complications for such an unimportant detail ;) The order does not really matter IMHO, but we should make the tooltip and the actual order consistent. For constructionsites as well. Will have a look at this soonish.

frankystone:
> So many complications for such an unimportant detail ;) 

:-D

> The order does not really matter IMHO, but we should make the tooltip and the actual order consistent. For constructionsites as well.

I would not spend any time to make the order consistent in this regard. The important information is: How many wares does this building cost/returns on dismantle? The order in the tooltips are absolutely unimportant, imho.

hessenfarmer:
+1 for letting everything like it is in the moment. Too much trouble for a more cosmetic change

gunchleoc:
FYI: saveloading and (and AFAIK networking) are done via the string names, not the indices.

Ware indices have not been designed to be deterministic, although they might be so by accident. For a stable order, you need to use the wares_order for the specific tribe. The tribes' units tables use integer indices rather than string indices to ensure that. If we had

```
wares_order = {
    -- Column in warehouses/economy settings
    {
        log = { <some data> },
        granite = { <some_data> }
    }
}
```

then Lua would load the keys in alphabetical order. So, in order to control the order, we need to define it like this:

```
wares_order = {
    -- Column in warehouses/economy settings
    {
        {
            name = "log",
            <some data>
        },
        {
            name = "granite",
            <some data>
        },
    }
}
```

@Noordfrees @hessenfarmer

PR #4313, opened 2020-10-01
Unload expeditions at once

### LINKED ISSUE #4311: unload expeditions immediately
As ships are loaded and unloadewd at once they should consistently unload their expedition wares at once as well.

See https://www.widelands.org/forum/topic/4911

Noordfrees:
Fixes #4311


gunchleoc:
Code LGTM, not tested yet.

Since this now happens instantly, should we remove the special case for "colonizing" in the ship statistics?

Noordfrees:
There is still a brief delay between creating the port csite and unloading the wares. If a fast player is allowed to click Cancel Expedition in this interval this would regress the bug that you can block a port space without unloading wares. So we need this special value internally to mark the ship and expedition as unsinkable/un-cancel-able during this delay; and also for savegame compatibility.

gunchleoc:
You misunderstand me - I mean in the ship **statistics** window only, as this state  is now not interesting to the player and will disappear too fast to filter for. We could keep it on "Port Space Found" or change to "Transporting wares".

Of course, we need to keep the internal state and the **ship** window as it is.

gunchleoc:
Tested and ready to go in :)

PR #4314, opened 2020-10-01
Build 21

Bjoern2Settler:
The Icons in the attack window can be overdone

- enlarge them to a nice size, because you click very often on them in singleplayer

- by default all units are attacking, because you click very often on them in singleplayer. If there is only one soldier attacking by default you need to change that order every single time

- "random attacks" in game setup. In order to not have always the same boring one hero attacks szenario, randomize it. Of all given soldiers in the area a random number of units will attack. It can be set to a special number like three quarters of the units or all.


Noordfrees:
What on earth are you doing with such a PR?

PR #4315, opened 2020-10-01
Handle GameDataError during preview rendering

### LINKED ISSUE #4306: Crash when loading invalid map preview
- Extract this map into your `.widelands/maps` folder: [ProblematicMap.wmf.zip](https://github.com/widelands/widelands/files/5306284/ProblematicMap.wmf.zip)
- Start Widelands, press `N`
- Highlight the problematic map

→ Crash back to main menu with 
```
terrain: Error while loading terrain type 'foreignplanet_rocks1': DescriptionManager::load_description: Object 'foreignplanet_rocks1' was not registered
```

Expected behaviour: Since this map is obviously not playable, perhaps just disable the OK button and show an error message in the right column (and no minimap of course). but don't cancel the mapselect screen.

matthiakl:
Fixes #4306 by showing the error message in the map detail section and disabling the Ok button.

### DIFF
--- a/src/wui/mapdetails.cc
+++ b/src/wui/mapdetails.cc
@@ -26,6 +26,7 @@
 #include "graphic/minimap_renderer.h"
 #include "graphic/text_layout.h"
 #include "io/filesystem/layered_filesystem.h"
+#include "logic/game_data_error.h"
 #include "logic/game_settings.h"
 #include "map_io/map_loader.h"
 #include "ui_basic/box.h"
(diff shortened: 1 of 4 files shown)


src/wui/mapdetails.cc
```diff
@@ -201,15 +201,27 @@ void MapDetails::update(const MapData& mapdata, bool localize_mapname) {
...
+					minimap_icon_.set_icon(minimap_cache_.at(last_map_).get());
+					minimap_icon_.set_visible(true);
+				}
+			} catch (const Widelands::GameDataError& e) {
+				// Put error message on top for better visibility
+				description =
+				   (boost::format("%s%s") % as_content(e.what(), style_) % description).str();
+				description =
+				   (boost::format("%s%s") % as_heading(_("Game data error"), style_) % description)
+				      .str();
```
Noordfrees, line 218:
```suggestion
				description =
				   (boost::format("%s%s%s") % as_heading(_("Game data error"), style_) % as_content(e.what(), style_) % description).str();
```

Noordfrees (approved):
LGTM, tested and working. One suggestion inline to need one boost::format less.

src/wui/mapdetails.cc
matthiakl: I simply adapted the previous style. If you want to reduce boost::format calls, I should apply this also to the other formatting and always combine content and heading in one call?

Noordfrees: If it's consistent with the other calls then it's up to you whether you want to change this or not :)

matthiakl: Then for readability I suggest to keep it.

PR #4321, opened 2020-10-04
revert checkbox

jmoerschbach:
Remove a check in Checkbox which hinders layouting when needed, see screenshots.
This was introduced with 4e9992f3a36ef5b04395a9ee5d67639d15c4b76f and I do not know/understand why @matthiakl @Noordfrees 

Before:
grafik

After:
grafik

### DIFF
--- a/src/ui_basic/checkbox.cc
+++ b/src/ui_basic/checkbox.cc
@@ -46,8 +46,7 @@ Statebox::Statebox(Panel* const parent,
    : Panel(parent, p.x, p.y, kStateboxSize, kStateboxSize, tooltip_text),
      flags_(Is_Enabled),
      pic_graphics_(pic),
-     rendered_text_(nullptr),
-     rendered_width_(0) {
+     rendered_text_(nullptr) {
 	uint16_t w = pic->width();
 	uint16_t h = pic->height();
 	set_desired_size(w, h);
@@ -73,7 +72,7 @@ Statebox::Statebox(Panel* const parent,
 
 void Statebox::layout() {
 	// We only need to relayout if we have text and the available width changed
-	if ((flags_ & Has_Text) && (rendered_width_ != get_w())) {
+	if ((flags_ & Has_Text)) {
 		int w = get_w();
 		int h = kStateboxSize;
 		int pic_width = kStateboxSize;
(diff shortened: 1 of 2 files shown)


matthiakl:
I introduced this check, because the text rendering in `Statebox::layout()` produced text with a width difference of 1 in subsequent calls, causing a call loop:
`Statebox::layout` -> `Panel::set_desired_size` -> `Box::update_desired_size` -> `Box::layout` -> `Box::update_positions` -> `Panel::set_size`-> `Statebox::layout`
This resulted in a stack overflow and crash. 

I can reproduce it by going into the options.

src/ui_basic/checkbox.cc
```diff
@@ -84,16 +83,15 @@ void Statebox::layout() {
...
 		                       text_width(get_w(), pic_width));
 		if (rendered_text_) {
 			w = std::max(rendered_text_->width() + kPadding + pic_width, w);
 			h = std::max(rendered_text_->height(), h);
 		}
-		rendered_width_ = w;
-		set_desired_size(rendered_width_, h);
-		set_size(rendered_width_, h);
+		set_desired_size(w, h);
+		set_size(w, h);
```
matthiakl: Remove the call to `set_size(w, h)`, which is eventually called during `set_desired_size(w, h)`. This fixes the options bug in a much simpler way.

src/ui_basic/checkbox.cc
```diff
@@ -84,16 +83,15 @@ void Statebox::layout() {
 		}
 		rendered_text_ =
 		   label_text_.empty() ?
-		      nullptr :
-		      UI::g_fh->render(as_richtext_paragraph(label_text_, UI::FontStyle::kLabel),
+            nullptr :
+            UI::g_fh->render(as_richtext_paragraph(label_text_, UI::FontStyle::kLabel),
```
matthiakl: Formatting shouldn't be necessary. (Maybe run utils/fix_formatting.py ?)

matthiakl (commented):
After some investigations, I found a different solution for the options crash. See comment inline.

jmoerschbach:
ok, let's go with `set_size` also being removed (this is indeed looking suspicious...) :) However, as I am not able to reproduce the bug in options (anything else I need to do other than going into options and clicking around?), there should be some more testing by other people, too I think :)

Noordfrees:
Code LGTM. I cannot reproduce that bug either, but this branch works fine for me both in the options and in the launch game screens :)

matthiakl:
It happens directly after clicking the options button. I tried it with a fresh build in both debug and release (debug log attached).
As you can't reproduce it, maybe it is related to graphics driver or library?

OS: Arch Linux (Kernel 5.8)
Graphics: OpenGL: Version "4.6.0 NVIDIA 455.23.04"
SDL version: 2.0.12

[crash.txt](https://github.com/widelands/widelands/files/5327924/crash.txt)

matthiakl (approved):

Issue #3209, opened 2019-09-09
Replace named integer aliases with structs


bunnybot:
Idea fromm SirVer:

> [PlayerSlot and other using = int decls]
> I am not a fan of this. An integer is an integer, the variable name should make it clear what it
> means. If you want type_safty (not accidentally use a PlayerSlot as a pixel length for example),
> introduce a struct PlayerSlot { int slot; }.

Other candidates for a refactoring are e.g. PlayerNumber, Time, Duration, DescriptionIndex. The structs could then also own some constants like INVALID_INDEX, neutral(), never(), ...

------------------------------------
Imported from Launchpad using lp2gh.

 * date created: 2017-09-02T13:59:56Z
 * owner: gunchleoc
 * the launchpad url was https://bugs.launchpad.net/bugs/1714680

PR #4325, opened 2020-10-07
Make `Time` and `Duration` typesafe

### LINKED ISSUE #1561: Gametime is sometimes uint32_t and sometimes int32_t
Gametime is sometimes uint32_t and sometimes int32_t. We should have a consistent data type here.

Start investigating from int32_t WLApplication::get_time()

 * date created: 2014-10-11T11:28:48Z
 * owner: gunchleoc
 * the launchpad url was https://bugs.launchpad.net/bugs/1380058

Noordfrees:
Fixes #1561 
Re #3209 

The only real change is in `logic/widelands.*`.

I added TODOs to a few places in AI code that look fishy to me @tibor95 @hessenfarmer


src/economy/ware_instance.cc
```diff
@@ -196,7 +196,7 @@ WareInstance::WareInstance(DescriptionIndex const i, const WareDescr* const ware
 WareInstance::~WareInstance() {
 	if (supply_) {
 		FORMAT_WARNINGS_OFF
-		molog(kNoTimestamp, "Ware %u still has supply %p\n", descr_index_, supply_);
+		molog(Time(kNoTimestamp), "Ware %u still has supply %p", descr_index_, supply_);
```
gunchleoc: Can `Time()` take over the function of `Time(kNoTimestamp)` in the logger now?

src/logic/widelands.h
```diff
@@ -68,15 +72,160 @@ using Quantity = uint32_t;  // e.g. the number of a type of ware in a warehouse.
...
+	}
+	void operator+=(const Duration& delta) {
+		assert(is_valid());
+		assert(delta.is_valid());
+		value_ += delta.get();
+	}
+	void operator-=(const Duration& delta) {
+		assert(is_valid());
+		assert(delta.is_valid());
+		assert(get() >= delta.get());
```
gunchleoc: It might be better to use NEVER_HERE() so we won't get hard to diagnose bugs in release builds due to overflow. Same for `-` operator below.

src/logic/widelands.h
```diff
@@ -68,15 +72,160 @@ using Quantity = uint32_t;  // e.g. the number of a type of ware in a warehouse.
...
+	void save(FileWrite&) const;
+
+private:
+	uint32_t value_;
+};
+
+// A time point, in milliseconds gametime.
+struct Time {
+	// The default-constructed Time is the special value "never"
+	constexpr explicit Time(uint32_t v = std::numeric_limits<uint32_t>::max()) : value_(v) {
```
gunchleoc: Both `Time` and `Duration` are unsigned int. Shall we use just 1 data type called `Time`?

gunchleoc (commented):
Added some design ideas

src/logic/widelands.h
Noordfrees: Type safety! Time points and durations are semantically different things. That's sort of the point of this refactoring ;)

src/economy/ware_instance.cc
Noordfrees: For this to work we'd need a `logic` dependency in the `base_log` library because we'd need to pass a `const Time&` to the log function and call it's `get()` and `is_valid()` functions. Currently `get()` is expanded in the macro definition so we can do without such a dependency.

gunchleoc: How about moving time and duration to the base library?

src/ai/defaultai.cc
```diff
@@ -633,12 +656,13 @@ void DefaultAI::late_initialization() {
...
 		bo.stocklevel_count = 0;
-		bo.stocklevel_time = 0;
-		bo.last_dismantle_time = 0;
+		bo.stocklevel_time = Widelands::Time(0);
+		bo.last_dismantle_time = Widelands::Time(0);
 		// this is set to negative number, otherwise the AI would wait 25 sec
 		// after game start not building anything
-		bo.construction_decision_time = -60 * 60 * 1000;
-		bo.last_building_built = kNever;
+		// TODO(Nordfriese): …but times can no longer be negative. Seems to be working fine though…
```
gunchleoc, line 653: We also have `const int32_t delay_time` above - should we make Time signed after all? Just a question, you can answer "no" to that.

I don't know the AI code well enough to judge how this comment should be resolved.

src/ai/defaultai.cc
```diff
@@ -2343,10 +2375,12 @@ bool DefaultAI::construct_building(uint32_t gametime) {
 	   !(player_statistics.get_player_power(pn) > player_statistics.get_old_player_power(pn));
 	inputs[24] = (highest_nonmil_prio_ > 18 + management_data.get_military_number_at(65) / 10);
 	inputs[25] = !(highest_nonmil_prio_ > 18 + management_data.get_military_number_at(65) / 10);
+	// TODO(Nordfriese): Passing a PlayerNumber to a function that expects
```
gunchleoc, line 2361: @tibor95 Can you help with this one?

src/ai/defaultai.cc
```diff
@@ -3970,23 +4006,27 @@ bool DefaultAI::create_shortcut_road(const Widelands::Flag& flag,
...
 				// this might be a new colonization port
 				if (bo.is(BuildingAttribute::kPort)) {
-					eco->dismantle_grace_time = gametime + 60 * 60 * 1000;  // one hour should be enough
+					eco->dismantle_grace_time =
+					   gametime + Widelands::Duration(60 * 60 * 1000);  // one hour should be enough
 				} else {  // other constructionsites, usually new (standalone) constructionsites
 					eco->dismantle_grace_time =
-					   gametime + 30 * 1000 +            // very shot time is enough
-					   (eco->flags.size() * 30 * 1000);  // + 30 seconds for every flag in economy
+					   gametime + Widelands::Duration(30 * 1000 +  // very shot time is enough
```
gunchleoc:
```suggestion
					   gametime + Widelands::Duration(30 * 1000 +  // very short time is enough
```
Typo

src/ai/defaultai.cc
```diff
@@ -4169,10 +4209,11 @@ bool DefaultAI::create_shortcut_road(const Widelands::Flag& flag,
 	if (flag.get_economy(Widelands::wwWORKER)->warehouses().empty()) {
 
 		// blocking only if latest block was less then 60 seconds ago or it is last attempt
-		if (eco->fields_block_last_time + kOneMinute < gametime || last_attempt_) {
+		if (eco->fields_block_last_time + Widelands::Duration(kOneMinute) < gametime ||
+		    last_attempt_) {
 			eco->fields_block_last_time = gametime;
 
-			const uint32_t block_time = last_attempt_ ? 10 * kOneMinute : 2 * kOneMinute;
+			const Widelands::Duration block_time(last_attempt_ ? 10 * kOneMinute : 2 * kOneMinute);
```
gunchleoc: How about moving kOneMinute to the Duration class?

src/ai/defaultai_seafaring.cc
```diff
@@ -111,8 +111,8 @@ uint8_t DefaultAI::spot_scoring(Widelands::Coords candidate_spot) {
 // and makes two decisions:
 // - build a ship
 // - start preparation for expedition
-bool DefaultAI::marine_main_decisions(const uint32_t gametime) {
-	if (gametime > last_seafaring_check_ + 20000U) {
+bool DefaultAI::marine_main_decisions(const Widelands::Time& gametime) {
+	if (gametime > last_seafaring_check_ + Widelands::Duration(20000)) {
```
gunchleoc:
```suggestion
	if (gametime > last_seafaring_check_ + Widelands::Duration(20 * 1000)) {
```

src/ai/defaultai_warfare.cc
```diff
@@ -360,12 +362,14 @@ bool DefaultAI::check_enemy_sites(uint32_t const gametime) {
 				              player_statistics.get_old60_player_power(pn)) ?
 				                3 :
 				                -3;
-				inputs[18] = (player_statistics.get_visible_enemies_power(pn) >
-				              player_statistics.get_old_visible_enemies_power(pn)) ?
+				// TODO(Nordfriese): Passing a PlayerNumber to a function that expects
```
gunchleoc, line 364: @tibor95  Can you help with this one?

src/base/times.h
```diff
@@ -0,0 +1,215 @@
...
+#include "base/wexception.h"
+
+class FileRead;
+class FileWrite;
+
+namespace Widelands {
+
+// The difference of two points in time, in milliseconds gametime.
+struct Duration {
+	// The default-constructed Duration is the special value "endless"
```
gunchleoc: The functions are called (in-)valid - change this comment or the function names, whichever you prefer.

src/base/times.h
```diff
@@ -0,0 +1,215 @@
...
+	explicit Duration(FileRead&);
+	void save(FileWrite&) const;
+
+private:
+	uint32_t value_;
+};
+
+// A time point, in milliseconds gametime.
+struct Time {
+	// The default-constructed Time is the special value "never"
```
gunchleoc: The functions are called (in-)valid - change this comment or the function names, whichever you prefer.

src/logic/game.h
```diff
@@ -38,7 +38,7 @@ class GameController;
 namespace Widelands {
 
 /// How often are statistics to be sampled.
-constexpr uint32_t kStatisticsSampleTime = 30000;
+constexpr Duration kStatisticsSampleTime = Duration(30000);
```
gunchleoc:
```suggestion
constexpr Duration kStatisticsSampleTime = Duration(30 * 1000);
```

src/logic/map_objects/bob.cc
```diff
@@ -1060,11 +1061,11 @@ void Bob::Loader::load(FileRead& fr) {
 				         animname.c_str(), bob.descr().name().c_str());
 			}
 
-			bob.animstart_ = fr.signed_32();
+			bob.animstart_ = Time(fr);
```
gunchleoc, line 1062: The data type has changed in the packet from signed to unsigned. We need to test that this is OK.

src/logic/map_objects/immovable.cc
```diff
@@ -696,7 +697,7 @@ void Immovable::save(EditorGameBase& egbase, MapObjectSaver& mos, FileWrite& fw)
 
 	// Animations
 	fw.string(descr().get_animation_name(anim_));
-	fw.signed_32(animstart_);
+	animstart_.save(fw);
```
gunchleoc, line 700: Test datatype change

src/logic/map_objects/tribes/battle.cc
```diff
@@ -348,7 +348,7 @@ void Battle::Loader::load(FileRead& fr) {
 
 	Battle& battle = get<Battle>();
 
-	battle.creationtime_ = fr.signed_32();
+	battle.creationtime_ = Time(fr);
```
gunchleoc, line 351: Test datatype change

src/logic/playercommand.h
```diff
@@ -276,7 +275,7 @@ struct CmdMilitarySiteSetSoldierPreference : public PlayerCommand {
 struct CmdStartOrCancelExpedition : public PlayerCommand {
 	CmdStartOrCancelExpedition() : PlayerCommand() {
 	}  // For savegame loading
-	CmdStartOrCancelExpedition(uint32_t const t, PlayerNumber const p, Building& b)
+	CmdStartOrCancelExpedition(const Widelands::Time& t, PlayerNumber const p, Building& b)
```
gunchleoc: Since we moved Time to base, it should no longer be in the `Widelands` namespace.

src/map_io/map_buildingdata_packet.cc
```diff
@@ -744,7 +744,7 @@ void MapBuildingdataPacket::read_productionsite(
 				productionsite.stack_[i].coord = read_coords_32_allow_null(&fr, game.map().extent());
 			}
 			productionsite.program_timer_ = fr.unsigned_8();
-			productionsite.program_time_ = fr.signed_32();
+			productionsite.program_time_ = Time(fr);
```
gunchleoc, line 748: signed -> unsigned

src/map_io/map_buildingdata_packet.cc
```diff
@@ -1187,7 +1187,7 @@ void MapBuildingdataPacket::write_militarysite(const MilitarySite& militarysite,
 
 	fw.unsigned_8(militarysite.didconquer_);
 	fw.unsigned_8(militarysite.capacity_);
-	fw.signed_32(militarysite.nexthealtime_);
+	militarysite.nexthealtime_.save(fw);
```
gunchleoc, line 1191: signed -> unsigned. 2 more below.

src/map_io/map_flagdata_packet.cc
```diff
@@ -65,7 +65,9 @@ void MapFlagdataPacket::read(FileSystem& fs,
 
 					//  Owner is already set, nothing to do from PlayerImmovable.
 
-					flag.animstart_ = fr.unsigned_16();
+					// TODO(Nordfriese): Change this line when we next increase the packet version:
```
gunchleoc: Please add "saveloading" and/or "Compatibility" to these comments so we can do a fulltext search on them later. I would definitely miss those otherwise.

gunchleoc (commented):
I have now finished going over the complete code.

Please review the saveloading too to ensure I didn't miss any changes from signed to unsigned - these will probably break savegame compatibility. For the networking packets, this is fine - shall we bump the protocol version?

src/ai/defaultai.cc
Noordfrees: `delay_time` refers to (I think) the difference between the actual gametime and the gametime at which the AI was intending to perform some job. So it's special. Except for the AI, Times should never be negative, so I want to keep it unsigned and use `signed int`s in the few places (AI) where negative times or intervals actually make sense.

src/ai/defaultai.cc
Noordfrees: it is used only 3 times so I'll just remove this constant

src/logic/map_objects/bob.cc
Noordfrees: This value was only ever assigned `0` or `egbase.get_gametime()` so it should be fine. Same below. Will try loading some older savegames that have all those objects in them to check.

Noordfrees: Tested, seems to be working fine

src/ai/defaultai.cc
hessenfarmer: the reason why this is initialized negative is that the AI is only allowing a building of each type every 25 seconds. The value gets updated as soon as a building of this type gets build. if this is initilaized with zero the AI won't build anything in the first 25 seconds. 
I see 2 solutions add an aditional check in every spot where the interval is checked whether it is the first building to build. or we change the semantic of the decision_time to be actual gametime +25 sec and check whether we have gametime bigger then decision_time

src/ai/defaultai.cc
hessenfarmer: From my perspective this is a very old bug accidentally discovered

src/ai/defaultai.cc
gunchleoc: The second solution sounds cleaner to me.

hessenfarmer: So I propose to postpone the change to a new branch as we probably need to fix / improve other stuff in AI as well we just need to open a new bug which should be marked as showstopping the next release

src/ai/defaultai.cc
gunchleoc: Shall we open an issue for this too, so we can merge this branch?

gunchleoc:
We now have merge conflicts fro the cppcheck branch

gunchleoc (approved):

hessenfarmer:
now I found a crash #4384

PR #4326, opened 2020-10-08
Improve testsuite error handling

### LINKED ISSUE #3716: Lua coroutine errors don't show up in regression tests output
When our testsuite fails due to a lua coroutine error, this error does not cause widelands to crash and therefore does not show up in the log output. 
On travis this leads to the build stalling without any log output that gives a clue to the failure type.

Suggestions:
- improve regression_test.py to identify coroutine errors in the output
- throw a wexception when a coroutine error is found

Noordfrees:
- Adds a command line switch to Widelands to crash on lua coroutine errors, which is enabled for the testsuite (fixes #3716 )
- Adds a command line switch to the testsuite to ignore error code 1, so it is possible to run the tests locally without ASan always returning error 1 due to the usual third-party-lib memory leaks

Sample output:
```
[00:00:00.101 game] ERROR: Error in Lua Coroutine
[00:00:00.101 game] ERROR: [../src/scripting/lua_errors.cc:22] [string "scripting/init.lua"]:191: attempt to index a nil value (global 'a')
[00:00:00.101 game] ERROR: Terminating Widelands.

Returned from Widelands, return code is -6
```
```
test/maps/plain.wmf/scripting/test_ui.lua ... 
  Running Widelands ... IGNORING error code 1
done.
ok
```


gunchleoc:
Have you tested that normal test failures and c++ crashes return with an error code other than 1?

Noordfrees:
Normal test failures are detected by `regressiontest.py` in a different way: The script checks not only the Widelands return code but also that the stdout contains "# All tests passed." and no references to "lua_errors.cc". So those work fine.
C++ crashes always returned error codes -6 (SIGABRT) or -11 (SIGSEGV) in my tests, but never ASan's favourite code 1.

gunchleoc:
Thanks for the explanation. Code looks good to me!

gunchleoc:
Tested and working :)

Please document the new command line switch on https://www.widelands.org/wiki/RegressionTests/ when you merge.

PR #4330, opened 2020-10-09
Handle additional expedition items in fri01/fri02

### LINKED ISSUE #4328: Adjust fri01 to work with additional expedition items
https://www.widelands.org/forum/post/34972/
Unfortunately the new additional-items feature makes this a bit confusing. The scenario is older than that feature, and the idea is that the soldiers have to be located in the port when the game ends to be carried over to the next scenario. I guess it would make sense to additionally preserve any wares, workers and soldiers that are taken on the ship in addition to the normal soldier takeover. Or disable taking additional items for this scenario.

Noordfrees:
Fixes #4328


hessenfarmer:
Well, I beleive this is the easier way to fix it however it is not necessarily consistent, as the player does not see a funtionality which he might have been used to see already.     
So for me we should aim at adjusting the ship capacity to the nr of soldiers given by the difficulty. and only save the soldiers on board in the persistence functions. 
Furthermore the additional items feature needs to be explained in the tutorial which would easily be doable so I'LL open a bug for this and provide the fix.

Noordfrees:
The AdditionalItemsDisplay can't handle non-default ship capacities – it assumes that all ships have at least the default capacity. If the ship has a greater-than-default capacity this just means that you cannot utilize all the space, and if it has less capacity the game might crash when the ship arrives. Ships are only assigned to an expedition when all wares have arrived in the dock so the UI does not know in advance how much capacity will be available.

Since launching the expedition is the last thing the player does in this scenario (and as there are no other expeditions there) it does not matter much if the feature is disabled here, IMHO.

>  Furthermore the additional items feature needs to be explained in the tutorial which would easily be doable so I'LL open a bug for this and provide the fix.

+1

hessenfarmer:
Well, why not alowing the new feature and encouraging the player to take some additional wares or sodiers with him, persisiting them and adding them to the second scenario. we can tell the player that additional soldiers in the port may finally get crammed together on board of the ship. 
In scenario 2 the only ware that would hurt would be iron, In case the player loads some iron this will not be taken as it got lost in the storm (was used by the smith on board) Anything else could be added to the scenario easily. If gold is taken we might need to check before firing the text about dismantling the port.

Noordfrees:
Ok good points. Iron, gold, iron ore, gold ore, all tools, miners, and master miners would be problematic in fri02. But for other wares and workers, why not. Will do that then

hessenfarmer:
I don't think gold is problematic. it is just a matter whether we need to dismatle the port or not. 
But you are right with iron, iron ore, picks, and all miners. 
textual solution for them are:
iron was used to keep the ship in a good mood by the ships smith / carpenter 
iron ore, maybe goldore as well was dropped as being to heavy in a hard sea. 
picks were rusted damaged by being exposed to salty water. by this the miners lost their tools

Noordfrees:
Dismantling the port should not be made dependent on whether the player brought gold. Too much of an easteregg IMHO. We can just say that all metals and ores had to be dropped in a storm on sea, good idea for that and the picks.
It seems to be currently not possible to get a list of the wares and workers present on a ship so I'll have to implement this first.

hessenfarmer:
treating metals all the same might do the trick. But I really would prefer to have more dedicated handling for 2 reasons.
1. I don't think it is kind of an easteregg (which we already have at least 2 I remember in the campaigns) but a non linear way for the campaign which adds variety and depth and increases the replayability (that was the reason for having the twist about vesta's temple in emp04) so having differnet ways to play a scenario is good imho.  
2. having different explanations for different wares adds depth and credibilty for the wares being lost / unusable.

having acces to a ships load via lua would add new possibilites in scripting so thanks for that.

data/campaigns/fri02.wmf/scripting/init.lua
```diff
@@ -20,19 +20,46 @@ map = game.map
...
 total_soldiers = {40, 25, 10}
 total_soldiers = total_soldiers[difficulty]
 takeover_soldiers = {15, 10, 5}
 takeover_soldiers = takeover_soldiers[difficulty]
+
 campaign_data = game:read_campaign_data("frisians", "fri01")
 if not campaign_data then
-   campaign_message_box(campaign_data_warning)
+   campaign_message_box({
+      title = _"Warning",
```
gunchleoc:
```suggestion
      -- TRANSLATORS: Message title for a piece of information
      title = _"Note",
```

I think this would sound friendlier.

data/campaigns/fri02.wmf/scripting/texts.lua
```diff
@@ -124,22 +124,30 @@ obj_defeat_both = {
...
 -- ==================
 
-campaign_data_warning = {
-   title = _"Warning",
-   body = p(_[[You appear not to have completed the previous mission. You may still play this scenario, but you will be at a disadvantage. I recommend that you complete the previous scenario first and then restart this mission.]]),
-   w = 450,
-   h = 150,
-}
+-- TRANSLATORS: Reebaud – Introduction 1
+local intro_1_start = _([[I began to believe the journey would never end. I have no idea how long we were on the sea, tossed about by envious storms… it must have been months.]])
```
gunchleoc:
```suggestion
local intro_1_start = _([[I began to believe the journey would never end. I have no idea how long we were at sea, tossed about by envious storms… it must have been months.]])
```

data/campaigns/fri02.wmf/scripting/texts.lua
```diff
@@ -124,22 +124,30 @@ obj_defeat_both = {
...
+-- TRANSLATORS: Reebaud – Introduction 1
+local intro_1_iron = _([[The storms have maintained a firm grip on the ship throughout our voyage. There was hardly a day on which we did not wonder whether it would be our last. We were compelled to drop the precious metals we have taken with us overboard to steady the ship, feeling as though it were a squalid sacrifice to the gods to give us hope to live a few hours longer.]])
```
gunchleoc:
```suggestion
local intro_1_iron = _([[The storms maintained a firm grip on the ship throughout our voyage. There was hardly a day on which we did not wonder whether it would be our last. We were compelled to drop the precious metals we had taken with us overboard to steady the ship, feeling as though it were a squalid sacrifice to the gods to give us hope to live a few hours longer.]])
```

data/campaigns/fri02.wmf/scripting/texts.lua
```diff
@@ -124,22 +124,30 @@ obj_defeat_both = {
...
-}
+-- TRANSLATORS: Reebaud – Introduction 1
+local intro_1_start = _([[I began to believe the journey would never end. I have no idea how long we were at sea, tossed about by envious storms… it must have been months.]])
+-- TRANSLATORS: Reebaud – Introduction 1
+local intro_1_iron = _([[The storms maintained a firm grip on the ship throughout our voyage. There was hardly a day on which we did not wonder whether it would be our last. We were compelled to drop the precious metals we had taken with us overboard to steady the ship, feeling as though it were a squalid sacrifice to the gods to give us hope to live a few hours longer.]])
+-- TRANSLATORS: Reebaud – Introduction 1
+local intro_1_picks = _([[The few tools we managed to salvage from Langeness did not cope well. The miners’ picks proved especially vulnerable, and we had to throw them in the ocean when it became evident that they had been corroded by the sea salt so much that we could never again mine anything with them.]])
+-- TRANSLATORS: Reebaud – Introduction 1
+local intro_1_end = _([[But we survived. Finally, we reached the mouth of a calmer fjord. We sailed upriver, and arrived here.]])
+local intro_1_body
```
hessenfarmer: suggestion is to shorten this with 
local intro_1_body = ""
if  lost_metals then
   intro_1_body = intro_1_body .. intro_1_iron .. paragraphdivider()
end
if lost_miners then
   intro_1_body = intro_1_body .. intro_1_picks .. paragraphdivider()
end

hessenfarmer (approved):
one small sugesstion 
Code and txts LGTM 
Thanks doing it in this direction

hessenfarmer:
as this is green now I'll do the merge

PR #4333, opened 2020-10-09
fix territorial messages by resetting last winning candidate values

### LINKED ISSUE #4327: territorial Win conditions messages broken
If a player has 50% of the land he starts to receive winning messages. however this is only cleared iof another player has more then 50% and not if nobody has 50% anymore.   
see https://www.widelands.org/forum/post/34971/

hessenfarmer:
if no winner candidate available
fixes #4327


gunchleoc (approved):
Code LGTM, not tested.

PR #4344, opened 2020-10-11
Chat recipients

### LINKED ISSUE #2082: Multiplayer: Chatting to all spectators only
It would be nice to have a special "spectators" chat, so all spectators of a game can view the message, but not the players.

Maybe this could be implemented like the personal comment '@username'. E.g. '@spectators' sends the message to all spectators but not to players.

 * date created: 2017-10-18T20:27:02Z
 * owner: franku
 * the launchpad url was https://bugs.launchpad.net/bugs/1724680

Notabilis:
Changes to the chat panel:
- Adding a dropdown to select the recipient of chat messages
- Allow to send messages to only the team members / the other spectators
- Auto-completion of player names by pressing space twice at the end of the line
- "team" is no longer a valid player name

Commits have been created for easier review. Two of them don't compile but the final state should be fine.

Fixes #2082


src/network/gamehost.cc
```diff
@@ -421,6 +422,8 @@ struct HostChatProvider : public ChatProvider {
 struct Client {
 	NetHostInterface::ConnectionId sock_id;
 	uint8_t playernum;
+	// TODO(Notabilis): usernum is int16_t while UserSettings::position is uint8_t.
+	//                  Should this be the same data type?
```
gunchleoc: See this note in `gamesettings.h`:

```
// PlayerSlot 0 will give us Widelands::PlayerNumber 1 etc., so we rename it to avoid confusion.
// TODO(GunChleoc): Rename all uint8_t to PlayerSlot or Widelands::PlayerNumber
using PlayerSlot = Widelands::PlayerNumber;
```

src/network/gamehost.cc
```diff
@@ -806,95 +820,146 @@ void GameHost::send_player_command(Widelands::PlayerCommand* pc) {
...
 void GameHost::send(ChatMessage msg) {
 	if (msg.msg.empty()) {
+		// No message: Nothing to do
 		return;
 	}
 
+	// The set containing all receivers of the message.
+	// Being a set ensures that each receiver only gets the message once
+	std::set<int32_t> recipients;
+	// Whether this is a public (0), personal (1), or team (2) message (see protocol definition)
```
gunchleoc: Sticking these into an enum class might make the code more readable

src/network/gamehost.cc
```diff
@@ -806,95 +820,146 @@ void GameHost::send_player_command(Widelands::PlayerCommand* pc) {
...
+	int msg_type = 0;
+	// Figure out who to send the message to
 	if (msg.recipient.empty()) {
```
gunchleoc: Reverse this `if (!msg.recipient.empty()) {` and then get rid of the else

src/network/gamehost.cc
```diff
@@ -806,95 +820,146 @@ void GameHost::send_player_command(Widelands::PlayerCommand* pc) {
...
+		if (msg.recipient != "team") {
+			// A single player is the recipient. Find it
+			const int32_t client_id = check_client(msg.recipient);
+			assert(client_id >= -2);
+			if (client_id == -1) {
+				// Error: Can't find given recipient, so we only will send
+				// an error message back to the sender
+				msg.sender.clear();
+				msg.playern = -2;
+				// TODO(Notabilis): Maybe make this a command so it can be localized on the client
```
gunchleoc: +1 to that idea :)

src/network/gamehost.cc
```diff
@@ -917,6 +982,10 @@ int32_t GameHost::check_client(const std::string& name) {
 	uint32_t client = 0;
 	for (; i < d->settings.users.size(); ++i) {
 		const UserSettings& user = d->settings.users.at(i);
+		if (user.position == UserSettings::not_connected()) {
+			// Ignore users thats are not yet fully or not anymore connected
```
gunchleoc:
```suggestion
			// Ignore users that are not fully connected yet or who lost/broke the connection
```

src/network/internet_gaming.cc
```diff
@@ -1043,13 +1042,18 @@ void InternetGaming::format_and_add_chat(const std::string& from,
 }
 
 /**
- * Check for vaild username characters.
+ * Check for vaild username characters and make sure its not "team".
```
gunchleoc:
```suggestion
 * Check for vaild username characters and make sure it's not "team".
```

gunchleoc (commented):
First round of review. Not tested yet.

src/network/gamehost.cc
Notabilis: I think these are different problems, or am I mistaken?
Not that I don't totally agree with your TODO.

src/network/gamehost.cc
Notabilis: An enum class is probably too strict for it since I have to write it as integers to the network packet.
I could introduce three `const uint8_t`, if that is acceptable? Or a normal enum, should work as well.

Notabilis:
Thanks for the first round of the review, 4 of 6 should be fixed now.

src/network/gamehost.cc
gunchleoc: I mean that the Datatype should be PlayerNumber or PlayerSlot, so the answer to your question is yes, this should be the same data type.

src/network/gamehost.cc
gunchleoc: Either constexpr or enum would work for me - the most important thing is that they have names.

gunchleoc:
I can't test the team chat because of https://github.com/widelands/widelands/issues/4367 but the PMs are working fine. DIdn't test the player name completion yet.

Niektory:
Name autocompletion works, except when the player's name contains an `@`. After typing the `@` in the name you can no longer autocomplete.

Double-`Space` is a little unusual, I would expect it to be done with the `Tab` key. But if that's difficult to reconcile with Tab-navigation then this solution works too.

Every time the chat window is opened in game, a message sound is played.

Team chat works, also for spectators.

There are some warnings:
```
/home/mnj/widelands/src/network/CMakeLists.txt:1: network misses DEPENDS on logic_constants, because it includes logic/widelands.h
/home/mnj/widelands/src/wui/CMakeLists.txt:12: wui_chat_ui misses DEPENDS on graphic_image_cache, because it includes graphic/image_cache.h
/home/mnj/widelands/src/wui/CMakeLists.txt:12: wui_chat_ui misses DEPENDS on network, because it includes network/participantlist.h
```

Notabilis:
Thank you for testing.

The @ character is a good catch, I wasn't aware that it is an allowed character. I will fix it for the autocomplete, but the problem remains for whispering (even in master): When there are two users named `@name` and `name`, it is not possible to figure out whether the user wants to whisper to `name` or just mention `@name` in its message. I would suggest forbidding the @ character at the beginning of names (or completely). What do you think about that?

I would prefer the Tab key as well, but that is taken for the navigation. Also, I can't directly receive an event for it from the edit box (I haven't checked whether I can get it elsewhere). After getting used to it double-space was fine for me, but it is indeed unusual (and probably deserves some tip/documentation somewhere).

> Every time the chat window is opened in game, a message sound is played.

Good point. That might be due to a "fix" I tried for a bug in master (chat history is not displayed ingame in the chat window). Guess I need to find another fix or just leave the bug open in this branch.

Regarding the warnings: Turns out fixing them in the wrong folder does not really help. ;-)

Niektory:
> The @ character is a good catch, I wasn't aware that it is an allowed character. I will fix it for the autocomplete, but the problem remains for whispering (even in master): When there are two users named `@name` and `name`, it is not possible to figure out whether the user wants to whisper to `name` or just mention `@name` in its message. I would suggest forbidding the @ character at the beginning of names (or completely). What do you think about that?

I wouldn't mind forbidding it. I don't remember seeing anyone using it so probably few people would miss it :-)

> > Every time the chat window is opened in game, a message sound is played.
> 
> Good point. That might be due to a "fix" I tried for a bug in master (chat history is not displayed ingame in the chat window). Guess I need to find another fix or just leave the bug open in this branch.

I think the sound is less annoying than missing history, so it's still an improvement :-P

Notabilis:
'@' as first letter is now forbidden, autocomplete works with '@' in the middle of the name.
The sound is gone, the text is still displayed.

gunchleoc:
> I would suggest forbidding the @ character at the beginning of names (or completely). What do you think about that?

+1

How about using Ctrl+Tab or Ctrl+Space for autocompletion?

src/network/participantlist.cc
```diff
@@ -0,0 +1,268 @@
...
+	// Check whether we are a player or a spectator
+	bool found_someone = false;
+	if (get_participant_type(my_index) == ParticipantType::kSpectator) {
+		// We are a spectator. Check if there are other spectators
+		for (int16_t i = 0; i < participant_counts_[0]; ++i) {
+			if (get_participant_type(i) != ParticipantType::kSpectator) {
+				continue;
+			}
+			if (found_someone) {
+				// The first one we find might be we. If we find a second one,
```
gunchleoc:
```suggestion
				// The first one we find might be us. If we find a second one,
```

src/network/participantlist.h
```diff
@@ -0,0 +1,227 @@
...
+	 * @param participant The index of the participant to get data about.
+	 * @return The RTT in milliseconds for this participant up to 255ms.
+	 */
+	uint8_t get_participant_ping(int16_t participant) const;
+
+	/// Called when the underlying data was updated and should be re-fetched.
+	boost::signals2::signal<void()> participants_updated;
+
+	/**
+	 * Called when the RTT for a participant changed.
```
gunchleoc, line 189: I don't do networking a lot, so I have no clue what RTT means.

src/network/participantlist.h
```diff
@@ -0,0 +1,227 @@
...
+	void update_participant_counts();
+
+	/// A pointer to the settings of the current game (running or being prepared).
+	const GameSettings* settings_;
+	/// A reference to the pointer of a currently runnning game.
+	Widelands::Game*& game_;
+	/// A reference to the user name of the human on this computer.
+	const std::string& localplayername_;
+	/// Counts of current participants: humans, AIs, total
+	int16_t participant_counts_[3];
```
gunchleoc: The code would be a lot easier to understand if you did this:

```
struct ParticipantCounts {
    int16_t ais;
    int16_t humans;
    int16_t total;
};
```

src/ui_fsmenu/netsetup_lan.cc
```diff
@@ -278,7 +278,8 @@ void FullscreenMenuNetSetupLAN::change_playername() {
 	if (!InternetGaming::ref().valid_username(playername_.text())) {
 		playername_.set_warning(true);
 		playername_.set_tooltip(_("Enter a valid nickname. This value may contain only "
-		                          "English letters, numbers, and @ . + - _ characters."));
+		                          "English letters, numbers, and @ . + - _ characters "
```
gunchleoc, line 285: Didn't we just disallow `@`?

src/wui/login_box.cc
```diff
@@ -164,7 +164,8 @@ void LoginBox::verify_input() {
 	} else if (!InternetGaming::ref().valid_username(eb_nickname->text())) {
 		eb_nickname->set_warning(true);
 		eb_nickname->set_tooltip(_("Enter a valid nickname. This value may contain only "
-		                           "English letters, numbers, and @ . + - _ characters."));
+		                           "English letters, numbers, and @ . + - _ characters "
```
gunchleoc: `@` is no longer allowed?

src/ui_fsmenu/netsetup_lan.cc
Notabilis: `@` in the middle of the name is still allowed, only at the beginning of the name its forbidden now. We can also forbid it completely if you prefer it that way.
I didn't updated the tooltips on purpose since the sentence is long enough even without writing about this corner case.

gunchleoc: Let's leave it for now.

gunchleoc (approved):

gunchleoc:
Do you still want to take a stab at using Ctrl+Space for autocompletion in this branch? If not, that's fine too.

Notabilis:
I like the idea of Ctrl+Space, but I still have to figure out how to receive the event. I will give it a try, if I fail I will merge this branch.

gunchleoc:
I have cancelled all other older builds now to get this in ASAP.

Niektory:
Ctrl+Space is sometimes used for switching input methods (in [fcitx](https://wiki.archlinux.org/index.php/Fcitx#Desktop_Environment_Autostart), and I think also in Chromebooks and older Windows versions), so you might want to keep double-Space as an alternative.

Notabilis:
Ctrl+Space and doubled space are now both available to trigger autocomplete.
The behavior has three bugs, can you find them? ;)
Okay, no idea how many bugs there are. But considering how many I already encountered there probably are some more still in there. So please test both completion triggers.

Niektory:
I tested for a bit and didn't run into any bugs. I'd include the `@` restriction in the tooltip though.

~~If you think that makes the tooltip too long, disallowing it completely would be fine too.~~ Scratch that, matching the rules to the website would probably be best.

gunchleoc:
Tested and working :)

As to the @, if we allow it on the website for usernames, we have to allow it here too. Otherwise, people who use @ in their username on widelands.org won't be able to log in for the internet lobby.

I think the best solution would be to show the exact reason for the failure below the input box. I'd say that's out of scope for this branch though.

frankystone:
@ is allowed in usernames:

[The django docs says:](https://docs.djangoproject.com/en/dev/ref/contrib/auth/#django.contrib.auth.models.User.username)

> 150 characters or fewer. Usernames may contain alphanumeric, _, @, +, . and - characters.

I remember at least one user to had a username like an e-Mail address.

Notabilis:
Hm, even if we only forbid @ as first character it differs to the website, right? Is it possible to disallow it as first character for the website as well?

frankystone:
> Hm, even if we only forbid @ as first character it differs to the website, right?

Yes.

> Is it possible to disallow it as first character for the website as well?

Hm, possibly. Probably with big effort.

I took a look at the usernames wich are registered on the website and which contain an @-sign. The result is:

* @-sign in front: 1 user
* @-sign in the middle: 5 users

Overall: 5028 users

Maybe the easiest thing is to add a hint in the registration form.

Niektory:
OK, then how about just allowing them like this: https://github.com/widelands/widelands/commit/c97694e879466507029a901e55183f7ccfd2c405
Messaging (also with autocompletion) such players will work fine if you put `@` before their names as usual. The only problem is that such a player's name at the beginning of a line sent to all players will be misinterpreted as a private message and will not autocomplete. Not ideal, but not a catastrophe either – can be worked around by typing anything before the name – and no worse from how it is in master.

frankystone:
Yes, i think that would be fine.

Notabilis:
Would be okay for me as well. I integrated it into this branch.
As far as I am concerned this can be merged when the tests are successful.

gunchleoc:
+1 from me too when the tests pass.

PR #4353, opened 2020-10-12
Remove ObjectManager::object_still_available

Niektory:
So I did this thing, but to be honest I can't reproduce the performance problem this is supposed to fix with any of my savegames. The save from [the original report](https://www.widelands.org/forum/topic/4938/?page=1#post-34444) won't load in current master so can't test with that one. There might be something funky going on with that Europeans tribe, or maybe you really have to play for 160 hours for the problem to appear :-) 

So it's not a high priority change in these circumstances.

This is reviewable commit by commit.


src/logic/map_objects/map_object.h
```diff
@@ -423,6 +421,7 @@ struct ObjectManager {
 
 /**
  * Provides a safe pointer to a MapObject
+ * Make sure the MapObject is initialized (has a serial) before using it in ObjectPointer!
```
Noordfrees, line 424: Please add an assert for that in the ObjectPointer and OPtr ctors

src/economy/waterway.cc
```diff
@@ -227,8 +226,8 @@ void Waterway::cleanup(EditorGameBase& egbase) {
 void Waterway::log_general_info(const EditorGameBase& egbase) const {
 	MapObject::log_general_info(egbase);
 
-	molog(egbase.get_gametime(), "Ferry %u\n", ferry_ ? ferry_->serial() : 0);
-	molog(egbase.get_gametime(), "FerryFleet %u\n", fleet_ ? fleet_->serial() : 0);
+	molog(egbase.get_gametime(), "Ferry %u\n", ferry_.get(egbase) ? ferry_.serial() : 0);
+	molog(egbase.get_gametime(), "FerryFleet %u\n", fleet_.get(egbase) ? fleet_.serial() : 0);
```
Noordfrees, line 230: No need to use get() here, just print the pointer's serial()

src/economy/waterway.cc
```diff
@@ -94,13 +94,12 @@ void Waterway::remove_worker(Worker& w) {
...
-	fleet->request_ferry(egbase, this);
-	fleet->init(egbase);
+	fleet->init(egbase, this);
 }
 
 void Waterway::assign_carrier(Carrier& c, uint8_t) {
-	if (ferry_) {
-		ferry_->set_location(nullptr);
+	if (ferry_.get(owner().egbase())) {
+		ferry_.get(owner().egbase())->set_location(nullptr);
```
Noordfrees: Duplicate calls to get() should be made more efficient by caching the return value, e.g.
```suggestion
	if (Ferry* f = ferry_.get(owner().egbase())) {
		f->set_location(nullptr);
```
There are more places in the diff

src/economy/ferry_fleet.cc
```diff
@@ -84,6 +84,14 @@ bool FerryFleet::init(EditorGameBase& egbase) {
 	return find_other_fleet(egbase);
 }
 
+bool FerryFleet::init(EditorGameBase& egbase, Waterway* waterway) {
```
Noordfrees, line 87: What is the purpose of this change?

src/economy/ferry_fleet.cc
```diff
@@ -237,7 +244,16 @@ void FerryFleet::remove_ferry(EditorGameBase& egbase, Ferry* ferry) {
 		              serial(), ferry ? ferry->serial() : 0);
 		return;
 	}
-	ferries_.erase(it);
+	while (it != ferries_.end()) {
+		ferries_.erase(it);
+		it = std::find(ferries_.begin(), ferries_.end(), ferry);
+		if (it != ferries_.end()) {
+			log_err_time(egbase.get_gametime(),
+			             "FerryFleet %u: Multiple instances of ferry %u were in the ferry fleet\n",
```
Noordfrees: Does this actually happen?
```suggestion
			log_warn_time(egbase.get_gametime(),
			             "FerryFleet %u: Multiple instances of ferry %u were in the ferry fleet",
```

src/logic/map_objects/tribes/worker.cc
```diff
@@ -1334,9 +1334,7 @@ void Worker::cleanup(EditorGameBase& egbase) {
 	}
 
 	if (ware) {
-		if (egbase.objects().object_still_available(ware)) {
-			ware->destroy(egbase);
-		}
+		ware->destroy(egbase);
```
gunchleoc, line 1337: No OPtr on purpose?

Niektory:
Alright, I managed to produce a savegame that reproduces the performance problem and loads in master: [ice_wars_lots_of_workers_test.wgf.zip](https://github.com/widelands/widelands/files/5377593/ice_wars_lots_of_workers_test.wgf.zip)
I'll address the reviews later.

src/logic/map_objects/tribes/worker.cc
Niektory: It's read from this OPtr:
https://github.com/widelands/widelands/blob/a47dc91293937a3f69d64a32a12ecb370412c047/src/logic/map_objects/tribes/worker.h#L294

src/economy/ferry_fleet.cc
Niektory: It shouldn't. But ships at least had this problem at some point, so I'm a little wary.

src/economy/ferry_fleet.cc
Niektory: `Waterway::request_ferry()` called `FerryFleet::request_ferry()` before `FerryFleet::init()`. This caused trouble when I switched `Waterway:fleet_` to `OPtr`, because it was being assigned a `FerryFleet` without a serial.

gunchleoc (approved):

PR #4354, opened 2020-10-12
fix player name

### LINKED ISSUE #4346: LaunchSPG: Switching player slot messes up player names
I was using the map [Merkur](https://www.widelands.org/maps/merkur/). Start a new singleplayer game with this map – player 1 is called "Menschheit" and player 2 is called "Die Außerirdischen". When switching to slot 2, both players are now called "Die Außerirdischen". The names will stay like this when switching back to slot 1.

jmoerschbach:
fixes #4346


Noordfrees (approved):
LGTM, fix confirmed :)

Issue #1406, opened 2019-09-09
Issues reported by cppcheck


bunnybot:
Attached you should find an up to date report of the issues found by cppcheck. This is generated by running the script utils/create_cppcheck_report. Be aware that this takes a while to run.

It currently lists various issues, the majority seem to be parameters which should be passed as reference, values not initialized by the constructor.

PS. For other warning reports, see bug 1258667 (Clang), bug 1278174 (flawfinder) and bug 1202101 (Visual Studio).

------------------------------------
Imported from Launchpad using lp2gh.

 * date created: 2012-04-21T16:21:07Z
 * owner: hjd
 * the launchpad url was https://bugs.launchpad.net/bugs/986611

bunnybot:
(by hjd)
Based on widelands r6346 (with a few additional fixes not yet merged) and cppcheck 1.54.

Note that while generating your own report will list roughly ~40 more issues, but these already have suggested patches, so please don't waste your time doing duplicate work.

Also note that cppcheck 1.49-1 (at least in Ubuntu 11.10) will list all issues each time it finds it, while later versions only lists each issue only once. Without the duplicate lines, the report is a lot less noisy and easier to read. (The difference was ~260 vs ~12000 lines)

bunnybot:
(by hjd)
Seems cppcheck 1.55 was recently released, so I generated a new report with that based on the latest revision (r6411). Some hightlights from the report:

* Found a couple of memory leaks...
* ...and some variables nulled, but not freed
* usleep is obsolete
* It notes some places where replacement functions which are recommended for threadsafe applications
* Various variables not initialized in the constructor.
* Unused variables (some of these seem like they could be false positives to me, so please check those carefully if you intend to fix them)
* Variables passed as value when they could have been passed as value instead of reference
* Exceptions caught as values
* many more...

bunnybot:
(by sirver)
I merged your branch. Likely this does not make this bug fixcommitted, does it :).

Thanks for your work hjd!

bunnybot:
(by hjd)
Unfortunately, no, so hopefully someone else will take a look at the report. ;)

It should resolve four out of ~300 issues though.

bunnybot:
(by qcumber-some)
Attached is the current report of cppcheck 1.55 at revision 6416.

Looks much better now :-)

bunnybot:
(by hjd)
A new version of cppcheck was recently released, so I tested lastet trunk with that to see whether it found any new issues:
* Some unused functions 
[...]

PR #4355, opened 2020-10-12
Fix simple cppcheck issues

matthiakl:
Re #1406
I worked through approx. half of the sources fixing simple cppcheck issues (mostly `const` and `explicit` declarations).
I also updated the `create_cppcheck_report` to work with newest version and github links.

[cppcheck_report.html](https://github.com/widelands/widelands/files/5367268/cppcheck_report.html.txt)


src/editor/map_generator.cc
```diff
@@ -846,7 +846,7 @@ int UniqueRandomMapInfo::map_id_char_to_number(char ch) {
 	} else if ((ch == '1') || (ch == 'l') || (ch == 'L') || (ch == 'I') || (ch == 'i') ||
 	           (ch == 'J') || (ch == 'j')) {
 		return 23;
-	} else if (ch >= 'A' && ch < 'O') {
+	} else if (ch >= 'A' && ch < 'Z') {
```
gunchleoc:
```suggestion
	} else if (ch >= 'A' && ch <= 'Z') {
```

src/logic/map_objects/tribes/soldier.cc
```diff
@@ -1231,9 +1231,6 @@ void Soldier::defense_update(Game& game, State& state) {
 	if (soldiers.empty() || (get_current_health() < get_retreat_health())) {
 		if (get_retreat_health() > get_current_health()) {
 			assert(state.ivar1 & CF_RETREAT_WHEN_INJURED);
-		}
-
-		if (get_current_health() < get_retreat_health()) {
```
gunchleoc, line 1236: Swap the molog calls below - the comparator is the other way around

gunchleoc:
Testing done, so when the 2 comments are fixed, this can go in.

src/logic/map_objects/tribes/soldier.cc
matthiakl: They are the same: the comparator is the other way around, but the operands are also switched. Both conditions match the one in line 1231, so I changed it to this notation for easier understanding. The `else` part corresponds to `soldiers.empty()`.
Thanks for the fast review :)

gunchleoc (approved):

gunchleoc:
We have a test failure now as well as a bunch of new compiler warnings https://travis-ci.org/github/widelands/widelands/jobs/735362661

matthiakl:
The compiler warning should be fixed, however I can't reproduce it locally (clang-10).
Testcase is also fixed. I added a comment to prevent future cppcheck reviews from the same problem.
+1 for the testsuite beating cppcheck :)

gunchleoc:
There's a merge conflict now

PR #4358, opened 2020-10-13
Add tooltips to tablerows

### LINKED ISSUE #4308: MP Lobby: Colummn for version is too small
**Describe the bug**
When entring the internet lobby the column for the version string is too small to fit the github version string:

mp_version_colummn

**To Reproduce**
1. Enter the online lobby with a github branch compiled

**Expected behavior**
Since the name of a game is usually very small, make the column for 'Game' smaller and the column for 'Version' bigger. Maybe make both 50% of the resulting width.

**Version:**
 - OS: archlinux
 - Widelands Version: [6e883c7@multithreading] (Debug)

jmoerschbach:
tables can now show tooltips for every row (always, never and default=when column width is too small for the text)

grafik

Fixes #4308 (simply increasing column width is not enough, there will never be enough space, especially in 800x600)

I more or less copied tooltip behaviour from Panel... currently, row tooltips are overlayed by subsequent rows because these are drawn after the tooltip, needs polishing. Worth the effort? Better ideas?


frankystone:
Maybe the version string can be shorten?

Say the string is `1.0~git341[322ee26@multithreading_new] (Debug)` the important parts are the branchcommit and branchname. Maybe also if it is a release or debug build. So cutting of parts will result in:

Before:
* 1.0~git341[322ee26@multithreading_new] (Debug)

After:
* 322ee26@multithreading_new

or
* 322ee26@multithreading_new (Debug)

But i am not sure how the version string for releases will look like in the future.

gunchleoc:
The release could just be `1.0 (Release)` or `1.0 (Debug)`. The sequential number is also needed to people can understand which build is newer - git hashes are not human readable.

> currently, row tooltips are overlayed by subsequent rows because these are drawn after the tooltip, needs polishing. Worth the effort? Better ideas?

Override the `handle_tooltip`  function from panel rather than adding a separate function.

frankystone:
> (simply increasing column width is not enough, there will never be enough space, especially in 800x600)

Yes, nevertheless the column width could be increased. Or exchanging the columns `Version` and `Game`?

Tooltips are also fine for me :-)

> The release could just be `1.0 (Release)` or `1.0 (Debug)`. 

The `(Release)` can be omitted, `(Debug)` is the most interesting part, imho. So everything without `(Debug)` will be a release build.

> The sequential number is also needed to people can understand which build is newer - git hashes are not human readable.

What is the sequential number?

The hash is important to see which revision of a branch this is.

jmoerschbach:
> 
> 
> > (simply increasing column width is not enough, there will never be enough space, especially in 800x600)
> 
> Yes, nevertheless the column width could be increased. Or exchanging the columns `Version` and `Game`?
> 
I disagree. The width is just fine for official builds (e.g. "build-21"). Widen it would waste space. Only git branches (which are only used by us developers...?!) have a "problem" ;)
>
> Tooltips are also fine for me :-)
> 
Good, I implemented them now correctly :)
> > The release could just be `1.0 (Release)` or `1.0 (Debug)`.
> 
I won't touch the version string itself because I think with the tooltip everything is set. Of course, this applies to not just in this screen but all potential places where tables are used, see screenshots:

 
grafik

grafik

Niektory:
I think it works nicely now. 👍

gunchleoc:
`TableRowTooltip::kDefault` Is the only mode used in the code right now. Do we need the other modes?

jmoerschbach:
> 
> 
> `TableRowTooltip::kDefault` Is the only mode used in the code right now. Do we need the other modes?

I don't know... I can remove it if you like

gunchleoc:
I think we better remove them - code paths that are never executed tend to accumulate bugs.

jmoerschbach:
> 
> 
> I think we better remove them - code paths that are never executed tend to accumulate bugs.

done

gunchleoc (approved):

Issue #4341, opened 2020-10-11
Tutorials: Automatic mouse move does not work with drop-up menu


frankystone:
In the first tutorial the section to show/hide buildhelp does not work correctly. The mouse pointer moves to the bottom show/hide menu but not walks up the menu entries. The mouse pointer sticks at the bottom menu.

**To Reproduce**
1. Start first tutorial
2. Do the things the tutorial wants you to do
3. The message box to toggle the buildhelp appears
4. after clicking OK you will see

**Expected behavior**
The mouse pointer should go up the menu entries until it reaches the 'Show Buildhelp' menu entry and should make a click there.

**Version:**
 - OS: archlinux
 - Widelands Version: [7a3a5ab@master]

flegmaatikko:
I can reproduce the bug. I think the incorrect behavior happens two times. The menu opens again after it was closed. But this may be the wanted behaviour.
There are also multiple clicking sounds when clicking the OK button of the  "Welcome the Widelands!" message box.

gunchleoc:
The dropdown entries are selected by simulating a keyboard shortcut. I decided against using mouse clicks there, because if we redesign anything in those menus that changes the height of the entries, it will stop working.

If somebody wants to implement detection for dropdown menu list position, width and entry height for Lua, that will be fine and certainly an improvement for this scenario.

frankystone:
I think for the menus we do not need such 'I'll show you how ...' with automatic mouse moves, maybe only 'I'll show where to find ...' and another message box describing the menu and the corresponding shortcuts. In this case the message box should request to press 'space' (or whatever one is wanted).

flegmaatikko:
Sounds good. The tutorial continues after the player has practiced enough and has closed the message box.
It requires some kind of selective user input blocker where only the corresponding shortcuts are active.

gunchleoc:
The tutorial continues when the player has done it, like for all other objectives. We use the actual state of the option for detection, so no problem there.

So, "I'll show you where the menu is" and click on the menu button, then let the user select the item from the dropdown or press the hotkeys as they see fit.

frankystone:
I think this is fixed...

PR #4363, opened 2020-10-14
Fast fix for the drop up menu bug #4341 

flegmaatikko:
Here is my first solution for the bug #4341 
I created a new method that shows where the target button is and did minor changed to the texts. There is also an unused method that opens the menu only. It can be removed if needed.

The mouse pointer moves over the menu but not over to the button. Instead of moving the mouse pointer I tried to show the position of target menu item by highlighting the buttons step by step from bottom to up.

This is just a fist draft and I look forward to get some feedback.

### DIFF
--- a/data/campaigns/tutorial01_basic_control.wmf/scripting/helper_functions_demonstration.lua
+++ b/data/campaigns/tutorial01_basic_control.wmf/scripting/helper_functions_demonstration.lua
@@ -71,12 +71,8 @@ end
 
 function select_item_from_dropdown(name, item)
    local blocker = UserInputDisabler:new()
-
-   wl.ui.MapView().dropdowns[name]:highlight_item(item)
-   sleep(5000)
    wl.ui.MapView().dropdowns[name]:select()
    sleep(3000)
-
    blocker:lift_blocks()
 end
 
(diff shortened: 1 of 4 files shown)


gunchleoc:
I like it, you can continue like that for the other dropdowns in this turorial :)

flegmaatikko:
Hopefully all the cases are fixed now.

data/campaigns/tutorial01_basic_control.wmf/scripting/helper_functions_demonstration.lua
```diff
@@ -71,12 +71,25 @@ end
...
+   blocker:lift_blocks()
+end
 
+function open_item_from_dropdown(name, item)
+   wl.ui.MapView().dropdowns[name]:open(item)
+end
+
+function show_item_from_dropdown(name, item)
+   local blocker = UserInputDisabler:new()
+   local ind = wl.ui.MapView().dropdowns[name]:get_no_of_items()
```
gunchleoc:
```suggestion
   local ind = wl.ui.MapView().dropdowns[name].no_of_items
```

src/scripting/lua_ui.cc
```diff
@@ -367,6 +367,7 @@ const MethodType<LuaDropdown> LuaDropdown::Methods[] = {
    METHOD(LuaDropdown, open),
    METHOD(LuaDropdown, highlight_item),
    METHOD(LuaDropdown, select),
+   METHOD(LuaDropdown, get_no_of_items),
```
gunchleoc: Remove this. It's available as a property and we don't need to add a method too.

gunchleoc (changes requested):

gunchleoc (approved):

PR #4369, opened 2020-10-16
Training wheels

gunchleoc:
Implements a system to gradually teach the player things in any non-scenario single player game, with any tribe, on any map, with any starting condition.

The concept is that we have a global conf file that records which things have been learned already, and to keep the teaching points in small, atomic scripts. Which scripts are active can be controlled by a dependency chain in the init.lua.

While scenarios don't run the training wheels themselves, they can still mark them as solved if they teach the same thing.

There is also a locking system to prevent concurrent execution.

One drawback: Like scenarios, replays will desync, so we can't write replays while training wheels are active.

Vector graphics for the pointy arrow thing: 
[arrow.svg.zip](https://github.com/widelands/widelands/files/5395624/arrow.svg.zip)

Screenshots:

shot0000
shot0001


Bjoern2Settler:
With regards to the motivationcurve of the player - maybe we should not disallow options, but give a nice motivationcurve of reaching the first upgrades and trainingssteps very easily. As imagined playing the game for the first time you want to aim for your first upgrade very heavily and then get the fun of trying out more.

gunchleoc:
> With regards to the motivationcurve of the player - maybe we should not disallow options, but give a nice motivationcurve of reaching the first upgrades and trainingssteps very easily. As imagined playing the game for the first time you want to aim for your first upgrade very heavily and then get the fun of trying out more.

This is completely off-topic for this pull request, which is implementing engine support. Please do stay on topic.

src/wui/story_message_box.cc
```diff
@@ -72,16 +72,27 @@ StoryMessageBox::StoryMessageBox(Widelands::Game* game,
 	}
 	move_inside_parent();
 	textarea_.focus();
+
+	if (!is_modal()) {
+		resume_game();
+	}
```
Noordfrees, line 78: Why not just skip pausing the game in the first place?

src/wui/interactive_player.cc
```diff
@@ -593,6 +605,23 @@ void InteractivePlayer::popup_message(Widelands::MessageId const id,
 	dynamic_cast<GameMessageMenu&>(*message_menu_.window).show_new_message(id, message);
 }
 
+void InteractivePlayer::set_training_wheel_indicator_pos(const Vector2i& pos) {
+	constexpr int kTrainingWheelArrowOffset = 5;
+	if (pos == Vector2i::invalid()) {
+		training_wheel_indicator_icon_.reset(nullptr);
+	} else {
+		// We create a new icon every time to paint it on top of the other child panels
```
Noordfrees, line 615: Won't it get hidden behind any new windows then? Perhaps the same technique as for the field arrow would be safer

src/scripting/lua_ui.cc
```diff
@@ -710,6 +769,21 @@ int LuaMapView::get_is_animating(lua_State* L) {
 	lua_pushboolean(L, get()->map_view()->is_animating());
 	return 1;
 }
+
+/* RST
+   .. attribute:: interactive_player
+
+      (RO) The player number for the interactive player.
+*/
+int LuaMapView::get_interactive_player(lua_State* L) {
```
Noordfrees: This is already a property in lua_root. Don't invent the wheel again ;)

Noordfrees (commented):
Tested this briefly. The arrow does not fit to Widelands in its style, IMHO. A few comments inline, I only had a very brief look at the code so far though.

src/logic/training_wheels.h
```diff
@@ -0,0 +1,108 @@
...
+	 * @brief load_objectives Check for training wheel objectives that can be run
+	 */
+	void load_objectives();
+	/**
+	 * @brief write Write configuration to file
+	 */
+	void write();
+
+	/**
+	 * @brief A training whell's script and filename
```
Noordfrees: typo

Noordfrees (commented):
It would be good to be able to reset the progress.
There could be something like this mockup:
shot0000
The Reset button would bring up a confirmation dialogue to clear the progress so the user can get all hints as if he only just started playing Widelands.

src/logic/training_wheels.cc
```diff
@@ -0,0 +1,131 @@
...
+#include "logic/training_wheels.h"
+
+#include <memory>
+
+#include "base/log.h"
+#include "io/filesystem/layered_filesystem.h"
+#include "logic/filesystem_constants.h"
+#include "scripting/lua_table.h"
+
+const std::string kTrainingWheelsFile = kSaveDir + g_fs->file_separator() + "training_wheels.conf";
```
Noordfrees: This doesn't belong in the `save` dir. Please create a new dir `~/.widelands/config` where this should go (and where the other config files should also be moved, but this could also be done later)

Noordfrees: And please move this to `filesystem_constants.h`

data/scripting/training_wheels/utils/lock.lua
```diff
@@ -0,0 +1,23 @@
+-- TODO(GunChleoc): Document more fully
+
+-- Sleep until we aqcuire permission to run
```
Noordfrees: typo

src/wui/story_message_box.cc
gunchleoc: Because the constructor does this:

```
	if (coords != Widelands::Coords::null()) {
		game_->get_ipl()->map_view()->scroll_to_field(coords, MapView::Transition::Jump);
	}
```

I didn't want player actions to interfere here, might result in unexpected behavior.

src/wui/interactive_player.cc
gunchleoc: Windows get painted on top of the map overlays, so this wouldn't work.

I'm not too bothered if the user decides to open a new window on top of the arrow - I expect them not to be interested in the arrow anyway if they do that. If it's us that wants the user to open a new window on top of it, we'll want to point to something in that new window instead anyway.

src/scripting/lua_ui.cc
gunchleoc: Thanks, I didn't see that one. Will fix.

gunchleoc:
If somebody can create a better arrow image, please do! My graphical skills are limited.

I agree with all the other suggestions that I didn't comment on.

For resetting the progress, we might even have a list of training wheels for the user to select individually. Good idea/bad idea?

Noordfrees:
> For resetting the progress, we might even have a list of training wheels for the user to select individually. Good idea/bad idea?

For now this would be overkill imho. But when the number of wheels has grown a lot in the future we could still implement this then

frankystone:
Will this replace the tutorials we have so far?

Just want to know if working on the (old) tutorials is useful atm.

gunchleoc:
I think we should suspend working on the old tutorials for a while until we have more of these new training wheels written. I'll have a better feel for things once we have that done. We could then still keep a version of the tutorials around, or turn training wheels into tutorial scenarios (maybe composed of them).

We definitely should get rid of the walls of texts in the military tutorial and move the more detailed information over to the encyclopedia.

src/logic/training_wheels.cc
gunchleoc: > This doesn't belong in the `save` dir. Please create a new dir `~/.widelands/config` where this should go (and where the other config files should also be moved, but this could also be done later)

This is actually more difficult than it looks, because we also support the XDG standard. This will need a wider refactoring.

Noordfrees (approved):
The wheel itself could do with some improvements as I already said on the forum, but code and concept LGTM now :)

src/logic/training_wheels.cc
Noordfrees: Let's keep it in the save dir for now then. I opened a new issue for this: https://github.com/widelands/widelands/issues/4383

gunchleoc:
The test suite was triggering the training wheels, so I had to change the implementation in game.cc

Issue #4234, opened 2020-09-07
Compiler warnings


PinPhreek:
**Describe the bug**
When I compile the newest version from git, I get compiler-warnings.

**To Reproduce**
1. update local repo
2. run the update-script

**Expected behavior**
There should be no error or warning.

**Screenshots**

**Version:**
 - OS: Arch Linux 5.4.62-1-lts
 - Widelands Version: Widelands Build `1.0~git38[07983f1@master](Debug)`

**Error-Log**
```
-- Configuring done
-- Generating done
-- Build files have been written to: /home/max/projects/widelands/build
[93/311] Building CXX object src/scripting/CMakeFiles/scripting_logic.dir/persistence.cc.o
In file included from ../src/scripting/eris.h:25,
                 from ../src/scripting/persistence.cc:26:
../src/third_party/eris/eris.h:150:13: warning: redundant redeclaration of ‘int luaopen_eris(lua_State*)’ in same scope [-Wredundant-decls]
  150 | LUA_API int luaopen_eris(lua_State* L);
      |             ^~~~~~~~~~~~
In file included from ../src/third_party/eris/lua.hpp:7,
                 from ../src/scripting/lua.h:25,
                 from ../src/scripting/persistence.h:23,
                 from ../src/scripting/persistence.cc:20:
../src/third_party/eris/lualib.h:51:17: note: previous declaration of ‘int luaopen_eris(lua_State*)’
   51 | LUAMOD_API int (luaopen_eris) (lua_State *L);
      |                 ^~~~~~~~~~~~
[113/311] Building CXX object src/ui_basic/CMakeFiles/ui_basic.dir/listselect.cc.o
In file included from /usr/include/c++/10.2.0/cassert:44,
                 from ../src/notifications/notifications_impl.h:26,
                 from ../src/notifications/notifications.h:26,
                 from ../src/graphic/note_graphic_resolution_changed.h:24,
                 from ../src/ui_basic/dropdown.h:27,
                 from ../src/ui_basic/listselect.cc:31:
../src/ui_basic/listselect.cc: In member function ‘virtual bool UI::BaseListselect::handle_key(bool, SDL_Keysym)’:
../src/ui_basic/listselect.cc:583:23: warning: suggest parentheses around comparison in operand of ‘^’ [-Wparentheses]
  583 |   assert(selected_idx <= max ^ selected_idx == no_selection_index());
      |          ~~~~~~~~~~~~~^~~~~~
[...]

PR #4377, opened 2020-10-19
Fix memset compiler warning

Niektory:
Fix one of the warnings from https://github.com/widelands/widelands/issues/4234#issuecomment-688375120:
```
[1246/1565] Building CXX object src/logic/CMakeFiles/logic_map.dir/map.cc.o
../src/logic/map.cc: In instantiation of ‘void Widelands::clear_array(std::unique_ptr<T []>*, uint32_t) [with T = Widelands::Field; uint32_t = unsigned int]’:
../src/logic/map.cc:525:44:   required from here
../src/logic/map.cc:509:8: warning: ‘void* memset(void*, int, size_t)’ clearing an object of non-trivial type ‘struct Widelands::Field’; use assignment or value-initialization instead [-Wclass-memaccess]
  509 |  memset(array->get(), 0, sizeof(T) * size);
      |  ~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from ../src/logic/map.h:30,
                 from ../src/logic/map.cc:20:
../src/logic/field.h:58:8: note: ‘struct Widelands::Field’ declared here
   58 | struct Field {
      |        ^~~~~
```
I'm looking into the other memset warning too.

### DIFF
--- a/src/logic/map.cc
+++ b/src/logic/map.cc
@@ -531,12 +531,6 @@ void Map::create_empty_map(const EditorGameBase& egbase,
 	filesystem_.reset(nullptr);
 }
 
-// Made this a separate function to reduce compiler warnings
-template <typename T = Field>
-static inline void clear_array(std::unique_ptr<T[]>* array, uint32_t size) {
-	memset(array->get(), 0, sizeof(T) * size);
-}
-
 void Map::set_origin(const Coords& new_origin) {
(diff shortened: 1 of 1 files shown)


gunchleoc (commented):
Code LGTM, not tested yet.

Noordfrees (approved):
The warning's gone :)

PR #4381, opened 2020-10-20
Draw road length tooltip in road building mode

Noordfrees:


gunchleoc:
I think this is useful for the waterways, because they have a limit. Do we really need this for land roads?

Noordfrees:
It happens to me frequently that I plan a long road, complete it with Ctrl down, and then discover that one of the resulting roads is 3 tiles long because the road length was odd. This is what I want to avoid by having a tooltip so can I see whether the road will have even or odd length.

src/wui/interactive_base.cc
```diff
@@ -750,6 +750,15 @@ void InteractiveBase::draw_overlay(RenderTarget& dst) {
 
 	Game* game = dynamic_cast<Game*>(&egbase());
 
+	if (in_road_building_mode() && tooltip().empty()) {
+		draw_tooltip(
+		   in_road_building_mode(RoadBuildingType::kRoad) ?
+		      (boost::format(_("Road length: %u")) % get_build_road_path().get_nsteps()).str() :
+		      (boost::format(_("Waterway length: %1$u (limit: %2$u)")) %
```
gunchleoc: I think we can make this one a bit more concise:

```suggestion
		      (boost::format(_("Waterway length: %1$u/%2$u")) %
```

Let's keep the road length.

gunchleoc (approved):

PR #4385, opened 2020-10-20
Table/Tabpanel Navigation

Noordfrees:
- Implement mousewheel support for TabPanel
- Fix arrow keys navigation for Table when no entry selected
- Implement PageUp, PageDown, Home, End for Table


gunchleoc:
Drive-by comment: The usage of arrow keys in tab panels steals them from moving the map. So, we need to replace that with something else - Ctrl+Tab, maybe?

Code LGTM, not tested yet.

gunchleoc (approved):

Issue #4384, opened 2020-10-20
crash with Attempt to add invalid Time or Duration


hessenfarmer:
latest commit crashed with:

Attempt to add invalid Time or Duration

setup was lokal network 4 AI game on full moon. 

Emergency save and stdout.txt attached
[stdout.txt](https://github.com/widelands/widelands/files/5410980/stdout.txt)
[2020-10-20T20.59.14.wgf.zip](https://github.com/widelands/widelands/files/5410982/2020-10-20T20.59.14.wgf.zip)

Noordfrees:
Do you have a replay? Without a backtrace it is hard to say from where in the code this comes

hessenfarmer:
[replays.zip](https://github.com/widelands/widelands/files/5411065/replays.zip)
maybe unrelated but in latest trunk savegames saved after 15 minutes or so fail to load with the following stdout.txt
[stdout.txt](https://github.com/widelands/widelands/files/5411072/stdout.txt)

Noordfrees:
Probably unrelated – `stol` is used only for saveloading in `map_io/map_players_view_packet.cc`

Noordfrees:
The problem is not reproduceable from the replay so it is probably an AI issue (or in game host code). Will try to catch this in gdb with the same setup

Noordfrees:
On the fifth try:
```
Thread 1 "widelands" hit Breakpoint 2, Time::operator+ (this=0x62b000400bc4, delta=...) at ../src/base/times.h:142
142				throw wexception("Attempt to add invalid Time or Duration");
(gdb) bt
#0  Time::operator+(Duration const&) const (this=0x62b000400bc4, delta=...) at ../src/base/times.h:142
#1  0x0000555556d69508 in AI::DefaultAI::check_building_necessity(AI::BuildingObserver&, AI::DefaultAI::PerfEvaluation, Time const&) (this=
    0x61d0001c7080, bo=..., purpose=AI::DefaultAI::PerfEvaluation::kForConstruction, gametime=...) at ../src/ai/defaultai.cc:5686
#2  0x0000555556d47ebc in AI::DefaultAI::construct_building(Time const&) (this=0x61d0001c7080, gametime=...) at ../src/ai/defaultai.cc:2607
#3  0x0000555556d2961f in AI::DefaultAI::think() (this=0x61d0001c7080) at ../src/ai/defaultai.cc:390
[...]

PR #4387, opened 2020-10-20
Fix `stol` out-of-range error on saveloading

Noordfrees:
@hessenfarmer can you test if this fixes the saveloading issue?

Re https://github.com/widelands/widelands/issues/4384#issuecomment-713087639


hessenfarmer (approved):
Yep that fixes the issue

gunchleoc (approved):

PR #4394, opened 2020-10-21
Remove unneeded NEVER_HERE in Vision

Niektory:
Apparently these cause compiler warnings: https://github.com/widelands/widelands/issues/4234#issuecomment-713853838


Noordfrees (approved):
The warning is gone :)

PR #4408, opened 2020-10-25
Reset individual training wheels

gunchleoc:
Add Options window for reactivating individual training wheels.
Lists only completed training wheels for brevity.


data/scripting/training_wheels/init.lua
```diff
@@ -16,7 +16,7 @@
 -- Their names need to be identical with their base file name (without the ``.lua`` file extension).
 -- For blocking the execution of a Training Wheel until the player has learned a prerequisite
 -- Training Wheel, list them in the Training Wheel's Lua table in the format
--- ``example = { "prerequisite1", "prerequisite1" },``.
+-- ``example = { descname = "Example", dependencies = "prerequisite1", "prerequisite1" } },``.
```
Noordfrees, line 19: No i18n markup for `descname` in the example?

Noordfrees:
Frankly this window looks a bit unpolished to me. I would prefer a list with *all* wheels, with a checkbox next to each one that indicates whether the wheel is solved or not. (This would have the advantage of giving the player a way to see how he has left to learn.) There should then be buttons "Cancel" (close window without applying any changes), "Reset" (reset the checkboxes to the initial state), "Unmark all" to untick all checkboxes, and "OK" to actually apply the changes.

Mockup:
grafik

src/ui_fsmenu/options.cc
```diff
@@ -315,8 +316,15 @@ FullscreenMenuOptions::FullscreenMenuOptions(FullscreenMenuMain& fsmm,
 	language_dropdown_.selected.connect([this]() { update_language_stats(); });
 	training_wheels_reset_.sigclicked.connect([this]() {
 		training_wheels_reset_.set_enabled(false);
-		Profile training_wheels_profile;
-		training_wheels_profile.write(kTrainingWheelsFile);
+		cancel_.set_enabled(false);
+		apply_.set_enabled(false);
+		ok_.set_enabled(false);
```
Noordfrees, line 378: No need for that. While the TrainingWheelOptions is running, the buttons are inaccessible anyway.

data/scripting/training_wheels/init.lua
gunchleoc: No, because the translation system would pick it up and put the string on Transifex.

src/ui_fsmenu/options.cc
gunchleoc: Yes they are, but this signals to the user that they are inaccessible. I find it more intuitive.

gunchleoc:
I have implemented the mockup now

training_wheels_ui

Noordfrees:
Looking much better, just a few nits UI-wise:
- The Edit Teaching Progress window is a child of the Options window, which looks bad when moving it around. It needs to be a child of the Options window's parent (the FsMenuMain) instead.
- It is off-center when created. It should be centered on the screen.
- Please give the buttons in the bottom row all the same size, they look too irregular.
- In the options window, perhaps add a bit of space above the Teach Me How To Play? The button is not visually connected to the checkbox because of the wide hgap between them.
- Also I'd suggest to disable the button when the checkbox is not ticked.

gunchleoc:
Thanks for your suggestions - all implemented :)

training_wheelsreset

Noordfrees:
UI looking good now, tested and working. Just one more nit: please stretch the checkboxes across the  whole width (like in the Options window) to make them more comfortable to click. Will look at the code soon.

gunchleoc:
> please stretch the checkboxes

Done :)

gunchleoc:
Do you still need to look over the code, or is this ready?

Noordfrees (approved):

PR #4412, opened 2020-10-26
Merge wl.World() and wl.Tribes()

gunchleoc:
Note: Renaming `tribes = wl.Tribes()` is deferred to a follow-up branch to keep the diff size manageable.

@MarkMcWire FYI


src/scripting/lua_root.h
```diff
@@ -131,38 +131,13 @@ class LuaWorld : public LuaRootModuleClass {
...
-	explicit LuaTribes(lua_State* L);
-
-	void __persist(lua_State* L) override;
-	void __unpersist(lua_State* L) override;
-
-	/*
-	 * Properties
-	 */
-
 	/*
```
hessenfarmer: shouldn't we move up this comment a few lines as the former world methods are the same thing now?

src/scripting/lua_root.cc
```diff
@@ -484,9 +498,19 @@ int LuaWorld::new_critter_type(lua_State* L) {
...
-// Documented in data/world/immovables/bush1/init.lua.
+// Documented in data/world/immovables/artifacts/artifact00/init.lua and
+// data/tribes/immovables/ashes/init.lua
 // See also the World and Tribes section in the Widelands Scripting Reference on the website.
-int LuaWorld::new_immovable_type(lua_State* L) {
+/* RST
+   .. method:: new_immovable_type{table}
+
+      Adds a new immovable type. Takes a single argument, a table with
+      the descriptions. See :ref:`lua_tribes_immovables` and :ref:`lua_world_immovables`
```
hessenfarmer: Do we really still need to document this in 2 separate places?

src/scripting/lua_root.cc
```diff
@@ -942,7 +879,7 @@ void luaopen_wlroot(lua_State* L, bool in_editor) {
 		add_parent<LuaGame, LuaBases::LuaEditorGameBase>(L);
 		lua_pop(L, 1);  // Pop the meta table
 	}
-	register_class<LuaWorld>(L, "", false);
-	register_class<LuaTribes>(L, "", false);
+	register_class<LuaDescriptions>(L, "", false);
+	register_class<LuaDescriptions>(L, "", false);
```
Noordfrees: Duplicate line

hessenfarmer (approved):
LGTM.
2 small nits inline

src/scripting/lua_root.cc
gunchleoc: I'm not married to it and I'd be fine with deleting all of those.

hessenfarmer: Ok if all means keeping one instance of documentation for all duplicate units

gunchleoc: @hessenfarmer please don't resolve until it has been implemented - it's easy to overlook otherwise.

hessenfarmer: Ok understood. However the change implemented showed that we misunderstood each other. What I meant was Why do we keep documentaion in a tribes section and in a world section. can be changed in a follow up branch though.

hessenfarmer:
Should be good to go now

PR #4418, opened 2020-10-26
Split website binaries test from testsuite

gunchleoc:


Noordfrees:
This appears to work but clicking through the build results the number of transient SDL errors seems pretty large for the web binaries, on a quick count I saw 27 out of 32 failures. There seem to be fewer failures in master, don't know why – perhaps sleeping a minute after starting Xvfb might help?

Noordfrees (approved):
Looks much better now :)

PR #4419, opened 2020-10-26
Get rid of code duplication in lua_root.cc

gunchleoc:
Wave goodbye to ca. 200 lines of code duplication :)


hessenfarmer (approved):
very nice catch 
lgtm

PR #4420, opened 2020-10-26
MSVC support

matthiakl:
I looked into compiling Widelands in Visual Studio, which apparently worked some time ago. Using vcpkg, the installation of the dependencies is quite simple, however, it takes some time.
The changes currently only "make it work". Some compiler flags should be added/removed for a better compiler output.
Tested on Windows 7 and Windows 10 (VirtualBox), feedback welcome.

Pros:
- Faster link-time
- Easier debugging @hessenfarmer 

Cons:
- Another compiler to support

---
# Building under MSVC 

## Prerequisites

- Install [python](www.python.org)
- Install [Visual Studio](https://visualstudio.microsoft.com) together with the English language pack (needed for vcpkg).
- Download and install [vcpkg](https://github.com/microsoft/vcpkg)
	+ Run `bootstrap-vcpkg.bat` first.
	+ Then run `vcpkg integrate install` to integrate vcpkg with Visual Studio.
	+ Install all dependencies  
	  `vcpkg install --triplet=x64-windows-static boost gettext libpng icu glbinding sdl2 sdl2-ttf sdl2-mixer[libvorbis,libflac,mpg123] sdl2-image[libjpeg-turbo,tiff] graphite2 harfbuzz opusfile libwebp`
	  Note the selected static triplet.
	+ If you want to save space, you can safely delete the `buildtrees` folder afterwards.
- Download the latest static release zip of [gettext](https://github.com/mlocati/gettext-iconv-windows/releases).	
	+ Extract the contained `bin` folder to `[vcpkg]/installed/x64-windows-static/bin`
- Remove the prefix `lib` from libmpg123.lib libiconv.lib libintl.lib libcharset.lib in `[vcpkg]/installed/x64-windows-static/lib` and `[vcpkg]/installed/x64-windows-static/debug/lib` (libmpg123.lib -> mpg123.lib).

## Project configuration

- Copy `utils/win32/CMakeSettings.json` to the top level directory to use with Visual Studio.
- Launch Visual Studio and select the Widelands root as workspace.
	+ CMake should automatically start parsing the project and picking up the vcpkg toolchain file.
	+ Then start compilation
- Copy `utils/win32/launch.vs.json` to the `.vs/` directory as start configuration.
	
## Common problems

- vcpkg fails to build a package (e.g. freetype)
	+ remove the package (or its dependency parent, e.g. sdl-ttf) from the install list and install it separately afterwards. vcpkg sometimes has problems resolving dependency chains.


src/network/network_lan_promotion.cc
```diff
@@ -178,7 +178,7 @@ bool LanBase::is_open() {
 	return socket_v4.is_open() || socket_v6.is_open();
 }
 
-ssize_t LanBase::receive(void* const buf, size_t const len, NetAddress* addr) {
+size_t LanBase::receive(void* const buf, size_t const len, NetAddress* addr) {
```
gunchleoc, line 181: Look for `if (receive(&info, sizeof(info), &addr) < static_cast<int32_t>(sizeof(info))) {` in function `void LanGameFinder::run()` and remove the `static_cast`, as this return type is no longer signed.

gunchleoc (commented):
I think this is a great idea :)

We can always drop support again down the road if we should have nobody to maintain it.

Can you drop the instructions fro this PR into an md file?

utils/win32/CMakeSettings.json
```diff
@@ -0,0 +1,95 @@
...
+          "type": "BOOL"
+        },
+        {
+          "name": "OPTION_BUILD_CODECHECK",
+          "value": "False",
+          "type": "BOOL"
+        },
+        {
+          "name": "OPTION_USE_GLBINDING",
+          "value": "True",
```
hessenfarmer, line 41: this setting is mutually exclusive with OPTION_BUILD_WINSTATIC as the static build works only with glew as glbinding doesn't provide a static library as far as I know at least for MSYS environment

hessenfarmer (dismissed):
not tested but one small comment

If the working build environment on MSYS is not affected we should go for this as it might be good to have more options.   
Will test this if I find time

utils/win32/CMakeSettings.json
matthiakl: This is only true for MSYS. With vcpkg glbinding works fine, but I got errors using glew.

matthiakl:
> Can you drop the instructions fro this PR into an md file?

Should the instructions really be in the tree? For msys and mac OS there are only wiki entries. So maybe that's a better place for them.

utils/win32/CMakeSettings.json
hessenfarmer: Ok but do you really get a static linked exe without the need of the glbinbding dll?

matthiakl: No glbinding dll needed. However, the debug build requires a c runtime dll, which is only installed with VS and not with the normal vc_redistributable. This should be omitted with /MTd. I will check, whether we can remove this dependency.

Edit: fixed in latest commit. Now a fully independent exe is generated.

gunchleoc:
> > Can you drop the instructions fro this PR into an md file?
> 
> Should the instructions really be in the tree? For msys and mac OS there are only wiki entries. So maybe that's a better place for them.

Wiki entry will be fine :)

gunchleoc (approved):
If @hessenfarmer is happy, this can go in. Thanks for this feature :)

Don't forget to update the wiki.

matthiakl:
> Don't forget to update the wiki.

Done: https://www.widelands.org/wiki/Building%20Widelands/#building-under-windows

gunchleoc:
Thanks!

PR #4423, opened 2020-10-27
Implement percent notation for animation sound probability

### LINKED ISSUE #4047: Refactor probability ranges

gunchleoc:
Closes https://github.com/widelands/widelands/issues/4047

I stuck the parser function into base/math. If somebody has a better idea where to put this, I'm all ears :)


gunchleoc:
@Noordfrees FYI this is the final API change and it's tiny, so you can continue with the Add-ons system any time you like now :)

The rest that I still have planned is to get rid of as many AI hints as possible, but that shouldn't hold you up.

src/base/math.h
```diff
@@ -43,6 +45,15 @@ template <typename T> T sqr(const T& a) {
 	return a * a;
 }
 
+static constexpr unsigned k100PercentAsInt = 10000U;
+
+/**
+ * @brief Reads a percentage
+ * @param input A percentage in the format 12%, 12.5% or 12.53%.
+ * @return Scaled precentage as integer, where 100% corresponds to kMaxProbability.
```
Notabilis: Ohh....

src/base/math.cc
```diff
@@ -17,4 +17,35 @@
...
+
+// This function has RST documentation in logic/map_objects/map_object_program
+unsigned read_percent_to_int(const std::string& input) {
+	std::smatch match;
+	std::regex re("^(\\d+)([.](\\d{1,2})){0,1}%$");
+	if (std::regex_search(input, match, re)) {
+		// Convert to range
+		uint64_t result =
+		   100U * std::stoul(match[1]) +
+		   (match[3].str().empty() ?
```
Notabilis: Checking for match[1] / match[3] instead of match[0] / match[2] looks a bit strange. If you know why that is so, can you please add two short comments about matches 0 and 2, which a reader of the code might expect?

doc/sphinx/source/animations.rst
```diff
@@ -36,7 +36,7 @@ and what your image files need to look like:
          fps = 4,
          sound_effect = {
             path = "sound/foo/bar",
-            priority = 50,
+            priority = "5.4%",
             allow_multiple = true, -- If this is given and true, multiple mapobjects can play the same sound
```
Notabilis: Maybe remove the comment about `allow_multiple`? It is also explained below.

src/base/math.h
```diff
@@ -43,6 +45,15 @@ template <typename T> T sqr(const T& a) {
 	return a * a;
 }
 
+static constexpr unsigned k100PercentAsInt = 10000U;
+
+/**
+ * @brief Reads a percentage
+ * @param input A percentage in the format 12%, 12.5% or 12.53%.
+ * @return Scaled precentage as integer, where 100% corresponds to k100PercentAsInt.
```
Notabilis: Ohh.... Typo `precentage`. That is unexpected, normally you are correcting my writing. ;-)

Notabilis (approved):
Code is looking good to me, not tested (not sure how/what except starting the game).
Three small comments regarding the documentation are inline.
I am fine with placing the code in base/math.

Notabilis:
I am not quite happy with your regex tweaks. Based on the documentation of the regex-function, `match[0]` is always the complete string which I could confirm by printing it. So when using `stoul[match[0])` we actually use the whole string and `stoul()` aborts parsing the "number" when it encounters `.` or `%`. With the old regex and the indices the part before the point is extracted and passed to `stoul()` without the rest of the string.
I think (but don't know) that both the new and the old variant will work fine. Personally I prefer the old variant, though.

gunchleoc:
> I am not quite happy with your regex tweaks. Based on the documentation of the regex-function, `match[0]` is always the complete string which I could confirm by printing it. So when using `stoul[match[0])` we actually use the whole string and `stoul()` aborts parsing the "number" when it encounters `.` or `%`. With the old regex and the indices the part before the point is extracted and passed to `stoul()` without the rest of the string.
> I think (but don't know) that both the new and the old variant will work fine. Personally I prefer the old variant, though.

I changed it to `match[0]` because

1. You asked me to, and
2. I think that less parentheses in the regex makes the regex itself more readable.

PR #4430, opened 2020-10-28
Set OpenGL swap interval to 0

Niektory:
This reduces graphics latency but may cause frame tearing on some systems. Please test (moving large windows, scrolling the map), both in fullscreen and not. If there are negative effects I'll make this an option.

PR #4435, opened 2020-10-28
Split minimap rendering into stages

Niektory:
Split minimap rendering into 2 stages:
1. Draw the static (independent of the view area) version of the minimap. This step is slow because the renderer iterates all pixels of the minimap and redraws them one by one, whether anything changed or not. So instead of rendering the entire minimap every frame, we now render just a part of it so it takes several frames to fully update it. Currently it is tuned so a 512x512 map takes 8 frames to fully update (smaller maps update faster).
2. Center the texture from Step 1 on the view area and draw the view window. This step is fast and is done every frame, so panning the map around is still smooth.

The result is lowered CPU use when drawing large minimaps, at the cost of slightly slower updates.
Note that due to some OpenGL swap interval/buffer synchronization voodoo the system might still report 100% CPU use – #4430 should address this.

Additional changes:
- Rendering zoomed in minimaps was slightly sped up by only calculating the color once per field instead of once per pixel.
- Immovable type checking was slightly sped up.
- Smoother minimap panning when zoomed in – it no longer snaps to the closest field edge. This was an accidental change but I'm not complaining. ^^


src/graphic/minimap_renderer.cc
```diff
@@ -67,16 +66,15 @@ inline RGBColor calc_minimap_color(const Widelands::EditorGameBase& egbase,
...
-			}
-
-			if (((layers & MiniMapLayer::Flag) && dynamic_cast<Widelands::Flag const*>(immovable)) ||
-			    ((layers & MiniMapLayer::Building) &&
-			     dynamic_cast<Widelands::Building const*>(immovable))) {
+		if (layers & (MiniMapLayer::Road | MiniMapLayer::Flag | MiniMapLayer::Building) &&
+		    f.field->get_immovable()) {
+			const Widelands::MapObjectType type = f.field->get_immovable()->descr().type();
+			if ((layers & MiniMapLayer::Flag && type == Widelands::MapObjectType::FLAG) ||
+			    (layers & MiniMapLayer::Building && type >= Widelands::MapObjectType::BUILDING)) {
```
Notabilis, line 73: Maybe check for the upper bound (<= TRAININGSITE) here as well?
Works fine for now, but might create bugs in a possible future.

src/graphic/minimap_renderer.cc
```diff
@@ -194,16 +207,20 @@ void do_draw_minimap(Texture* texture,
...
 				const auto& field = player->fields()[i];
 				vision = field.vision;
 				owner = field.owner;
 			}
 
 			if (vision != Widelands::VisibleState::kUnexplored) {
-				texture->set_pixel(x, y,
-				                   calc_minimap_color(egbase, f, layers, owner,
-				                                      vision == Widelands::VisibleState::kVisible));
+				RGBAColor color = calc_minimap_color(
```
Notabilis: Could probably be made `const RGBAColor`.

src/graphic/minimap_renderer.cc
```diff
@@ -241,32 +258,89 @@ std::unique_ptr<Texture> draw_minimap(const Widelands::EditorGameBase& egbase,
...
 
 	// Center the view on the middle of the 'view_area'.
-	const bool zoom = layers & MiniMapLayer::Zoom2;
-	Vector2f top_left =
-	   minimap_pixel_to_mappixel(map, Vector2i::zero(), view_area, minimap_type, zoom);
-	const Widelands::Coords node =
-	   MapviewPixelFunctions::calc_node_and_triangle(map, top_left.x, top_left.y).node;
+	switch (minimap_type) {
+	case MiniMapType::kStaticViewWindow: {
+		uint16_t move_x =
```
Notabilis: const?

src/graphic/minimap_renderer.cc
```diff
@@ -162,25 +160,40 @@ void draw_view_window(const Widelands::Map& map,
...
 // Does the actual work of drawing the minimap.
-void do_draw_minimap(Texture* texture,
+void do_draw_minimap(Texture& texture,
                      const Widelands::EditorGameBase& egbase,
                      const Widelands::Player* player,
                      const Vector2i& top_left,
-                     MiniMapLayer layers) {
+                     MiniMapLayer layers,
+                     const bool draw_full = true,
+                     uint16_t* rows_drawn = nullptr) {
```
Notabilis: You *could* interpret `rows_drawn == nullptr` as order to draw_full. But its more expressive the current way.

src/graphic/minimap_renderer.h
```diff
@@ -73,12 +73,38 @@ Vector2f minimap_pixel_to_mappixel(const Widelands::Map& map,
...
+// given texture. Can update the entire texture or just a part of it,
+// depending on 'draw_full'.
+// If 'draw_full' is false, 'rows_drawn' has to be passed to keep track of
+// the row number where we stopped drawing.
+void draw_minimap_static(Texture& texture,
+                         const Widelands::EditorGameBase& egbase,
+                         const Widelands::Player* player,
+                         MiniMapLayer layers,
+                         const bool draw_full = true,
+                         uint16_t* rows_drawn = nullptr);
```
Notabilis, line 100: Maybe add a comment that `*rows_drawn` should be initialized as 0?
You are catching too large values, so its no bug, only unclear/undefined documentation.

src/wui/minimap.h
```diff
@@ -88,13 +88,18 @@ class MiniMap : public UI::UniqueWindow {
...
+
 		// This needs to be owned since it will be rendered by the RenderQueue
 		// later, so it must be valid for the whole frame.
-		std::unique_ptr<Texture> minimap_image_;
+		std::unique_ptr<Texture> minimap_image_final_;
 
 	public:
 		MiniMapLayer* minimap_layers_;
 		MiniMapType* minimap_type_;
+		bool started_drawing_;
```
Notabilis: Could be replaced by checking whether `minimap_image_final` is a valid pointer.

Notabilis (approved):
Code is looking good, some small suggestions are in the comments. But nothing which delays merging as far as I am concerned.

Testing hasn't shown any errors. The minimap no longer feels laggy on large maps, which is quite nice.

src/graphic/minimap_renderer.cc
Niektory: Other parts of the codebase also don't check for the upper bound. I think a better way to make it future-proof would be doing all the comparisons in helper functions in `MapObjectType`. But it's a job for another branch.

src/graphic/minimap_renderer.cc
Niektory: So is it good or bad? I'm getting mixed signals here :-P

src/graphic/minimap_renderer.h
Niektory: The initial value doesn't really matter. I'll try to clarify it.

src/graphic/minimap_renderer.cc
Notabilis: Pick whichever you prefer. ;-)
It would be a small, unneeded optimization(?) to remove `draw_full`. But then the caller has to read the documentation to figure out what `nullptr` means. So ... do what you want, both choices are fine. (Personally, I would remove `draw_full`. But its your code and its up to you.)

PR #4436, opened 2020-10-28
Tweak low productivity threshold in building statistics

Niektory:
- ~~Remove the space between "Low productivity" and "Low productivity threshold". I think they shouldn't be separated because they're closely related.~~
- ~~Show "Low productivity threshold" only when "Low productivity" is shown.~~
- Change the capitalization from "Low Productivity" to "Low productivity" for consistency.
- Center the "Low productivity threshold" label and spinbox.
- Make the spacing more consistent by removing padding between invisible elements in a Box.
- Minor Spinbox padding tweaks.

~~There is still some dead space when no building is selected. It seems that elements take up some space even when hidden.~~

**Before (left) and after (right):**
shot0126shot0123
shot0128shot0125
shot0127shot0124


gunchleoc (commented):
Code & Screenshots LGTM, not tested.

Noordfrees:
I deliberately left the spinbox always visible because this is a global setting. It affects not the selected building type but all building icons. I think it should stay visible.
The extra space is because spaces added by `add_(inf)_space()` as well as the box's inner padding (not sure about that) cannot be hidden. We'd need to replace them with spacer panels for that.

Niektory:
> I deliberately left the spinbox always visible because this is a global setting. It affects not the selected building type but all building icons. I think it should stay visible.

Oh, I see, it changes the colors of text below the building icons. I thought it only affected the "Low productivity" counter. So, back to being always visible:

shot0129shot0130shot0131

frankystone:
It looks a bit cluttered, because the last row isn't visually connected to any other row. Can there be a small line above the last two rows and center both the string 'Low productivity threshold' and the spinners?

gunchleoc:
How about left-aligning the spinbox?

Niektory:
I restored the space, but made it more consistent – invisible elements no longer add padding. A line would likely look nicer, but I don't know at the moment how to add one (or how to change the spinbox's alignment).

gunchleoc:
We can't do separator lines in the UI yet, so not having a line is fine.

The horizontal alignment of the spinbox is still neither fish now fowl. It needs to be aligned to something.

Niektory:
left-aligned spinbox center-aligned spinbox

Which one? I think the centered version looks better.

frankystone:
I didn't used this setting so far but afaik this setting affects all buildings. The former three left aligned rows affects the marked building, the last rows affects all buildings, so there should be be a visual distinction imho. I would vote for the right one (centered).

Noordfrees:
+1 for centered

gunchleoc (approved):

Niektory:
Noticed some minor problems with the spinbox tweaks, don't merge yet.

Niektory:
Should be fine now.

PR #4439, opened 2020-10-29
Optimize drawing overlapping workareas

Niektory:
Just speeding things up.


gunchleoc (approved):
More performance is always good :)

PR #4451, opened 2020-10-31
Clear readability-redundant-string-cstr

gunchleoc:
Re https://github.com/widelands/widelands/issues/3223


Notabilis (approved):
Thanks for cleaning up!

PR #4462, opened 2020-11-02
Send localized cheating messages

Noordfrees:
When one player plays in german and another in chinese they will probably appreciate seeing these messages in english as well


gunchleoc:
How about sending it in English and translating on the client side?

Noordfrees:
Do we already have any chat messages that behave like this? If not we'd have to implement this into the protocol to send the info that the message needs to be translated along with the message. Bit overkill IMHO, not to mention I'd have no idea where to program this.

gunchleoc:
Take a look at `InternetGamingMessages` in networking.

Noordfrees:
Done :)

gunchleoc:
Error on server:

`[00:00:59.761 real] ERROR: GenericSaveHandler::save_file: data could not be written to file replays/2020-11-15T13.14.56_nethost.wrpl.wgf: basic_string::substr: __pos (which is 8) > this->size() (which is 2)
`

On client:

`[00:00:32.870 real] ERROR: GenericSaveHandler::save_file: data could not be written to file replays/2020-11-15T13.14.56_netclient_1.wrpl.wgf: basic_string::substr: __pos (which is 13) > this->size() (which is 6)`

Noordfrees:
I cannot reproduce this. What did you do exactly?

gunchleoc:
1. Compile debug build
2. create LAN game from map
3. run second Widelands instance from same directory
4. change language
5. join game on "localhost" start game

Noordfrees:
No luck reproducing, and the resulting replays also load and run fine. Did you check that this does not happen for you on other branches?

frankystone:
Forgot it, this was on master... :-D

Niektory:
I also can't reproduce.

matthiakl:
Is there an easy way to test the translation? I added the string to po/widelands/de.po and rebuild the locale dir, but it only shows the English version.

Noordfrees:
You need to add it to `po/widelands/widelands.pot` as well, then it should work

matthiakl (approved):
> You need to add it to `po/widelands/widelands.pot` as well, then it should work

It does :). No complaints here.

Noordfrees:
Thanks for the review :)

Issue #4271, opened 2020-09-15
enhanced productionsite help not working properly


hessenfarmer:
the new collects part of the productionsite help is indicating wrong items to be collected from the map.

reason is:
the information is wrongly determined from the immovables findobject program, rather then from their modificating programs like (remove) 

best visible in the amazons branch for the wilderness Keeper's tent where trees are listed to be collected which sre only searched for to release a game there, but remain untouched.

gunchleoc:
This will need more elaborate parsing, because the "remove" program doesn't name the object. So, first record the information from the findobject program, then confirm it with the remove program.

We'll only be able to test this fully with the amazons branch, but it can be implemented on master.

gunchleoc:
Actually, the trees should still be listed, because the building needs to be built close to them. I think we should keep it simple for now, but maybe replace the "Collects:" header with something else. The header should also fit the Frisian Beekeeper.

PR #4013, opened 2020-07-23
WIP: Collection help, Workarea overlaps + AI tweaks

gunchleoc:
Proposing this now for the formatting bot, so I can split off the workarea overlap code.

Help is mostly done too, AI is unfinished and regression tests are missing.

Re. https://github.com/widelands/widelands/issues/3989

PR #4465, opened 2020-11-02
Get rid of production support AI hints and add support for multiple rangers/lumberjacks

gunchleoc:
- Deprecate `collects_ware_from_map` AI hint
- Deprecate `supports_production_of` AI hint
- Add support for multiple lumberjack/ranger pairs
- Get rid of dirty hack for rare tree cutter/plantation. `amazons_rare_tree_plantation` no longer needs supporters, but the `rare_tree_cutters_hut` still does
- Fix miscategorization of Frisian Fishers being supported by clay pits
- `frisians_reed_farm` is now detected as supporting the `frisians_beekeepers_house`
- Hotfix for Amazon trainingsites
- New property `needed_attributes` for productionsites when the immovable doesn't get removed from the map
- Get rid of hard-coded work area radius when detecting useful immovables near a building

- If a building has no input but some output wares and creates no immovables, but collects some:
   - If it produces a construction material
     - If it has supporters -> lumberjack
     - If it has no supporters -> quarry
  - If it produces ware that is not a construction material -> needs berry
- A hunter collects and bob from the map. This adds `amazons_hunter_gatherers_hut` to the list
- If needs water and collects "resource_fish" -> fisher
- If a building has no input and it produces "water" -> well
- If a building supports a lumberjack, it's a ranger

Re. https://github.com/widelands/widelands/issues/3989
Re. https://github.com/widelands/widelands/issues/4271

Supersedes https://github.com/widelands/widelands/pull/3999
Supersedes https://github.com/widelands/widelands/pull/4013

@tibor95 @hessenfarmer


src/ai/defaultai.cc
```diff
@@ -841,29 +853,140 @@ void DefaultAI::late_initialization() {
...
-			}
-			// quarries
-			if (bh.collects_ware_from_map() == "granite") {
-				bo.set_is(BuildingAttribute::kNeedsRocks);
+			// Some important buildings are identified
+
+			// Woodcutters/Lumberjacks, Quarries and Collectors are a buildings that collect an
+			// immovable attribute and create no immovables themselves
+			Widelands::MapObjectDescr::AttributeIndex tree_attribute =
+			   Widelands::MapObjectDescr::get_attribute_id("tree");
```
hessenfarmer: Here we hardcode something into ccc that is free definable in lua.

src/ai/defaultai.cc
```diff
@@ -841,29 +853,140 @@ void DefaultAI::late_initialization() {
...
+			if (prod.created_immovables().empty()) {
+				Widelands::MapObjectDescr::AttributeIndex rocks_attribute =
+				   Widelands::MapObjectDescr::get_attribute_id("rocks");
```
hessenfarmer: same thing hardcoded values that shouldn't be hardcoded

src/ai/defaultai.cc
```diff
@@ -748,8 +735,33 @@ void DefaultAI::late_initialization() {
...
+				   gametime, "AI %d detected supporting producer: %s", player_number(), bo.name);
+				for (const auto& supp : prod.supported_productionsites()) {
+					log_dbg_time(gametime, "  -> %s", supp.c_str());
+				}
 				bo.set_is(BuildingAttribute::kSupportingProducer);
+				/* Buildings detected at the time of writing:
+				 *
+				 *   frisians_clay_pit  -> frisians_aqua_farm, frisians_charcoal_burners_house
+				 *   frisians_farm      -> frisians_beekeepers_house
+				 *   frisians_reed_farm -> frisians_beekeepers_house
```
hessenfarmer: frisians berry farm is missing

src/ai/defaultai.cc
```diff
@@ -6237,13 +6367,19 @@ void DefaultAI::consider_productionsite_influence(BuildableField& field,
...
-		++field.collecting_producers_nearby.at(bo.get_collected_map_resource());
+	for (const auto& supported_building : bo.supported_producers) {
+		// Using std::map syntax to create entry on demand.
+		++field.supported_producers_nearby[supported_building.first];
 	}
 
-	if (!bo.production_hints.empty()) {
-		for (auto ph : bo.production_hints) {
-			++field.supporters_nearby.at(ph);
+	if (bo.type == BuildingObserver::Type::kProductionsite) {
```
hessenfarmer: this should have been already a prerequisite for being here.

hessenfarmer (changes requested):
I believe this is going in the wrong direction as we now hardcode lua attribute into AI code without mentioning this in the documentation. the former Ai hints could be explained to work for a lot of scripters probably. this won't work without deep knowledge of the Ai code anymore. Furthermore it might screw up things in amzons which I haven't checked yet. 
So I clearly vote for having amazons first and afterwards doing an attempt on this.  
Personally I could live with the additojnal work to specify these things in lua and maintain the freedom to desing things. 
If we may detect depenencies without any hardcoding this would be a different thing though.

gunchleoc:
> I believe this is going in the wrong direction as we now hardcode lua attribute into AI code without mentioning this in the documentation. the former Ai hints could be explained to work for a lot of scripters probably. this won't work without deep knowledge of the Ai code anymore. Furthermore it might screw up things in amzons which I haven't checked yet.
> So I clearly vote for having amazons first and afterwards doing an attempt on this.
> Personally I could live with the additojnal work to specify these things in lua and maintain the freedom to desing things.
> If we may detect depenencies without any hardcoding this would be a different thing though.

This branch does not have more hard-coded unit names than master. E.g. we used to have "log" hard-coded, now we have "tree" hard-coded.

I definitely want to continue working on this and to get rid of all hard-coding, but that would make a pretty big branch if I do it all now. If you prefer a big branch and for me to do it all at once, I could certainly do that. I think reviewing that one item at a time would make things easier though.

src/ai/defaultai.cc
gunchleoc: Should but isn't. I'll move the check to the calling code:
```

if (player_immovable->descr().type() == Widelands::MapObjectType::PRODUCTIONSITE) {
  BuildingObserver& bo = get_building_observer(player_immovable->descr().name().c_str());
  consider_productionsite_influence(field, imm_found.coords, bo);
} else if (upcast(Widelands::ConstructionSite const, constructionsite, player_immovable)) {
  const Widelands::BuildingDescr& target_descr = constructionsite->building();
  // Check goes here
  BuildingObserver& bo = get_building_observer(target_descr.name().c_str());
  consider_productionsite_influence(field, imm_found.coords, bo);
}
```

src/ai/defaultai.cc
gunchleoc: If it is, this is already a bug in master. Here is a log from master for the Frisians:

```
[00:00:26.749 real] DEBUG: NOCOM supporting producer: frisians_clay_pit
[00:00:26.749 real] DEBUG: NOCOM supporting producer: frisians_farm
```

The reed farm is added in this branch, which is a change but correct, because it does support the beekeeper.

hessenfarmer: i was wrong as berry farm is only a supporter and not a supporting producer

src/ai/ai_hints.h
```diff
@@ -114,17 +102,15 @@ struct BuildingHints {
...
 	const uint32_t basic_amount_;
 	const int32_t forced_after_;
 	const int8_t mines_percent_;
 	const int16_t very_weak_ai_limit_;
 	const int16_t weak_ai_limit_;
 	const int16_t normal_ai_limit_;
-	const bool requires_supporters_;
 	int trainingsites_max_percent_;
-	std::set<std::string> supported_production_;
+	std::set<Widelands::DescriptionIndex> supports_buildings_;
```
hessenfarmer: couldn't find any usage of this

hessenfarmer (commented):
Well, the difference of the hardcoded values between master and now is that one can not influence the detection wihtin the design of the building anymore. In master we hardcoded the detection of Building hints while now we don't have this psossibiltiy anymore.
As this will have great influence on amazons I'll vote for having them first. and work on that afterwards. 
There is defintily a high chance that they are incompatible with this branch in their current design.

gunchleoc:
The amazons definitely need more tlc, so I have to make this branch bigger to address things.

hessenfarmer:
Tested this branch yesterday evening. Found a weird behaviour that is already in trunk and not related to this branch but maybe to the Issue introduced by the typesafety of time change I need to fix this to get a proper reference for the behaviour of these changes.   
However one noticeable thing was that there is a difference in tribes building foresters. Where they are build latest for amazons which need them most.   
Will keep trying to analyse and afterwards might do some review / propose changes.

hessenfarmer:
Just a short explanation as it seems you might not have understood the dirty hack for the rare tree buildings in amazons:
the rare tree cutter is an upgraded building. So if you want to have a working chain you need to place a normal woodcutter next to rare tree plantation and afterwards upgrade it. The Ai can't do this as it does not know about this relation. Therefore i needed the hack to have the plantation require the rare tree cutter as a supporter by having the dummy program in the plantation and the supports hint in the rare tree cutters. 
this achieved the following sequence the AI upgrades the first wood cutter after a while and after an experienced woodcuter is availabe to rare trees cutter. Now the plantation can be built in the vicinity as it requires a rare tree cutter as a supporter. Other wise we can't guarantee they are in close vicinity. 
Hope this helps in the efforts to make this work. Maybe we need to redesign the AI detections completely and take more dependencies into account for the upgrade process as well.

gunchleoc:
I am finished with my changes now. I have annotated the categorization changes with NOCOM in the code.

The rare tree cutters are fine, because they are lumberjacks now. The rare tree plantations can arrive a little bit late, but not before the rare tree cutter gets dismantled. I guess the reasons for this can be lack of lianas. The cutter requires the plantation as supporter. If we keep both as requiring each other as supporter, the AI won't build them correctly at all.

BTW this also fixes a bug where the Frisian fishers thought that they were being supported by the clay pit.

hessenfarmer:
> I am finished with my changes now. I have annotated the categorization changes with NOCOM in the code.
> 
> The rare tree cutters are fine. The rare tree plantations can arrive a little bit late, but not before the rare tree cutter gets dismantled. I guess the reasons for this can be lack of lianas. The cutter requires the plantation as supporter. If we keep both as requiring each other as supporter, the AI won't build them correctly at all.        

The problem is you fixed the wrong site of this hack. The supporter attribute is only evaluated if a new building is to be scored against a specific buildable field. the rare tree cutter hut however is an upgrade and by this never placed directly on the map and therefore not evaluated in this sense. upgrades are only dependent on overall productivity of the building type. So the plantation is the only building that gets evaluated for needing a supporter or not. As now it doesn't need a supporter it doesn't get a better score if a cutters hut is near. (maybe I have overread a change to this though, in this case please direct me to the line changed)     
 
Furthermore the identification of the rare tree plantation as a ranger is dangerous as well as, cause in this case it is subject to the wood policy decisions of the AI which it should not be.  If there is plenty of wood available all rangers get stopped or even dismanteled according to this policy.

I am sorry but if we keep these identifcations semantically correct the AI can't handle them without major rework. That is why I was in favor of keeping control in the tribes definitions. But I will have a look how I can get these changes to work again with amazons and all other possibly affected buildings.    
> 
> BTW this also fixes a bug where the Frisian fishers thought that they were being supported by the clay pit.

gunchleoc:
Can you point me to the wood policy decision in the AI code, so that I can take a look? The amazons expanded really well in my tests, so it's hard for me to see where a problem might be. I am not familiar with the scoring code, and it's a lot of lines.

Keep in mind that the new code knows that the rare trees are not the same as normal trees. So, when immovables are counted for scoring a ranger/lumberjack, it will count only the tree attributes associated with the particular building type.

hessenfarmer:
you may start here for wood policy https://github.com/gunchleoc/widelands/blob/d7f787a498f7657afe5f3ed37228c22378ee171f/src/ai/defaultai.cc#L2699

however other problems are bigger according to my tests. Liana cutters are not built early anymore but only if the first woodcutter has been upgraded. this might lead to a severe deadlock with ropes.  
the link between rare tree cutter and plantation is broken as expected so it is pure chance if they get placed in the vicinity of each other.

gunchleoc:
Thanks for pointing me to the wood policy. I expect we should have the same type of policy for the rare trees.

I'll have to look into the liana cutters.

> the link between rare tree cutter and plantation is broken as expected

Why are you expecting that this link is broken? It works exactly the same as with in master lumberjack/ranger, it's 100% the same code. If it's pure chance they are in each other's vicinity, it's pure chance in master that rangers and lumberjacks are in each other's vicinity.

hessenfarmer:
> 
> 
> Thanks for pointing me to the wood policy. I expect we should have the same type of policy for the rare trees.      
     
No we shouldn't this is already handled in the lua definitions of the plantation.    
   
> 
> I'll have to look into the liana cutters.
> 
> > the link between rare tree cutter and plantation is broken as expected
> 
> Why are you expecting that this link is broken? It works exactly the same as with in master lumberjack/ranger, it's 100% the same code. If it's pure chance they are in each other's vicinity, it's pure chance in master that rangers and lumberjacks are in each other's vicinity.     
    
As already explained in my last three comments the difference is tht the rare tree cutter can't be directly placed on the map and so their need for supporters isn't evaluated. Please read again the explanations I gave as I can't explain it better. For an upgraded building the need for a supporter isn't evaluated and can't be without a lot of effort. So the only chance is to have the plantation built near a cutter rather then the other way round. May be it was an error to define something that looked like this for the rare tree cutter in the beginning.

src/logic/map_objects/tribes/worker_program.cc
```diff
@@ -660,6 +660,12 @@ void WorkerProgram::parse_callobject(Worker::Action* act, const std::vector<std:
 
 	act->function = &Worker::run_callobject;
 	act->sparam1 = cmd[0];
+
+	// TODO(Gunchleoc): We might need to dig into the called object's program too, but this is good
+	// enough for now.
+	if (!needed_attributes_.empty()) {
```
hessenfarmer, line 666: this would deliver a false positive for frisian pond, which is technically not collected but emptied or filled. however from the mechanics needed it doesn't matter

src/ai/defaultai.cc
```diff
@@ -748,8 +740,41 @@ void DefaultAI::late_initialization() {
...
+			if (!bo.ware_outputs.empty() && !prod.supported_productionsites().empty()) {
+				log_dbg_time(
+				   gametime, "AI %d detected supporting producer: %s", player_number(), bo.name);
+				for (const auto& supp : prod.supported_productionsites()) {
+					log_dbg_time(gametime, "  -> %s", supp.c_str());
+				}
 				bo.set_is(BuildingAttribute::kSupportingProducer);
+				/* Buildings detected at the time of writing:
+				 *
+				 *   amazons_rare_tree_plantation -> amazons_liana_cutters_hut
```
hessenfarmer, line 752: this is a false positve as it is not really supporting anything. This is due to the dummy program inserted there to maintain the dirty hack for the link to the rare tree cutter.
In reality it is a pure supporter

data/tribes/buildings/productionsites/amazons/liana_cutters_hut/init.lua
```diff
@@ -26,7 +26,6 @@ descriptions:new_productionsite_type {
       very_weak_ai_limit = 1,
       weak_ai_limit = 2,
       basic_amount = 2,
-      collects_ware_from_map = "liana",
       requires_supporters = true,
```
hessenfarmer, line 29: this could probably be deleted as well as with amazons we can expect trees to cut liana from nearly to be everywhere.   
First tests with this showed good results.

hessenfarmer:
in this branch junglepreservers are very late, even if declared for basic economy. this might lead to a deadlock on maps with not many trees around the starting positions .   
Needs to be analysed

hessenfarmer:
The review can be delayed as my dev machine broke down yesterday without any preindication :-( So I can't do any testing, although I have ideas how to solve it.
in my latest tests I made one rare tree plantation a basic building and a supporter for normal lumberjack by changing the dummy program to callworker=plant. This should ensure at least on plantation in the vicinity of the first 2 lumberjacks if the delay mentioned in my last comment is solved. I also made one rare tree cutter a basic building with some changes in the code of upgrading which allows upgrading a building before the 10 min lim given there. Changes need to be made here (https://github.com/gunchleoc/widelands/blob/d7f787a498f7657afe5f3ed37228c22378ee171f/src/ai/defaultai.cc#L4567) and here (https://github.com/gunchleoc/widelands/blob/d7f787a498f7657afe5f3ed37228c22378ee171f/src/ai/defaultai.cc#L4621). 
If we can get a check in there which checks for supporters_nearby and only upgrade if there are at least one present this might work.   
Additional changes liana cutters can safelky loose their requires_supporters hint. 1 Jungle preserver should be a basic building

src/ai/defaultai.cc
```diff
@@ -1628,39 +1781,39 @@ void DefaultAI::update_buildable_field(BuildableField& field) {
...
+						continue;
+					}
+				}
+				uint8_t amount = map.find_immovables(
+				   game(),
+				   Widelands::Area<Widelands::FCoords>(map.get_fcoords(field.coords), kProductionArea),
+				   nullptr, Widelands::FindImmovableAttribute(attribute_category.first));
+
+				// adding 2 if rocks found
+				if (attribute_info.building_attribute == BuildingAttribute::kNeedsRocks) {
```
hessenfarmer, line 1803: I beleive this line is always adding 2 regardless whether there has been found some rocks in the vicinity this should be dependent on amount found as it was in previous code.

src/ai/defaultai.cc
```diff
@@ -2803,7 +2994,7 @@ bool DefaultAI::construct_building(const Time& gametime) {
 					prio += bo.primary_priority;
 
 					// keep wells more distant
-					if (bf->collecting_producers_nearby.at(bo.get_collected_map_resource()) > 2) {
+					if (number_of_supported_producers_nearby > 2) {
```
hessenfarmer, line 2997: the former code identified other wells while the new code doesn't.
This doesn't matter currently as water is pretty much indefinite but it is not correct.

src/ai/defaultai.cc
```diff
@@ -2865,9 +3057,8 @@ bool DefaultAI::construct_building(const Time& gametime) {
 						prio *= 2;
 					}
 
-					// to prevent to many quaries on one spot
-					prio =
-					   prio - 50 * bf->collecting_producers_nearby.at(bo.get_collected_map_resource());
+					// to prevent too many quarries on one spot
+					prio = prio - 50 * number_of_supported_producers_nearby;
```
hessenfarmer, line 3061: again this does not identify other quarries I believe

src/ai/defaultai.cc
```diff
@@ -2823,31 +3014,32 @@ bool DefaultAI::construct_building(const Time& gametime) {
...
 					prio += std::abs(management_data.get_military_number_at(26)) *
-					        (bf->trees_nearby - trees_nearby_treshold_) / 10;
+					        (bf->immovables_by_name_nearby[bo.name] - trees_nearby_treshold_) / 10;
 
 					// consider cutters and rangers nearby
-					prio += 2 * bf->supporters_nearby.at(bo.get_collected_map_resource()) *
+					prio += 2 * number_of_supporters_nearby *
 					        std::abs(management_data.get_military_number_at(25));
-					prio -= bf->collecting_producers_nearby.at(bo.get_collected_map_resource()) *
+					prio -= number_of_supported_producers_nearby *
```
hessenfarmer, line 3028: as far as I understand the code above this is calculated from the current BuildingObserver which means it is zero for lumberjacks and not identifiying other lumberjacks as it should

src/ai/defaultai.cc
```diff
@@ -2779,14 +2962,22 @@ bool DefaultAI::construct_building(const Time& gametime) {
...
-					}
-					if (supporters_nearby == 0) {
+					if (bf->supporters_nearby.count(bo.name) == 0) {
 						continue;
 					}
+					number_of_supporters_nearby += bf->supporters_nearby.at(bo.desc->name());
+				}
+
+				// Priorities will be adjusted according to nearby buildings needing support
+				uint8_t number_of_supported_producers_nearby = 0;
```
hessenfarmer, line 2974: As this is replacing the former detection of the vicinity to buildings of the own type I believe this algorithm or the usage of this variable is not correct. 
Edit: the usage is sometimes wrong see dedicated comments.

src/ai/defaultai.cc
```diff
@@ -2882,10 +3073,9 @@ bool DefaultAI::construct_building(const Time& gametime) {
 					// Overdue priority here
 					prio += bo.primary_priority;
 
-					prio += bf->supporters_nearby.at(bo.get_collected_map_resource()) * 5;
+					prio += number_of_supporters_nearby * 5;
 
-					prio += (bf->critters_nearby * 3) - 8 -
-					        5 * bf->collecting_producers_nearby.at(bo.get_collected_map_resource());
+					prio += (bf->critters_nearby * 3) - 8 - 5 * number_of_supported_producers_nearby;
```
hessenfarmer, line 3078: another wrong usage of number_of_supported_producers_nearby here we should have identified other hunters

src/ai/defaultai.cc
```diff
@@ -2899,17 +3089,16 @@ bool DefaultAI::construct_building(const Time& gametime) {
 
 					// Overdue priority here
 					prio += bo.primary_priority;
-
-					prio -= bf->collecting_producers_nearby.at(bo.get_collected_map_resource()) * 20;
-					prio += bf->supporters_nearby.at(bo.get_collected_map_resource()) * 20;
+					prio -= number_of_supported_producers_nearby * 20;
```
hessenfarmer, line 3092: same issue : this line does not penalize other fishers anymore

src/ai/defaultai.cc
```diff
@@ -3120,15 +3300,15 @@ bool DefaultAI::construct_building(const Time& gametime) {
...
+							prio += 1 - bf->immovables_by_name_nearby[bo.name] / 4;
 							// we attempt to cluster space consumers together
 							prio += bf->space_consumers_nearby * 2;
 							// and be far from rangers
 							prio += 1 - bf->rangers_nearby *
 							               std::abs(management_data.get_military_number_at(102)) / 5;
 						} else {
 							// leave some free space between them
-							prio -= bf->collecting_producers_nearby.at(bo.get_collected_map_resource()) *
+							prio -= number_of_supported_producers_nearby *
```
hessenfarmer, line 3311: here again this does not identify the same building type anymore

src/ai/defaultai.cc
```diff
@@ -2211,6 +2368,32 @@ void DefaultAI::update_productionsite_stats() {
 	}
 }
 
+unsigned DefaultAI::find_immovables_nearby(
```
hessenfarmer, line 2371: this funtion should be usable for detection of supportinmg buildings nearby in the upgrade building algorithm section.

gunchleoc:
>  So the only chance is to have the plantation built near a cutter rather then the other way round. May be it was an error to define something that looked like this for the rare tree cutter in the beginning.

Exactly, the plantation is placed near the cutter instead of placing the cutter near the plantation. This means that they are still near, and not by chance. Does it matter which way around it is?

I'll need to take a look at your suggestions in detail. I was wondering whether liana cutter - jungle preserver should also be a lumberjack/ranger pair and whether this will help with the lack of jungle preservers.

hessenfarmer:
> 
> 
> > So the only chance is to have the plantation built near a cutter rather then the other way round. May be it was an error to define something that looked like this for the rare tree cutter in the beginning.
> 
> Exactly, the plantation is placed near the cutter instead of placing the cutter near the plantation. This means that they are still near, and not by chance. Does it matter which way around it is?

in this case it does matter as the rare cutters can't be placed free, as they are only upgraded. However in the past the requires supporter flaf was only taken into account by the supported buiding not the supporters. So we needed to inverse this for the rare tree plantation which I did with making it produce ironwood in a dummy program and let the cutter support this production. 
Now I do need to reevaluate how this is done for the woodcutter forester combination. I am already in the process of fixing things but I am still not happy eith the result.     

> 
> I'll need to take a look at your suggestions in detail. I was wondering whether liana cutter - jungle preserver should also be a lumberjack/ranger pair and whether this will help with the lack of jungle preservers.     
   
In later games they are not lacking there are rather plenty of them. only problem left with the jungle preservers they are built too late which might lead to a deadlock on maps without any trees. So if a jungle preserver is defined a basic building he needs to be built very early which isn't the case in this branch's current state.

hessenfarmer:
perhaps you might understsand better my arguments if you try to watch the Ai rather then looking at the graphs only.  Currently I use stone rivers map for this. I take a locla lan game with me as spectator and 5 AI (each tribe) and watch what they are building and when.  
For evaluating the no tree scenario an empire strikes back might be a good map.

hessenfarmer:
made a quick test having them be just rangers for the rare tree cutters. It does not work as intended as the Ai builds plantations far away from the rare tree cutters. 
Having them defined as rangers is problematic I guess as the complete code referring to the kRanger is log centric and was not written to support multiple rangers.

src/ai/defaultai.cc
hessenfarmer: forget this comment Ifound a differnet solution

src/ai/defaultai.h
```diff
@@ -375,10 +380,16 @@ struct DefaultAI : ComputerPlayer {
...
+	// Record with immovable attribute is created or collected by which building names and attributes
+	std::map<Widelands::MapObjectDescr::AttributeIndex, std::set<ImmovableAttribute>>
+	   buildings_immovable_attributes_;
+
 	// this is a bunch of patterns that have to identify weapons and armors for input queues of
 	// trainingsites
-	std::vector<std::string> const armors_and_weapons = {
-	   "ax", "armor", "helm", "lance", "trident", "tabard", "shield", "mask", "spear"};
+	// TODO(GunChleoc): Get rid of this hard-coding
+	std::vector<std::string> const armors_and_weapons = {"ax",     "armor",  "boots",   "helm",
```
hessenfarmer, line 390: until we get rid of this we should incorporate the keywords sword and garment to deal with frisians Keyword lance can be deleted though

src/ai/defaultai.cc
```diff
@@ -923,47 +1058,65 @@ void DefaultAI::late_initialization() {
...
-		   "This is the building that has 'collects_ware_from_map = \"water\"' in its AI hints.",
-		   tribe_->name().c_str());
-	}
-	if (count_buildings_with_attribute(BuildingAttribute::kLumberjack) != 1) {
-		throw wexception(
-		   "The AI needs the tribe '%s' to define 1 type of lumberjack's building. "
-		   "This is the building that has 'collects_ware_from_map = \"log\"' in its AI hints.",
-		   tribe_->name().c_str());
+		log_warn("The AI needs the tribe '%s' to define 1 type of well. "
+		         "This is the building that produces te ware 'water' and has no inputs.",
```
hessenfarmer, line 1105: typo te -> the

src/ai/defaultai.cc
```diff
@@ -841,31 +866,141 @@ void DefaultAI::late_initialization() {
...
+				 * */
 			}
-			// and collectors
-			if (bh.collects_ware_from_map() == "fruit") {
-				bo.set_is(BuildingAttribute::kNeedsBerry);
+
+			// wells
+			if (prod.input_wares().empty()) {
+				for (Widelands::DescriptionIndex ware_index : prod.output_ware_types()) {
+					if (tribe_->get_ware_descr(ware_index)->name() == "water") {
```
hessenfarmer, line 970: this definition is now different then before as well. in trunk only buildings that mine water from the ground are defined as well now this is dependent on the output. Does not do any harm currently as the amazon water gatherer was faked to be such a well but if we can lower the restriction to have one well defined we might need to change this again

src/ai/defaultai.cc
```diff
@@ -6237,14 +6406,18 @@ void DefaultAI::consider_productionsite_influence(BuildableField& field,
 	for (const Widelands::DescriptionIndex& i : bo.ware_outputs) {
 		++field.producers_nearby.at(i);
 	}
-	if (bo.has_collected_map_resource()) {
-		++field.collecting_producers_nearby.at(bo.get_collected_map_resource());
+	for (const auto& supported_building : bo.supported_producers) {
+		// Using std::map syntax to create entry on demand.
+		++field.supported_producers_nearby[supported_building.first];
```
hessenfarmer, line 6411: I believe this is not doing what it should according to the variable name. 
As far as I understand this we are adding the supported buildings of the building we found here, while we should add the building itself as the goal is to identifiy the number of exisitng dedicated buildings in the vicinity of a single field. So we need to count thze building we found rather then the ones that are supported by our building found.

src/ai/defaultai.cc
```diff
@@ -6237,14 +6406,18 @@ void DefaultAI::consider_productionsite_influence(BuildableField& field,
...
-	if (!bo.production_hints.empty()) {
-		for (auto ph : bo.production_hints) {
-			++field.supporters_nearby.at(ph);
+	const Widelands::ProductionSiteDescr* productionsite =
+	   dynamic_cast<const Widelands::ProductionSiteDescr*>(bo.desc);
+	for (const auto& supported : productionsite->supported_productionsites()) {
+		if (field.supporters_nearby.count(supported) != 1) {
+			field.supporters_nearby[supported] = 0;
 		}
+		++field.supporters_nearby[supported];
```
hessenfarmer, line 6420: same here as above. we need to count the building found at this coordinate not the buildings supported by this one.

hessenfarmer: forget this one as here it is correct we count the building that was found as supporter for each building it supports

hessenfarmer:
started fixing issues
can't push
however here are my current changes
[fixes.zip](https://github.com/widelands/widelands/files/5569855/fixes.zip)

src/ai/defaultai.cc
```diff
@@ -2929,45 +3118,42 @@ bool DefaultAI::construct_building(const Time& gametime) {
...
-						}
+						prio += management_data.neuron_pool[49].get_result_safe(
+						           bf->immovables_by_name_nearby[bo.name], kAbsValue) /
+						        5;
+
+						prio += number_of_supported_producers_nearby * 5 -
+						        (expansion_type.get_expansion_type() != ExpansionMode::kEconomy) * 15 -
+						        bf->space_consumers_nearby *
+						           std::abs(management_data.get_military_number_at(102)) / 5 -
+						        bf->immovables_by_name_nearby[bo.name] / 3;
```
hessenfarmer, line 3133: here previously rocks were counted now we count trees.

src/ai/defaultai.cc
```diff
@@ -2929,45 +3118,42 @@ bool DefaultAI::construct_building(const Time& gametime) {
...
+
+						prio += number_of_supporters_nearby * 3;
```
hessenfarmer, line 3135: this needs to identify other rangers but doesn't in this state .

src/ai/defaultai.cc
```diff
@@ -4843,13 +5008,12 @@ bool DefaultAI::check_productionsites(const Time& gametime) {
 			return false;
 		}
 
-		const uint32_t trees_in_vicinity = map.find_immovables(
-		   game(), Widelands::Area<Widelands::FCoords>(map.get_fcoords(site.site->get_position()), 5),
-		   nullptr,
-		   Widelands::FindImmovableAttribute(Widelands::MapObjectDescr::get_attribute_id("tree")));
+		unsigned trees_nearby = find_immovables_nearby(site.site->descr().collected_attributes(),
```
hessenfarmer, line 5011: as this is dealing with rangers we need to find created immovables rather then collected ones.

hessenfarmer:
ok I solved most of the issues except one that I will mark in the code in about an hour. 
here are my changes made.
I keep on working on this tonight. Found some AI issues on the way which I fixed as well
[changes.zip](https://github.com/widelands/widelands/files/5581435/changes.zip)

src/ai/defaultai.cc
```diff
@@ -4613,18 +4782,16 @@ bool DefaultAI::check_productionsites(const Time& gametime) {
...
-		   nullptr,
-		   Widelands::FindImmovableAttribute(Widelands::MapObjectDescr::get_attribute_id("tree")));
-
 		if (site.site->get_statistics_percent() >
 		    std::abs(management_data.get_military_number_at(117)) / 2) {
 			return false;
 		}
 
-		if (remaining_trees > trees_nearby_treshold_ / 3) {
+		unsigned trees_nearby = find_immovables_nearby(site.site->descr().collected_attributes(),
```
hessenfarmer, line 4790: this does not deliver the same number as the old code. But so far I don't know why

src/ai/defaultai.cc
hessenfarmer: same here as with lumberjacks even when using created attributes the code delivers different numbers as in trunk. which results in rangers not stopped even if totally surrrounded with trees.

src/ai/defaultai.cc
hessenfarmer: Well here I was wrong. After inserting some debugging logs I confirm this line is working.

src/ai/defaultai.cc
hessenfarmer: Ok I found the problem we can't count collected Immovables from a ranger as he has none. We can't count created immovables neither as they are saplings so I made a solution to count the collected attributes from the suppoprted buildings which delivered the correct attributes and numbers. Will upload my changes again tonight.

src/ai/defaultai.cc
hessenfarmer: will fix this in another branch after this goes in. together with some other optimizations especially for amazons

hessenfarmer:
so here are my final changes. with them in this should be ready to go in.   
[final changes.zip](https://github.com/widelands/widelands/files/5592719/final.changes.zip)

hessenfarmer:
@gunchleoc: I hope you are well. As I had a working solution for this, which is included in the zip shall I Push it to the main repository and put it under review, so this can get in before v1.0?
@Noordfrees Any thoughts on this?

Noordfrees:
> As I had a working solution for this, which is included in the zip shall I Push it to the main repository and put it under review, so this can get in before v1.0?

+1

hessenfarmer:
As the superseeding PR got merged I am closing this now

PR #4470, opened 2020-11-03
Add support for running training wheels in scenarios

gunchleoc:


src/logic/training_wheels.cc
```diff
@@ -115,11 +130,16 @@ bool TrainingWheels::acquire_lock(const std::string& objective) {
 		current_objective_ = objective;
 		return true;
 	}
-	return current_objective_ == objective;
+	const bool result = current_objective_ == objective;
+	if (result) {
+		log_dbg("Training wheel '%s' acquired the lock", objective.c_str());
```
Noordfrees:
```suggestion
		log_info("Training wheel '%s' acquired the lock", objective.c_str());
```

src/logic/training_wheels.cc
```diff
@@ -115,11 +130,16 @@ bool TrainingWheels::acquire_lock(const std::string& objective) {
...
+	}
+	return result;
 }
 
 // TODO(Gunchleoc): Optionally add training wheel name so that it can be put back in the queue
 void TrainingWheels::release_lock() {
+	log_dbg("Training wheel lock released");
```
Noordfrees:
```suggestion
	log_info("Training wheel lock released");
```

src/scripting/lua_game.cc
```diff
@@ -752,6 +753,31 @@ int LuaPlayer::mark_training_wheel_as_solved(lua_State* L) {
 	return 0;
 }
 
+/* RST
+   .. method:: run_training_wheel(name[, force])
+
+      Trigger running a training wheel while skipping the dependency check.
+      Previously solved training wheels will be ignored if ``force == true``.
```
Noordfrees: Add a note that the wheel's dependency requirements are ignored regardless of whether `force` is `true` or `false`

Noordfrees (commented):
Three comments inline, not tested yet

Noordfrees (commented):
Tested and working

Noordfrees (approved):

PR #4478, opened 2020-11-05
Allow hiding fields that are seen by units

### LINKED ISSUE #4422: wl.Game.Player:hide_fields() does not work anymore
hide_fields() was previously used to hide fields of a headquarters or a ship and reveal the fields through [field_animations.lua](https://www.widelands.org/documentation/autogen_auxiliary_field_animations/), e.g. reveal_concentric(). This does not work anymore, because the fields revealed by a headquarters can't be hid.

**To Reproduce**
1. start attached scenario

Or use one of the campaigns, e.g. emp01.

One can also use the debug console within the attached scenario and try to run `pl:hide_fields(pl_sf:region(12), true)`

**Expected behavior**
The visible area shouldn't be visible until it get revealed again.

**Version:**
 - OS: archlinux
 - Widelands Version: [cf3f7f6@master]

[test_sleep.zip](https://github.com/widelands/widelands/files/5443654/test_sleep.zip)

Niektory:
Fields can now be either Revealed (always visible no matter what), Hidden (always invisible/unexplored no matter what) or Normal (visible if seen by units).

The Lua functions work as follows:
- `reveal_fields(fields)` - make fields Revealed
- `hide_fields(fields, unexplore = false)` - make fields Normal if they were Revealed
- `hide_fields(fields, unexplore = true)` - make fields Hidden

This should make the old scripts work without changes, which is probably desired. However, there is no function to make Hidden fields Normal without making them Revealed first, so perhaps some change/addition is needed. Thoughts?

Fix #4422
Only lightly tested, naming and documentation not updated yet.


frankystone:
Yeah, the old scripts work again :-) Additionally military buildings can also be hid now.

I did some small testing and found the function `hide_fields` with `false` as argument has no effect (the area didn't get dark)? But i remember this was already in build21. So i am not sure if arguments are needed at all. But i think i do not understand the exact definition of 'Normal'. That's might be the reason why i have no meaning about this:

> However, there is no function to make Hidden fields Normal without making them Revealed first, so perhaps some change/addition is needed. Thoughts?

I am in hurry now... probably i will do some more testing this afternoon.

Niektory:
All fields are "Normal" when starting the game. "Normal" basically means that regular gameplay rules apply. So a "Normal" field is visible if it's in vision range of any of your buildings or workers, and not visible (either black or darkened) if it's not in vision range of any of them.

By contrast, "Revealed" and "Hidden" states override the normal gameplay rules – in these states fields are always visible or always hidden respectively, and whether they're in range of your buildings or workers doesn't matter.

So using `hide_fields` with `false` results in no immediate change if the fields are in vision range of your buildings or workers. If they are not in range, the fields will be slowly darkened.

Examples:
All fields are Normal:
shot0153

All fields are Hidden:
shot0152

All fields are Revealed:
shot0154

src/logic/vision.h
```diff
@@ -113,35 +113,56 @@ struct Vision {
...
 		if (reveal == is_revealed()) {
 			return;
 		}
 		if (reveal) {
-			value_ = std::max(value_, uint16_t(2)) + 1;
+			override_ = Override::kRevealed;
+		} else {
+			override_ = Override::kNormal;
+		}
+		if (value_ < 2) {
```
gunchleoc: This would be more compact:
`
value_ = std::max(value_, 2U);`

gunchleoc (commented):
1 code nit. Not tested yet.

frankystone:
Many thanks for the explanation of the term 'Normal' :)

I think all works as expected now again.

> However, there is no function to make Hidden fields Normal without making them Revealed first, so perhaps some change/addition is needed. Thoughts?

So far those functions were used to hide fields and reveal them close afterwards again. So i think this shouldn't be a problem. @hessenfarmer ?

gunchleoc (approved):

PR #4491, opened 2020-11-08
Fix tooltip position

### LINKED ISSUE #4180: Tooltip layouting issues

jmoerschbach:
fixes #4180

Not sure if the workaround for the unwanted line break is ok. Maybe it better should be fixed in the text rendering component itself (but that looks rather complicated to me imho)?

grafik

### DIFF
--- a/src/ui_basic/panel.cc
+++ b/src/ui_basic/panel.cc
@@ -1288,11 +1288,20 @@ bool Panel::draw_tooltip(const std::string& text, const PanelStyle style) {
                                                                 UI::FontStyle::kFsTooltip);
 	}
 
-	constexpr uint32_t kTipWidthMax = 360;
+	constexpr int kTipWidthMax = 360;
 	std::shared_ptr<const UI::RenderedText> rendered_text =
 	   g_fh->render(text_to_render, kTipWidthMax);
+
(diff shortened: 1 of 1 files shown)


Noordfrees (approved):
LGTM, tested and working

Just one glitch: When the text is so wide that it doesn't fit on the screen, the text overlays the right edge of the tooltip. Not a very big bug though, not sure if this is worth fixing.

jmoerschbach:
> 
> 
> LGTM, tested and working
> 
> Just one glitch: When the text is so wide that it doesn't fit on the screen, the text overlays the right edge of the tooltip. Not a very big bug though, not sure if this is worth fixing.

Not from my side, no (this is a corner case of a corner case...^^)

Issue #4300, opened 2020-09-29
Minimap far too large


Noordfrees:
When starting a game on a 64×512 map (low width, big height) at 1024×768 window resolution, the minimap is scaled up to many times the screen height. It should be small enough to fit on the screen completely.
![shot0003](https://user-images.githubusercontent.com/48293446/94602174-cfd21880-0294-11eb-849b-c8032037e56e.png)

frankystone:
There was a similar issue when clicking on 2x in the minimap window with big maps and small screen resolution, but i can't find it anymore. Anyway this seems related to me.

matthiakl:
Scaling the map only considers the width and not the height:
https://github.com/widelands/widelands/blob/9919ef5bf50f6873f36c594eb506d32e702d666d/src/graphic/minimap_renderer.cc#L273-L287

PR #4495, opened 2020-11-09
Scale minimap based on the maximum of width and height of map.

flegmaatikko:
These changes fix the bug #4300

The Minimap window scaling didn't work properly with map size of
64x512 pixels. The scaling resulted too high minimap window.
The window scales better when the maximum of width and height is used.

The minimap scaling has tested in the editor with maps of 64x144, 64x160, 64x288, 64x304, 64x512, 144x64, 160x64, 288x64, 304x64 and 512x64 size by using 800x600 and 2560x1600 Window resolutions. The scaling has also tested with the 2x zoom of minimap (the zoom is disabled if height or width of a map is greater than 300px).
All the combinations seems to work quite well.


Niektory (commented):
Should the width and the height be treated the same? The width can probably be slightly larger than the height since I think it's supposed to fit into 800x600.

`MiniMap::View::can_zoom()` needs to be adjusted too. A 192x144 map can be zoomed in, but 144x144 cannot, despite being smaller.

The buttons at the bottom are tiny when the map is 64x512, but it's still an improvement over the previous behavior.

flegmaatikko:
I think the reason why the 144x144 can not be zoomed comes from the `-60` value in the `MiniMap::View::can_zoom()` method. The scale is calculated without the windows borders and the button height in `scale_map`. Without the `-60` zoom works but the minimap window is slightly higher than 600px. I can try to figure out how to handle that case.

I'm not sure should  the width and the height be treated the same. I have to think more about that.
You are right, that buttons are very tiny with the map of  64x512. The width of the minimap should be wider in that point of view. What you think about idea of using two rows for the buttons when the width is smaller than the width of 6 buttons? Should we create a new issue for the button case?

Niektory:
> What you think about idea of using two rows for the buttons when the width is smaller than the width of 6 buttons? Should we create a new issue for the button case?

Yeah, a new issue sounds good.

flegmaatikko:
The scale of a map is calculated without the borders and the button height of the Minimap now. The `MiniMap::View::can_zoom` method also tests the width of the scaled map.
The result is not perfect but at least the Minimap stays inside the window and 2x zoom works up to a map of 270px resolution.

src/graphic/minimap_renderer.cc
```diff
@@ -346,16 +347,19 @@ std::unique_ptr<Texture> draw_minimap_final(const Texture& input_texture,
...
 		if (zoom) {
-			return (600 - (600 % map_w)) / map_w;
-		} else if (map_w > 150) {
-			return (400 - (400 % map_w)) / map_w;
+			return (540 - (540 % max)) / max;
+		} else if (max > 135) {
+			return (370 - (370 % max)) / max;
 		} else {
-			return (300 - (300 % map_w)) / map_w;
+			return (270 - (270 % max)) / max;
```
Niektory: I don't get these formulas. Wouldn't `return 270 / max;` and so on suffice?

src/wui/minimap.cc
```diff
@@ -98,8 +98,10 @@ void MiniMap::View::set_zoom(const bool zoom) {
 bool MiniMap::View::can_zoom() {
 	const Widelands::Map& map = ibase_.egbase().map();
 	// The zoomed MiniMap needs to fit into: height - windows boarders - button height. -> 60px
-	return !(scale_map(map, true) == 1 ||
-	         map.get_height() * scale_map(map, true) > ibase_.get_h() - 60);
+	auto scale = scale_map(map, true);
```
Niektory: Make it `const`.

src/wui/minimap.cc
```diff
@@ -98,8 +98,10 @@ void MiniMap::View::set_zoom(const bool zoom) {
...
+	return !(scale == 1 ||
+		 map.get_width() * scale > ibase_.get_w() - 60 ||
+		 map.get_height() * scale > ibase_.get_h() - 60);
```
Niektory: Maybe it's just me, but I think this would be easier to understand without the negation (`scale != 1 && foo <= bar && foo <= bar`)

Niektory (approved):
Looks good. Added a few minor code suggestions.

src/graphic/minimap_renderer.cc
flegmaatikko: Yes, simpler is better Even if the max is a floating point number the  `270 / max` results an integer. I just did minimal changes without changing the logic. I can do the changes if you want?

src/wui/minimap.cc
flegmaatikko: Yes, it is easier and also no need to check  `foo <= bar`  if `scale != 1` fails. Going even further we should use `scale > 1` but not really necessary because `scale_map` function always returns  >= 1.

src/graphic/minimap_renderer.cc
gunchleoc: Better to do them now before we forget.

Niektory (approved):

Niektory:
I'd merge but Travis has issues (unrelated to this PR) and I don't know how to get around it.

Issue #4490, opened 2020-11-07
different translations for tribes of the same string is not possible anymore.  


hessenfarmer:
Previously it was possible to translate the same string diferently for different tribes.
e.g. soldiers was translated as "Soldaten" (soldiers) for atlanteans and empire but "Krieger" (Warriors) for barbarians and frisians.    

This seems to be not possible anymore e.g. for the miitary Buildings with the same name.

Do we need / want to change the source strings to be able to do that as well? or is there anaother way of achieving this.

gunchleoc:
We can change the translation markups in `units.lua` and in the militarysites to use a tribe-specific pgettext context. Best grep for "soldier" so we won't miss anything.

What will remain as "Soldier(s)" when we do that are the census strings and statistics windows.

hessenfarmer:
Census and statistics would be ok. Although census would be nice as well especially with amazons

gunchleoc:
I had a look on Transifex. For C++ changes, we have the following categories:

- **Census** strings. Where do we define it in Lua? Could be at the soldier worker, at the military sites or in the units.lua. IAt militarysites sounds best to me, although this will generate some duplication. What do you think?
- **Production program** strings (would need to be defined via Trainingsite Lua
- **UI** strings (toolbar menu, attack box, building statistics menu, soldierlist) - can stay as they are

hessenfarmer:
> 
> 
> I had a look on Transifex. For C++ changes, we have the following categories:
> 
>     * **Census** strings. Where do we define it in Lua? Could be at the soldier worker, at the military sites or in the units.lua. IAt militarysites sounds best to me, although this will generate some duplication. What do you think? 

Milsites should be ok

> 
>     * **Production program** strings (would need to be defined via Trainingsite Lua

ok 

> 
>     * **UI** strings (toolbar menu, attack box, building statistics menu, soldierlist) - can stay as they are

agreed

hessenfarmer:
Unfortunately it isn't fixed
see
https://www.transifex.com/widelands/widelands/translate/#de/tribes/65218455

hessenfarmer:
maybe we need to have an own TD for each tribe

gunchleoc:
> Unfortunately it isn't fixed

Those are different strings, they just need a pgettext too

> maybe we need to have an own TD for each tribe

Not possible because of shared wares & immovables

PR #4499, opened 2020-11-10
Disambiguate helptexts for soldier-related buildings

gunchleoc:
Re https://github.com/widelands/widelands/issues/4490

More work needs to be done on the engine side for closing the issue.

@hessenfarmer


hessenfarmer (approved):
LGTM.
I was not aware that the fix will be that easy although I was suspicious about these text parts

PR #4502, opened 2020-11-10
Disambiguate militarysite statistics for tribes

### LINKED ISSUE #4490: different translations for tribes of the same string is not possible anymore.

gunchleoc:
Closes https://github.com/widelands/widelands/issues/4490

This is a bit more complicated because of npgettext.

@hessenfarmer


hessenfarmer (approved):
LGTM. except the .poi file which might be a remainder odf some test I think they will be created by the translation later on

gunchleoc:
Yep, the PO file was an accidental commit.

gunchleoc:
@hessenfarmer New strings are up on Transifex

PR #4503, opened 2020-11-10
Fix readability-else-after-return in UI code

gunchleoc:
Re. https://github.com/widelands/widelands/issues/3223

I also streamlined the logic a bit in some places.

Splitting this check up into multiple branches to make code review easier.


Noordfrees (approved):
This one check doesn't deserve the name "readability" in my opinion, but the code looks OK

gunchleoc:
Personally, I do find it more readable, because it's easier to spot where the code will return. Maybe not that obvious from looking at a diff.

PR #4504, opened 2020-11-10
Fix MacOS compiler warnings on Travis

gunchleoc:
I don't have a Mac, so we'll need to wait for the Travis logs to see if this branch works.


Noordfrees (approved):
Please add a note about the change to gettext to `src/third_party/README`. Otherwise LGTM.

gunchleoc:
That file is heavily customized already - I don't think it needs a separate change log.

PR #4505, opened 2020-11-10
Fix readability-else-after-return in logic

gunchleoc:
Re. #3223


src/logic/player.cc
```diff
@@ -1828,11 +1825,11 @@ void Player::set_attack_forbidden(PlayerNumber who, bool forbid) {
...
 	if (forbid ^ (it == forbid_attack_.end())) {
 		return;
-	} else if (forbid) {
+	}
+	if (forbid) {
 		forbid_attack_.emplace(who);
-	} else {
-		forbid_attack_.erase(it);
 	}
+	forbid_attack_.erase(it);
```
Noordfrees: This does not look right…

gunchleoc: Eek. Fixed.

Noordfrees (approved):

PR #4506, opened 2020-11-11
Update Europa Map to v1.2

Noordfrees:
Update Europa map from version 1.1 to 1.2 by @MarkMcWire.

Additionally deleted whitespaces from map file names.


gunchleoc:
While you're fixing directory names, how about removing the map version number from them? This way, we won't have to `git mv` every time we update a map.

gunchleoc (approved):

PR #4507, opened 2020-11-11
Amazons string fixes

gunchleoc:
Found these while translating the Tribes textdomain.

Spelling for ship names taken from the English Wikipedia.


hessenfarmer (approved):
LGTM.  
Maybe you could address the remaining issue on transifex (about using the correct unicode for -) as a side catch as well

gunchleoc:
> Maybe you could address the remaining issue on transifex (about using the correct unicode for -) as a side catch as well

I'd rather do that as a separate branch so that we can get this one in. I'll want to check all the mathematical symbols while I'm at it.

PR #4515, opened 2020-11-12
Allow savegames anywhere

### LINKED ISSUE #3776: XDG can unexpectedly turn WL_INSTALL_DATADIR useless
Since commit d1b47e73af34bee0f2a38ade58d9ef575afff59b there is support for XDG_DATA_DIR which was later enabled by default in commit ebe684c674149acd3e8991312899063b363cad36
Now the XDG code silently overrides (WL_)INSTALL_DATADIR in [src/wlapplication.cc](https://github.com/widelands/widelands/blob/master/src/wlapplication.cc#L975) if the path given there is not the first one contained in XDG_DATA_DIR.

This has two implications: 
- The first one is, that the built `widelands` binary will never look for its data files in the correct path unless called with the `--datadir` command line option because the data got installed into the path given with WL_INSTALL_DATADIR during build.
- The second one is that even if we happen to have chosen a path for WL_INSTALL_DATADIR that is contained in XDG_DATA_DIR the built `widelands` binary will only find its data files if the path happens to be the first in XDG_DATA_DIR. Imagine what happens if a user manually adds another path as first one to his XDG_DATA_DIR environment variable.

Why do I even come up with WL_INSTALL_DATADIR? Because the top [CMakeLists.txt](https://github.com/widelands/widelands/blob/master/CMakeLists.txt#L70) file contains a comment that explains what happens if this variable is unset or if it contains a relative or absolute path.
This variable is quite handy for package maintainers if the `widelands` binary is supposed to get installed into `/usr/bin` for example. Now I can put an absolute path into WL_INSTALL_DATADIR (like `/usr/share/widelands/data`) and the binary could always find its data files no matter where it gets installed but only if there wasn't this strange behavior if XDG is enabled.

This has already been reported in issue #3490 but that one got closed without any real resolution.

As a long term solution I'd like to suggest using [GNUInstallDirs](https://cmake.org/cmake/help/v3.16/module/GNUInstallDirs.html) in the cmake files for installation into the Linux file system.

**To Reproduce**
1. `cmake -DCMAKE_INSTALL_PREFIX=/usr -DOPTION_BUILD_WEBSITE_TOOLS=OFF -DUSE_XDG=ON -DWL_INSTALL_BASEDIR=/usr/share/widelands -DWL_INSTALL_DATADIR=/usr/share/widelands/data`
2. build the project and move the binary to `/usr/bin/`
3. Run it and watch the error message about missing `i18n/fonts.lua` file

**Expected behavior**
The binary finds its data files

**Version:**
 - Gentoo Linux current ~amd64
 - Widelands Version: 5841c3146ea4e7e15479e6ca1592bfbea97fceb3

### LINKED ISSUE #4202: Allow savegame files anywhere
When double-clicking on a Widelands save in Windows, Widelands will crash.

I can reproduce that on Linux like this:

```
username@machine:~/sources/widelands/github/widelands-codereview $ ./widelands --loadgame=test.wgf

<snip>
ERROR: Fatal exception: [../src/io/filesystem/layered_filesystem.cc:245] LayeredFileSystem: unable to create sub filesystem for existing directory: /home/username/sources/widelands/github/widelands-codereview/test.wgf
ERROR: I have tried the following path(s):
ERROR:     /home/username/.widelands//home/username/sources/widelands/github/widelands-codereview/test.wgf
ERROR:     /media/username/Linux/sources/widelands/github/widelands-codereview/data//home/username/sources/widelands/github/widelands-codereview/test.wgf
ERROR: FATAL ERROR - game crashed. Attempting emergency save.
terminate called after throwing an instance of 'WException'
  what():  [../src/io/filesystem/layered_filesystem.cc:245] LayeredFileSystem: unable to create sub filesystem for existing directory: /home/username/sources/widelands/github/widelands-codereview/test.wgf
I have tried the following path(s):
    /home/username/.widelands//home/username/sources/widelands/github/widelands-codereview/test.wgf
    /media/username/Linux/sources/widelands/github/widelands-codereview/data//home/username/sources/widelands/github/widelands-codereview/test.wgf

```
1. It would be nice to have support for savegames anywhere
2. Maybe we can have some user-facing GUI for this error if we don't implement 1

matthiakl:
Fixes #4202: 
Load save/replay from anywhere and allow file extension of .wgf/.wrpl in file browser to work. GUI error message included.

Partially fixes #3776:
WL_INSTALL_DATADIR is only overridden by XDG if it is a relative path. Still `/widelands` is appended to XDG_DATA_DIR.


Niektory:
I think the error message is misleading if the file doesn't exist. "The file format seems to be incompatible." implies that the file exists, but is incompatible. So the message for non-`GameDataError` exceptions should be different.

src/wlapplication.h
```diff
@@ -272,6 +272,10 @@ struct WLApplication {
 	/// Prevent toggling fullscreen on and off from flickering
 	uint32_t last_resolution_change_;
 
+	/// Content for popup error messagebox
+	std::string messagetitle;
+	std::string message;
```
Niektory: `message` and `messagetitle` could be parameters to the `WLApplication::mainmenu()` call instead of member variables. If you think they should be member variables, add underscores to their names.

src/wlapplication.cc
```diff
@@ -980,7 +994,14 @@ void WLApplication::parse_commandline(int const argc, char const* const* const a
 
 		// Are we looking at an option at all?
 		if (opt.compare(0, 2, "--")) {
-			throw ParameterError();
+			if (argc == 2) {
```
Niektory, line 1050: Maybe also check if it ends with a .wgf extension here? I'm not sure.

src/wlapplication.h
Noordfrees: `WLApplication::mainmenu()` will be removed soon, so local variables and local handling of error messages would probably be better. I'm not looking forward to the merge conflicts if this is merged before #4518… ;)

src/wlapplication.cc
matthiakl: I omitted additional checks here because you will get a GUI message, if it isn't a wgf ("incompatible file").

src/wlapplication.h
matthiakl: Local variables sounds good. I will wait for #4518 and then handle the merge.

src/wlapplication.cc
Niektory: I was thinking of command line use. Typing `./widelands badargument` will attempt to load `badargument` instead of printing the usage information. But maybe that's fine.

Speaking of usage information (the one printed with `--help`), do add information there about this behavior.

matthiakl: Ah, ok. The argument now requires a wgf extension.

src/wlapplication_messages.cc
```diff
@@ -170,6 +172,12 @@ void show_usage(const std::string& build_id, const std::string& build_type) {
 	          << endl;
 	std::cout << _(" --verbose            Enable verbose debug messages") << endl << endl;
 	std::cout << _(" --help               Show this help") << endl << endl;
+	std::cout
+	   << _(" <save.wgf>/<replay.wrpl> Directly loads the given savegame or replay. Useful for\n"
+	        "                      .wgf/.wrpl file extension association. Does not work with\n"
+	        "                      other options. Also see --loadgame/--replay.")
```
Noordfrees: Let the explanation begin on a new line for consistency

src/wlapplication.cc
```diff
@@ -503,30 +503,46 @@ void WLApplication::run() {
...
+			title = _("File system error");
 		} catch (const std::exception& e) {
 			emergency_save(nullptr, game, e.what());
+			message = e.what();
+			title = _("Error message:");
+		}
+		g_sh->change_music("menu");
+		FsMenu::MainMenu m(true);
+		if (!message.empty()) {
+			m.show_messagebox(title, message);
```
Noordfrees, line 543: Please print the message to stdout as well (either here or in `show_messagebox()`)

src/ui_fsmenu/main.cc
```diff
@@ -97,7 +97,7 @@ int16_t MainMenu::calc_desired_window_y(const UI::Window::WindowLayoutID id) {
 	return (get_h() - calc_desired_window_height(id)) / 2 - UI::Window::kTopBorderThickness;
 }
 
-MainMenu::MainMenu()
+MainMenu::MainMenu(bool skip_init)
```
Noordfrees:
```suggestion
MainMenu::MainMenu(const bool skip_init)
```

Noordfrees (commented):
Tested and working, only a few suggestions inline

Noordfrees (approved):

matthiakl:
Testsuite failure is #4575

PR #4517, opened 2020-11-13
4508 fix argument position in campaign_message_box

### LINKED ISSUE #4508: Argument 'position' of campaign_message_box does not hold the description
The documentation of campaign_message_box says:

>position - A string that indicates at which border of the screen the message box shall appear. Can be “top”, “bottom”, “right”, “left” or a combination (e.g. “topright”). Overrides posx and posy. Default: Center. If only one direction is indicated, the other one stays centered.

But some of the setting does not work correct:

1. "top" shows the message box at top left. According to the documentation it should be horizonally centered 
2. "bottom" shows the message box bottom left. According to the documentation it should be horizontally centered 
3. "right" shows the message box top right. According to the documentation it should be vertically centered 
4. "left" shows the message box top left. According to the documentation it should be vertically centered.

This happens also if the settings in options for "Dock window to edges" or "Snap windows when overlapping" are unmarked.

**To Reproduce**
The attached scenario opens windows at all 4 positions. Just load it.

**Expected behavior**
When using "top" the window should be horizontally centered.
When using "bottom" the window should be horizontally centered.
When using "left" the window should be vertically centered.
When using "right" the window should be vertically centered.

**Screenshots**
This screeny uses the new option `modal = false` to show every window at once, but the issue is the same if this option is not used and one clicks through all windows.
 
messagebox_position

**Version:**
 - OS: archlinux
 - Widelands Version: [1878bf2@master] (Debug)

[message_position.zip](https://github.com/widelands/widelands/files/5526042/message_position.zip)

frankystone:
Fixes #4508 

This fixes the issue with LUA. Maybe there is another way to fix it on the c++ side...

This solution removes also some magic numbers for moving the window(s) to the right or to the bottom by getting the width and height of the main window through wl.ui.MapView().


data/scripting/messages.lua
```diff
@@ -139,16 +139,26 @@ function campaign_message_box(message, sleeptime)
    message.w = message.w or 450
 
    if message.position then
+      local window = wl.ui.MapView()
       if string.find(message.position,"top") then
-         -- Set it a bit lover than 0 to prevent overlap with game speed text
```
Niektory, line 143: "bit lover" :-D

Niektory (commented):
Wouldn't it be simpler to do it like this?
```lua
if string.find(message.position,"top") then
   message.posy = 25
elseif string.find(message.position, "bottom") then
   message.posy = window.height - message.h
else
   message.posy = math.floor((window.height - message.h) / 2)
end
if string.find(message.position, "left") then
   message.posx = 0
elseif string.find(message.position, "right") then
   message.posx = window.width - message.w
else
   message.posx = math.floor((window.width - message.w) / 2)
end
```

frankystone:
Yep, thanks :-)

Niektory (approved):
LGTM. Just add back the comment about the top position, it's useful.

PR #4521, opened 2020-11-17
Update partially seen buildings' visibility when placed/removed

### LINKED ISSUE #4480: Unable to load savegame (vision packet)
The attached two savegames were created with Version 1.0~git25021[fd4744c@master] (Release). They both fail to load with the same version:
```
[00:00:08.743 real] ERROR: Fehler mit de Speel-Daten
[00:00:08.743 real] ERROR: view: [../src/map_io/map_players_view_packet.cc:174] Read 183 unseen fields but detected 182 when the packet was written
```
```
[00:03:31.803 real] ERROR: Fehler mit de Speel-Daten
[00:03:31.803 real] ERROR: view: [../src/map_io/map_players_view_packet.cc:174] Read 1133 unseen fields but detected 1132 when the packet was written
```
[savegames.zip](https://github.com/widelands/widelands/files/5497565/savegames.zip)

Niektory:
Fix #4480

Partially seen buildings were not being discovered during gameplay when built at the edge of vision, but they were during loading. This caused a mismatch.

Now they should be discovered immediately when placed.

This doesn't fix broken savegames but you should be able to resave from a replay to get a good one.

### DIFF
--- a/src/logic/map_objects/immovable.cc
+++ b/src/logic/map_objects/immovable.cc
@@ -95,7 +95,9 @@ void BaseImmovable::set_position(EditorGameBase& egbase, const Coords& c) {
 	}
 
 	// Needed so players can update partially visible buildings' vision.
-	Notifications::publish(NoteFieldTerrainChanged{f, map->get_index(f)});
+	if (get_size() >= BIG) {
+		Notifications::publish(NoteFieldTerrainChanged{f, map->get_index(f)});
+	}
 }
 
 /**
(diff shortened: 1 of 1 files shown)


Noordfrees (commented):
Tested, loading the replay and resaving works fine now. Thanks for fixing :)

How is the performance of this fix? Rediscovering a node every time an immovable is created or removed might be performance-hungry, perhaps this should be done only for immovables with multiple positions

Niektory:
Rediscovering is pretty fast so I don't think it makes much of a difference. But OK, I restricted it to big buildings only, since that's all we need now.

src/logic/map_objects/immovable.cc
```diff
@@ -115,6 +120,11 @@ void BaseImmovable::unset_position(EditorGameBase& egbase, const Coords& c) {
 
 	assert(f.field->immovable == this);
 
+	// Needed so players can update partially visible buildings' vision.
+	if (get_size() >= BIG) {
+		Notifications::publish(NoteFieldTerrainChanged{f, map->get_index(f)});
+	}
+
 	f.field->immovable = nullptr;
```
Noordfrees, line 128: Shouldn't the field rather be rediscovered *after* unlinking the immovable from the field?

Niektory: The notification only triggers rediscovery of fields that are visible for a given player. The building's main position is out of sight, so it can only be rediscovered indirectly through rediscovering another part of the building that's in sight. Which only works before unlinking.

Yes, this is hacky.

Noordfrees: Right, good point. LGTM :)

Noordfrees (approved):

PR #4522, opened 2020-11-18
Get rid of the std::move warnings without triggering the other ones

Niektory:
Hopefully.


Noordfrees (approved):
My compiler likes it :) 
Someone else should also test whether both warnings are gone.

klaus-halfmann (approved):
Code  is straightforward
Compiling with clang version 9.0.1  on macOS does not trigge that warning in master.
compiles and regression_tests are ok

PR #4527, opened 2020-11-20
Speed up Economy::find_best_supply()

Niektory:
Experimental. Cut searches earlier and cache the calculated costs to better sort future searches.

### DIFF
--- a/src/economy/economy.cc
+++ b/src/economy/economy.cc
@@ -302,6 +302,11 @@ void Economy::remove_flag(Flag& flag) {
  * This is called from the merge code.
  */
 void Economy::do_remove_flag(Flag& flag) {
+	for (auto& cost_group : cached_costs_) {
+		cost_group.second.erase(flag.serial());
+	}
+	cached_costs_.erase(flag.serial());
+
 	flag.set_economy(nullptr, type_);
 
 	// fast remove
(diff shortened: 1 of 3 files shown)


tibor95:
Those cached_costs are tied to request or ware? Because everytime you are pairing request and supply it is something different.

Niektory:
> Those cached_costs are tied to request or ware? Because everytime you are pairing request and supply it is something different.

I'm not sure I understand the question. Each cached cost is tied to:
- the economy
- the destination flag
- the source flag
If these match, the cached cost is used instead of air distance to sort the searches. This hopefully lets us more often search the least costly ones first so the longer ones are cancelled early.

tibor95:
Yes, this was my question. Is there any lifetime on this cached costs? Removal of target flags removes cached entry for that flag - ok. But how it copes with other changes in road network?

Niektory:
The cached costs are currently only removed when flags are removed or when the game is exited (they are not saveloaded which might or might not be a problem). Changes in the road network don't trigger any immediate changes to the cached costs, but they are updated whenever a search between the associated flags is run. If the search completes the new cost is cached. If it's cancelled because the cost_cutoff was reached then it's increased to cost_cutoff+1 if a lower cost was previously cached.

tibor95:
Did you consider memory consumption on big maps?

Niektory:
It's hard to estimate. It would have to be measured. I didn't do it yet.

hessenfarmer:
@widelands/elders 
This one is stale as well I am afraid

PR #4536, opened 2020-11-23
Use two rows if there are less than 120 pixels available for six buttons.

### LINKED ISSUE #4528: The buttons of the Minimap are tiny when the map is 64x512
**Describe the bug**
A button width of the Minimap is much smaller than its height (20px) with the map of 64x512px. A button is hard to click and texture is not visible.
The pull request #4495  fixed the bug #4300 but at the same time created this problem.

**To Reproduce**
1. Editor > New Map 
2. Select width 64 and height 512'
3. Press Create Map button
4. Press M key

**Expected behavior**
Button width should be at least the same as height. All the buttons should be easier to click. The texture of buttons should be more visible.

**Screenshots**
wl-minimap-with-map-64x512px

**Version:**
 - Widelands Version: 41e8f57a and after

flegmaatikko:
Problem is that one button takes space at least 20 pixels. Six buttons do
not fit nicely into the Minimap if its width is smaller than 120 pixels.
This happens with maps that are narrower than 120 pixels with or without
zoom.

The commit fixes the problem and the bug #4528.

An interesting map size is 64x192px. Its width is narrower than 120px
without zoom and over it with 2X zoom when 800x600px window resolution
is used.

### DIFF
--- a/src/wui/minimap.cc
+++ b/src/wui/minimap.cc
@@ -123,10 +123,12 @@ destructor
 ===============
 */
 inline uint32_t MiniMap::number_of_buttons_per_row() const {
-	return 6;
+	// Six buttons need at least 120 pixels.
+	return view_.get_w() < 120 ? 3 : 6;
 }
 inline uint32_t MiniMap::number_of_button_rows() const {
-	return 1;
+	// Use two rows if there are less than 120 pixels available.
+	return view_.get_w() < 120 ? 2 : 1;
 }
 inline uint32_t MiniMap::but_w() const {
 	return view_.get_w() / number_of_buttons_per_row();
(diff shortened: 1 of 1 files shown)


flegmaatikko:
64x512px
wl-minimap-with-map-64x512px-improved

src/wui/minimap.cc
```diff
@@ -235,18 +237,20 @@ void MiniMap::toggle(MiniMapLayer const button) {
 
 void MiniMap::resize() {
 	view_.set_zoom(*view_.minimap_layers_ & MiniMapLayer::Zoom2);
-	set_inner_size(view_.get_w(), view_.get_h() + number_of_button_rows() * but_h());
+	// Read number of rows after the zoom.
+	const auto rows = number_of_button_rows();
```
Noordfrees:
```suggestion
	const uint32_t rows = number_of_button_rows();
```
No `auto` for primitive types please for readability

Noordfrees (approved):
Tested and working, one comment inline

PR #4538, opened 2020-11-26
Fix curl on AppVeyor

Noordfrees:


hessenfarmer (approved):
first build is green. I tested it locally as well so this shuld be good to go. 
please remove the branch from appveyor.yml before merging
thanks

PR #4543, opened 2020-11-29
Remove Travis

### LINKED ISSUE #4535: Mac builds fail on Travis
Since a couple of days (around 17.11.) the Travis build for Mac has a good chance of failing.
Comparing the logs of [this](https://travis-ci.org/github/widelands/widelands/jobs/745089920) and [that](https://travis-ci.org/github/widelands/widelands/jobs/745090537) build, the problem seems to be a new dependency, which requires a newer cmake version. Building this package takes just about the declared stall time of 10 min.

Noordfrees:
closes #4535


Noordfrees:
Installing the dependencies takes *hours* on travis now. I am strongly in favour of dropping support for it.

gunchleoc:
I'd prefer finding out what the problem is over summarily removing the builds.

How about configuring the Travis builds to allow to fail while we're looking for the issue?

Noordfrees:
The problem is that we install a number of dependencies on travis via homebrew. brew in turn installs further dependencies, and the `make` commands for each of the sub-dependencies (invoked by `brew`) easily takes more than the 10-minutes timeout threshold. This can be fixed by using `travis_wait`, but the dependencies alone are then still not installed after 60 minutes. So even if we extend the timeout to more than that, it is likely that we will hit the absolute timeout (what is it, 120 minutes or such?) before compilation had enough time to complete.

stonerl:
macOS 10.13 is neither supported by Apple nor by homebrew anymore. So I support removing it from the tests.

Noordfrees:
Since matthiakl succeeded in fixing travis we can keep it around after all. 
*…until it breaks again… ;)*

PR #4551, opened 2020-12-09
Fix some encyclopedia strings for the Amazons

gunchleoc:


hessenfarmer (approved):
LGTM although there are more of them, as reported on transifex

gunchleoc:
There probably are - I haven't had the time to finish looking through them and wanted to at least get those in that I had already fixed.

PR #4559, opened 2020-12-14
Fix two info panel crashes

Noordfrees:
The crashes fixed by this PR can be reproduced by saving a map in the editor (1) and by right-clicking a Game/Map Saved notification (2).


matthiakl (approved):
Tested and working :)

PR #4562, opened 2020-12-16
Adjust field size for mountains

### LINKED ISSUE #4553: Immovables do not restrict buildingspace on mountains
Immovables wich have a size do restrict building space on arable terrains, but not on mineable terrains:

immovables_mineable

The standing stone on mineable terrain shows that only one  buildingspace is lost, whereas on arable terrain all surrounding buildingspace is lost.

**To Reproduce**
1. Start the editor
2. Place some mineable terrains
3. Enable buildhelp
4. Place some immovables on mineable and arable terrains

**Expected behavior**
Immovables placed on mineable terrains should restrict buildingspace for mines in the same way like with arable terrains. Since mines have only one size, every immovable wich have a size should mark all neighboring fields as unmineable if it get placed on mountains.

**Version:**
 - OS: archlinnux
 - Widelands Version: [7c074cf@master] (Debug)

matthiakl:
Fixes #4553


Noordfrees (commented):
LGTM, tested and working.
Did you check whether any of the official maps may need updating after this change?

matthiakl:
> Did you check whether any of the official maps may need updating after this change?

This shouldn't be necessary, the buildsize is calculated during map loading and not stored in the map itself. I created and saved a map before the modification and the building spaces were correct afterwards, without changing the map.

Noordfrees:
I meant, whether any of the official maps contains a mining plot which is no longer available after this change. If there is a mountain patch with very few plots it is possible that the map can then no longer be played as it is meant to.

frankystone:
I think the map Dolomites is the most affected one. It has many of the immovable called 'debris' on the mountains.

I do not believe other maps are affected, because immovables are mostly placed on ground terrains.

matthiakl:
I looked through the maps, and I found three, that are affected by this change:

- **Dolomites** and **River Explorers** don't need an update, because affected areas are surrounded by large mineable areas.
- **Swamp Island** should be changed, because some mining plots are now completely obstructed.

Should I include a change in this PR or let some professional map builders do it ;) ?

Noordfrees:
Please do it here :)

Noordfrees (approved):

PR #4571, opened 2020-12-22
Fix crash when worker's terraform program fails

### LINKED ISSUE #4570: Game Crash randomly
widelands: ../src/logic/map_objects/bob.h:301: Widelands::Bob::State& Widelands::Bob::top_state(): Zusicherung »stack_.size()« nicht erfüllt.

I used this build with europeans tribe (as player tribe) and barbarians and empire as AI tribes on the build-in Europa 1.2 map in an endless game. I haven't changed the source code of the game. The game randomly crashes when the game time is above 24 hours. After a certain point in time, a saved game can no longer be loaded. The game crashes seconds after loading.

It should be noted that between the start of the game and the time of the crash I changed a few things in my tribe, such as production programs for workers and buildings or build / enhancement costs. I have no idea how that affects this bug.

Noordfrees:
Fixes #4570


hessenfarmer:
@MarkMcWire 
It would be nice if you could confirm that this PR fixes the mentioned issue, as you know how to reproduce it

MarkMcWire:
It fix the problem. No crash with this exception anymore since I use your code change.

hessenfarmer (approved):
I would rely on @MarkMcWire in his judgement that this fixes the issue 
so it should be merged imho

PR #4573, opened 2020-12-28
fix new static referenced libraries for win builds

hessenfarmer:


Noordfrees (approved):

hessenfarmer:
thanks for the review

Issue #4598, opened 2021-01-07
Prepare for clang-format 11


matthiakl:
This is a heads up, that the `.clang-format` should be reviewed for clang-format 11. The Github actions currently run version 10, but my local version 11 found lots of format changes. To prevent a complete formatting by Github, we should check this, before the runners are updated.
See
https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md
https://github.com/actions/virtual-environments/issues/2130

Noordfrees:
Couldn't we just change the apt line in https://github.com/widelands/widelands/blob/master/.github/workflows/format.yaml#L60 to install `clang-format-10` instead of `clang-format`
Then we can keep the current format version until we choose to migrate to the later version (ideally at a time after the release and when fewer conflicting PRs are open)

matthiakl:
Oh... You're right. That's so much simpler :)

Edit: ...but this would probably brake the dependency to the preinstalled clang version. But the linked issue proposes a solution for that.

Noordfrees:
Argh… the format bot has just formatted 117 files on master, even though it still does use clang and clang-format version 10 according to the logs…

matthiakl:
That's bad. So the fix did not work correctly.
The log only shows version 10 for clang explicit, but not for clang-format:
`update-alternatives: using /usr/bin/clang-10 to provide /usr/bin/clang (clang) in auto mode`
Other than that I don't see any evidence which version is used.

As we already have the proposed mitigation way implemented, maybe we should report this problem over on `virtual-environments#3235`?

PR #4601, opened 2021-01-08
Require clang-format version 10

Noordfrees:
Re #4598

### DIFF
--- a/.github/workflows/format.yaml
+++ b/.github/workflows/format.yaml
@@ -61,6 +61,8 @@ jobs:
         pip install pyformat
         # Use clang-format 10 as default until we choose to migrate to 11
         sudo update-alternatives --remove-all clang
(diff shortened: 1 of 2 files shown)


.github/workflows/format.yaml
```diff
@@ -57,8 +57,11 @@ jobs:
     - name: Installing formatting tools
       run: |
         sudo apt-get update
-        sudo apt-get install clang-format
+        sudo apt-get install clang-format-10
         pip install pyformat
+        # Use clang-format 10 as default until we choose to migrate to 11
+        sudo update-alternatives --remove-all clang
```
matthiakl, line 63: I think you should either remove clang-format only or install the clang/clang++ links as well. Not sure whether clang-format calls clang or not.

matthiakl (approved):
That should do it for now

PR #4618, opened 2021-01-11
Ai production support

hessenfarmer:
finishes the work of https://github.com/gunchleoc/widelands/tree/ai-production-support
- introduce number_of_same_nearby to identify nearby b uildings of same type
- if upgrade building needs supporters take this into consideration
- fixes some values in training to make the ai train amazons
- minor updates of basic amounts and other lua values
- all fixes of GunChleoc that enable the detection of supported buildings and their supporters from work programs
- Gunchleoc concept of having multiple rangers

Re. #3989
Re. #4271

Supersedes #3999
Supersedes #4013
Supersedes #4465


src/ai/defaultai.cc
```diff
@@ -748,10 +740,44 @@ void DefaultAI::late_initialization() {
...
 				bo.set_is(BuildingAttribute::kSupportingProducer);
-			}
-
+				/* Buildings detected at the time of writing:
+				 *
+				 *   amazons_rare_tree_plantation -> amazons_liana_cutters_hut
+				 *                                -> amazons_rare_tree_cutters_hut
+				 *                                -> amazons_wilderness_keepers_tent
+				 *
+				 *   amazons_rare_tree_cutters_hut, amazons_woodcutters_hut NOCOM removed, they are
```
Noordfrees: There are a lot of NOCOMs in those comments (more below), should be cleaned up

src/ai/defaultai.cc
```diff
@@ -748,10 +740,44 @@ void DefaultAI::late_initialization() {
...
+
+			bo.supported_producers.clear();
+			for (const std::string& supported_building_name : prod.supported_productionsites()) {
+				Widelands::DescriptionIndex supported_building_index =
+				   tribe_->building_index(supported_building_name);
+				bo.supported_producers.insert(std::make_pair(
+				   supported_building_index, dynamic_cast<const Widelands::ProductionSiteDescr*>(
+				                                tribe_->get_building_descr(supported_building_index))));
+			}
+			// TODO(hessenfarmer): we need to fiind a solution for AMAzons here as they do not mine
```
Noordfrees: typos

src/ai/defaultai.cc
```diff
@@ -7007,7 +7223,7 @@ void DefaultAI::print_stats(const Time& gametime) {
 		   static_cast<uint32_t>(mines_.size()),
 		   static_cast<uint32_t>(warehousesites.size() - num_ports), num_ports);
 	}
-	if (false) {  // NOLINT
+	if (true) {  // NOLINT
```
Noordfrees: revert

Noordfrees (commented):
Code LGTM, not tested yet

src/ai/defaultai.h
```diff
@@ -375,10 +380,16 @@ struct DefaultAI : ComputerPlayer {
 	// id of iron as resource to identify iron mines in mines_per_type map
 	int32_t iron_resource_id = Widelands::INVALID_INDEX;
 
+	// Record with immovable attribute is created or collected by which building names and attributes
+	std::map<Widelands::MapObjectDescr::AttributeIndex, std::set<ImmovableAttribute>>
+	   buildings_immovable_attributes_;
```
tibor95, line 385: I must say the variable name is difficult to understand (what it is about) also description above is not very helpful.

hessenfarmer: ok I will try to find a better name then.

tibor95: It would be good. I am trying to read the code and all the time I just stuck and try to guess what a variable can contain. 
When I look at ImmovableAttribute, then building_immovable_attributes is something like:
<Building index: \<Building name : Building Attribute>> 
correct?

hessenfarmer: Well, I admit I had difficulties to understand all of this when I tried to understand the changes of gunchleoc. but as my changes are about 2 month old (hoped she could incorporate them by herself) I need to dig into this by myself again. 
I believe to remember this is an index (a set) of all immovable attributes that a building is interacting with.
This means really the attributes not the immovables itself I think

tibor95: Because building name vs building index is 1:1, so this does not bring any new information to have both. But I can admit that there may be reasons that this approach is the best

hessenfarmer: > 
> When I look at ImmovableAttribute, then building_immovable_attributes is something like:
> <Building index: <Building name : Building Attribute>>
> correct?

No. 
Instead it is 
AttributeIndex: (Building name : Building Attribute)
where Attribute index is the index of an immovable attribute that is created, collected or needed by the building. eg the index of the attribute "tree" or the attribute "pond_dry"
Building Attribute is something like kNeedsRocks

hessenfarmer: How about renaming it to buildings_per_immovable_attributes_ as it lists the buildings (and their building attributes) for each immovable attribute they interact with

tibor95: the word "attribute" seems not good to me. It mean ~'map object', 'collectable object' or 'surface resource', doesnt it? Tings like "colected_attributes" might be renamed to collected_map_objects or so.

Also do we need Building Attribute in buildings_immovable_attributes_, because its "attribute" and "building attribute" refers to the same things berry <-> kNeedBerry.

What about "collectable_to_collectors_map" or "buildings_per_collected_objects" or similar (instead of buildings_immovable_attributes_)?

I know that it is only cosmetic but it would increase readability of code IMHO.

hessenfarmer: for the first question we need the building attribute and the name as with these changes we have multiple rangers (currently only for the amazons: jungle_preserver and rare_tree_plantation) and we can have multiple other types theoretically (currently a warning is thrown if we have multiple of them) so sometimes in the code we use the name and sometimes the attribute.

having collect in the variable name isn't recommendable as not only collected immovables are listed but also created immovables (i.e. we have immovables with the attribute sapling where the normal ranger or the amazon rare tree plantationb is assigned, we have the attrib pole where amazon ranger is assigned, ) furthermore even immovables that are only needed and neither collected or created are to be recorded there (eg. the amazon wilderness_keeper needs a tree to free game)

tibor95: I see that so much stuff was changed since I worked on AI, I would perhaps need to put it on paper to understand it properly. But makes more sense now when you explained it to me, thanks

hessenfarmer: no worries. take your time. However all I tried in this branch was fixing some logical errors in GunChleocs attempt to read some former AI hints from the production programs and make them work properly. for some particular changes it might be worth reading my comments on the original branch (#4465)

src/ai/defaultai.cc
```diff
@@ -850,31 +857,76 @@ void DefaultAI::late_initialization() {
...
+					if (tribe_->is_construction_material(output_idx)) {
+						produces_construction_material = true;
+					} else {
+						produces_non_construction_material = true;
+					}
+				}
+				if (produces_construction_material) {
+					// TODO(GunChleoc): We should lose the hard distinction between quarry and
+					// lumberjack, so that a building can be both
+					if (prod.supported_by_productionsites().empty()) {
```
tibor95, line 876: Why dont we unify this logic below, I mean match attribute from collected_attributes directly to (strings?) rocks, logs and berries. Right now here is rather a indirect guessing than direct identification of what collected_attribute it is.

Also prod.collected_attributes can return more than one item, correct? Can one building collect more than one type of objects from map?

hessenfarmer: for the first question: this was designed by @GunChleoc so I cn only tell what might have been the reason. I guess it was because this would result in some hardcoding again while this attempt should work without hardcoding the relations you mentioned. 
Second question: yes there are some buildings: amazons huntergatherer collects a ressource (fish) and bobs (game).  Frisians charcoal burner hut "collects" empty ponds and ponds with charcoal.

tibor95: I must say I am quite lost in this logic. This is also because I do not know new tribes.

hessenfarmer: you can have a look in GunChleocs original branch where there was some debug code to print the buildings identified. I removed this for review however this could give you a good impression of this logic

Noordfrees (commented):
From my side this is good to go if @tibor95 is also fine with this

hessenfarmer:
I really would like this to be in 1.0 as it is the last of a series of changes to get rid of AI hints that can be read from the lua definitions. As the others already got merged master is now in an intermediate situation where some AI Information is read from the programs and other obtainable information needs to be defined in AI hints in addition.

hessenfarmer:
It would be good if someone could at least confirm that this solution is not performing worse then actual trunk.      
It can be tested in a pure AI Game with all 5 tribes on a balanced 5 player map (pure AI is a local Lan game with 5 AI and the tester as spectator).  
Points to watch for is the relation of woodcutters and the respective foresters (especially for amazons)
other pairs of buildings that are dependent (especially frisian berry collectors and planters)

As this was the last of a series of refactorings started by @gunchleoc it really would make sense to have this in the upcoming release.

klaus-halfmann:
OK, I will spend some CPU Cycles on this one.

klaus-halfmann (commented):
Testing on Astoria with all 5 tribes:

The Amazones AI builds a watergatherer with now reachable water, can this be fixed here?

hessenfarmer:
> 
> 
> Testing on Astoria with all 5 tribes:
> 
> The Amazones AI builds a watergatherer with now reachable water, can this be fixed here?

will have a look. do you have a replay available?

klaus-halfmann:
Here the first Part[2021-02-22T11.07.57.zip](https://github.com/widelands/widelands/files/6022158/2021-02-22T11.07.57.zip)
The Amazones perform very good Military wise. ATL have a very poor performance.

klaus-halfmann (commented):
I got a heap-buffer-overflow  when quitting them game after perhaps 8 hours of Gametime 
[Crash.txt](https://github.com/widelands/widelands/files/6023681/Crash.txt)
no idea if this is related

Noordfrees:
looks like #4645

klaus-halfmann:
OK here is the Replay, its so big that I had to upload it to some cloud storage:
https://www.magentacloud.de/share/ygwso1qmey
Do we need an addtional savegame, too?

hessenfarmer:
> 
> 
> Here the first Part[2021-02-22T11.07.57.zip](https://github.com/widelands/widelands/files/6022158/2021-02-22T11.07.57.zip)
> The Amazones perform very good Military wise. ATL have a very poor performance.

Thanks for the replays. The weaker performance of Atlanteans is somewhat dependent on them not having any gold in their starting area. They have to expand relatively far to reach some gold. But in general they are performing a bit low in the beginning due to their massive need for logs. but this should settle down after a while and then they are performing better.

hessenfarmer:
> Testing on Astoria with all 5 tribes:
> 
> The Amazones AI builds a watergatherer with now reachable water, can this be fixed here?       
     
Ok this is due to the water gatherer being handled like a normal well for the moment. This is the same as in current trunk. It can be and should be fixed but this would make this complicated diff even more complicated so I would prefer doing it in a separate branch after this gets merged
The change would include not handle it as a well and modify the check for the amount of wells to be defined for Ai to <=1 instead of ==1. Furthermore it would need evaluating the AI hint needs_water properly (this isn't taken into account currently which could irritate tribe designers in other occasion as well.

matthiakl:
I noticed, that in all tribes the wood production is stalled at some point. Frisians and Amazons have too few woodcutters, while the others lack foresters. In comparison to master I get no military interactions by the end of the game, while on master one player already is defeated at 3h.
Bildschirmfoto vom 2021-02-24 21-59-46
Bildschirmfoto vom 2021-02-24 22-00-53
Here is the [replay](https://github.com/widelands/widelands/files/6038917/replay.zip)
I used the same settings as in #4699, where you can compare the performance on master.

hessenfarmer:
@matthiakl 
Many thanks for testing this. However as the performance of the AI is not deterministic it might be the case that you expereinced one of very low performance in wood buildings placement. For example I did a run in your setup myself and wood was not the problem but amazons lacked water badly. But nevertheless all tribes (except frisians due to the setup) went into battle after around 3 hours.   
Problems I noticed which might not be related to this branch was that all but atlanteans build only one stonemine which is especially bad for amazons if there is no other stone available anymore. 

to keep it short: May I ask you for doing some more runs in the same setup to have a better comparison. I will do so for myself as well.

matthiakl:
> Problems I noticed which might not be related to this branch was that all but atlanteans build only one stonemine which is especially bad for amazons if there is no other stone available anymore.

This is also only a non deterministic appearance. I did not experience it and the amazons build up to three stone mines.

> May I ask you for doing some more runs in the same setup to have a better comparison. I will do so for myself as well.

I did some more runs in both master and this branch. Without knowing what to watch, I would say the performance is equal; sometimes an AI gets stuck as I mentioned earlier, but this appeared in both branches.
When I pay attention on the wood economy I would say the trees now are used more efficiently than in master (i.e. amount of woodcutters and foresters match better).

hessenfarmer:
> 
> 
> > Problems I noticed which might not be related to this branch was that all but atlanteans build only one stonemine which is especially bad for amazons if there is no other stone available anymore.
> 
> This is also only a non deterministic appearance. I did not experience it and the amazons build up to three stone mines.

ok good to know it is nothing systematic

> 
> > May I ask you for doing some more runs in the same setup to have a better comparison. I will do so for myself as well.
> 
> I did some more runs in both master and this branch. Without knowing what to watch, I would say the performance is equal; sometimes an AI gets stuck as I mentioned earlier, but this appeared in both branches.

then I am happy as this branch should achieve exactly the same as in master without the need of defining some Ai hints anymore. 

> When I pay attention on the wood economy I would say the trees now are used more efficiently than in master (i.e. amount of woodcutters and foresters match better).

that might be the case indeed as in this area there was alittle tweak in the code.

matthiakl:
So from the functionality this should be good? I can't say anything about the code, because I hardly understand the whole AI stuff.

Noordfrees:
Watched a few more AI-only matches with this branch. Amazons always built their first water gatherer very late (~2 hours) resulting in a poor economy start. I have the impression that it is not that bad in master (though this might be just chance, I did not watch that many matches). Do you want to look into this here? 
Other than that all seems fine.

hessenfarmer:
> Watched a few more AI-only matches with this branch. Amazons always built their first water gatherer very late (~2 hours) resulting in a poor economy start. I have the impression that it is not that bad in master (though this might be just chance, I did not watch that many matches). Do you want to look into this here?
> Other than that all seems fine.

I will cjeck this myself and then I'll decide what would be best. As far as I can tell nothing has chnaged in wells handling. But other buildings might have higher priorities now.

hessenfarmer:
> 
> 
> Watched a few more AI-only matches with this branch. Amazons always built their first water gatherer very late (~2 hours) resulting in a poor economy start. I have the impression that it is not that bad in master (though this might be just chance, I did not watch that many matches). Do you want to look into this here?
> Other than that all seems fine.

Did a quick check with 5 Amazons on stoneriver the first one completed a watergather after approx. 25 minutes. All AI had a watergatherer started at about 35 min but some were lacking rubber to complete it. 
Would beglad to see a replay, else I would say no need to change something for watergatherers

Noordfrees:
[replays.zip](https://github.com/widelands/widelands/files/6097843/replays.zip)
Player 2 builds the first water gatherer after 24 minutes but the place could hardly be chosen worse. Player 1 builds his first at 1:18 hours. It is especially strange that the extreme water shortage does not induce the AI to build more for some hours, and then in very poorly chosen places.

hessenfarmer:
> 
> 
> [replays.zip](https://github.com/widelands/widelands/files/6097843/replays.zip)
> Player 2 builds the first water gatherer after 24 minutes but the place could hardly be chosen worse. Player 1 builds his first at 1:18 hours. It is especially strange that the extreme water shortage does not induce the AI to build more for some hours, and then in very poorly chosen places.      
    
the places are choosen poorly, because the Ai code can't handle them properly yet but this would make this PR even more complicated. It is treated the same way as a normal well however the Ai dismantle them if they are completely unproductive after a while.
We could try to make water more precious for amazons and hope they get a better rating and will be a bit more preferred. However AI games often end with amazons winning already so I believe this is not really necesary for v1.0 cause it would need a lot of testing to not overdo this.
After this PR gets merged, I will try to implement an own handling for the amazon water gathering in the next version.    

Edit: the poor handling is the same in trunk as well
Edit 2: in your replay Player 1 has 2 water gatherer at around 0:35 directly south of the hQ it has a first productive one at 1:00 north of the lake and strts the 4th at 1:18. So apart from the placement which is a known weakness it is doing ok I would say.
Furthermore Crater is hard for the Ai as it has very early contact and still I believe an Amazonian Ai can be a challenge for a not so experienced player.   
So I really think there is no particular issue with the watergatherer that we should take care of in this branch. If you would deem it necessary to fix the placement for 1.0 I would prioritize this in my contribution effort.

Noordfrees:
Okay let's keep it like it is then for 1.0. Code LGTM, though I don't know the AI code very well. Do you want to wait for someone else to do a code review? If not feel free to merge :) 

---

By the way, the amazons AI is performing extremely poorly in my fri04 branch, could you perhaps have a look at that as well when you have time?

hessenfarmer:
> 
> 
> Okay let's keep it like it is then for 1.0. Code LGTM, though I don't know the AI code very well. Do you want to wait for someone else to do a code review? If not feel free to merge :)

The problem is there is hardly anybody else then @tibor95 who could do this probably. So if it is ok for both of you I'll merge this, so that Tibor can do some AI training afterwards. and we postpone the amazon_water_gatherer fix to v1.1. Would be fine if you could confirm this by approving the branch or writing back.   

> 
> By the way, the amazons AI is performing extremely poorly in my fri04 branch, could you perhaps have a look at that as well when you have time?  

Of course. I'd love to. I will pull thge branch and see what I can do about. But I think this has to wait as I think I'll focus myselkf in bugfinding and fixing for the v1.0 first

Noordfrees (approved):

PR #4619, opened 2021-01-12
Fix S2 map loader to load stones correctly

hessenfarmer:
I belive this was just a small typo that prevented Stones being present in S2 maps
probably fixes the issue reported on the forum
https://www.widelands.org/forum/topic/5100/

### DIFF
--- a/src/map_io/s2map.cc
+++ b/src/map_io/s2map.cc
@@ -573,8 +573,8 @@ void S2MapLoader::load_s2mf(Widelands::EditorGameBase& egbase) {
 	//   owner == 4 -> grey
 	//   owner == 6 -> green
 	//   owner == 6 -> orange
-	// TODO(hessenfarmer): this comment is not correct section 6 determines how the values of section 5
-	// are to be interpreted. Solution is working more the less though
+	// TODO(hessenfarmer): this comment is not correct section 6 determines how the values of section
+	// 5 are to be interpreted. Solution is working more the less though
 	section = load_s2mf_section(fr, mapwidth, mapheight);
 	if (!section) {
 		throw wexception("Section 6 (Ways) not found");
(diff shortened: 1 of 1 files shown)


Noordfrees (commented):
Could you also add another two `case` clauses for water and fish? According to the comments above those are also available in s2 maps but we don't seem to load them

hessenfarmer:
> 
> 
> Could you also add another two `case` clauses for water and fish? According to the comments above those are also available in s2 maps but we don't seem to load them

I was thinking about this. However from the documentation I found for the s2 maps (link inside a comment of s2map.cc) fish and water are infinite in s2, so we would need to add default values for the amount. But this is what our map loader in the game does anyway so this info is only missing in the editor not in the map itself. If we set them to one default value this would override our terrain dependent default values in the game. So I am undecided about this. 
Furthermore I need to think about scaling the amopunt of ressources avaiulable in s2 (1-7) to our max ressopurce value, but there I am undecided as well. 
This would need some testing which I am planning to do tonight.

Noordfrees:
> > Could you also add another two `case` clauses for water and fish? According to the comments above those are also available in s2 maps but we don't seem to load them
> 
> I was thinking about this. However from the documentation I found for the s2 maps (link inside a comment of s2map.cc) fish and water are infinite in s2, so we would need to add default values for the amount. But this is what our map loader in the game does anyway so this info is only missing in the editor not in the map itself. If we set them to one default value this would override our terrain dependent default values in the game. So I am undecided about this.

Good point, let's keep it as it is then. Please add a comment why we can ignore this info on loading though

> Furthermore I need to think about scaling the amopunt of ressources avaiulable in s2 (1-7) to our max ressopurce value, but there I am undecided as well.

That's already done below in line 747f (though it uses a hardcoded scaling factor instead of the ResourceDescr's max amount)

Noordfrees (approved):
LGTM (don't have an s2 map to test with)

hessenfarmer:
tested it successfully so merging thanks for the review

PR #4625, opened 2021-01-14
empire 4 script fixes

hessenfarmer:
this fixes the resource names in the starting script which removes the default water and fish from the map to make the small buildings empty. 
furthermore one of the original mines could not be placed anymore due to new mine spots calculation. so this was adapted to avoid a no space error message in the users inbox


Noordfrees (approved):
LGTM & works fine.
I now get a `WARNING: Training site 'empire_trainingcamp1' is lacking its 'messages' table` do you want to fix this here as well?

hessenfarmer:
Done.
Thanks for the review

PR #4633, opened 2021-01-18
Show initial status message for win conditions

### LINKED ISSUE #4024: Show initial status message for win conditions
We reduced the number of win condition messages to make them less noisy. This means that we now don't have an initial message, which makes the win conditions harder to understand.

So, there should be 1 additional status message at 0:00 gametime.

matthiakl:
Fixes #4024


Noordfrees (approved):
LGTM, tested and working

PR #4634, opened 2021-01-19
v1.0 Maps Review

### LINKED ISSUE #967: Desert tournament is unfair
This map has a couple of issues:

- The left two sides seem to have more space then the right two sides.
- They also seem to have less trees to get to the middle.
- Most importantly, the yellow player cannot build a big building in the middle, while all other players can.

Generically speaking, the map is poorly designed: Choke points are not a good map element in Widelands. The concept of a "save starting island" for each player paired with a center place were everybody meets will give long drawn out games. A Widelands map should have at least two - three meeting points for each player or even be very wide open.

 * date created: 2015-02-19T07:28:53Z
 * owner: sirver
 * the launchpad url was https://bugs.launchpad.net/bugs/1423464

Noordfrees:
https://www.widelands.org/forum/topic/5112

- New map *Lesser Ring* by PkK
- Updated map *Wide World*
- Deleted map *Sun of Fire* (it is now available in an add-on "Formerly Official Maps")
- Marked *Desert Tournament* as unbalanced and renamed it to a name that does not imply any sort of balance (this closes #967)


frankystone (approved):
LGTM, not tested

PR #4653, opened 2021-02-05
Fix statistics string cache

### LINKED ISSUE #4537: Wrong militarysite soldier statistics strings
In this screenshot the milsite displays the wrong soldier info: It says there are 0 soldiers present but in fact 1 soldier is inside.
shot0002
[Savegame](https://github.com/widelands/widelands/files/5603589/soldierstatsbug.wgf.zip)

matthiakl:
Fixes #4537

### DIFF
--- a/src/logic/map_objects/tribes/militarysite.cc
+++ b/src/logic/map_objects/tribes/militarysite.cc
@@ -397,8 +397,7 @@ void MilitarySite::update_statistics_string(std::string* s) {
 					cr->push_arg(capacity_);
 					cr->resume();
 					std::string new_string = cr->pop_string();
-					statistics_string_cache_[idx].insert(
-					   std::make_pair(cache_key, new_string));
+					statistics_string_cache_[idx].insert(std::make_pair(cache_key, new_string));
 					return new_string;
 				} catch (LuaError& err) {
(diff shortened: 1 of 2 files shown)


src/logic/map_objects/tribes/militarysite.cc
```diff
@@ -396,7 +396,8 @@ void MilitarySite::update_statistics_string(std::string* s) {
 					cr->push_arg(capacity_);
 					cr->resume();
 					std::string new_string = cr->pop_string();
-					statistics_string_cache_[idx].insert(std::make_pair(stat, new_string));
+					statistics_string_cache_[idx].insert(
+					   std::make_pair((pres << 8) + (stat << 4) + capacity_, new_string));
```
Noordfrees: Good catch :) 
But wouldn't milsites with large soldier capacities still show the bug if the capacity can occupy more than four bits? Perhaps use std::tuple<int,int,int> as key instead of plain int to ensure all keys are unique.

matthiakl: I thought the current milsites don't have more than 12 soldiers. But for future expansions (and Addons) you are right.

Noordfrees (approved):
Fix confirmed, code LGTM :)

PR #4658, opened 2021-02-05
Fix LAN shortcut in main menu

Noordfrees:


hessenfarmer (approved):
Code looks like the rest aka good.
tested and working

PR #4660, opened 2021-02-06
Select text via mouse

jmoerschbach:
select text with mouse (left button must be pressed while cursor is moved)
also fixes a  not-yet reported bug regarding selecting text and then clicking with mouse in master -> cursor is repositioned but selection is still active


matthiakl (approved):
Tested and working. Nice and simple.

PR #4667, opened 2021-02-09
Generalize addon packager

matthiakl:
Split the addon packager into logic and UI parts and allow integration of other addon categories.
It now also supports theme and campaign addons by automatically including template files.
Other per category actions can be added easily.

(Sorry @Noordfrees for overhauling your code, but after the addons_packager.cc went into master I had the desire to untangle it)


Noordfrees:
Gave this a quick test – very good :) 

The autogenerated campaings.lua could use some improvements. How about generating this via C++ with a FileWrite instead of copying a template file? Then the textdomain and the campaign's name and description can already be set to default values taken from the add-on's data, and the scenarios table can be filled in with all maps in the add-on with the correct add-on prefixes and paths.

Writing a new theme add-on causes Widelands to hang for quite a while. The first time I wondered whether there was an endless loop somewhere, before it finally completed. A progress bar to indicate the number of files written and remaining would be nice if possible.

matthiakl:
Both implemented. A dynamic progress bar would be too complex for this, I simply added a short waiting info. 
The campaigns.lua is now filled during the first save and allows to directly use the addon.

src/logic/mutable_addon.cc
```diff
@@ -0,0 +1,307 @@
...
+	case AddOnCategory::kCampaign:
+		return std::unique_ptr<CampaignAddon>(new CampaignAddon(a));
+	case AddOnCategory::kWinCondition:
+		return std::unique_ptr<WinCondAddon>(new WinCondAddon(a));
+	case AddOnCategory::kStartingCondition:
+		return std::unique_ptr<StartingCondAddon>(new StartingCondAddon(a));
+	case AddOnCategory::kTheme:
+		return std::unique_ptr<ThemeAddon>(new ThemeAddon(a));
+	default:
+		return std::unique_ptr<MutableAddOn>(new MutableAddOn(a));
```
Noordfrees: Does this ever happen? Better throw an exception "invalid category %u"

src/logic/mutable_addon.cc
```diff
@@ -0,0 +1,307 @@
...
+	return requires;
+}
+
+bool MutableAddOn::write_to_disk() {
+	// Step 1: Gather the requirements of all contained maps
+	std::string requires = parse_requirements();
+
+	// Step 2: Create the `addon` file.
+	g_fs->ensure_directory_exists(directory_);
+	// Write profile
```
Noordfrees: Add a newline and prefix the comment with "Step 3:"
Also, no need to enclose the profile code in {} now.

Noordfrees (commented):
LGTM, just two comments inline. And one other thing, the tribe dropdown for campaigns is in a weird place IMHO. Perhaps move it below the dirstruct list and turn it into a regular textual dropdown? The Create Directory should also be visible in the campaign creator for consistency, no reason to forbid this.

src/ui_fsmenu/addons_packager_box.cc
```diff
@@ -0,0 +1,335 @@
...
+                   "dropdown_tribe",
+                   0,
+                   0,
+                   50,
+                   8,
+                   kButtonSize,
+                   _("Tribe"),
+                   UI::DropdownType::kTextual,
+                   UI::PanelStyle::kWui,
+                   UI::ButtonStyle::kWuiSecondary) {
```
Noordfrees:
```suggestion
                   UI::PanelStyle::kFsMenu,
                   UI::ButtonStyle::kFsMenuSecondary) {
```

Noordfrees:
The `campaigns.lua`'s `scenarios` section now does not list maps in subdirectories.

---

And something else I thought of now, how about extending the template file with a difficulty stages template like this?
```
         difficulties = {
            -- TRANSLATORS: A campaign difficulty (easiest of three options)
            _("Easy"),
            -- TRANSLATORS: A campaign difficulty (medium of three options)
            _("Medium"),
            -- TRANSLATORS: A campaign difficulty (hardest of three options)
            _("Hard"),
         },
```
If you want to be fancy there could even be a multiline-editbox or such where each line corresponds to one autogenerated difficulty level.

matthiakl:
> The campaigns.lua's scenarios section now does not list maps in subdirectories.

There I was a bit to fast with committing the code :)

I added the difficulties to the template. The in-game editing I would like to do in a separate branch next and do it so, that an existing campaign can be modified, as I currently can only insert data once into the template.

Noordfrees (approved):
Great work :)

PR #4678, opened 2021-02-13
Campaign editing

matthiakl:
Extends the addon packager for basic configuration of campaigns.
Also fixes a bug, where editbox tooltips were only shown on click.


src/logic/mutable_addon.cc
```diff
@@ -245,15 +268,44 @@ bool MapsAddon::write_to_disk() {
...
+
 	return true;
 }
 
+CampaignAddon::CampaignAddon(const AddOnInfo& a)
+   : MapsAddon(a),
+     // (?:^|$|\\r\\n|\\n|.) is the c++ equivalent for multiline regex matching of .
+     // Find difficulties.descname
+     rex_difficulty_("(difficulties(?:^|$|\\r\\n|\\n|.)*descname\\s*=\\s*_\")(.*)(?=\"(?:^|$|"
+                     "\\r\\n|\\n|.)*image)"),
```
Noordfrees, line 281: Wow, regex magic :) I didn't check these thoroughly but it seems to work

src/ui_fsmenu/addons_packager.cc
```diff
@@ -449,6 +449,7 @@ void AddOnsPackager::clicked_write_changes() {
 	for (const auto& str : addons_with_changes_) {
 		msg = (boost::format(_("%1$s\n· %2$s")) % msg % str.first).str();
 	}
+	msg += _("\n\nBe careful with this, if you have modified addon files yourself!");
```
Noordfrees: Should be more precise, e.g.
```suggestion
	msg += _("\n\nWarning: If you manually edited any files in this add-on, your changes may be overwritten!");
```

src/ui_fsmenu/addons_packager_box.cc
```diff
@@ -308,29 +312,73 @@ CampaignAddOnsPackagerBox::CampaignAddOnsPackagerBox(MainMenu& mainmenu, Panel*
...
+	short_desc_.changed.connect([this]() { edited(); });
+	maps_box_.set_modified_callback([this]() { edited(); });
+
+	difficulty_.set_tooltip(_("The campaigns difficulty. One word."));
+	short_desc_.set_tooltip(_("Short description, which will be appeded to the difficulty."));
+
+	desc_tribe_box_.add(
+	   new UI::Textarea(&desc_tribe_box_, UI::PanelStyle::kFsMenu,
+	                    UI::FontStyle::kFsMenuInfoPanelHeading, _("Difficulty:"), UI::Align::kRight),
+	   UI::Box::Resizing::kFullSize);
```
Noordfrees: Using `kFullSize` or `kExpandBoth` for Textareas in a horizontal box causes the text to be vertically off-center. You probably need to use `kAlign`-`kCenter` alignment.

Noordfrees (commented):
Working well, a few nits inline.

The difficulty image needs to be adjusted manually. It seems unexpected that I can set the difficulty to "Super-hard" and the image defaults to Easy. How about a dropdown next to the editboxes where the user gets to choose between the four difficulty icons?

Also the line seems too cramped at low resolutions, please pull out the tribe dropdown as a separate line and align the left difficulty-editbox with the four editboxes above.

Noordfrees:
Very good :)
Three more nits:
- Typo in the right editbox's tooltip: appeded → appended
- The ordering of the difficulties in the dropdown is alphabetical which makes no sense here. Hardcode the order as Easy-Medium-Hard-Challenging please.
- Feature suggestion: When selecting an entry from the difficulty dropdown and the text in the left editbox is identical to the previously selected difficulty's name, replace it with the new difficulty's name.

Noordfrees (commented):
Excellent :)
Just one more thing, when selecting an entry in the difficulty icon dropdown the text should also be updated if the left editbox is currently empty.

Noordfrees (approved):

matthiakl:
Thanks for the review :)

Issue #4258, opened 2020-09-11
Sound Effect Disconnect


josephomorrow:
After saving the game, SoundHandler errors begin.

**To Reproduce**
1. Begin campaign
2. Run follow some instructions
3. Save game
4. Game sounds stop, while console screen continuous scrolls error messages

**Expected behavior**
Nothing should change after game saved.

**Screenshots**
n/a

**Version:**
 - Fedora 32
 - Widelands Version: 1.0~git1[c3fc079@master] (debug)

**Additional context**
Partial readout from console after saving game:

[00:06:41.974 real] ERROR: SoundHandler: Sound effect 7 does not exist!
[00:06:41.977 real] ERROR: SoundHandler: Sound effect 7 does not exist!
[00:06:41.977 real] ERROR: SoundHandler: Sound effect 6 does not exist!
[00:06:41.977 real] ERROR: SoundHandler: Sound effect 2 does not exist!
[00:06:41.979 real] ERROR: SoundHandler: Sound effect 1 does not exist!
[00:06:41.979 real] ERROR: SoundHandler: Sound effect 1 does not exist!
[00:06:41.979 real] ERROR: SoundHandler: Sound effect 2 does not exist!
[00:06:41.979 real] ERROR: SoundHandler: Sound effect 1 does not exist!
[00:06:41.979 real] ERROR: SoundHandler: Sound effect 1 does not exist!
[00:06:41.979 real] ERROR: SoundHandler: Sound effect 1 does not exist!
[00:06:41.979 real] ERROR: SoundHandler: Sound effect 1 does not exist!
[00:06:41.979 real] ERROR: SoundHandler: Sound effect 7 does not exist!
[00:06:41.979 real] ERROR: SoundHandler: Sound effect 1 does not exist!
[00:06:41.979 real] ERROR: SoundHandler: Sound effect 1 does not exist!
[00:06:41.979 real] ERROR: SoundHandler: Sound effect 1 does not exist!
[00:06:41.979 real] ERROR: SoundHandler: Sound effect 1 does not exist!
[00:06:41.979 real] ERROR: SoundHandler: Sound effect 1 does not exist!
[00:06:41.979 real] ERROR: SoundHandler: Sound effect 1 does not exist!
[00:06:41.979 real] ERROR: SoundHandler: Sound effect 1 does not exist!
[00:06:41.979 real] ERROR: SoundHandler: Sound effect 1 does not exist!

gunchleoc:
[...]

PR #4201, opened 2020-09-01
Map preview on selection screen

matthiakl:
Fixes #1913 
In the map selection screen a minimap will be shown below the description. In order to support older maps which do not already include an image file, the map is rendered dynamically. For faster rendering time, only the bare minimum of the map is loaded.

Also fixes an error in UI::Icon, where the image was scaled twice, if it does not fit.

Some possible additions:
Is it desirable to show starting positions (as mentioned in the issue) or other additional information?
Should other map selection screens (save loading, campaign selection) use a dynamic minimap?

Issue #4509, opened 2020-11-11
Map is loaded on Load Map screen before pressing OK


Niektory:
Selecting a map in the "Load Map" window in the editor loads it immediately (without pressing OK).

It can also lead to a crash if a minimap window is opened at the time:

```
widelands: ../src/graphic/texture.cc:244: void Texture::set_pixel(uint16_t, uint16_t, const RGBAColor&): Assertion `x < width()' failed.

Thread 1 "widelands" received signal SIGABRT, Aborted.
0x00007ffff4f4a615 in raise () from /usr/lib/libc.so.6

(gdb) bt
#0  0x00007ffff4f4a615 in raise () from /usr/lib/libc.so.6
#1  0x00007ffff4f33862 in abort () from /usr/lib/libc.so.6
#2  0x00007ffff4f33747 in __assert_fail_base.cold () from /usr/lib/libc.so.6
#3  0x00007ffff4f42bf6 in __assert_fail () from /usr/lib/libc.so.6
#4  0x000055555663b277 in Texture::set_pixel (this=0x606001ebea60, x=256, y=128, color=...) at ../src/graphic/texture.cc:244
#5  0x0000555556ea27c7 in (anonymous namespace)::do_draw_minimap (texture=..., egbase=..., player=0x0, top_left=..., 
    layers=(MiniMapLayer::Terrain | MiniMapLayer::Owner | MiniMapLayer::Flag | MiniMapLayer::Road | MiniMapLayer::Building | MiniMapLayer::ViewWindow), 
    draw_full=false, rows_drawn=0x61f0001a0b40) at ../src/graphic/minimap_renderer.cc:220
#6  0x0000555556ea38bd in draw_minimap_static (texture=..., egbase=..., player=0x0, 
    layers=(MiniMapLayer::Terrain | MiniMapLayer::Owner | MiniMapLayer::Flag | MiniMapLayer::Road | MiniMapLayer::Building | MiniMapLayer::ViewWindow), 
    draw_full=false, rows_drawn=0x61f0001a0b40) at ../src/graphic/minimap_renderer.cc:287
#7  0x00005555563e453e in MiniMap::View::draw (this=0x61f0001a0a38, dst=...) at ../src/wui/minimap.cc:66
#8  0x0000555556131143 in UI::Panel::do_draw_inner (this=0x61f0001a0a38, dst=...) at ../src/ui_basic/panel.cc:882
[...]

PR #4686, opened 2021-02-15
Do not instantiate egbase multiple times

### LINKED ISSUE #4676: SoundHandler: Sound effect xy does not exist!
When saving a map the console prints endlessly:
~~~~
[00:54:08.273 real] ERROR: SoundHandler: Sound effect 12 does not exist!
[00:54:08.273 real] ERROR: SoundHandler: Sound effect 11 does not exist!
[00:54:08.273 real] ERROR: SoundHandler: Sound effect 9 does not exist!
[...]
~~~~

**To Reproduce**
1. Start editor and load map Kings and Queens
2. Save it

I got this endless prints also if i had create a new map and placed some critters, but i can't reproduce it this way.

**Version:**
 - OS: archlinux
 - Widelands Version: [ddcc509@3849_win_conditions] (Debug)

**Additional context**
A similar bugreport was #4258 but this was after saving a game.

matthiakl:
Fixes #4676 
Re #4258
All caused by me instantiating multiple egbases in #4201

### DIFF
--- a/src/editor/ui_menus/main_menu_load_map.cc
+++ b/src/editor/ui_menus/main_menu_load_map.cc
@@ -52,10 +52,6 @@ void MainMenuLoadMap::clicked_ok() {
 		set_current_directory(mapdata.filename);
 		fill_table();
 	} else {
-		// Prevent description notes from reaching a subscriber
-		// other than the one they're meant for
-		egbase_.delete_world_and_tribes();
-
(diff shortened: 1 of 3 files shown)


Noordfrees (changes requested):
This regresses #4509. We need a separate egbase object here to store the temporary information.

matthiakl:
Indeed. I now understand what is going on during loading. I will look for a different fix. Thanks for the pointer.

PR #4695, opened 2021-02-20
Building animation for empire barracks by DragonAtma

Noordfrees:


frankystone (approved):
LGTM. Tested.

It would be nice to have the training dummy only visible if the trainer is inside the building...

PR #4703, opened 2021-02-23
No message box on "CLIENT_LEFT_GAME"

matthiakl:
Since #4692 a message box is shown even when a client voluntarily leaves a game setup resulting in a glitched UI.
So the showmsg parameter should be respected.


src/network/gameclient.cc
```diff
@@ -1217,6 +1212,15 @@ void GameClient::disconnect(const std::string& reason,
...
+	}
+
+	// TODO(Klaus Halfmann): Some of the modal windows are now handled by unique_ptr resulting in a
+	// double free.
+	if (d->modal) {
+		if (d->modal->is_modal()) {
+			d->modal->end_modal<FsMenu::MenuTarget>(FsMenu::MenuTarget::kBack);
+		} else {
 			d->modal->die();
 		}
```
Noordfrees:
```suggestion
		d->modal->die();
```
This can be simplified now because `die()` for modal panels simply calls `end_modal(kBack)`.

Noordfrees:
Tested and working :)

Noordfrees (approved):

PR #4705, opened 2021-02-24
discovery and new world start conditions for amazons

hessenfarmer:


Noordfrees:
Diff LGTM, will test this now

Noordfrees (approved):
Found both very easy to beat. The fact that amazons don't need to set up a full mining economy before they can start producing tools makes a huge difference compared to the other tribes' hard starts.

PR #4710, opened 2021-02-26
Reduce crashes from missing add-on dependencies

matthiakl:
- Disable the Ok button for maps, replays and saves when add-on requirements are not satisfied
- Refresh remote add-on list only every 10 minutes, to allow quick enabling/disabling


src/logic/addons.h
```diff
@@ -66,6 +66,7 @@ bool is_newer_version(const AddOnVersion& base, const AddOnVersion& compare);
 
 // Required add-ons for an add-on, map, or savegame with the recommended version
 using AddOnRequirements = std::vector<std::pair<std::string, AddOnVersion>>;
+using AddOnConflict = std::pair<std::string, bool>;
```
Noordfrees: This needs some documentation what the two parts mean, e.g.
```suggestion
using AddOnConflict = std::pair<std::string /* localized message */, bool /* critical */>;
```

src/network/net_addons.cc
```diff
@@ -93,7 +97,8 @@ static inline std::string get_addons_server_name() {
 }
 
 constexpr unsigned kCurrentListVersion = 3;
-std::vector<AddOnInfo> NetAddons::refresh_remotes() {
+constexpr uint32_t kRefreshPeriod = 600000;  // Only refresh every 10min
```
Noordfrees:
```suggestion
constexpr uint32_t kRefreshPeriod = 10 * 60 * 1000;  // Only refresh every 10min
```
Readability. Also add some blank lines please.

Noordfrees (commented):
I'm not sure about the caching of the remotes list. IMHO it would be better to fetch the list of add-ons afresh when selecting the Browse tab for the first time if the cache was being used until then. Whenever an add-on is being updated on GitHub, users will experience lots of checksum mismatches or 404 errors if they attempt to download or update something at the wrong time, and need to refresh the lists to fix this.
The Refresh Remotes button now does not do what the tooltip says it should. It should force an update when clicked.

src/ui_fsmenu/addons.cc
```diff
@@ -416,6 +416,11 @@ AddOnsCtrl::AddOnsCtrl(MainMenu& fsmm, UI::UniqueWindow::Registry& reg)
 		refresh_remotes();
 		tabs_.activate(1);
 	});
+	tabs_.sigclicked.connect([this]() {
+		if (remotes_.empty() && tabs_.active() == 1) {
```
Noordfrees:
```suggestion
		if (tabs_.active() == 1 && remotes_.size() <= 1) {
```
to retry more often if a network error occurred previously

Noordfrees (commented):
Better, though now there's a glitch: When you open the manager and click the Uninstall button for a local add-on, the confirmation dialogue will warn that the add-on can't be downloaded again. Not sure how this should best be addressed – do we need this warning or could it be removed again?

Noordfrees (approved):

PR #4712, opened 2021-02-27
Shorten the ripe stage of barley fields

Niektory:
This shortens the ripe stage of barley fields by half, from about 40 minutes to about 20 minutes.

See the forum discussion: https://www.widelands.org/forum/topic/4920/?page=1#post-34312


hessenfarmer:
it would be ok from my side however this should be reviewed by @Noordfrees, because he is the spiritus rector of this tribe

Noordfrees (approved):

PR #4713, opened 2021-02-27
Add an option for preserving wares

### LINKED ISSUE #3981: CTRL click behavior for enhancing/dismantling a building
When using CTRL click for enhancing or destroying a building, all wares in buildings stock are lost.

Change this so that 'preserve wares' is default when using CTRL click.

See discussion round about [this post](https://www.widelands.org/forum/post/33669/). At least for me this post is the most important argument to change this:

> Ctrl normally just skips the "are you sure" dialog in building interaction. so as the default settin in this dialog is preserve wares I would have expected that the result is the same as just clicking ok without any modification.
> So the current behaviour is not what I have expected due to long time experience in the game.
> 
> By this I strongly support to change this to keep the wares if ctrl + enhance or ctrl + dismantle is pressed.

**Version:**
 - OS: all
 - Widelands Version: build 21

Niektory:
Add an option to preserve wares if Ctrl is pressed when dismantling or upgrading a building.

Fixes #3981


matthiakl:
Better use `false` as default value, so the current behavior does not get changed. Otherwise it is confusing for those who used it.

hessenfarmer:
> 
> 
> Better use `false` as default value, so the current behavior does not get changed. Otherwise it is confusing for those who used it.

I still don't see the sense in defulting to loose the wares on ctrl click while the standard normal behaviour is that they are preserved (the tick in the box is defaulting to true)

Noordfrees:
IMHO yet another option is overkill for such a small feature. Why not just change the modifiers instead, Ctrl+click to discard wares and Ctrl+Shift+Click or GUI+Click to keep them?

hessenfarmer:
> 
> 
> IMHO yet another option is overkill for such a small feature. Why not just change the modifiers instead, Ctrl+click to discard wares and Ctrl+Shift+Click or GUI+Click to keep them?

As the problem (at least for me) is that the behaviour is not consistent with the normal route this would not solve the problem while an option does so I'll vote for the option or to implement the behaviour the other way round (ctrl click preserves the wares and ctrl shift click discards them)

matthiakl:
> I still don't see the sense in defulting to loose the wares on ctrl click while the standard normal behaviour is that they are preserved (the tick in the box is defaulting to true)

I agree with you, but I also understand that inverting the current behavior will confuse players who used it a lot in b21. Therefore an option should allow everyone to adapt to different play styles.

> IMHO yet another option is overkill for such a small feature. Why not just change the modifiers instead, Ctrl+click to discard wares and Ctrl+Shift+Click or GUI+Click to keep them?

Since we cannot get a consensus on this, a toggleable option can be a solution. Introducing another modifier shortcut would be confusing to me (are there currently any other Ctrl+*+Click?)

frankystone:
> > I still don't see the sense in defulting to loose the wares on ctrl click while the standard normal behaviour is that they are preserved (the tick in the box is defaulting to true)
> 
> I agree with you, but I also understand that inverting the current behavior will confuse players who used it a lot in b21. 

I guess there are not many players who used this feature. Probably most players did not notice that the wares are lost when CTRL-Clicking...

> Therefore an option should allow everyone to adapt to different play styles.

 An option is the same overkill than modifiers, imho.

Just make CTRL click the same behavior for dismantling/upgrading as we have for other windows where CTRL click is the same as when clicking OK, just without the confirmation dialogue. This is logical UI behavior, imho.

Niektory:
I picked `true` for the default because it seems it's what most people prefer and I believe is more intuitive. I don't think it matters much in any case, I'd be fine with `false` as the default as well.

Niektory:
I flipped the default to not preserve wares, so there is no change in behavior without changing the option.

frankystone:
Why the hell (sorry) do we need an option for this? I do not understand that.

Sorry, but sometimes i have the feeling that the opinion "the more options a game has, the better a game is" takes precedence. But this is not ever true. Some options are needed, but some are superfluous, and the one here will be one of them.

PR #4730, opened 2021-03-02
Update stats python3

### LINKED ISSUE #4729: Python error in translations update script
The translations update script no longer works for me (neither before nor after the changes of #4728 by which I tried to fix it). The step that fails is the `update_translation_stats` program.
```
Fetching translation stats .Something went wrong:
Traceback (most recent call last):
  File "utils/update_translation_stats.py", line 140, in main
    result = generate_translation_stats(po_dir, output_file)
  File "utils/update_translation_stats.py", line 64, in generate_translation_stats
    if 'ERROR' in stats_output:
TypeError: a bytes-like object is required, not 'str'
```

frankystone:
fixes #4729

At least this works locally here. I couldn't test the errors.

### DIFF
--- a/data/i18n/translation_stats.conf
+++ b/data/i18n/translation_stats.conf
@@ -1,5 +1,5 @@
 [global]
-total=68812
+total=68812 
 
 [ar]
 translated=8053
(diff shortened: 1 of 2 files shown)


Noordfrees (approved):
Tested and working. Thanks for fixing :) 

edit: no need to wait for the builds here

PR #4735, opened 2021-03-03
GitHub action for automatic translations updates

### LINKED ISSUE #4598: Prepare for clang-format 11

### LINKED ISSUE #4933: Add a check for developers.json
Every time a comma is missing, or a table entry ends with a comma, the website throws lots of server errors.

So a check for validity for this file would be fine.

The website code fails e.g. here https://github.com/widelands/widelands-website/blob/master/mainpage/views.py#L104

Noordfrees:
Suggestion: Instead of manually pulling and pushing the translations occasionally, how about having it done automatically by a GitHub action every night?
@gunchleoc 

If we want this, we should create a Transifex account for bunnybot, make him an admin, and add his TX API token as a secret to the organization for use in this action. A similar action is already implemented and functional in https://github.com/widelands/wl_addons_server/blob/master/.github/workflows/i18n.yaml (it currently uses my personal API token though which I'm not entirely happy about).

---

Given no response for a long time we should just do that.

Additionally some minor other action tweaks which fix #4598 and fix #4933.


Noordfrees:
Anyone to review this please? We won't get a much more convenient time than now for the formatting change.

The only real changes are in the three YAML files; the rest is formatting only.

.github/workflows/build.yaml
```diff
@@ -6,17 +6,23 @@ on:
...
     runs-on: ubuntu-latest
     steps:
     - name: Checkout
       uses: actions/checkout@v2
       with:
         fetch-depth: 1
+    - name: Validate JSON files
+      run: |
+        python3 utils/validate_json.py data/txts/
+        python3 utils/update_authors.py
```
matthiakl, line 20: The outcome of this script is not pushed back into git. Is this intended?

matthiakl (commented):
> If we want this, we should create a Transifex account for bunnybot, make him an admin, and add his TX API token as a secret to the organization for use in this action. 

Is this already configured?

.github/workflows/build.yaml
Noordfrees: The updated files are pushed frequently by the i18n updater workflow already. Here we only check whether any of these programs return an error code in which case the branch author is expected to fix the json syntax.

Noordfrees:
> > If we want this, we should create a Transifex account for bunnybot, make him an admin, and add his TX API token as a secret to the organization for use in this action.
> 
> Is this already configured?

I'll configure this tomorrow morning or this evening

matthiakl (approved):
Let's have it then.
I'm not really happy about the format changes, but it looks like our old format isn't supported anymore.

Noordfrees:
…should be set up now. We'll see tomorrow morning if it worked.
Thanks for the review :)

klaus-halfmann:
There is a nice tool called jq that can be used for all kind of jsoe manipulations (including validation).
Would this help us?

Noordfrees:
We already have a [json validator](https://github.com/widelands/widelands/blob/master/utils/validate_json.py)
No need to reinvent the wheel ;)

PR #4737, opened 2021-03-04
Fix crash when using debug console in editor

Noordfrees:


matthiakl:
Your fix does work. But wouldn't it make more sense to disable the keyboard shortcut in the editor? Currently the "Cheat mode enabled" message is displayed, which is not quite correct for the editor because it doesn't change anything AFAIK.

Noordfrees:
I was thinking about the debug console. I found this bug by attempting to run a lua script in the editor to modify the terrains automatically, which crashed with unreachable code reached. But you're right I'll additionally disable the cheat mode hotkey and overlay for the editor (the features unlocked by cheat mode are permanently enabled in the editor anyway).

matthiakl (approved):
LGTM

PR #4752, opened 2021-03-08
Two more string fixes

Noordfrees:


hessenfarmer (approved):
LGTM. first one is a nice catch

PR #4753, opened 2021-03-08
fix AI handling of amazon_water_gatherer

hessenfarmer:
the fix was easy and minimal invasive. Now the amazons do not waste any building material for the try and error method for water gatherer huts.


Noordfrees (approved):
Code LGTM, and the placement of water gatherers is indeed much better now.

Noordfrees:
On a side note, there's a lot of debug output in the logs, most of it AI-related. I want to clean this up soonish so it's shown only with ``--verbose`` enabled, any objections?

hessenfarmer:
> 
> 
> On a side note, there's a lot of debug output in the logs, most of it AI-related. I want to clean this up soonish so it's shown only with `--verbose` enabled, any objections?

no objections, this would make the log much cleaner.

PR #4756, opened 2021-03-09
Clean up helptexts files

hessenfarmer:


data/tribes/immovables/ashes/init.lua
```diff
@@ -21,10 +23,6 @@
 --
 --    **descname**: The translatable display name. Use ``pgettext`` to fetch the string.
 --
---    **helptext_script**:  The full path to the ``helptexts.lua`` script for this immovable.
---    **NOTE: Deprecated. Immovable helptexts have been shifted to tribes initialization in the current
---    development version.**
---
```
Noordfrees, line 27: Keep the deprecation notice around until after the release

Noordfrees (commented):
Good catches :) LGTM, one comment inline

data/tribes/immovables/ashes/init.lua
hessenfarmer: I decided to delete them to be consistent with buildings and workers where the passage has already been deleted. Furthermore with this change there is no instance anymore left where this is used so it might only confuse people. 

But if you insist I'll readd this but then for all entities

Noordfrees: There are still a couple of other references:
```
.../doc/sphinx$ grep -r helptexts.lua source/
source/lua_tribes_other.rst.org:  help of each unit's and tribe's ``helptexts.lua`` script.
source/lua_tribes_units.rst.org:``data/tribes``. Each subdirectory needs to have a ``register.lua``, an ``init.lua`` and a ``helptexts.lua``
source/lua_tribes_units.rst.org:* ``helptexts.lua`` will provide the information used by the in-game help
source/lua_tribes_workers.rst.org:**helptext_script**: The full path to the ``helptexts.lua`` script for this worker.
source/lua_tribes_workers.rst.org:Each worker has a ``helptexts.lua`` script, which is located in the same directory as its ``init.lua`` script.
source/lua_tribes_buildings.rst.org:Buildings are defined in their ``init.lua`` file. They also have a corresponding ``helptexts.lua`` file that contains their help texts.
source/lua_tribes_buildings.rst.org:    *Mandatory*. The full path to the ``helptexts.lua`` script for this building.
```

hessenfarmer: Arrrgh.  I did only check .lua files to speed up the search I didn't think of the rst files. Thanks for pointing this out. I will clean them up as well

Noordfrees (commented):
LGTM now. I get a testsuite failure on the market map though:
```
[00:00:03.386 real] ERROR: Scenario not started: Game data error: Error while loading tribe 'barbarians': [../src/scripting/lua_errors.cc:22] tribe barbarians: Failed adding ware 'ax_sharp': Duplicate definition of ware
[00:00:03.386 real] ERROR: stack traceback:
[00:00:03.386 real] ERROR: 	[string "tribes/initialization/barbarians/units.lua"]:10: in main chunk
```

hessenfarmer:
> 
> 
> LGTM now. I get a testsuite failure on the market map though:
> 
> ```
> [00:00:03.386 real] ERROR: Scenario not started: Game data error: Error while loading tribe 'barbarians': [../src/scripting/lua_errors.cc:22] tribe barbarians: Failed adding ware 'ax_sharp': Duplicate definition of ware
> [00:00:03.386 real] ERROR: stack traceback:
> [00:00:03.386 real] ERROR: 	[string "tribes/initialization/barbarians/units.lua"]:10: in main chunk
> ```

according to the deleted Helptext the ware was replaced with a custom definition originally this does not work anymore so I have commented out the conflicting part and left a Todo.

Noordfrees:
That can't be right, it works fine in master…

That's what we have a _regression_ test suite for ;)

Noordfrees:
I checked this again with another branch, and you don't need to add the replaced ware type to the tribe modification table at all – it correctly replaces the default ware simply by existing:
grafik

hessenfarmer:
Yes I know therefore I commented out the whole thing in the init.lua However with the new design we can only replace the name and the image, while we can't exchange preciiousness, default quantity and especially helptexts anymore. Therefore I did not delete it completely and left the TODO.
at the moment I deleted the deprecated values which are not taken into account anymore I just tried to add the information back in not knowing that this is not possible anymore. But this is not urgent or anything near to urgent. Just take it as a reminder.

Noordfrees (approved):
Let's have it then :)

PR #4773, opened 2021-03-15
reworked RST for lua_game.cc

### LINKED ISSUE #4761: Documentation errors
over the past time some errors have been emerged in our docuzmentation suite.  this Issue is to track them and hunt them down:

- [x] introduction.rst: Initializations should be renamed to starting conditions, a Link to the starting condition documentation should be added. Example should be deleted as it is outdated and is contained in the recently introduced SC documentation. Win condition example is outdated and should be updated with a real one.

- [x] Productionsite Programs (documented in logic/map_objects/tribes/production_program.cc) Deprecation notice for program work can be deleted, in progrma mine headline the parameter "resources" needs to be renamed "yield"

frankystone:
re #4761 

Not really sure about wl.game.report_result. I guess this was implemented when widelands had used an online gaming service which was meant to propose game results from online multiplayer games (didn't know the name anymore, but this service has been down for several years|decades). @SirVer ?
I have removed the hint to PlayerEndStatus. As far i understand this offers only one status, namely "score"?

### DIFF
--- a/src/scripting/lua_game.cc
+++ b/src/scripting/lua_game.cc
@@ -326,7 +326,7 @@ int LuaPlayer::set_hidden_from_general_statistics(lua_State* L) {
 /* RST
    .. method:: send_to_inbox(title, message[, opts])
 
-      Send a message to the players inbox. One should prefer 
+      Send a message to the players inbox. One should prefer
       :meth:`send_to_inbox` from `messages.lua`, which
       has an option to wait before sending the message until the player leaves
       the roadbuilding mode.
@@ -579,7 +579,7 @@ int LuaPlayer::seen_field(lua_State* L) {
       :arg what: either "all" or an array containing the names of the allowed
          buildings
       :returns: :const:`nil`
-      
+
       The opposite function is :meth:`forbid_buildings`
 */
 int LuaPlayer::allow_buildings(lua_State* L) {
@@ -868,13 +868,13 @@ int LuaPlayer::get_buildings(lua_State* L) {
    .. method:: get_constructionsites(which)
 
       which can be either a single name or an array of names. In the first
-      case, the method returns an array of all constructionsites that the 
+      case, the method returns an array of all constructionsites that the
       player has of this kind. If which is an array, the function returns a
       table of (name,array of buildings) pairs.
 
       :type which: name of constructionsites building or array of building names
       :rtype which: :class:`string` or :class:`array`
-      :returns: information about the players constructionsites, 
+      :returns: information about the players constructionsites,
          see :class:`wl.map.ConstructionSite`
       :rtype: :class:`array` or :class:`table`
 */
@@ -1502,7 +1502,7 @@ const Widelands::Message& LuaInboxMessage::get(lua_State* L, Widelands::Game& ga
 /* RST
 .. function:: report_result(plr, result[, info = ""])
 
-   Shows the window indicating the end of the game. The window contains a 
+   Shows the window indicating the end of the game. The window contains a
    table showing the results of all players and teams (if there are any).
    For mutliplayer games this reports the game ending also to the metaserver
    if this is an Internet network game.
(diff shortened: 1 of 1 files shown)


hessenfarmer (approved):
LGTM

src/scripting/lua_game.cc
```diff
@@ -326,8 +326,8 @@ int LuaPlayer::set_hidden_from_general_statistics(lua_State* L) {
 /* RST
    .. method:: send_to_inbox(title, message[, opts])
 
-      Send a message to the players inbox.
-      There is also a function :meth:`send_to_inbox` in `messages.lua`, which
+      Send a message to the players inbox. One should prefer
```
```suggestion
      Send a message to the player’s inbox. One should prefer
```

src/scripting/lua_game.cc
```diff
@@ -1499,18 +1502,24 @@ const Widelands::Message& LuaInboxMessage::get(lua_State* L, Widelands::Game& ga
 /* RST
 .. function:: report_result(plr, result[, info = ""])
 
-   Reports the game ending to the metaserver if this is an Internet
-   network game. Otherwise, does nothing.
+   Shows the window indicating the end of the game. The window contains a
```
Noordfrees: I'm not sure this is correct. I believe it notifies the game of the result for 1 player, and the window is shown when a result has been reported for every player.

src/scripting/lua_game.cc
```diff
@@ -1248,9 +1251,9 @@ int LuaObjective::set_visible(lua_State* L) {
    .. attribute:: done
 
       (RW) defines if this objective is already fulfilled. If done is
-      :const`true`, the objective will not be shown to the user, no matter what.
+      `true`, the objective will not be shown to the user, no matter what.
       :attr:`visible` is set to. A savegame will be created when this attribute
-      is changed to :const`true`.
+      is changed to `true`.
```
Noordfrees: use double backticks for all constants so they are shown ``in code style`` and not *in italics*

frankystone: There were more occurrences like this, changed them all.

There should be general rules for the documentation, imho. Will start to make a wiki article to archive this.

src/scripting/lua_game.cc
frankystone: Chnaged this so it should be more clear

Noordfrees (approved):

Noordfrees:
@frankystone After merging this, the format bot broke the RST indentation in master, which I fixed manually now. Your fork's action logs show that the format action is not working for you because your PAT secret seems to have a trailing whitespace in it.

frankystone:
Hm. I hate such things.

I remember you proposed to add such a PAT for something. But i can't find your post about that. 
GitHub gives me the possibility to regenerate this token but the warning makes me worry to do that. And i am not sure if this would solve the issue.

Do i need this token at all? The [wikipage](https://www.widelands.org/wiki/GitPrimer/) doesn't mention it...

Is this the PAT which makes trouble at all?

Lost at GitHub :-D

Noordfrees:
You have created such a token already – otherwise there would be a warning in the [logs](https://github.com/frankystone/widelands/runs/2124234634?check_suite_focus=true) under `Fetching branch info` advising you to create one:
grafik

See https://www.widelands.org/forum/topic/5043/

The problem is that the secret in your fork contains the token plus a trailing whitespace.

If you are using this PAT only for the format bot, it is safe to regenerate the token and replace the secret's contents with it. If you are using the same token for other purposes as well then it would be better to create a new token and replace the secret's contents with that.

frankystone:
OK, i have regenerated the token and copied it to the secrets of my fork. Can i test if works now?

Noordfrees:
Just open a draft PR where you messed up the formatting in a few files (e.g. adding whitespaces after parentheses, removing tabs, etc)

frankystone:
I have another branch with documentation fixes... let's see if it works then.

Thanks for your support :-)

PR #4780, opened 2021-03-21
Fix MP dropdown conditions

### LINKED ISSUE #4778: MP setup changes tribes of savegame players
Select a savegame in the LaunchMPG setup. It is possible to share a saved slot with a different player, even if that player uses a different tribe, resulting in the shared slot to be assigned a different tribe. It is also possible to assign an AI to a player with New Tribe, in which case the tribe is changed to Amazons.

matthiakl:
Fixes #4778


Noordfrees (approved):
LGTM, tested and working :)

PR #4793, opened 2021-03-31
Fix amazons fortified village

Noordfrees:
https://www.widelands.org/forum/topic/5170/?page=1#post-36272


hessenfarmer (approved):
change looks good.
Thanks for fixing up my errors

Issue #4334, opened 2020-10-10
Options screen freezes graphics


gunchleoc:
My computer does not like the options screen since the resolution dropdown was reworked. It will often (but not always) freeze my MATE desktop to the point where I can move the mouse but do nothing else except kill my system with the power button.

Niektory:
Uh, that's bad. Does it only happen in the options screen? Immediately after opening it or after a while?

gunchleoc:
Yes, only on the options screen. The screen doesn't show at all, it freezes when I hit the button.

I'm still catching up - it might have been fixed with https://github.com/widelands/widelands/pull/4321 - I was not on the master branch.

My graphics are a bit wonky sometimes and I get freezes from other programs, but until recently Widelands was fine once I had started it successfully once. There might be an SDL bug somewhere - I can't tell at this point.

Noordfrees:
Does this still happen? I cannot reproduce (Ubuntu MATE 20.04). If yes, please run in gdb and create a backtrace while it's frozen

gunchleoc:
I can't create a backtrace when it's frozen, because everything is frozen, including typing something into GDB.

I'll have to debug this by uncommenting code, then running widelands with logging into logfile, kill and start my operating system when it freezes, rinse and repeat until I find the line that's causing it. I haven't had time for that yet.

Noordfrees:
I'm getting a total freeze on my Ubuntu MATE like every half hour ever since the upgrade to 20.04. And from what I read on the Internet lots of people have that issue, with the problem occurring randomly. For me, they usually happen while the compiler is running, never yet in Widelands. So I would guess it's not related to Widelands at all

Niektory:
If the mouse still moves then the system might be recoverable. Does switching TTYs with `Ctrl+Alt+F1...F7` work? Or at least `Ctrl+Alt+SysRq,R,E,I,S,U,B` for a gentler reboot.

gunchleoc:
It's definitely a general graphics problem on my system, but it wasn't triggered until the recent changes to the options screen so I want to investigate this when I have time. Others might have the same buggy type of system.

gunchleoc:
I have created a branch with more log output, but the crash doesn't reproduce today. Will keep trying.

gunchleoc:
It reproduced today. I annotated it like this:

log_dbg("Filling resolution dropdown");

```
// Fill in data
// Interface options
[...]

PR #4798, opened 2021-04-02
Fix documentation 3

### LINKED ISSUE #4761: Documentation errors

frankystone:
Re #4761 

Should fix the [issues mentioned by hessenfarmer](https://github.com/widelands/widelands/issues/4761#issue-827956747) I liked the given idea of a win condition so i adapted it and made it work. For beginners this might be too complicated now, suggestions to remove complexity are welcome.

Additionally i reread tutorial.rst and added a chapter "Scopes of variables".

Here are the html-files for the documentaion in website style, in case one wants to view the made changes:
[current.zip](https://github.com/widelands/widelands/files/6253097/current.zip)


doc/sphinx/source/introduction.rst
```diff
@@ -3,6 +3,14 @@ Introduction
 
 This documentation describes the usage and possibilities of Lua to extend Widelands.
 
+.. note::
+
+   Files of type ``.lua`` are very basic text files. Do not use any fancy word
+   processor (Word, OpenOffice and their like) because they produce the wrong 
+   file format. Make sure to use a plain text editor (like Notepad under 
+   Windows, nedit under Linux and TextEdit under Mac OS X). If you have
+   programmed before, you likely already have found your favorite editor...
```
Noordfrees:
```suggestion
   programmed before, you likely already have found your favorite editor.
```

doc/sphinx/source/introduction.rst
```diff
@@ -14,102 +22,86 @@ Scenarios
...
 a game. The definitions of win conditions is very similar to defining
-initializations: We have to create a Lua script in
-``/data/scripting/win_conditions`` which returns an array with ``name``,
-``description`` and ``func``. Let's also make up a quick example: The first
-player to have 200 logs in his HQ wins the game. All others loose. Save the
-following file as ``/data/scripting/win_conditions/havest_logs.lua``.
+starting conditions which returns an array with ``name``, ``description`` and ``func``. 
+Let’s make up a quick example: The first player to have 200 logs in his HQ 
+wins the game. All others loose. Save the following file as 
+``/data/scripting/win_conditions/havest_logs.lua``.
```
Noordfrees:
```suggestion
``data/scripting/win_conditions/harvest_logs.lua``.
```

doc/sphinx/source/introduction.rst
```diff
@@ -14,102 +22,86 @@ Scenarios
...
+         while not winner do                            -- Iterate all players, check if he is the winner
+            sleep(5000)                                 -- Do this every 5 seconds
+            for idx, p in ipairs(plrs) do               -- Iterate the players array
+               
+               local sf = map.player_slots[p.number].starting_field    -- The starting field of this player
+               local hq = sf.immovable                  -- The headquarters of this player
+               if hq:get_wares("log") >= 200 then       -- Check if more than 200 logs are stored
+                  winner = p                            -- This player is the winner
+                  losers = plrs                         -- Store all players and ...
+                  table.remove(losers,idx)              -- .. remove the winner from the losers table
```
Noordfrees:
```suggestion
                  table.remove(losers,idx)              -- ... remove the winner from the array of losers
```

doc/sphinx/source/introduction.rst
```diff
@@ -144,10 +135,10 @@ back.
 Debug console
 ^^^^^^^^^^^^^
 
-In widelands debug builds you can open a debug console by pressing ``F6``. You
-can enter Lua commands here that act in the global environment: That is if you
-are in a scenario you can access the global variables and alter all Lua
-objects that are in the global scope:
+In widelands debug builds you can open a debug console by pressing 
+``SHIFT+CTRL+SPACE``. You can enter Lua commands here that act in the global
```
Noordfrees:
```suggestion
``Ctrl+Shift+Space``. You can enter Lua commands here that act in the global
```

doc/sphinx/source/introduction.rst
```diff
@@ -160,7 +151,24 @@ This makes for excellent cheating in debug builds, but note that this is for
...
+.. code-block:: lua
+
+   function all_players()
+      for idx, player in ipairs(wl.Game().players) do
+        print("Player:" player.name, player.number, player.tribe.name) 
+      end
+   end
+
+Save this as ``tests.lua``. Now start a normal game, open the debug console 
+by pressing ``SHIFT+CTRL+SPACE`` and enter ``dofile("/full/path/to/tests.lua")``.
```
Noordfrees: here as well

doc/sphinx/source/introduction.rst
```diff
@@ -160,7 +151,24 @@ This makes for excellent cheating in debug builds, but note that this is for
...
+Now you can run the function ``all_players()``. If the output is not what you
+expected just change the function, load the file again with ``dofile()`` and
```
Noordfrees:
```suggestion
expected just change the function, load the file again with ``dofile`` and 
```
otherwise people might think this function needs to be called without arguments to reload the file

doc/sphinx/source/introduction.rst
```diff
@@ -160,7 +151,24 @@ This makes for excellent cheating in debug builds, but note that this is for
...
+        print("Player:" player.name, player.number, player.tribe.name) 
+      end
+   end
+
+Save this as ``tests.lua``. Now start a normal game, open the debug console 
+by pressing ``SHIFT+CTRL+SPACE`` and enter ``dofile("/full/path/to/tests.lua")``.
+Now you can run the function ``all_players()``. If the output is not what you
+expected just change the function, load the file again with ``dofile()`` and 
+call the function again. For convenience you can get the last 5 commands back by
+pressing ``↑``.
```
Noordfrees:
```suggestion
pressing the Up arrow key.
```

doc/sphinx/source/tutorial.rst
```diff
@@ -195,21 +163,21 @@ again. Let's dive into an example right away:
...
 .. code-block:: lua
 
-   use("aux", "coroutine")
+   include "scripting/coroutine.lua"
 
    function print_a_word(word)
       while 1 do
          print(word)
-         sleep(1000)
+         sleep(1000)         -- always sleep a bit, otherwise the system got stuck
```
Noordfrees:
```suggestion
         sleep(1000)         -- always sleep a bit in loops, otherwise Widelands will hang
```

doc/sphinx/source/tutorial.rst
```diff
@@ -195,21 +163,21 @@ again. Let's dive into an example right away:
 
 .. code-block:: lua
 
-   use("aux", "coroutine")
+   include "scripting/coroutine.lua"
 
    function print_a_word(word)
       while 1 do
```
Noordfrees:
```suggestion
      while true do
```

doc/sphinx/source/tutorial.rst
```diff
@@ -331,16 +343,27 @@ Then you still need to format the string with your variable:
...
+two separate strings for translators. We don't have a "break after x characters"
+rule for this; please use common sense here. It is easier for translators to
+translate smaller chunks, and if you should have to change the string later on,
+e.g. to fix a typo, you will break less translations. The strings will be put
+into the translation files in the same order as they appear in the source code,
+so the context will remain intact for the translators.
+
+Also, please hide all formatting control characters from our translators. This
+includes HTML tags as well as new lines in the code! For an example, have a
+look at ``data/campaigns/atl01.wmf/scripting/texts.lua``.
```
Noordfrees: Also add a note that one must not simply concatenate two sentences but instead use bformat or a localized-concatenation function (don't remember what it's called and where it's defined), unless there's a paragraphdivider or other richtext markup between the sentences.

Edit: and replace "HTML" with "richtext" – we are not using HTML anywhere in Widelands, though our richtext markup syntax looks similar

Noordfrees (commented):
Looking good, a few comments inline

doc/sphinx/source/tutorial.rst
frankystone: It's not just widelands... i got sometimes there loosing control over the whole system. No mouse, no taskmanager, not able to switch windows... the only thing i could do was to press the main power button...

doc/sphinx/source/tutorial.rst
frankystone: I have not enough knowledge about this localization stuff to write such a note.

Changed HTML with a link to richtext.lua.

doc/sphinx/source/tutorial.rst
frankystone: done

doc/sphinx/source/introduction.rst
frankystone: done

doc/sphinx/source/introduction.rst
frankystone: done

doc/sphinx/source/introduction.rst
frankystone: Done, except that i did not change the order of keys. For me it's handy to have this ordering.

Same for below.

doc/sphinx/source/introduction.rst
frankystone: done

doc/sphinx/source/introduction.rst
frankystone: done

doc/sphinx/source/introduction.rst
frankystone: Why so strict?

doc/sphinx/source/introduction.rst
Noordfrees: It's inconsistent with the UI though, where the ordering of modifier keys is always the same:
grafik

doc/sphinx/source/introduction.rst
Noordfrees: the typo in the filename is still there

doc/sphinx/source/introduction.rst
Noordfrees: An ellipsis looks out of place here to me. You can also keep it if you prefer it, but then please use the `'…'` character.
(`…` should be used instead of `...` everywhere except in monospaced code markup.)

doc/sphinx/source/tutorial.rst
Noordfrees: I find this hard to believe since operating systems should surely never grant a single process full access to all the CPU power for more than a few milliseconds at a time? #4334 is another such weird thing. I'm not convinced that it is Widelands that effects such a total system freeze; I've seen a fair share of such freezes that seem to coincide suspiciously often with a particular program and yet cannot possibly be related.

I've tried out entering an endless loop in the debug console now, and while Widelands got hung up as expected, the rest of the system kept behaving just fine.

doc/sphinx/source/tutorial.rst
Noordfrees:
```
Note that simply concatenating two translatable strings with ``..``
does not work in all languages. To achieve the desired effect, use
the function :meth:`join_sentences` with the two sentences as
arguments, or :meth:`paragraphdivider` if you wish the text to
continue on a new paragraph.

Never assemble a single localized sentence with the ``..``
operator. Use :meth:`string.bformat` with appropriate
placeholder substitutions for such cases.
```

doc/sphinx/source/tutorial.rst
frankystone: Can't reproduce this in the debug console also. But surely i got there in the past while testing scenarios …

Anyway: The warning should be as hard as it can be, imho. It's both a situation which is bad, not all users know how to kill a process.

Noordfrees: Okay. But then how about putting this info prominently in a `.. note::` section instead of mentioning it on the side in a code example? E.g.
```
.. note:: 
   Always add :meth:`sleep` statements to lengthy loops.
   Widelands will hang while the code between two :meth:`sleep`
   calls is executed, resulting in a risk that the program – or
   even the whole operating system – may stall for an extended
   period of time, or even until Widelands is force-closed.
```

doc/sphinx/source/introduction.rst
frankystone: I decided to remove the last sentence.

doc/sphinx/source/tutorial.rst
frankystone: Since there was already a not, i wrote a mix of my own words, the already written note and your suggestion.

doc/sphinx/source/introduction.rst
frankystone: sorry, overlooked this, done

doc/sphinx/source/tutorial.rst
frankystone: Thanks, done.

doc/sphinx/source/introduction.rst
frankystone: ok, done

doc/sphinx/source/tutorial.rst
```diff
@@ -195,22 +163,22 @@ again. Let's dive into an example right away:
 
 .. code-block:: lua
 
-   use("aux", "coroutine")
+   include "scripting/coroutine.lua"
 
-   function print_a_word(word)
-      while 1 do
+   function print_a_word(word)         -- this function is a coroutine
```
Noordfrees: Strictly this is not correct (and the paragraph above is also overly simplistic). `print_a_word` is a *function*. The call to `run(f)` then creates a new coroutine object from the function `f`. A coroutine, in Widelands's Lua implementation, is a wrapper *object* around a function in a *Lua State*, with its own *Lua stack* and set of local variables. Most of the time this is a fine distinction, but when you want more control over your script's workflow it becomes a very important one.
We don't need to explain it in such detail in the beginner's tutorial but we shouldn't give incorrect information either.

frankystone: " -- this function will get a coroutine" ?

Noordfrees: "get" might be interpreted as "receive" in this context. It's a word with many different meanings which we're overfond of using when translating German into English ;D

"Function from which a coroutine will later be constructed" or "A function we'll use to create a coroutine" would be clearer imho. 
This could be clarified further by changing the comment on the `run(...)` line to "Constructs a new coroutine from the function `print_a_word` and an argument, and immediately launches this coroutine."

frankystone: Good explanation :+1: 

Done

Noordfrees (commented):
Two more nits found while building the doc locally.
- At the start of the tutorial there's a reference to "the nozip option in the advanced options inside of Widelands". This should be updated to the current tab and option name.
- In the introduction section " Regression testing infrastructure", there are a number of markups where single backticks need to be exchanged with double backticks.

Noordfrees (approved):

frankystone:
Thanks for your support!

Let's wait for hessenfarmer, maybe he has ideas to make the example of the wincondition easier.

doc/sphinx/source/introduction.rst
```diff
@@ -14,102 +21,86 @@ Scenarios
...
+   local wc_version = 1
+   local wc_desc = _("The first player with 200 logs in his headquarter wins!")
+
+   local wc_definition = {                              -- The array defining the win condition
+      name = wc_name,
+      description = wc_desc,
+      peaceful_mode_allowed = true,
+      func = function()                                 -- The function processing the win condition
+         local plrs = wl.Game().players                 -- All players
+         broadcast_objective("win_condition",           -- Set objective with win condition for all players
```
hessenfarmer, line 67: I believe the broadcast should be the first command in the function.
Furthermore the necessary parameters up to this broadcast should be explained in the Text after the Code example as they are the bare minimum. Especially the use of peaceful_mode needs to be explained. 
The endless-game Wincondition would serve as a better example to explain the minimum required so we could use the endles game as first code block, explain the settings and give the harvest log example to show a more sophisitcated example of code to determine a winner.

hessenfarmer (commented):
Sorry for being quiet for a while but I was busy and tired. 
This PR looks already good and would be goot to go however as being asked about a less complex example my suggestion for a 2 stepped explanation approach to not loose the Harvest logs example inline.

doc/sphinx/source/introduction.rst
frankystone: Sorry, i have trouble with github E-mails for some time, so i overlooked your answer here.

`peaceful_mode`: I thought also to explain this, but in the end i thought it is self explaining. A user who wants to dive into LUA has probably played a lot of games so he might know the meaning of this. I am open for a suggestion explaining this though

Having two examples shown here is a bit overkill, imho. The endless_game win condition can be opened by everyone so i do not feel the need to show it here again. And if the wincondition will be changed, the documentation might get outdated again because no one thinks about this.
But i will add a sentence that one can look at the endless_game win condition for an easy example and add explanations to some of the variables.

hessenfarmer: What I meant was to explain the basic parameters in an explanatory text after the code example rather then in the code comments.
However this would be only the cherry on the cake and the current version is good to go as well.
So let's have it.

PR #4799, opened 2021-04-02
Fix textdomain for keyboard shortcuts

Noordfrees:
The shortcut-related game tips shown during game loading use the wrong textdomain on master


matthiakl (approved):
Good catch :)

PR #4810, opened 2021-04-13
Fix initial info panel text visibility

Noordfrees:
If a game is started with the info panel initially in Hidden state, the text areas are initially visible even though they're meant to be always invisible when the panel is set to Hidden.


matthiakl (approved):
Fix confirmed :)

PR #4811, opened 2021-04-14
Don't store Script add-ons in savegames

Noordfrees:
It is not necessary as it's already dumped in the scripting packet and ignored on loading.


frankystone (approved):
Works as expected

PR #4818, opened 2021-04-16
Add ordered iteration for lua tables

### LINKED ISSUE #4816: HQ Hunter replays desyncs
https://www.widelands.org/forum/topic/5181

Replays with the HQ hunter win condition desync shortly after an HQ is destroyed. Replay created with 1.0~git25267[296b275@master] (Release) attached. It desyncs after the light green player player (amz) is defeated by violet (emp).
```
[11:47:05.223 game] ERROR: REPLAY: Lost synchronization at time 42425223
[11:47:05.223 game] ERROR: I have:     1d063393fa259a3ede3684cdbdae9cd2
[11:47:05.223 game] ERROR: Replay has: ec40265db26d661629b5e6c565a9ba87
```
[replays.zip](https://github.com/widelands/widelands/files/6324546/replays.zip)

matthiakl:
Fixes #4816 
Implements an `ordered_pairs` function, to iterate lua tables deterministically.
Taken from http://lua-users.org/wiki/SortedIteration
There might be more occurences, there an ordered iteration should be used to prevent similar issues, but I'm not familiar with the lua code.

To verify the fix:
- Start a pure AI game on crater with HQ Hunter
- Let it finish
- Save
- Run the replay
- On master it desyncs, with `ordered_pairs` it works fine

Note: This is my first lua code submission, so I hope I matched the style.

### DIFF
--- a/data/scripting/table.lua
+++ b/data/scripting/table.lua
@@ -31,3 +31,71 @@ function array_combine(...)
    end
    return t
 end
+
+
+-- RST
+-- .. function:: __gen_ordered_index(t)
+--
+--    Generate a sorted index from a table
+--    Taken from http://lua-users.org/wiki/SortedIteration
+--
+--    :arg t: a table
+--    :returns: a table containing t's keys in alphabetic order
+function __gen_ordered_index( t )
+    local ordered_index = {}
+    for key in pairs(t) do
+        table.insert( ordered_index, key )
+    end
+    table.sort( ordered_index )
+    return ordered_index
+end
+
+-- RST
+-- .. function:: ordered_next(t, state)
+--
+--    Equivalent of the next function, but returns the keys in the alphabetic
+--    order. We use a temporary ordered key table that is stored in the
+--    table being iterated.
+--    Taken from http://lua-users.org/wiki/SortedIteration
+--
+--    :arg t: a table
+--    :arg state: current iteration state
+--    :returns: the next key,value pair
+function ordered_next(t, state)
+
+    local key = nil
+    if state == nil then
+        -- the first time, generate the index
+        t.__ordered_index = __gen_ordered_index( t )
+        key = t.__ordered_index[1]
+    else
+        -- fetch the next value
+        for i = 1,#t.__ordered_index do
+            if t.__ordered_index[i] == state then
+                key = t.__ordered_index[i+1]
+            end
+        end
+    end
+
+    if key then
+        return key, t[key]
+    end
+
+    -- no more value to return, cleanup
+    t.__ordered_index = nil
+    return
+end
+
+-- RST
+-- .. function:: ordered_pairs(t)
+--
+--    Equivalent of the pairs() function on tables. 
+--    Allows to iterate in deterministic order.
+--    Taken from http://lua-users.org/wiki/SortedIteration
+--
+--    :arg t: a table
+--    :returns: a key,value iterator
+function ordered_pairs(t)
+    return ordered_next, t, nil
+end
+
--- a/data/scripting/table.lua
+++ b/data/scripting/table.lua
@@ -45,7 +45,7 @@ function __gen_ordered_index(t)
 end
 
 -- Helper function for ordered_pairs(t)
--- Equivalent of the next function, but returns the keys in the alphabetic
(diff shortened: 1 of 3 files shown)


data/scripting/table.lua
```diff
@@ -31,3 +31,71 @@ function array_combine(...)
...
+end
+
+-- RST
+-- .. function:: ordered_pairs(t)
+--
+--    Equivalent of the pairs() function on tables. 
+--    Allows to iterate in deterministic order.
+--    Taken from http://lua-users.org/wiki/SortedIteration
+--
+--    :arg t: a table
```
Noordfrees:
```suggestion
--    Equivalent of the :func:`pairs()` function on tables. 
--    Allows to iterate in deterministic order.
--    Taken from http://lua-users.org/wiki/SortedIteration
--
--    :arg table t: The table to iterate over
```

Noordfrees (commented):
Code looking good, will test this now.

Stylewise, please remove the whitespaces inside parentheses (e.g. `( t )` → `(t)`). Also, the functions `__gen_ordered_index` and `ordered_next` seem to be only helper functions for `ordered_pairs` and not meant to be used elsewhere, so they shouldn't have RST comments (a brief usage comment with the note that they're for internal use only is enough).

Edit: Tested and working :)

hessenfarmer:
not tested yet. however there are more instances of pairs table in HQ Hunter.lua and winconditions_functions.lua that should be safeguarded. 
I assume ipairs usage does not need the safeguard as it is already indexed. (I don't know if it is ordered as well)

data/scripting/table.lua
frankystone: > --    Equivalent of the :func:`pairs()` function on tables. 

I guess this will create a warning when creating the documentation, because there isn't a defined function called pairs() in the docs. I would use 
>  --  Equivalent of the Lua-function ``pairs()``  ... (markup: ` ``pairs()`` `)

Same for mentioning the next function in line 56:

> --  Equivalent of the Lua-function ``next()``, ...

matthiakl:
From https://stackoverflow.com/a/55109411:
> - pairs() returns key-value pairs and is mostly used for associative tables. key order is unspecified.
> - ipairs() returns index-value pairs and is mostly used for numeric tables. Non numeric keys in an array are ignored, while the index order is deterministic (in numeric order).

So the preferred way to iterate should be `ipairs()` which is only possible for continuous numeric keys. If the keys are not numeric and the outcome of the `for` loop depends on the iteration order the new `ordered_pairs` should be used.
I searched the lua files for calls of `pairs` and most of them could be replaced with `ipairs` to guarantee the order. But I'm not sure, how to determine the key type in the code, so I suggest to postpone this refactoring after the release (or someone with more lua experience does it ;).

Currently we profit from a "common behavior in Lua":
> I believe that the ordering is guaranteed with pairs(), and it will go up through the numeric indexes and then will go to the other non numeric values in a random order. 

> @BeastCoder2: At lua.org/manual/5.3/manual.html#pdf-next, it says "The order in which the indices are enumerated is not specified, even for numeric indices." What you describe might be common behavior in Lua, but it could change in any bugfix release.

frankystone:
You can use `type()` :
~~~~
> t= {"one", two=2}
> t[1]
one
> for k,v in pairs(t) do print(k .. " type: ".. type(k), v, "type: " .. type(v)) end
1 type: number  one     type: string
two type: string        2       type: number
> f = function (v) print(v) end
> table.insert(t, f)
> for k,v in pairs(t) do print(k .. " type: ".. type(k), v, "type: " .. type(v)) end
1 type: number  one     type: string
2 type: number  function: 0x559d6b9f9c50        type: function
two type: string        2       type: number
> t[2]("hello")
hello
>
~~~~

One has to loop over all keys in a table to determine if the whole table is an indexed array.

Noordfrees (approved):

data/scripting/table.lua
matthiakl: Fixed.

Noordfrees:
I opened a follow-up issue for the `pairs/ipairs/ordered_pairs` refactoring. Thanks for fixing :)

PR #4822, opened 2021-04-19
Fix documentation wlmaps

frankystone:
Not ready yet.

I have a question about MapObjectDescription, which confuses me all the time. From my understanding this class provides access to descriptions of all objects a tribe can use. Regardless if the object is used, or allowed, in the actual game or on the map. So this class provides __general__ information of all objects a tribe can have.
What confuses me is the name: _**Map**ObjectDescription_ which makes me assume this is something related to map objects (in a game). And this (probably wrong) assumption makes me unsure if i write the correct things…

What i want to archive is to make a clear distinction between MapObjectDescription and MapObject.descr. 

Can one take look please to what i have written in this regard?


Noordfrees:
On the whole, correct.
`MapObject.descr` returns an object of type `MapObjectDescription` (or rather, a subclass), which is shared by all existing map objects of the same type. E.g. there's *one* singleton instance of the "barbarians_carrier" `WorkerDescription`. If you have an actual barbarian carrier `c`, `c.descr` returns that instance. If `x` is a different barbarian carrier, `x.descr` also returns the same instance. And it is the same one returned by `wl.Game():get_worker_description("barbarians_carrier")`.
The point is that all attributes of map objects which are identical for all actual instances and independent of the game state are gathered in a unique and never-changing instance of a subclass of `MapObjectDescription`, to allow accessing them independent from any instance. 
The subclass naming convention is `Xyz` for an actual instance and `XyzDescription` for the corresponding Description type.

> What i want to archive is to make a clear distinction between MapObjectDescription and MapObject.descr.

I am not sure what exactly you mean by this though. `MapObjectDescription` is a class, and `MapObject.descr` returns an instance of this class. That's all there is to this distinction…

frankystone:
Thanks :-)

The problems i had recently was to know when to use MapObject.descr. It took me some time to get the conclusion "Use MapObject.descr whenever you want to access descriptions for an existing object in a game."

It is just confusing if the description of [MapObjectDescription](https://www.widelands.org/documentation/autogen_wl_map/#mapobjectdescription) states:
"A static description of a tribe’s map object, …"

This all (the name and the description) makes a reader believe this could be used for existing map objects wich is not strictly true. It can be used for existing map objects, but only when using MapObject.descr. This is what i want to make clear to the reader. Of course without weaken the information for developers.

frankystone:
I think the function [helptext](https://www.widelands.org/documentation/autogen_wl_map/#wl.map.MapObjectDescription.helptext) of MapObjectDescription can't be invoked and should be removed from the documentation.

Or am i wrong?

Noordfrees:
It seems there's a typo in the method declaration:
```
METHOD(LuaImmovableDescription, helptexts),
```
should be
```
METHOD(LuaMapObjectDescription, helptexts),
```

and another one in the RST: helptext → helptexts

frankystone:
Ok, but since it seems this was never working and the helptexts are fetched in another way: Is this needed?

BTW, in the current documentation the description is not complete. [The source](https://github.com/widelands/widelands/blob/master/src/scripting/lua_map.cc#L2268) mentions what is returned exactly.

Anyway i changed the declaration tested it and it works, although it seems a bit wrong to me. Something like `decription_object:helptexts()` (without tribename as argument) which would return a helptext for the object, if there are any, would make more sense imho.

frankystone:
Although the changes are mainly minimal, i guess the diff is far to big for a review.

Should i split this in two branches or create another branch with only two or three commits?

Noordfrees:
It's only 1100 lines, I've opened PRs with many times that ;P 
For me it's fine to keep this as one PR

frankystone:
Ok then, this is ready for review.

Every time i reread this file i find more to fix, but i am tired now of this file...

src/scripting/lua_map.cc
```diff
@@ -2107,17 +2121,26 @@ MapObjectDescription
...
-   A static description of a tribe's map object, so it can be used in help files
-   without having to access an actual object on the map.
-   This class contains the properties that are common to all map objects such as buildings or wares.
+   A static description of a tribe’s object, so it can be used without
+   having to access an actual object on the map. This class contains the
+   properties that are common to all objects a tribe has, such as buildings or
+   wares.
+
+   The dynamic class :class:`MapObject` corresponding to this class is the base
+   class for all Objects in widelands, including immovables and Bobs. This
```
Noordfrees:
```suggestion
   class for all objects in Widelands, including immovables and bobs. This
```

I don't understand the term "dynamic class"

src/scripting/lua_map.cc
```diff
@@ -2107,17 +2121,26 @@ MapObjectDescription
...
+   class can't be instantiated directly, but provides the base for all others.
+
+   Accessing descriptions for an existing map object in a game is done via
+   the class :class:`MapObject` and the attribute :attr:`MapObject.descr`.
 
-   The dynamic MapObject class corresponding to this class is the base class for all Objects in
-   widelands, including immovables and Bobs. This class can't be instantiated directly, but provides
-   the base for all others.
+   To access static descriptions without being an existing map object one can use
```
Noordfrees: "being"?

src/scripting/lua_map.cc
```diff
@@ -2178,96 +2201,116 @@ int LuaMapObjectDescription::get_name(lua_State* L) {
...
+
+          * :class:`building <BuildingDescription>`, the base class
+            for all buildings
+          * :class:`constructionsite <ConstructionSiteDescription>`,
+            an actual building is being constructed here,
+          * :class:`dismantlesite <DismantleSiteDescription>`, an
+            actual building is being dismantled here,
+          * :class:`warehouse <WarehouseDescription>`, a warehouse
+            can store wares and workers. Headquarters and ports are
+            special tapes of warehouses, but they belong to the same
```
Noordfrees: typo

src/scripting/lua_map.cc
```diff
@@ -2178,96 +2201,116 @@ int LuaMapObjectDescription::get_name(lua_State* L) {
...
+            on water terrain where ships can load/unload wares and
+            workers. A portdock is invisible to the player and one is
+            automatically placed next to each port building.
+
+      * **Abstract:** These types are abstract map objects that are used by the engine and are
+        not visible on the map.
+
+        * :class:`battle <BattleDescription>`, holds information
+          about two soldiers in a fight,
+        * :class:`fleet <FleetDescription>`, holds information for
```
Noordfrees: Please double-check if this is still correct. There are now classes ShipFleet and FerryFleet and I don't know how the Lua interface for these looks now.

src/scripting/lua_map.cc
```diff
@@ -2178,96 +2201,116 @@ int LuaMapObjectDescription::get_name(lua_State* L) {
...
+
+      .. code-block:: lua
+
+         -- get tribe description
+         local tribe_descr = wl.Game():get_tribe_description("barbarians")
+
+         -- get building descriptions of this tribe
+         local buildings = tribe_descr.buildings
+
+         -- loop all building descriptions
```
Noordfrees:
```suggestion
         -- iterate over all building descriptions
```

src/scripting/lua_map.cc
```diff
@@ -2366,11 +2406,20 @@ int LuaImmovableDescription::get_becomes(lua_State* L) {
 /* RST
    .. attribute:: terrain_affinity
 
-         returns the terrain affinity values for this immovable
+      (RO) For trees: A :class:`table` containing numbers labeled as pickiness (uint),
```
Noordfrees: Bushes also have terrain affinity

Noordfrees: "uint" should probably have type markup?

src/scripting/lua_map.cc
```diff
@@ -2366,11 +2406,20 @@ int LuaImmovableDescription::get_becomes(lua_State* L) {
...
+      preferred_fertility (uint), preferred_humidity (uint), and preferred_temperature (uint),
+      or nil if the immovable has no terrain affinity.
```
Noordfrees: `` :const:`nil` ``

src/scripting/lua_map.cc
```diff
@@ -2442,13 +2490,15 @@ int LuaImmovableDescription::has_attribute(lua_State* L) {
...
 
 /* RST
-   .. method:: probability_to_grow
+   .. method:: probability_to_grow(terrain_description)
 
-      :arg terrain: The terrain that we are checking the probability for.
-      :type terrain: :class:`wl.map.TerrainDescription`
+      (RO) For trees: Returns a double describing the probability that this tree will grow on the
+      given terrain. Returns nil if this tree has no terrain affinity (all trees should
+      have one).
```
Noordfrees: 2× type markup

bushes also have terrain affinity

src/scripting/lua_map.cc
```diff
@@ -2573,8 +2622,8 @@ int LuaBuildingDescription::get_enhanced(lua_State* L) {
 /* RST
    .. attribute:: enhanced_from
 
-         (RO) returns the building that this was enhanced from, or nil if this isn't an enhanced
-         building.
+      (RO) The :class:`~wl.map.BuildingDescription` that this was enhanced from, or nil if this
```
Noordfrees: type markup

src/scripting/lua_map.cc
```diff
@@ -2683,8 +2733,8 @@ int LuaBuildingDescription::get_vision_range(lua_State* L) {
 /* RST
    .. attribute:: workarea_radius
 
-         (RO) the first workarea_radius of the building as an int,
-              nil in case bulding has no workareas
+      (RO) The first workarea_radius of the building as an :class:`integer`,
+      nil in case bulding has no workareas.
```
Noordfrees: type markup

src/scripting/lua_map.cc
```diff
@@ -2747,11 +2793,9 @@ ProductionSiteDescription
    Child of: :class:`MapObjectDescription`, :class:`ImmovableDescription`,
    :class:`BuildingDescription`
 
-   A static description of a tribe's productionsite, so it can be used in help files
-   without having to access an actual building on the map.
-   This class contains the properties for productionsites that have workers.
-   For militarysites and trainingsites, please use the subclasses.
-   See the parent classes for more properties.
+   A static description of a tribe's productionsite. This class contains
+   the properties for productionsites that have workers. For militarysites and trainingsites,
```
Noordfrees: trainingsites are also "productionsites that have workers", and militarysites are not productionsites at all

src/scripting/lua_map.cc
```diff
@@ -2853,9 +2897,9 @@ int LuaProductionSiteDescription::get_collected_immovables(lua_State* L) {
 /* RST
    .. attribute:: collected_resources
 
-      (RO) An array with :class:`ResourceDescription` containing the resources that
+      (RO) An :class:`array` of :class:`ResourceDescription` containing the resources that
       this building will collect from the map, along with the maximum percentage mined and the
-      chance to still find some more after depletion. For example, a Fishers's House will collect:
+      chance to still find some more after depletion. E.g. for a Fisher's hut this will be:
```
Noordfrees: either "Fisher's Hut" or "fisher's hut"

src/scripting/lua_map.cc
```diff
@@ -3113,13 +3158,16 @@ int LuaProductionSiteDescription::get_working_positions(lua_State* L) {
 }
 
 /* RST
-   .. attribute:: consumed_wares_workers
+   .. method:: consumed_wares_workers(program_name)
+
+      (RO) Returns a :class:`table` of ``{{ware_name},ware_amount}`` for the wares consumed by this
```
Noordfrees: RO doesn't make sense for methods

src/scripting/lua_map.cc
```diff
@@ -3148,13 +3196,15 @@ int LuaProductionSiteDescription::consumed_wares_workers(lua_State* L) {
 }
 
 /* RST
-   .. attribute:: produced_wares
+   .. method:: produced_wares(program_name)
+
+      (RO) Returns a :class:`table` of ``{ware_name=ware_amount}`` for the wares produced by this
```
Noordfrees: dito

Noordfrees: and more below

src/scripting/lua_map.cc
```diff
@@ -3295,8 +3347,8 @@ int LuaTrainingSiteDescription::get_food_attack(lua_State* L) {
 /* RST
    .. attribute:: food_defense
 
-      (RO) A table of tables with food ware names used for Defense training,
-           e.g. {{"barbarians_bread"}, {"fish", "meat"}}
+      (RO) A :class:`table` of tables with food ware names used for Defense training,
```
Noordfrees: The training attributes sometimes start with Uppercase and sometimes with lowercase here and below. This should be made consistent.

src/scripting/lua_map.cc
```diff
@@ -3633,9 +3681,7 @@ WareDescription
 
    Child of: :class:`MapObjectDescription`
 
-   A static description of a tribe's ware, so it can be used in help files
-   without having to access an actual instance of the ware on the map.
-   See also :class:`MapObjectDescription` for more properties.
+   A static description of a tribe's ware. See the parent class for more properties.
```
Noordfrees: Unlike buildings, workers, and ships, a ware can belong to multiple tribes.

src/scripting/lua_map.cc
```diff
@@ -3847,7 +3889,7 @@ int LuaWorkerDescription::get_employers(lua_State* L) {
 /* RST
    .. attribute:: buildable
 
-      (RO) returns true if this worker is buildable
+      (RO) Returns true if this worker is buildable.
```
Noordfrees: Not a function, so no "returns". Also type markup.

src/scripting/lua_map.cc
```diff
@@ -4321,8 +4371,12 @@ Economy
 -------
 .. class:: LuaEconomy
 
-   Provides access to an economy. A player can have multiple economies;
-   you can get an economy from a :class:`Flag`.
+   Provides access to an economy. An economy will be created each time a player places a
+   flag on the map. As soon this flag get connected to other flags the economies of all flags
+   will be joined to a single economy. A player can have multiple economies each with it's own
+   economy-settings.
```
Noordfrees:
```suggestion
   flag on the map. As soon this flag is connected to another flag, their two economies will
   be merged into a single economy. A player can have multiple economies, each of which has
   its own set of economy target settings.
```

src/scripting/lua_map.cc
```diff
@@ -4402,10 +4456,10 @@ int LuaEconomy::target_quantity(lua_State* L) {
       through placing/deleting roads and flags, you must get a fresh economy
       object every time you use this function.
 
-      :arg workername: the name of the worker type.
+      :arg workername: The name of the worker type.
       :type workername: :class:`string`
 
-      :arg amount: the new target amount for the worker. Needs to be >= 0.
+      :arg amount: The new target amount for the worker. Needs to be >= 0.
```
Noordfrees: type markup?

src/scripting/lua_map.cc
```diff
@@ -4640,9 +4703,12 @@ int LuaMapObject::destroy(lua_State* L) {
 }
 
 /* RST
-   .. method:: has_attribute(string)
+   .. method:: has_attribute(attribute)
+
+      (RO) Returns :const:`true`, if the map object has this attribute, else :const:`false`.
```
Noordfrees:
```suggestion
      Returns :const:`true` if the map object has this attribute, :const:`false` otherwise.
```

src/scripting/lua_map.cc
```diff
@@ -5329,6 +5397,10 @@ int LuaBuilding::get_flag(lua_State* L) {
    .. method:: dismantle([keep_wares = false])
 
       Instantly turn this building into a dismantlesite.
+
+      :arg keep_wares: Optional: If :const:`true` the wares in this buildings stock get destroyed.
+         If :const:`false` (default) the wares in this buildings stock will be preserved.
```
Noordfrees: Isn't it the other way around?

src/scripting/lua_map.cc
```diff
@@ -6287,8 +6373,9 @@ int LuaProductionSite::get_valid_workers(lua_State* L) {
 
       (RO) Returns whether this productionsite is currently active or stopped
 
-      :returns: true if the productionsite has been started,
-         false if it has been stopped.
+      :returns: :const:`true` if the productionsite has been started,
+         :const:`false` if it has been stopped.
```
Noordfrees: not the other way around?

src/scripting/lua_map.cc
```diff
@@ -6870,13 +6959,14 @@ int LuaShip::get_last_portdock(lua_State* L) {
 /* RST
    .. attribute:: state
 
-   Query which state the ship is in:
+      (RW) Query which state the ship is in. Can be either of:
```
Noordfrees: Is this really writeable?

src/scripting/lua_map.cc
```diff
@@ -7194,12 +7279,51 @@ int LuaShip::build_colonization_port(lua_State* L) {
...
 
       Turns this ship into an expedition ship without a base port. Creates all necessary
-      wares and a builder plus, if desired, the specified additional items.
-      Any items previously present in the ship will be deleted.
-
+      wares and a builder plus, if desired, the specified additional **items**.
       The ship must be empty and not an expedition ship when this method is called.
 
-      :returns: nil
+      Note that the :attr:`capacity` is not adjusted if you give additional **itmes**.
```
Noordfrees: typo

src/scripting/lua_map.cc
```diff
@@ -7194,12 +7279,51 @@ int LuaShip::build_colonization_port(lua_State* L) {
...
+      If the amount of additional items exceed the capacity, the game don't like it.
```
Noordfrees: exceed**s**
"don't" → "won't" or "doesn't"

src/scripting/lua_map.cc
```diff
@@ -7194,12 +7279,51 @@ int LuaShip::build_colonization_port(lua_State* L) {
...
+
+      See also :any:`launch_expeditions` which adjusts :attr:`capacity`
+      depending on the given wares and workers.
+
+      :arg items: Optional: Additional items to the ones that are needed to build a port.
+      :type items: :class:`table` of ``{ware_or_worker=amount}`` pairs.
+
+      :returns: :const:`nil`
+
+      Example with check for sufficent capacity:
```
Noordfrees: sufficient

src/scripting/lua_map.cc
```diff
@@ -7194,12 +7279,51 @@ int LuaShip::build_colonization_port(lua_State* L) {
...
+         -- finally subtract one slot for the builder
+         free_capacity = free_capacity - 1
+
+         -- adjust capacity
+         if free_capacity < 13 then
+            ship.capacity = ship.capacity + 13
+         end
+
+         -- create expedition with additional 13 items
+         ship:make_expedition({barbarians_soldier = 10, fish=3})
```
Noordfrees: Use consistent spacing around the =

Noordfrees (commented):
I'm through. Looks good, a number of comments inline.

Thanks for doing this huge task :)

src/scripting/lua_map.cc
frankystone: The [original text](https://www.widelands.org/documentation/autogen_wl_map/#wl.map.MapObjectDescription) was:

> The dynamic MapObject class corresponding to this class is the base class for all Objects in widelands, including immovables and Bobs. This class can’t be instantiated directly, but provides the base for all others.

Please suggest a better rephrasing.

src/scripting/lua_map.cc
frankystone: There are no Lua interfaces about `fleet` or `ShipFleet` or `'FerryFleet`. The original documentation says `Fleet` and `BattleDescription` are 'abstract'.

src/scripting/lua_map.cc
frankystone: done

src/scripting/lua_map.cc
frankystone: done

src/scripting/lua_map.cc
frankystone: Since Lua has no `uint` i removed them all.

src/scripting/lua_map.cc
frankystone: done, also other occurences of nil

src/scripting/lua_map.cc
frankystone: Removed 'For Trees'

src/scripting/lua_map.cc
frankystone: removed the sentence

src/scripting/lua_map.cc
frankystone: done

src/scripting/lua_map.cc
frankystone: done

src/scripting/lua_map.cc
frankystone: made all lowercase

src/scripting/lua_map.cc
frankystone: There are some occurences like this. If we are strict in this way we have to remove also all markup `:returns: …` for attributes.

src/scripting/lua_map.cc
frankystone: done

src/scripting/lua_map.cc
frankystone: For the expression? Changed to ``>=0``

src/scripting/lua_map.cc
frankystone: done

src/scripting/lua_map.cc
frankystone: oh yes...

src/scripting/lua_map.cc
frankystone: oh yes...

src/scripting/lua_map.cc
frankystone: no

src/scripting/lua_map.cc
frankystone: done

src/scripting/lua_map.cc
frankystone: done

src/scripting/lua_map.cc
frankystone: done

frankystone (commented):
Fixed most things

src/scripting/lua_map.cc
Noordfrees: The other sentence can also go as all productionsites need a worker

src/scripting/lua_map.cc
Noordfrees: Let's remove the "dynamic" and then the sentence is fine.

"Widelands" should always be Uppercase as it's a name.

A few lines above I dislike the phrasing "description of a **tribe**’s object" – critters, rocks, trees etc are also map objects. Perhaps rephrase as "A static description of an object on the map, so it can be used without having to access an actual instance of an object on the map."

src/scripting/lua_map.cc
Noordfrees: The corresponding strings are "ferry_fleet" and "ship_fleet" (defined in logic/map_objects/map_object_type.cc). Please add a note that no properties are currently implemented and remove the (probably broken) class links.

Also, for roads, the map object type strings are "roadbase" (abstract generic road/waterway), "road", and "waterway" (three different strings even though they're represented by the one and the same Lua class!), please update that as well.

src/scripting/lua_map.cc
Noordfrees: Okay let's keep it like this for now

src/scripting/lua_map.cc
```diff
@@ -2058,7 +2072,7 @@ int LuaTribeDescription::get_workers(lua_State* L) {
 /* RST
    .. method:: has_building(buildingname)
 
-      Returns true if buildingname is a building and the tribe can use it.
+      Returns true if **buildingname** is a building and the tribe can use it.
```
Noordfrees: Type markup for true/false. Several more below.

src/scripting/lua_map.cc
```diff
@@ -2178,96 +2201,116 @@ int LuaMapObjectDescription::get_name(lua_State* L) {
...
 
 /* RST
-   .. method:: helptext(tribename)
+   .. method:: helptexts(tribename)
+
+      (RO) Returns a :class:`table` of helptexts if it exists for the given tribe, an
+      empty table otherwise.
 
-      Returns the tribe-specific helptexts for this object.
+      Keys are ``lore``, ``lore_author``, ``purpose``, ``note``, ``performance``, all of them
```
Noordfrees, line 2314: Strictly, these only are the keys *commonly found*. You can also define `foobar="helloworld"` in the helptext definition in the units.lua, and this  should be present in the resulting table here.

src/scripting/lua_map.cc
```diff
@@ -4248,7 +4292,9 @@ int LuaTerrainDescription::get_default_resource_amount(lua_State* L) {
 /* RST
    .. attribute:: fertility
 
-         (RO) the :class:`uint` fertility value for this terrain
+      (RO) The fertility value for this terrain as :class:`uint`.
```
Noordfrees: Still some `uint`s left here and below

src/scripting/lua_map.cc
```diff
@@ -7194,12 +7279,51 @@ int LuaShip::build_colonization_port(lua_State* L) {
...
 
       Turns this ship into an expedition ship without a base port. Creates all necessary
-      wares and a builder plus, if desired, the specified additional items.
-      Any items previously present in the ship will be deleted.
-
+      wares and a builder plus, if desired, the specified additional **items**.
       The ship must be empty and not an expedition ship when this method is called.
 
-      :returns: nil
+      Note that the ships :attr:`capacity` is not adjusted if you give additional **itmes**.
```
Noordfrees: itmes → items

src/scripting/lua_map.cc
```diff
@@ -2442,13 +2490,14 @@ int LuaImmovableDescription::has_attribute(lua_State* L) {
 }
 
 /* RST
-   .. method:: probability_to_grow
+   .. method:: probability_to_grow(terrain_description)
 
-      :arg terrain: The terrain that we are checking the probability for.
-      :type terrain: :class:`wl.map.TerrainDescription`
+      Returns a :class:`double` describing the probability that this immovable will grow on the
```
Noordfrees, line 2500: It would be good to have a note that working with `double` values is desync-prone because floating point arithmetic is hardware- and compiler-dependent. Not sure if you want to add this here.

hessenfarmer: Currently this is only used (and was introduced for the help) using this in normal scripting might be dangerous. A Note should clarify this imho.

frankystone: Lua does not know any of the types `integer`, `uint`, `double`, `float`, … . Any numerical value in Lua is of type `number`.

So i don't see a reason why this should be desync-prone or dangerous when using this **value** in scripts. At least it would be not a  fault of a scripts author.

I think the original author of this text just wanted to say: "This value is a floating point number in a range of +x.y until -x.y", not "This is value is of type double".

What about:
> Returns a number in form of a `double` describing …

or:
> Returns a value in form of a `double` describing …

or:
> Returns a floating point number describing …

We should always think about that the Lua documentation is mainly for people who wants to do scripting. The documentation shouldn't be wrong, but also easy to understand for people who are inexperienced in programming. Difficult task though.

src/scripting/lua_map.cc
frankystone: Since this [hadn't worked before](https://github.com/widelands/widelands/pull/4822) (ff) i am in favor of removing this entire function.

src/scripting/lua_map.cc
frankystone: I read it over and over but didn't saw this misspelling :-D

src/scripting/lua_map.cc
frankystone: fixed

src/scripting/lua_map.cc
frankystone: What i do not understand is: If MapObject is the base class of all objects, isn't MapObjectDescription a subclass of MapObject? Shouldn't there be a statement like: "Child of: MapObject"?

src/scripting/lua_map.cc
frankystone: removed all occurences of uint

src/scripting/lua_map.cc
frankystone: I have to think about this... my first impression is: This documentation is for Lua-scripting. Not for explaining the widelands Lua api, which should be part of the c++ documentation.

src/scripting/lua_map.cc
hessenfarmer: The problem about doubles or floating point arithmetics is that if you do them on different platforms the results might differ, due to rounding effects resulting from the internal precision / implementation.      
If you use fractional numbers in a script the result might differ from one computer executing the script to the other which will lead to a desync. 
e.g. 300 * 0.07 < 21 might be true or false depending on the machines floating implementation, while 300 * 0.07 == 21 most probably will be false on both of them.
for this reason we got rid of floating point calculation as far as we could by simply using large integers and doing integer arithmetics on them.
in Lua every number is internally represented as double normally (therefore I assume this is the case in our implementation of lua as well). But the programmer needs to take care to ensure the results are integer. therefore especially in skripting it might be recomendable to use the integer floor division "//" operator which returns the smallest integer close to the result.

see chapter 3.4.1 of https://www.lua.org/manual/5.3/manual.html for example.
see also https://github.com/widelands/widelands/blob/3f3c18dea19ff2cfa449603398c4953ed8b59ecf/src/logic/map_objects/terrain_affinity.cc#L34

I hope that helps. 
therfore my recommendation is twofold we might test to switch this function to return integers (which will be just percentages without any fractional part e.g. 95% instead of 95.45 or 95.95%) as the precision is not needed for the usage in the terrain and immovable help imho, or we just might add a statement to warn the scripter about the possible implications of double arithmetics.

src/scripting/lua_map.cc
Noordfrees: No, there's no inheritance relation between MapObject and MapObjectDescr. We have one inheritance tree for actual instances of objects on the map and one congruent tree for the description of such objects. 
```
         MapObject
        /         \
       Bob         BaseImmovable
   /    |    \           |     \
Ship Critter Worker Immovable PlayerImmovable
               |              /        |     \
             Carrier      PortDock  Building  RoadBase
               |                   /     |…        |  \
             Ferry   ProductionSite Warehouse    Road  Waterway
```
And the same tree *again*, independent from it, with `Description` appended to every name.

src/scripting/lua_map.cc
Noordfrees: The info about the roads and waterways needs to go here because the information stated is plainly wrong (incorrect typename strings). For fleets and battles you could also just write a brief note like

> Ferry fleets ("ferry_fleet"), ship fleets ("ship_fleet"), and battles ("battle") are also map objects, but they can not be accessed from Lua currently.

src/scripting/lua_map.cc
```diff
@@ -4259,7 +4304,9 @@ int LuaTerrainDescription::get_fertility(lua_State* L) {
 /* RST
    .. attribute:: humidity
 
-         (RO) the :class:`uint` humidity value for this terrain
+      (RO) The humidity value for this terrain as.
```
Noordfrees: The sentence seems. 

                    ^ incomplete ;)

src/scripting/lua_map.cc
frankystone: Such an (graphical) overview of all classes is a thing that i really miss in the whole documentation. For years i have tried to make such a graph, but i gave up after some time.

I have rephrased the text. If my rephrasing is not correct, please feel free to correct it.

src/scripting/lua_map.cc
frankystone: I feel not be able to do it correctly. Please feel free to make the changes you think would be more correct.

src/scripting/lua_map.cc
frankystone: I am really tired of this file. I did my best to use correct markup and fixed some things in a way that i think is correct. Because i have just a little knowledge of all the internals i have let the texts as they are, maybe i did some rephrasing.

Sorry, now i feel one makes me responsible for this texts, but i am not. And i feel i am not able to write the correct things.

If some texts are not correct to reflect the internals, or make problems (like the double thing), i would dedicate the task of "proofreading correctness" to people who have more knowledge than me :-)

src/scripting/lua_map.cc
frankystone: fixed.

src/scripting/lua_map.cc
```diff
@@ -2107,17 +2121,22 @@ MapObjectDescription
 
 .. class:: MapObjectDescription
 
-   A static description of a tribe's map object, so it can be used in help files
-   without having to access an actual object on the map.
-   This class contains the properties that are common to all map objects such as buildings or wares.
+   A static description of a map object, so it can be used without
+   having to access an actual object on the map. This class contains the
+   properties that are common to all objects a tribe has, such as buildings or
+   wares. E.g. the tribal encyclopedia is build upon this class.
```
Noordfrees: build → built

src/scripting/lua_map.cc
Noordfrees: Yes this would be very helpful. I could draw such a graph if you like

src/scripting/lua_map.cc
Noordfrees: I have pushed a [commit](https://github.com/widelands/widelands/commit/60dde177adf4a655fe9f81fd211eec287138cd68) to update this and also the dangers of `double`. Please check if you agree with my changes.

src/scripting/lua_map.cc
hessenfarmer: > 
> 
> I am really tired of this file. I did my best to use correct markup and fixed some things in a way that i think is correct. Because i have just a little knowledge of all the internals i have let the texts as they are, maybe i did some rephrasing.
> 
> Sorry, now i feel one makes me responsible for this texts, but i am not. And i feel i am not able to write the correct things.

I am terribly sorry that I may have caused such feelings. I really appreciate the work you do for on the documentation for the better of all scripters. The problem of this special function is surely not your responsibility or your fault. I just wanted to help with some inforamtion so you might understand the issue. (which took me a long time as well to do so). In the end it is good that this was discovered due to your work and we can leave a hint for ll scripters to avoid problems. So again I am glad you put all this effort in this and due to this we are close to a real step forward in documentation

src/scripting/lua_map.cc
```diff
@@ -2491,6 +2496,9 @@ int LuaImmovableDescription::has_attribute(lua_State* L) {
       Returns a :class:`double` describing the probability that this immovable will grow on the
       given terrain. Returns :const:`nil` if this immovable has no terrain affinity.
 
+      Note that floating-point arithmetic is platform-dependent. Using :class:`double`
+      values to make any decisions in the script logic will result in desyncs.
```
hessenfarmer: will --> might result

src/scripting/lua_map.cc
frankystone: No need to feel sorry, hessenfarmer. I apologize for my harsh words.

We all are trying to do our best for this game, each in his/her scope of experience.

hessenfarmer:
@frankystone as Nordfrees agreed to your proposal we might get this ready if you incorporate the last few agreed changes I believe.

frankystone:
Yep, this is today on my ToDo  list. I just needed a pause ;)

src/scripting/lua_map.cc
frankystone: Added a sentence that the table may can contain other keys as well.

src/scripting/lua_map.cc
frankystone: done

src/scripting/lua_map.cc
frankystone: done

Noordfrees (approved):
Great work :)

hessenfarmer:
> 
> 
> Great work :)

I agree a very big task and great work to achieve it. Many thanks

frankystone:
Many thanks for the reviews of this big task!

PR #4826, opened 2021-04-21
Don't let invisible panels handle keys

Noordfrees:
Fixes the bug that the game speed doesn't react to <kbd>Page ↑</kbd>/<kbd>Page ↓</kbd> after closing an in-game window with the <kbd>Esc</kbd> hotkey


matthiakl (approved):
Fix confirmed :)

PR #4830, opened 2021-04-26
add missing pgettext calls in emp04 - fixes #4829

### LINKED ISSUE #4829: Untranslated strings in emp04 (Surprise, Surprise!)
The following strings should use pgettext (in data/campaigns/emp04.wmf/scripting/tribes, rgrep descname | grep -v pgettext):

foresters_house1/init.lua:   descname = "Forester’s House",
foresters_house1/init.lua:         descname = "planting trees",
lumberjacks_house1/init.lua:   descname = "Lumberjack’s House",
lumberjacks_house1/init.lua:         descname = "felling trees",
trainingcamp1/init.lua:   descname = "Training Camp",
trainingcamp1/init.lua:         descname = "sleeping",
trainingcamp1/init.lua:         descname = "upgrading soldier attack from level 0 to level 1",
trainingcamp1/init.lua:         descname = "upgrading soldier attack from level 1 to level 2",
trainingcamp1/init.lua:         descname = "upgrading soldier attack from level 2 to level 3",
trainingcamp1/init.lua:         descname = "upgrading soldier attack from level 3 to level 4",
trainingcamp1/init.lua:         descname = "upgrading soldier health from level 0 to level 1",
trainingcamp1/init.lua:         descname = "upgrading soldier health from level 1 to level 2",
trainingcamp1/init.lua:         descname = "upgrading soldier health from level 2 to level 3",
trainingcamp1/init.lua:         descname = "upgrading soldier health from level 3 to level 4",
mill1/init.lua:         descname = "grinding wheat",
well1/init.lua:         descname = "working",
brewery1/init.lua:         descname = "brewing beer",
farm1/init.lua:   descname = "Farm",
farm1/init.lua:         descname = "working",
farm1/init.lua:         descname = "planting wheat",
farm1/init.lua:         descname = "harvesting wheat",

tothxa:
I'm not sure if I need to care about po template update... also this is my first ever pull request, so I can only hope I'm doing it right...

### DIFF
--- a/data/campaigns/emp04.wmf/scripting/tribes/brewery1/init.lua
+++ b/data/campaigns/emp04.wmf/scripting/tribes/brewery1/init.lua
@@ -53,7 +53,7 @@ descriptions:new_productionsite_type {
 
    programs = {
       main = {
-         descname = "brewing beer",
+         descname = pgettext("empire_building", "brewing beer"),
          actions = {
             "sleep=duration:30s",
             "return=skipped unless economy needs beer",
--- a/data/campaigns/emp04.wmf/scripting/tribes/farm1/init.lua
+++ b/data/campaigns/emp04.wmf/scripting/tribes/farm1/init.lua
@@ -4,7 +4,7 @@ push_textdomain("scenario_emp04.wmf")
 
 descriptions:new_productionsite_type {
    name = "empire_farm1",
-   descname = "Farm",
+   descname = pgettext("empire_building", "Farm"),
    icon = dirname .. "menu.png",
    size = "big",
 
@@ -39,21 +39,21 @@ descriptions:new_productionsite_type {
 
    programs = {
       main = {
-         descname = "working",
+         descname = pgettext("empire_building", "working"),
          actions = {
             "call=plant",
             "call=harvest",
          }
       },
       plant = {
-         descname = "planting wheat",
+         descname = pgettext("empire_building", "planting wheat"),
          actions = {
             "sleep=duration:14s",
             "callworker=harvest"
          }
       },
       harvest = {
-         descname = "harvesting wheat",
+         descname = pgettext("empire_building", "harvesting wheat"),
          actions = {
             "sleep=duration:4s",
             "callworker=harvest"
--- a/data/campaigns/emp04.wmf/scripting/tribes/foresters_house1/init.lua
+++ b/data/campaigns/emp04.wmf/scripting/tribes/foresters_house1/init.lua
@@ -4,7 +4,7 @@ push_textdomain("scenario_emp04.wmf")
 
 descriptions:new_productionsite_type {
    name = "empire_foresters_house1",
-   descname = "Forester’s House",
+   descname = pgettext("empire_building", "Forester’s House"),
    icon = dirname .. "menu.png",
    size = "small",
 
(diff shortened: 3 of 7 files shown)


Noordfrees (approved):
LGTM, thanks for fixing :) 

Updating the PO/POT files and syncing them with Transifex is done every couple of days by me or GunChleoc, so this shouldn't be done in pull requests.

hessenfarmer (commented):
Using pgettext for the programs look weird. I belioeve a leading "_" would have done the trick.

tothxa:
The other the lua scripts in the campaign's tribes directory all use pgettext. Though the campaign's texts.lua indeed uses _, as well as the "original"/main building init.lua-s.

hessenfarmer:
First of all thank you for cleaning this up. 
What I wanted to express is the following:
We should use pgettext for the buildings names as in the main buildings. For the descname of their working programs (e.g. brewing) we use normally a _.

data/campaigns/emp04.wmf/scripting/tribes/farm1/init.lua
```diff
@@ -4,7 +4,7 @@ push_textdomain("scenario_emp04.wmf")
 
 descriptions:new_productionsite_type {
    name = "empire_farm1",
-   descname = "Farm",
+   descname = pgettext("empire_building", "Farm"),
```
hessenfarmer, line 7: this is the buzildings name. Pgettext is correct

data/campaigns/emp04.wmf/scripting/tribes/brewery1/init.lua
```diff
@@ -53,7 +53,7 @@ descriptions:new_productionsite_type {
 
    programs = {
       main = {
-         descname = "brewing beer",
+         descname = pgettext("empire_building", "brewing beer"),
```
hessenfarmer, line 56: this is a name of a working program --> we should use _

Noordfrees:
We actually have a mix of both types. Most programs use `_()` but quite a lot use `pgettext`. Just do a `grep -r 'descname = pgettext' data/tribes/buildings/` and there'll be a lot of program definitions in the output. `pgettext` is actually better for those than `_` because it allows tribe-specific translations which we need if a ware/worker may have different names (soldier, bread, …).

data/campaigns/emp04.wmf/scripting/tribes/brewery1/init.lua
tothxa: Oh, I see. Well, I wasn't sure about using the empire_building context for them, but all strings seemed to use it, like e.g. "No soldier to train!" in trainingcamp1.

hessenfarmer:
Ok, I undestand the reason for using pgettext in the trainingsites now. Shgouldn't we use it for all programs than? first to be consistent second we don't know whether there might be tribe specifics in some language. This would increase the translatable strings though.

tothxa:
I was wondering though if there was a way to make these scripts use the strings directly from the original buildings. That would not only save translation work, but also ensure consistency in case of edits to the originals.

Noordfrees:
This could be done easily by pushing the "tribes" textdomain in these files, but our i18n script will still pick the markup up so the strings will be translated on Transifex anyway. Individual file excludes would be doable but very maintainer-hostile (and therefore likely to accumulate new bugs).

PR #4832, opened 2021-04-27
4746 vision fix

### LINKED ISSUE #4746: Lua: Setting visibility of fields does not work correct after hide_fields()
Hey,

I played the campaign Neptunes Revenge and couldn't finish it because I couldn't find the last artefact. After looking into the map in the editor I found it at (31,3). The point is this area is for some reason never "discovered" (e.g. the fog is  never removed) neither after I tried to discover it with an expedition nor after building a military building close to it.

I attached two screenshots. Looks for me like a bug in the -- I think you call it -- vision,  right?

I'm playing
`Widelands Version 1.0~bzr9796-202102282232 (Release)` from the daily build launchpad repository.

neptunes-revenge-last-artefact
neptunes-revenge-last-artefact-2

hessenfarmer:
added a new lua function unreveal_fields which does the job needed for the scenario. 
needs testing of saveloading and check of other Scenarios.

fixes #4746


data/scripting/field_animations.lua
```diff
@@ -84,7 +87,11 @@ function hide_randomly(plr, region, time)
    if delay < 1 then delay = 1 end
    while #region > 0 do
       local id = math.random(1, #region)
-      plr:hide_fields({region[id]},true)
+      if permanent then
+         plr:hide_fields({region[id]},true)
+      else
+         plr:unexplore_fields({region[id]})
+      end
```
Noordfrees: If you make the new switch the default then this affects all other occurrences of the function. Please make the old behaviour (`permanent=true`) the default, and use the new functionality only for emp03.

hessenfarmer: Well, the hide randomly is only used in emp03. and the new function is exactly what was the default before the vision change from Niektory, so from my point of view I just reinstated the former default, which was changed by the vision code changes. 
Only instance of hide concentric is in emp01 which should be fine as well to be rexplorable.

src/logic/player.cc
```diff
@@ -1547,6 +1547,15 @@ void Player::hide_or_reveal_field(const Coords& coords, HideOrRevealFieldMode mo
...
 		assert(field.vision == VisibleState::kUnexplored);
 		break;
+
+	case HideOrRevealFieldMode::kUnexplore:
+		hide_or_reveal_field(coords, HideOrRevealFieldMode::kUnreveal);
+		if (field.vision == VisibleState::kPreviouslySeen) {
+			field.vision = VisibleState::kUnexplored;
+		}
+		assert(!field.vision.is_revealed());
+		assert(field.vision == VisibleState::kUnexplored);
```
Noordfrees: These assertions do not hold if the fields are still seen by a bob. To reproduce just comment out the `ship:remove()` and you get
```
widelands: ../src/logic/player.cc:1557: void Widelands::Player::hide_or_reveal_field(const Widelands::Coords&, Widelands::HideOrRevealFieldMode): Assertion `field.vision == VisibleState::kUnexplored' failed.
```

data/scripting/field_animations.lua
```diff
@@ -60,7 +60,7 @@ function reveal_randomly(plr, region, time)
 end
 
 -- RST
--- .. function:: hide_randomly(player, region, time)
+-- .. function:: hide_randomly(player, region, time, permanent)
```
Noordfrees:
```suggestion
-- .. function:: hide_randomly(player, region, time[, permanent=false])
```
same below

data/scripting/field_animations.lua
```diff
@@ -74,8 +74,11 @@ end
 --    :arg time: Optional. The time the whole animation will run.
 --               Defaults to 1000 (1 sec)
 --    :type time: :class:`integer`
+--    :arg permanent: Optional. Set to true for permanently hide. Set to false for unexplore.
+--               Defaults to false
```
Noordfrees:
```suggestion
--    :arg permanent: Optional. Set to ``true`` to hide the fields permanently. Set to ``false`` to mark the fields as unexplored.
--                    Defaults to ``false``.
```
same below

and needs a note how this function behaves if a bob currently sees the fields

src/logic/player.cc
hessenfarmer: ok, the first one will hold the second one not. this was a wrong assumption from me. will take care of that tonight. same for the documentation.

frankystone:
If i understand correctly this will add functionality to a (Lua) function, with a new Lua-function, which is not the culprit of the issue. This feels more like a workaround than a real fix.

I do not understand why we can't have `plr:hide_fields(fields, ["hidden"|"normal"|"has_seen"])`. With my little knowledge of this coding stuff i would assume adding the (new) `mode` `kUnexplore` to [plr:hide_fields](https://github.com/widelands/widelands/blob/master/src/scripting/lua_game.cc#L693) and exchange the ternary operator for distinguishing between the three states would do the trick, without the need for a new Lua function.

Anyway, if this is not possible, i am struggling with the wording:

> Set to ``true`` to hide the fields permanently. 

I know what _permanently_ means because i am involved. For a new reader this is unclear, imho. Maybe better:

>> If this is ``true`` the fields will remain black and do not reveal if they get explored e.g. by military buildings or workers. They can only be revealed again by using :meth: `~wl.game.Player.reveal_fields`.

And here i am struggling with _unexplored_:
> Set to ``false`` to mark the fields as unexplored. Defaults to ``false``.

Maybe better:
>> Set to ``false`` to mark the fields as unexplored, so they can be revealed by e.g. military buildings or workers. This is the default.

tothxa:
Is this a candidate for merging? Should I test it?

hessenfarmer:
@frankystone 
you are right this is more a workaround rather than a completely consistent and well elaborated fix. My c++ skils are remarkably low so I took the easy road. But I will give it a try to put everything to hide_fields.

Noordfrees:
This doesn't compile:
```
-- Version of Widelands Build is 1.0~git25290[6c74d4b@HEAD](Debug)
-- Configuring done
-- Generating done
-- Build files have been written to: /home/benedikt/wl/multithreading/build
[219/226] Building CXX object src/logic/CMakeFiles/logic.dir/player.cc.o
FAILED: src/logic/CMakeFiles/logic.dir/player.cc.o 
/usr/bin/g++  -DBOOST_ALL_NO_LIB -DBOOST_SYSTEM_DYN_LINK -DBOOST_UNIT_TEST_FRAMEWORK_DYN_LINK -DUSE_XDG -DWL_USE_GLVND -I../src -Isrc -isystem /usr/include/SDL2 -g    -fext-numeric-literals -std=c++11  -g -DDEBUG -fsanitize=address -fno-omit-frame-pointer -Wredundant-decls -Wall -Wpedantic -Wdeprecated-declarations -Wextra -Wformat -Wformat-nonliteral -Wformat-security -Wformat-y2k -Winit-self -Winvalid-pch -Wlogical-op -Wmissing-include-dirs -Wno-undef -Wold-style-cast -Woverlength-strings -Wpacked -Wpointer-arith -Wsign-promo -Wsync-nand -Wtrampolines -Wundef -Wunused -Wunused-macros -Werror=format-security -Werror=return-type -Werror=shadow -Werror=uninitialized -MD -MT src/logic/CMakeFiles/logic.dir/player.cc.o -MF src/logic/CMakeFiles/logic.dir/player.cc.o.d -o src/logic/CMakeFiles/logic.dir/player.cc.o -c ../src/logic/player.cc
In file included from /usr/include/c++/10/cassert:44,
                 from ../src/sound/sound_handler.h:23,
                 from ../src/logic/player.cc:59:
../src/logic/player.cc: In member function ‘void Widelands::Player::hide_or_reveal_field(const Widelands::Coords&, Widelands::HideOrRevealFieldMode)’:
../src/logic/player.cc:1557:10: error: no match for ‘operator!’ (operand type is ‘Widelands::Vision’)
 1557 |   assert(!field.vision == VisibleState::kPreviouslySeen);
      |          ^~~~~~~~~~~~~
../src/logic/player.cc:1557:10: note: candidate: ‘operator!(bool)’ (built-in)
../src/logic/player.cc:1557:10: note:   no known conversion for argument 1 from ‘Widelands::Vision’ to ‘bool’
../src/logic/player.cc:1557:24: warning: logical not is only applied to the left hand side of comparison [-Wlogical-not-parentheses]
 1557 |   assert(!field.vision == VisibleState::kPreviouslySeen);
      |                        ^~
[222/226] Building CXX object src/scripting/CMakeFiles/scripting_logic.dir/lua_map.cc.o
ninja: build stopped: subcommand failed.
```

Noordfrees:
In bar02 I now get an
```
[00:04:03.752 game] ERROR: Error in Lua Coroutine
[00:04:03.752 game] ERROR: [../src/scripting/lua_errors.cc:22] [string "scripting/mission_thread.lua"]:101: bad argument #2 to 'hide_fields' (string expected, got no value)
[00:04:03.752 game] ERROR: Send message to all players and pause game
```
Version 1.0~git25292[ad9b601@HEAD] (Debug)
The testsuite also doesn't like it.

---

This function is used in a lot of scenarios. Instead of adjusting them all and risking to miss some occurrences, please fall back to the old behaviour if the new argument is `nil`.

src/scripting/lua_game.cc
```diff
@@ -667,21 +667,24 @@ int LuaPlayer::reveal_fields(lua_State* L) {
 }
 
 /* RST
-   .. method:: hide_fields(fields[, unexplore = false])
+   .. method:: hide_fields(fields[, mode = "seen"])
```
frankystone: Please use the same name for parameter. Here this is `mode`, in the field info list (`:arg ...` and `:type ....:`) this is `state`.

src/scripting/lua_game.cc
```diff
@@ -667,21 +667,24 @@ int LuaPlayer::reveal_fields(lua_State* L) {
...
       Undo the effect of :meth:`reveal_fields` on these fields for the current player and
-      optionally completely hide them.
+      optionally completely hide them or reset them to unexplored.
       See also :ref:`field_animations` for animated hiding.
 
       :arg fields: The fields to hide
       :type fields: :class:`array` of :class:`fields <wl.map.Field>`
 
-      :arg unexplore: *Optional*. If  ``true``, the fields will be marked as completely unexplored
+      :arg state: *Optional*. If  ``permanent``, the fields will be marked as completely hidden
```
frankystone: Please add parentheses to the values -> `"permanent"`

"permanent" sounds a bit illogical against "seen" and "explorable", imho.

src/scripting/lua_game.cc
```diff
@@ -667,21 +667,24 @@ int LuaPlayer::reveal_fields(lua_State* L) {
...
          and will not be seen by buildings or workers until they are revealed again
          by :meth:`reveal_fields`.
-         If `false`, They will no longer be permanently visible, but can still be seen by
-         buildings or workers (own or allied), and the player will remember the last seen state.
-      :type unexplore: :class:`boolean`
+         If ``seen``, They will no longer be permanently visible, but can still be seen by
+         buildings or workers (own or allied), and the player will remember the last state that
+         they had been seen. this is the default
```
frankystone: `seen`-> `"seen"`

src/scripting/lua_game.cc
```diff
@@ -667,21 +667,24 @@ int LuaPlayer::reveal_fields(lua_State* L) {
...
+         If ``explorable``, They will no longer be visible, but can still be rediscovered by
+         buildings, ships or workers (own or allied).
+      :type state: :class:`string`
```
frankystone, line 689: `explorable` -> `"explorable"`

As far i understand this is the initial state for undiscovered fields. I think this should be the default (or if no value is given), instead of `"seen"`.

frankystone (changes requested):
Just some nits for the documentation.

src/scripting/lua_game.cc
```diff
@@ -690,8 +693,11 @@ int LuaPlayer::hide_fields(lua_State* L) {
 	Widelands::Player& p = get(L, game);
 
 	luaL_checktype(L, 2, LUA_TTABLE);
-	const Widelands::HideOrRevealFieldMode mode = (!lua_isnone(L, 3) && luaL_checkboolean(L, 3)) ?
+	const std::string state = luaL_checkstring(L, 3);
+	const Widelands::HideOrRevealFieldMode mode = (!lua_isnone(L, 3) && state == "permanent") ?
```
frankystone: Shouldn't something like this do work? (Sorry for the python-like syntax)

```	
        if state == "permanent" then 
              mode = Widelands::HideOrRevealFieldMode::kHide;
        elseif state == "seen" then 
              mode = Widelands::HideOrRevealFieldMode::kUnreveal;
        else 
             mode = Widelands::HideOrRevealFieldMode::kUnexplore;
```

Those nested ternary operators are difficult to understand for me...

hessenfarmer:
> 
> This function is used in a lot of scenarios. Instead of adjusting them all and risking to miss some occurrences, please fall back to the old behaviour if the new argument is `nil`. 

That is exactly what I tried but it seems I failed. Don't know why yet. Help would be appreciated.

src/scripting/lua_game.cc
hessenfarmer: This would require changing all occurences of hide fields then, which was not desired. The previously seen stste was the default until this change if no boolean was provided.

src/scripting/lua_game.cc
hessenfarmer: I don't understand this comment which occurence of seen do you mean

src/scripting/lua_game.cc
```diff
@@ -690,8 +693,11 @@ int LuaPlayer::hide_fields(lua_State* L) {
...
 	                                                 Widelands::HideOrRevealFieldMode::kHide :
+	                                                 (!lua_isnone(L, 3) && state == "explorable") ?
+	                                                 Widelands::HideOrRevealFieldMode::kUnexplore :
 	                                                 Widelands::HideOrRevealFieldMode::kUnreveal;
```
Noordfrees: I guess here's the problem. You assume there's a string in position 3 before checking whether position 3 is nil.
```suggestion
	const std::string state = lua_isnone(L, 3) ? "SOME_DEFAULT_VALUE" : luaL_checkstring(L, 3);
	const Widelands::HideOrRevealFieldMode mode = (state == "permanent") ?
	                                                 Widelands::HideOrRevealFieldMode::kHide :
	                                                 (state == "explorable") ?
	                                                 Widelands::HideOrRevealFieldMode::kUnexplore :
	                                                 Widelands::HideOrRevealFieldMode::kUnreveal;
```

src/scripting/lua_game.cc
hessenfarmer: ok I forgot this one as I needed to change the name due to duplication

src/scripting/lua_game.cc
hessenfarmer: *Head Desk* you are right thanks for the hint

Noordfrees:
What will happen when I call the function with an invalid argument, e.g. `p1:hide_fields(f:region(2), "helloworld")`?

hessenfarmer:
> 
> 
> What will happen when I call the function with an invalid argument, e.g. `p1:hide_fields(f:region(2), "helloworld")`?    
   
in this case state will be "helloworld and this would result in the default beahviour as we only explicitly check for "permanent" and "explorable". At least it is supposed this way

Noordfrees:
Hm, if I wanted to make fields "explorable" and mistyped it as "explrable" then this would result in very hard to find bugs. IMHO it would be safer to report a coroutine error when the script passes an unsupported argument.

src/scripting/lua_game.cc
frankystone: You made it just what my comment was meant to be :-)

src/scripting/lua_game.cc
frankystone: Maybe i do not understand enough of this vision things.

"seen" sounds to me they will be half-visible, like here:

half-visible

If this i not true, which value is used to get something like that?

src/scripting/lua_game.cc
frankystone: I would vote for renaming "permanent" to "hidden". A reader can better distinguish the values then, imho.

src/scripting/lua_game.cc
frankystone: Forget it, i got it now :-)

src/scripting/lua_game.cc
```diff
@@ -667,21 +667,24 @@ int LuaPlayer::reveal_fields(lua_State* L) {
...
       :type fields: :class:`array` of :class:`fields <wl.map.Field>`
 
-      :arg unexplore: *Optional*. If  ``true``, the fields will be marked as completely unexplored
+      :arg state: *Optional*. If  ``"permanent"``, the fields will be marked as completely hidden
          and will not be seen by buildings or workers until they are revealed again
          by :meth:`reveal_fields`.
-         If `false`, They will no longer be permanently visible, but can still be seen by
-         buildings or workers (own or allied), and the player will remember the last seen state.
-      :type unexplore: :class:`boolean`
+         If ``"explorable"``, They will no longer be visible, but can still be rediscovered by
```
frankystone:
```suggestion
         If ``"explorable"``, they will no longer be visible, but can still be rediscovered by
```

src/scripting/lua_game.cc
```diff
@@ -667,21 +667,24 @@ int LuaPlayer::reveal_fields(lua_State* L) {
...
+         buildings, ships or workers (own or allied).
+         If ``"seen"``, They will no longer be permanently visible (fading to foggy), but can
```
frankystone:
```suggestion
         If ``"seen"``, they will no longer be permanently visible (fading to foggy), but can
```

frankystone (commented):
Two typos

src/scripting/lua_game.cc
hessenfarmer: well I tried to find parameter names that work together with the verb hide. So hide permanent made sense to me. Furthermore permanent has more focus on the effect, especially in comparison with explorable.

frankystone: ok then :-)

Thanks for your work!

frankystone (approved):

src/scripting/lua_game.cc
```diff
@@ -690,9 +693,15 @@ int LuaPlayer::hide_fields(lua_State* L) {
 	Widelands::Player& p = get(L, game);
 
 	luaL_checktype(L, 2, LUA_TTABLE);
-	const Widelands::HideOrRevealFieldMode mode = (!lua_isnone(L, 3) && luaL_checkboolean(L, 3)) ?
+	const std::string state = lua_isnone(L, 3) ? "seen" : luaL_checkstring(L, 3);
```
Noordfrees: I'm not happy that you change the parameter type. If there's any scenario anywhere that uses `p1:hide_fields(fields, true|false)` this will suddenly crash now. We need to keep supporting a boolean argument in v1.0 (with a deprecation notice). Check for lua_isboolean and translate true/false to the appropriate string value.

data/scripting/field_animations.lua
```diff
@@ -132,7 +143,7 @@ function reveal_concentric(plr, center, max_radius, hide, delay)
 end
 
 -- RST
--- .. function:: hide_concentric(player, center, max_radius, delay)
+-- .. function:: hide_concentric(player, center, max_radius, delay, permanent)
```
Noordfrees: here too mark the argument as optional
```suggestion
-- .. function:: hide_concentric(player, center, max_radius, delay[, permanent=false])
```

Noordfrees (commented):
Did you test that *The Green Plateau* and *Smugglers* still work as they should after this change? They both use `hide_fields`.

hessenfarmer:
> 
> 
> Did you test that _The Green Plateau_ and _Smugglers_ still work as they should after this change? They both use `hide_fields`.

I did not test them but looked at the code. They use Hide_fields without any parameter resulting in KUnreveal as before. From the code using the kUnreveal mode seems to be fine as the relevant spots are intended to be previously_seen from my understanding.

Noordfrees (commented):
LGTM now, only one more phrasing suggestion

edit: argh, github deleted my inline comment…

src/scripting/lua_game.cc
```diff
@@ -667,21 +667,26 @@ int LuaPlayer::reveal_fields(lua_State* L) {
...
 /* RST
-   .. method:: hide_fields(fields[, unexplore = false])
+   .. method:: hide_fields(fields[, state = "seen"])
 
       Undo the effect of :meth:`reveal_fields` on these fields for the current player and
-      optionally completely hide them.
+      optionally completely hide them or reset them to unexplored.
       See also :ref:`field_animations` for animated hiding.
+      *Note*: hide_fields(fields[, unexplore = false]) is deprecated and will be converted to
+              ``"permanent"`` if true and ``"seen"`` if false.
```
```suggestion
      .. note:: Passing a :class:`boolean` as the *state* argument is deprecated.
         Use ``"permanent"`` instead of ``true`` and ``"seen"`` instead of ``false``.
```

Noordfrees (approved):
Can go in when the checksuite errors are fixed

hessenfarmer:
testsuite green except broken windows. 
As I compile with Msys2 myself I'll merge now

PR #4844, opened 2021-05-02
Empire 04 text improvements

hessenfarmer:
see https://www.widelands.org/forum/topic/5212/

Added help for creating heroes.
added a hint to the clear road network objective
deleted a superfluos hint


data/campaigns/emp04.wmf/scripting/texts.lua
```diff
@@ -597,6 +609,17 @@ amalea_23 = {
...
 }
 
+amalea_24 = {
+   title=_"Amalea in a Hurry",
+   body= amalea1(_"Amalea is very busy…",
+      -- TRANSLATORS: Amalea
+      _([[Brother, as you can see, a lot of things seem to be wrong in this economy. I’ll be very busy to examine everything, so I won’t be here to provide as detailed advice as I used to do.]])
+      .. paragraphdivider() ..
+      -- TRANSLATORS: Amalea
+      _([[But I am very confident, that you will handle everything properly all by yourself. I will report and advice immediately, if I am going to discover new problems, of course.]])),
```
Noordfrees:
```suggestion
      _([[But I am very confident that you will handle everything properly all by yourself. Of course I will report and advise immediately whenever I discover new problems.]])),
```

data/campaigns/emp04.wmf/scripting/mission_thread.lua
```diff
@@ -72,6 +72,8 @@ end
 -- the chaotic road network has to be cleared
 function clear_roads()
    local o = add_campaign_objective(obj_clear_roads)
+   local o1 = add_campaign_objective(obj_clear_roads_2)
+   o1.done = true
```
Noordfrees: This design looks a bit weird to me. IMHO it would be cleaner to not define `o1` at this point yet (`local o1 = nil`), then change line 107 use `o1 = add_campaign_objective(...)` and line 113 to `if o1 then ... done`.

Or even better: According to the doc, `LuaObjective::body` is RW so instead of having two separate objectives you could simply change the first objective's text if needed.

hessenfarmer: Well, changing the text of an objective in mission_thread.lua that is defined in texts.lua would be not very understandable imho. 
And I made good experiences with switchinmg objectives visibility based on external events.
implementing your first suggestion would make no difference though as we would have the 2 objectives and need to control their visibility.

Noordfrees: The way you're doing it here still triggers an additional autosave though doesn't it? The filename of this autosave will suggest that the Clear Roads objective has already been fulfilled which is misleading.

For the clarity you could define the new body as a global variable in texts.lua (just like the objectives) and simply do `o.body = obj_clear_roads_body_2` in the mission_thread. Or stick this in a function `elaborate_clear_roads_objective_text` which is defined in texts.lua.

hessenfarmer: ok, 
I forgot about the autosaves. so I will change this as suggested

hessenfarmer: fixed a second instance of this superfluous savegame as well.

data/campaigns/emp04.wmf/scripting/mission_thread.lua
```diff
@@ -31,16 +31,15 @@ function farm_plans()
    local f = map:get_field(47, 10)
    local farmclick = false
    local count = 0
-   local o1 = add_campaign_objective(obj_click_farmbuilding)
-   o1.done = true
+   local o1 = nil
```
Noordfrees, line 34: Good catch :) 
When you set `o1.done = true` below in line 47 you probably need to check whether `o1` is not `nil` though.

Noordfrees:
Do you want this in v1.0? As nearly all major bugs are fixed by now I'm growing more reluctant about adding new strings that are not necessary for 1.0, so I'd rather to schedule this for 1.1. But if you think this change is important we can have it in 1.0.

data/campaigns/emp04.wmf/scripting/mission_thread.lua
hessenfarmer: implemented. Always easy to find your own wrong habits. (that was the case I mentioned before)

hessenfarmer:
> 
> 
> Do you want this in v1.0? As nearly all major bugs are fixed by now I'm growing more reluctant about adding new strings that are not necessary for 1.0, so I'd rather to schedule this for 1.1. But if you think this change is important we can have it in 1.0.

Well it was a reaction to the forum discussion which made me reevaluate my strings. I believe it improves the scenario and helps the players. As 1.0 is deemed the first ever release I would like to incorporate this. Translation stats are low anyhow (this is a big problem for me but a different one) so I would vote to have this.

Noordfrees (approved):
LGTM, tested and working :)

data/campaigns/emp04.wmf/scripting/texts.lua
```diff
@@ -57,6 +57,11 @@ obj_clear_roads = {
    ),
 }
 
+obj_clear_roads_hint = rt(objective_text(_"Clear Road Network",
+   li(_[[Resolve the chaotic road network by clearing all unnecessary roads.]]) ..
+   li_arrow(_[[Ensure there are not more than 3 dead ends or flags with more than 4 roads.]])
```
tothxa, line 62: At first I interpreted this as "not more than 3 (dead ends or flags with more than 4 roads)", and so I thought it was different from amalea_20 "and not more than one flag with more than four roads". It would be better if it used the same wording.

tothxa: By the way, I still think it would be nice to clarify here (it's not necessary in Amalea's text) that building entrances count as roads.

tothxa:
Oh, and thanks to both of you for letting these go in 1.0.

data/campaigns/emp04.wmf/scripting/texts.lua
hessenfarmer: I tried to shorten the message in the objectives. During translation to german I already realized this could be misunderstood. However the player already received Amaleas reminder when this is displayed so he should know if he read the message. we might change it to ... dead ends or 1 flag with .... though. regarding the explanation of the house entrances, I believe if at alol this should go to Amaleas text and not the objective.

PR #4845, opened 2021-05-02
Skip ca-cert update

### LINKED ISSUE #4838: Github Windows CI is broken
Currently the windows CI Buolds on github are broken.
The error is due to a failed attempt to upgrade ca_certificates while doing the MSYS2 upgrades.
Error message says fatal error - cygheap base mismatch detected   

no clue so far how to solve this

matthiakl:
Quick fix for #4838

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -204,7 +204,7 @@ jobs:
         } else {
           $env:bits = "32"
         }
-        $env:curl = "7.76.1"
+        $env:curl = "7.74.0"
         $env:Path = "C:/msys64/mingw$env:bits/include;C:/msys64/mingw$env:bits/lib;C:/msys64/usr/bin;C:/msys64/mingw$env:bits/bin;$env:Path"
         curl -Lo curl.zip "https://curl.haxx.se/windows/dl-$env:curl/curl-$env:curl-win$env:bits-mingw.zip"
         unzip curl.zip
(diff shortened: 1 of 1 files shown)


hessenfarmer:
In my own installation using the curl 7.76 binaries failed so I am curious whether it will work here (I am using still 7.75 in appveyor using 7.73 is also working). Also I am not sure whether curl will work without the cert update anyway. 
But it is worth a try at least.

matthiakl:
I tested it in my fork, and at least the dependencies were successful and the build started.
As the current mingw curl version is 7.76.1, I also bumped our version,

hessenfarmer:
> I tested it in my fork, and at least the dependencies were successful and the build started.  

Ok, lets wait for the result then.    

> As the current mingw curl version is 7.76.1, I also bumped our version,

that is right but in the 7.76.1 there are a lot more dependencies needed for linking which I didn't get to work. So i still use the libcurl.a from 7.75 which works fine.   
Appveyor still uses 7.73 and it works as well

matthiakl:
> that is right but in the 7.76.1 there are a lot more dependencies needed for linking which I didn't get to work. 

Ok, let's skip this update for now and just get the runner working again.

hessenfarmer (approved):
It works so we should have it for now. 
For the libcurl we could bump the version to download to 7.75.0 to be not too much behind but it is not an urgent thing. As said Appveyor is still on 7.73.

PR #4847, opened 2021-05-04
another typo in emp04

tothxa:
I see everything's getting cleared so I may not be able to finish translating this in time... so here's another typo I have found.


Noordfrees:
Well spotted :) 
No need to wait for the checksuite here

PR #4848, opened 2021-05-05
emp04 clarify objective to clear roads

hessenfarmer:
see comments on original PR #4844


Noordfrees (commented):
Now this could be interpreted as "Ensure there are not more than 3 dead ends and ensure there is one flag with more than 4 roads" ;)

How about? "Ensure there are not more than three dead ends, nor more than one flag with more than four roads."

hessenfarmer:
> 
> 
> Now this could be interpreted as "Ensure there are not more than 3 dead ends and ensure there is one flag with more than 4 roads" ;)
> 
> How about? "Ensure there are not more than three dead ends, nor more than one flag with more than four roads."

Much better. thanks a lot

PR #4852, opened 2021-05-09
more story text change suggestions to emp04

tothxa:
Amalea text: talking about UI elements is just not right for an in-game character, leave it to the objective text (it is already there).

Saledus text: this is the same as what was removed from the objective for being too beginner level.

@hessenfarmer Feel free to change them to your taste.

@Noordfrees Can this still go in 1.0 to finish off the previous changes?

I completed the translation of this scenario, so I don't think I'll have any more.


hessenfarmer (approved):
Ok from my side.

Noordfrees:
We're going into Winter Time Freeze on the 15th of May. Until then string changes are still accepted; afterwards only critical fixes.

PR #4855, opened 2021-05-12
Don't jump to undefined landmarks

### LINKED ISSUE #4854: Bookmarks (Keys1-9) jump to "undefined location" if not set.

**Describe the bug**

With the new Keyboard mapping on macOS the "Bookmark Buttons are CMD-1..9.
Out of old habit I used Ctrl-1..9 (which did not work). But when I then pressed
1 he mao moved to a "random position" Looks like 0,0

**To Reproduce**
1. Open any single or mutiplayer game without bookmarks
2. Press 1..9
3. Map moves to Coordinate 0,0

**Expected behavior**

* The Map will not move until the Bookmark is actually set.
* OR The Game responds with a Message like "Bookmark 2 not set"

**Version:**
 - OS:  macOS BigSur 11.3.1 (20E241)
 - Widelands Version: [1.0~git25300[5a9471a@master]

**Comment**

No need to fix this for Version 1.0 I think this was around since quite a while

Noordfrees:
Fixes #4854
This was introduced by accident during the configurable keyboard shortcuts refactoring. If reviewed before winter-time freeze (15.5.) we can have this fix in v1.0.

### DIFF
--- a/src/wui/quicknavigation.cc
+++ b/src/wui/quicknavigation.cc
@@ -70,7 +70,8 @@ bool QuickNavigation::handle_key(bool down, SDL_Keysym key) {
 			set_landmark(i, current_);
 			return true;
 		}
-		if (landmarks_[i].set && matches_shortcut(
+		if (landmarks_[i].set &&
+		    matches_shortcut(
 		       static_cast<KeyboardShortcut>(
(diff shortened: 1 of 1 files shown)


klaus-halfmann (commented):
This does not really look bad, stll will do a small test, later

klaus-halfmann (approved):
Works as designed, lets go!

PR #4859, opened 2021-05-15
Add Spinbox background

### LINKED ISSUE #4795: Give spin boxes a background
Spin boxes should have a background similar to drop down boxes.

This should make the visual relation between value and arrows more visible.

Mockup:

bg_sliders

In this mockup the borders are missing but i think there should be some borders. So the value is visually in a sunken area, just like the drop downs.

jmoerschbach:
fixes  #4795 by (mis)using an always disabled button for displaying the current spinner's value


Noordfrees (approved):
LGTM, tested and working :) 

Since it's feature freeze merging this will need to wait until after the release.

PR #4860, opened 2021-05-15
Blinking caret

jmoerschbach:
Blinking caret for EditBox and MultilineEditBox as one is used to from text input widgets


src/graphic/wordwrap.cc
```diff
@@ -63,12 +63,17 @@ int text_height(int ptsize) {
 
 namespace UI {
 
+static const int CARET_BLINKING_DELAY = 500;
```
Noordfrees: Please use `constexpr` instead of `static const`. Would it be possible to place this variable somewhere where both `WordWrap` and `EditBox` can access it, so we need it only once?

Noordfrees (changes requested):
This crashes when closing the editor map options with verbose output enabled:
```
terminate called after throwing an instance of 'boost::wrapexcept<boost::io::too_many_args>'
  what():  boost::too_many_args: format-string referred to fewer arguments than were passed
```
Best add code to the ScopedTimer destructor to skip printing a message if the message string is empty

src/graphic/wordwrap.cc
jmoerschbach: changed, I dont see a common place for that constant...

jmoerschbach:
> 
> 
> This crashes when closing the editor map options with verbose output enabled:
> 
> ```
> terminate called after throwing an instance of 'boost::wrapexcept<boost::io::too_many_args>'
>   what():  boost::too_many_args: format-string referred to fewer arguments than were passed
> ```
> 
> Best add code to the ScopedTimer destructor to skip printing a message if the message string is empty

done

Noordfrees:
This needs a master merge now to clean up the clutter in the diff left by the temporarily broken format bot

Noordfrees:
The format bot seems to refuse work on your repo :( 
https://github.com/jmoerschbach/widelands/runs/2651327672
Perhaps the access token secret is outdated?

jmoerschbach:
> 
> 
> The format bot seems to refuse work on your repo :(
> https://github.com/jmoerschbach/widelands/runs/2651327672
> Perhaps the access token secret is outdated?

yeah, thx for the hint. looks good now :)

Noordfrees:
Just a compiler warning left now
```
../src/ui_basic/editbox.h:86:7: warning: 'focus' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
        void focus(bool topcaller = true);
             ^
../src/ui_basic/panel.h:277:15: note: overridden virtual function is here
        virtual void focus(bool topcaller = true);
                     ^
```
Also the blinking looks just a bit too hectic to me. Could you increase the delay to ~1000?

otherwise LGTM

jmoerschbach:
all done :)

Noordfrees (approved):

PR #4865, opened 2021-05-19
Remove pseudo classes HasXy's

frankystone:
I am not happy with the naming of the sections, because they contain functions _and_ attributes. But naming them "Functions and attributes for objects implementing x and y" is too long, imho.

Suggestions welcome :-)

Here are the zipped html files in website style: [current.zip](https://github.com/widelands/widelands/files/6510763/current.zip)


Noordfrees:
I think the section names sound a bit misleading. Suggestions.
- Functions for objects implementing wares → Functions for objects with ware storage
- Functions for objects implementing workers → Functions for objects requiring workers
- Functions for objects implementing soldiers → Functions for objects garrisoning soldiers
- Functions for objects implementing wares and workers → Functions for objects requiring production inputs

Instead of "Functions" there is also possible: (Common) Functionality; (Common) Interface; Shared/Common properties

IMHO it's not necessary to mention that `TrainingSite` has inputs since it inherits these methods from `ProductionSite`.

Would it be possible to hide the `wl.map.object.` part of the method names? It looks strange as there is no "real" module called `wl.map.object`.

frankystone:
Thanks for the suggestions.

Will change them to "Common properties for objects …" and the suggestions you made.

> IMHO it's not necessary to mention that `TrainingSite` has inputs since it inherits these methods from `ProductionSite`.

This does not harm, imho.

> Would it be possible to hide the `wl.map.object.` part of the method names? It looks strange as there is no "real" module called `wl.map.object`.

First i had only `object.…` as the first part, because i thought it gives more clarity to the reader: _"Aha, this function/attribute can be accessed through an object"_. Then i thought this applies to map objects, so i changed it to `map.objects` and then to `wl.map.objects`.

Without changing the module-name the functions and attributes will be prefixed with `wl.map.` which seemed wrong to me. 

Completely removing the prefix is possible by moving the lines:

~~~~
.. module:: wl.map
   :synopsis: Provides access to Fields and Objects on the map

.. moduleauthor:: The Widelands development team
~~~~

to the section "Module Classes", so the name of the module is set below the "Common functions". But this will tear up the main [section :mod:wl.map](https://www.widelands.org/documentation/_sources/autogen_wl_map.rst.txt).

So what to do? 
I am in favor having `object.` as prefix, since the headings also refers to 'objects'. Although there is no module named _"object"_.

Noordfrees:
Could it be changed to `wl.map.MapObject.function_name()`? This would be clearer IMHO as these are functions that are available for some (though not all) MapObjects. If that's not possible let's keep it as `object.` then.

frankystone:
> Could it be changed to `wl.map.MapObject.function_name()`? This would be clearer IMHO as these are functions that are available for some (though not all) MapObjects.

Changed it like so. All other things are also done.

Noordfrees (approved):
LGTM :)

frankystone:
Thanks for your support :-)

Issue #4867, opened 2021-05-20
Broken mingw github builds


matthiakl:
Some update in the github runner broke the mingw build. The working directory is not resolved correctly:
`/usr/bin/python: can't open file '/d/a/widelands/widelands/D:/a/widelands/widelands/utils/detect_revision.py': [Errno 2] No such file or directory`

See for example [this log](https://github.com/widelands/widelands/pull/4865/checks?check_run_id=2627316439#step:5:98)

PR #4869, opened 2021-05-21
Use fallback WL_VERSION when revdetect is broken

Noordfrees:
Re #4867
Does not fix the issue, but a broken revdetect should never be fatal

(Let's see whether the windows builds work in this PR)

doesn't work


hessenfarmer:
I just encounterted another problem on Mingw with libtiff update 
will investigate this evening

hessenfarmer:
Just checked libtiff now requires jbigkit.
so lets see whther I can fix this tonight

PR #4871, opened 2021-05-21
Fix appveyor again

hessenfarmer:


Noordfrees:
#4870 and 4871 individually fail but together should fix it :) 
I'll apply this fix to #4870 to see whether they do work together

PR #4883, opened 2021-05-24
Remove invisible scrollbar in dropdown

### LINKED ISSUE #4317: Dropdown list not responding to mouse move
The dropdown's listselect does not respond to mouse moves when the mouse is moving over the hotkey column.

To reproduce: In the main menu, open the SP dropdown, highlight the first entry, move the mouse to the right until it's over the 'N', then move it straight down → doing so should set the highlight and tooltip to a lower entry, but the highlight remains on the first line and the tooltip always says "Begin a new game".
wl-dropdown-issue

matthiakl:
Fixes #4317


Noordfrees (approved):
LGTM, tested and working :)

PR #4884, opened 2021-05-24
Multiplayer template game

### LINKED ISSUE #2120: Ability to start network game from cli
It would be nice if we could start new network game purely from command line. Specifically what we need is to specify tribes of players, so CLI might look like:

widelands --map [map file] --players frisians,barbarians,barbarians,frisians

There is a lot to be clarified...

It would help for AI training...

 * date created: 2018-10-04T08:40:50Z
 * owner: tiborb95
 * the launchpad url was https://bugs.launchpad.net/bugs/1796042

matthiakl:
Closes #2120 
Extends `--new_game_from_template` with pseudo-multiplayer options to start AI training games.
New options `closed` and `ai` can also be used for finer configuration of single-player games.
See attached template for an example with 3 AIs, 1 closed slot and the player as spectator.
[mp_template.txt](https://github.com/widelands/widelands/files/6534766/mp_template.txt)


src/wlapplication.cc
```diff
@@ -618,29 +646,40 @@ void WLApplication::init_and_run_game_from_template() {
...
 			std::string key = "player_";
 			key += std::to_string(p + 1);
-			init_one_player_from_template(p, settings, profile.pull_section(key.c_str()), map);
+			bool human = p == playernumber - 1;
+			init_one_player_from_template(p, human, settings, profile.pull_section(key.c_str()), map);
 		}
 	}
 
+	if (!settings->can_launch()) {
+		wexception("Inconsistent game setup configuration. Cannot launch.");
```
Noordfrees: *throw

data/templates/new_game_template
```diff
@@ -6,11 +6,15 @@
 
 # Section `global`: General game settings.
 [global]
+# Whether a multiplayer game should be configured. Default: false
+# Note, that this cannot be used for a real multiplayer game, because the host will be the only possible human player
```
Noordfrees: Sounds a bit unclear IMHO. Suggestion:
```suggestion
# Note that this cannot be used for a real multiplayer game, because the game will not be registered on
# the metaserver or in the local network, and therefore no clients or spectators will be able to join.
```

Noordfrees:
If I set `interactive_player="3"` in your template (with slot 3 remaining closed) then a nameless AI is used for player 3 in the game and I'm made a spectator

Noordfrees:
Now setting `interactive_player="3"` gives the expected error, but if I set interactive_player="2" and leave everything else as it is then I'm made a spectator anyway.

It seems that exceptions are not caught anywhere though:
```
Caught exception (of type '10WException') in outermost handler!
The exception said: [../src/wlapplication.cc:515] Cannot close interactive player slot.

This should not happen. Please file a bug report on version 1.0~git25317[350dcf9@HEAD](Release).
```
Instead of throwing exceptions perhaps just add a `log_err()` and `exit(2)`

matthiakl:
> It seems that exceptions are not caught anywhere though

Yes, that's right, but because I use a debug build, I only got a short error message, which I thought was fine:
```
terminate called after throwing an instance of 'WException'
  what():  [/mnt/storage/Programme/widelands/src/wlapplication.cc:515] Cannot close interactive player slot.
```
I will add a more user friendly handling to it.

Noordfrees (approved):
LGTM, working well now

PR #4893, opened 2021-05-26
Add add-on authors to developers

frankystone:
Re: https://github.com/widelands/widelands/discussions/4878
Added Add-on authors to developers.json.

I have to ask some people how they will be named there:
* OIPUN
* the-x
* MarkMcWire
* matthiakl

matthiakl isn't mentioned yet anywhere in this list. I would add him also to the "Coders" section.


data/txts/developers.json
```diff
@@ -373,6 +373,24 @@
 				}
 			]
 		},
+		{
+			"heading": "Add-Ons",
+			"image": "images/players/genstats_player.png",
```
Noordfrees: How about using an image with a scroll (for Scripting) here? E.g. like for the homepage

frankystone: Something like this?

addon_authors

frankystone: Or bronwish:

addon_authors_brown

I think this is better.

In which folder should this be saved? No of the current folders are appropriate, imho.

Noordfrees: I like the first one better. Perhaps place the image in `data/images/ui_fsmenu` since it'll be displayed in the FsMenu's About window.

frankystone: Created a new folder add_ons… maybe we need some more add-on specific images.

frankystone: Sorry didn't saw your post. I have to wait for OIPUN anyway and make the changes then.

data/txts/developers.json
```diff
@@ -373,6 +373,24 @@
 				}
 			]
 		},
+		{
+			"heading": "Add-Ons",
+			"image": "images/add_ons/addons.png",
```
Noordfrees:
```suggestion
			"image": "images/ui_fsmenu/addon_authors.png",
```

Noordfrees (approved):
LGTM :) 
Merging should wait until after the release as it adds a new translatable string

PR #4909, opened 2021-06-02
disable buildhelp while choosing custom starting position

markusu49:


Noordfrees:
I don't think this is desirable. You can still toggle between buildhelp on and off with the Space (or other user-defined shortcut) key. Whether it's shown by default is determined by your start-of-game options settings, and overriding them would be inconsistent. I'd consider it a bug if my settings were ignored sometimes.

Late edit: and one needs to be able to see how the entire buildable land looks around the potential starting positions. It would not do to pick a nice spot and only then realize you have taken the only big plot for miles around.

markusu49:
Yes, those are valid points. The solution is probably not optimal.

But the problem persists: You cannot see valid starting positions while buildhelp is on.

Another idea: a different symbol for valid starting positions? Not an ordinary green building?

frankystone:
Whats the status here?

I guess the two thumbs-ups were related to the other idea:

> Another idea: a different symbol for valid starting positions? Not an ordinary green building?

?

@markusu49 : Do you want to implement this idea?

markusu49:
> @markusu49 : Do you want to implement this idea?

I don't think I can... I don't have any graphics skills and my knowledge of C++ is also very limited.

Maybe I can try the coding part, but no promises and not within the next months.

hessenfarmer:
we would be happy if you could try on the code. For the graphics I would propose to just recolor the current big building plot and use this. A greyish color shoudl work on nearly all maps I believe

tothxa:
I think we should use the HQ marker from the editor for the graphics. As for coding, I once wanted to replace this PR, but I couldn't find the code responsible for showing the possible HQ positions.

tothxa:
Replaced by #6017

PR #4910, opened 2021-06-02
fix graphic glitch in Barbarian flag

tothxa:


Noordfrees (approved):
Thanks :)
Tested and working. A small change so this can go in 1.0

PR #4920, opened 2021-06-04
Fix documentation wl

frankystone:
Reworked documentation for lua_root and lua_bases.

One question: What is the purpose of [set_loading_message](https://www.widelands.org/documentation/autogen_wl_bases/#wl.bases.EditorGameBase.set_loading_message)? Since the progress-bar has been improved (:+1: ) is this function valid anymore (didn't test it)?


Noordfrees:
`set_loading_message` is no longer used anywhere, but it is still functional, and it might be useful for add-ons with a very long postload time.

src/scripting/lua_bases.cc
```diff
@@ -162,11 +167,12 @@ int LuaEditorGameBase::get_players(lua_State* L) {
 /* RST
    .. function:: get_immovable_description(immovable_name)
 
-      :arg immovable_name: the name of the immovable
-
-      Returns the ImmovableDescription for the named object.
+      Returns the description of the named immovable depending on the type of the immovable.
```
Noordfrees: What do you mean by "type"? This is only used for the derived Immovable type such as trees, rocks, wheatfields, but *not* for BaseImmovables such as buildings. an object returned by this function should always have `type_name "immovable"`.

src/scripting/lua_bases.cc
```diff
@@ -186,11 +192,11 @@ int LuaEditorGameBase::get_immovable_description(lua_State* L) {
 /* RST
    .. function:: immovable_exists(immovable_name)
 
-      :arg immovable_name: the name of the immovable
+      Returns whether the descriptions know about an ImmovableDescription for the named
+      object. This is either :const:`true` if the named immovable exists, :const:`false` otherwise.
```
Noordfrees: Linkify ImmovableDescription. 

Grammatical nitpicking: "either" requires an "or", so it should be shortened to "This is true if […], false otherwise"

Oh and the :returns: directive seems to be missing in several places

src/scripting/lua_bases.cc
```diff
@@ -442,16 +442,18 @@ static void save_table_recursively(lua_State* L,
...
 
-      Saves information that can be read by other scenarios.
+      :arg campaign_name: The name of the current campaign, e.g. "empiretut" or "frisians".
+      :type campaign_name: :class:`string`
+      :arg scenario_name: The name of the current scenario, e.g. "emp04" or "fri03".
+      :type scenario_name: :class:`string`
+      :arg data: If an array is used, the data will be saved in the correct order. Arrays may not
+         contain :const:`nil` values. If a table is used, all keys have to be strings. Tables may
+         contain subtables of any depth. Cyclic dependencies will cause Widelands to crash. Values
+         have to be of type :const:`string`, :const:`integer` or :const:`boolean`
```
Noordfrees: …or table

A '.' missing at the end of the last sentence

src/scripting/lua_bases.cc
```diff
@@ -614,7 +619,7 @@ PlayerBase
 
 .. class:: PlayerBase
 
-   The Base class for the Player objects in Editor and Game
+   The Base class for the Player objects in Editor and Game.
```
Noordfrees: Base → base

src/scripting/lua_bases.cc
```diff
@@ -955,14 +974,13 @@ int LuaPlayerBase::place_ship(lua_State* L) {
...
       player already. This will conquer the fields no matter who owns it at the
       moment.
 
-      :arg f: center field for conquering
-      :type f: :class:`wl.map.Field`
-      :arg radius: radius to conquer around. Default value makes this call
-         conquer 7 fields
+      :arg f: Center field for conquering.
+      :type f: :class:`~wl.map.Field`
+      :arg radius: (Optional) Radius to conquer around. Defaults to conquer 7 fields.
```
Noordfrees: 7 fields is misleading as the sentence before is talking about a radius. To the casual reader this sounds like the default is a radius of 7 fields. Perhaps rephrase "Defaults to conquer a radius of 1 (7 fields in total)"?

src/scripting/lua_bases.cc
frankystone: > Oh and the :returns: directive seems to be missing in several places

If it is mentioned in the description i see no reason why we should add it again. If you insist to have *always* the `:returns:` directive i have to rework several places (with the danger of implementing lots of failures :-D ).

frankystone:
Except the `:returns:` directive all should be fixed.

Will refactor the documentation for wl.game also in this pr.

frankystone:
Should be ready now...

This is a never ending story ;)

src/scripting/lua_bases.cc
```diff
@@ -854,16 +864,25 @@ int LuaPlayerBase::place_road(lua_State* L) {
...
+
+      :arg name: The internal name of the building to create.
+      :type name: :class:`string`
+      :arg field: The field to place the building.
+      :type field: :class:`~wl.map.Field`
+      :arg cs: (Optional) If :const:`true` the building is not created directly,
+         instead a constructionsite for this building is placed.
+      :type cs: :class:`boolean`
+      :arg force: (Optional) If :const:`true` the building is forced into
+         existence: The same action is taken as for :meth:`place_flag` when force
```
Noordfrees: missing highlight for "force"

src/scripting/lua_game.cc
```diff
@@ -1047,15 +1074,16 @@ int LuaPlayer::is_attack_forbidden(lua_State* L) {
 }
 
 /* RST
-   .. method:: set_attack_forbidden(who, forbid)
+   .. method:: set_attack_forbidden(playernumber, forbid)
 
       Sets whether this player is forbidden to attack the player with the specified
-      player number. Note that setting this to `false` does not necessarily mean that this
+      **playernumber**. Note that setting this to `false` does not necessarily mean that this
```
```
`false` → :const:`false`
```

Noordfrees (commented):
two more nits then this is good to go :)

frankystone:
All done :-)

Thanks for your support!

Noordfrees (approved):

PR #4935, opened 2021-06-09
Create spritesheets

### LINKED ISSUE #4919: Can't get WL_create_spritesheet to run on a Windows System
see titlle. I ahve compiled this succesfully however it doe not run.   
Do I have to run it from a special directory?   
Does it need to be installed registered in some path?     

I usually use a buildfolder different from the sourcefolder does this matter?   
do I need to use -- for the options like in the normal game?

hessenfarmer:
creates spritesheets for existing animations to reduce number of datafiles.    

fixes #4919


src/website/create_spritesheet.cc
```diff
@@ -374,6 +377,9 @@ void write_animation_spritesheets(Widelands::EditorGameBase& egbase,
  */
 
 int main(int argc, char** argv) {
+	#ifdef _WIN32
+		set_logging_dir();
+	#endif
```
matthiakl: This should also be added to the other website tools, in case some one wants to use them on Windows.
Alternatively put it into the `initialize()` function and output the Usage messages of the website tools directly to std::cerr or std::cout

test/maps/lua_testsuite.wmf/scripting/path.lua
```diff
@@ -22,12 +22,12 @@ function _array_contains(list, string)
 end
 
 function path_test:test_list_files()
-   local dirname = "tribes/buildings/productionsites/amazons/woodcutters_hut/"
-   assert_equal(3, #path.list_files(dirname .. "idle_?.png"))
-   assert_equal("tribes/buildings/productionsites/amazons/woodcutters_hut/idle_1.png", path.list_files(dirname .. "idle_?.png")[1])
-   assert_equal("tribes/buildings/productionsites/amazons/woodcutters_hut/idle_4.png", path.list_files(dirname .. "idle_?.png")[3])
+   local dirname = "tribes/buildings/productionsites/barbarians/barracks/"
+   assert_equal(4, #path.list_files(dirname .. "idle_*.png"))
```
Noordfrees: `path.list_files` can only detect files of the sort `basename_?? with a running number from 0 onwards – the system needed for the old multiple-files animations definitions but obsolete for spritesheets. The function is marked deprecated. This test could just be removed.

hessenfarmer: Well, as far as I undrstood the code it should detect every file with the basename correctly matching the placeholder as an "f" placeholder does not match the 0.5 file it breaks. so with an "*" it should work, to detect the files.    
The test can be removed if we remove the lua function. If I can't get it to work with arrays i will simplify it for this branch as it is already to big a diff

src/website/create_spritesheet.cc
hessenfarmer: +1 but in another branch as this grew already too big

test/maps/lua_testsuite.wmf/scripting/path.lua
Noordfrees: No it only matches this one case and does not work for files in general:

https://github.com/widelands/widelands/blob/408e06450c445d9d1af63929ba224f27ad72d7a7/src/scripting/lua_path.cc#L31-L79
https://github.com/widelands/widelands/blob/408e06450c445d9d1af63929ba224f27ad72d7a7/src/scripting/lua_path.cc#L148-L165

hessenfarmer: ok so first I make it simple to get the tests go green. and we should remove all of the deprecated stuff afterwards.

hessenfarmer:
I finished all buildings and all workers except soldiers.   
Leftovers will be done in a separate branch.
What a huge amount of work

Noordfrees:
This does not start…
```
[00:00:03.291 real] ERROR: ##############################
[00:00:03.291 real] ERROR:   FATAL EXCEPTION: player info: Error while loading tribe 'atlanteans': [../src/scripting/lua_errors.cc:22] tribe atlanteans: Failed adding worker 'atlanteans_toolsmith': [../src/scripting/lua_errors.cc:22] MO: Missing directional animation: Unknown animation: walk_ne for atlanteans_toolsmith
[00:00:03.291 real] ERROR: stack traceback:
[00:00:03.291 real] ERROR: 	[string "tribes/workers/atlanteans/toolsmith/init.lua"]:5: in main chunk
[00:00:03.291 real] ERROR: 	[C]: in method 'new_tribe'
[00:00:03.291 real] ERROR: 	[string "tribes/initialization/atlanteans/units.lua"]:274: in main chunk
[00:00:03.291 real] ERROR: stack traceback:
[00:00:03.291 real] ERROR: 	[string "tribes/initializ
[00:00:03.291 real] ERROR: ##############################
```

Noordfrees:
>  What a huge amount of work

Certainly true. Thanks for your work!

PR #4940, opened 2021-06-11
Write config after sound initialization

### LINKED ISSUE #4939: Sound settings reset to defaults after crash
This has regressed again in current master (03d27be), probably since #4806. Does not happen in 1.0~rc1.
1. Open Widelands and change the sound settings in the options menu.
2. Close and reopen Widelands.
3. Open the options menu and check that the sound settings still are what you just changed them to.
4. Force-close Widelands.
5. Reopen Widelands. The sound settings are now back to the presets.

matthiakl:
Fixes #4939 
Write config not after reading the global section, but after the sound handler is also initialized.


Noordfrees (approved):
Tested and working, code LGTM :)

PR #4960, opened 2021-06-19
Lua class graphs

frankystone:
This is the working solution i have for now for creating lua class graphs for the website (in conjunction with [the website related branch](https://github.com/frankystone/widelands-website/tree/lua_class_graphs))

One can use the commandline switch `-graphs` applied to `extract_rst.py`, e.g. `extract_rst.py -graphs` to create the sphinxdoc directive directly in the *.rst-files.
Currently this does not work if one wants to create such graphs for widelands only (without the website branch). To do so one has to install graphviz and change in `doc/sphinx/source/conf.py`:
~~~~
`extensions = ['sphinx.ext.mathjax']
~~~~
to:
~~~~
extensions = ['sphinx.ext.mathjax', 'sphinx.ext.graphviz']
~~~~
and add the graphviz option:
~~~~
graphviz_output_format = 'svg'
~~~~
This will create the graphs as svg images.

I guess there are more changes needed to the make file for create the documentation at home. I didn't investigated this much. Should i make it work for creating the documentation at home (without the website stuff)?

In make_class_diagram.py are two main settings which restrict the classes in the graph:
~~~~
MAX_CHILDS = 1            # how many child classes are shown
￼MAX_PARENTS = MAX_CHILDS # how many parent classes are clickable
~~~~

Here are the html-files of the current state in website style: [current.zip](https://github.com/widelands/widelands/files/6681243/current.zip)


frankystone:
There is still some work to do:

Since the String `Child of: class_x, class_y` in the documentation is error prone and i have those information in this brach i want to exchange this string automatically with correct values.

And of course some refactoring is surely needed :-D

Noordfrees:
Had a quick look – very good :) 

Multiline tooltips are not working though:
grafik

I find the font hard to read, could you switch to either a sans-serif or a monospaced one, and also make it larger?

frankystone:
> Had a quick look – very good :)

Thanks :)

> 
> Multiline tooltips are not working though:
> grafik

Yes i noticed this also but didn't payed too much attention to this (for now). But i have an idea how to solve this.

> I find the font hard to read, could you switch to either a sans-serif or a monospaced one, and also make it larger?

I just made them smaller :-D Just because i am playing around with the width of the whole graph (which depends on MAX_CHILDS). But i am also in favor to use a non-serif font.

frankystone:
The Problem with bigger fontsize is the overall width of the graph, especially for BuildingDescription:

lua_class_graphs_width


What i can try to make it smaller is:
1. truncation of words, e.g. `ConstructionSiteD...`. Since the full word is inside the tooltip i think this is acceptable (here with MAX_CHILDS=2, showing two rows of children):
    lua_class_graphs_width_2_childs

2. make a different [layout](http://magjac.com/graphviz-visual-editor/?dot=graph%20%7B%0A%20%20main%20%5Bcolor%20%3D%20red%5D%0A%20%20main%20--%20child1%0A%20%20main%20--%20child2%20%5Bminlen%3D2%5D%0A%20%20main%20--%20child3%0A%20%20main%20--%20child4%20%5Bminlen%3D2%5D%0A%20%20main%20--%20child5%0A%7D)

frankystone:
I am fiddling with the css now. While FireFox and Opera renders fine, OtterBrowser has problems rendering MapObjectDescription:

mapobjectdescription_on_otterbrwoser

@hessenfarmer : Can you please check how it looks on windows with some browsers (e.g.: Edge)?
If anyone uses Chrome or Chromium please check as well.

Another thing to look at: Please check if the chapter of `ConstructionSiteDescription` is as height as the image.

Here are current html-files: [current.zip](https://github.com/widelands/widelands/files/6682461/current.zip)
Just unpack them, open an HTML-file in the respective browser and browse to the classes mentioned above.

Noordfrees:
Some units have a trailing '…' that seems to be superfluous:
grafik
Except for "ConstructionSit…" none of them need it, and would "ConstructionSite" not be shorter than "ConstructionSit…"?

Is there a way to exclude the Market and MarketDescription from the graphs as long as their Lua interface is hidden?

ConstructionSiteDescription section has the same height as the image for me

frankystone:
> Some units have a trailing '…' that seems to be superfluous:
> grafik
> Except for "ConstructionSit…" none of them need it, and would "ConstructionSite" not be shorter than "ConstructionSit…"?

Will fix...

> Is there a way to exclude the Market and MarketDescription from the graphs as long as their Lua interface is hidden?

Commenting the definition of LuaMarket and LuaMarketDescription in `lua_map.h` should work.

> 
> ConstructionSiteDescription section has the same height as the image for me

Which browser?

Noordfrees:
Tested with Firefox and ABrowser

Noordfrees:
In the last website-style zip there's a glitch with the graph height for wl.editor.Player:
grafik
(browser: firefox)

And in the locally built version of the latest commit (57c8612) every derived class has an additional "Child of (inserted): …" tag:
grafik

Also without the website styling the lines connecting the boxes are the same colour as the background, is there a way to fix this without having to add a custom CSS for local testing?

frankystone:
> In the last website-style zip there's a glitch with the graph height for wl.editor.Player:
> grafik
> (browser: firefox)

I know why, but have to think how to solve this.

> And in the locally built version of the latest commit ([57c8612](https://github.com/widelands/widelands/commit/57c8612d94e239089c0224dab0ccc531e9e15b55)) every derived class has an additional "Child of (inserted): …" tag:

:-D These are the strings that get _inserted_ automatically. Just wanted to compare this strings with the ones which are still inside the RST-comments, which i want to remove before creating a pull request (sorry for not marking this as a draft)

> grafik
> 
> Also without the website styling the lines connecting the boxes are the same colour as the background, is there a way to fix this without having to add a custom CSS for local testing?

Currently the color is set by [edge [color=white]](https://github.com/widelands/widelands/pull/4960/files#diff-08e556bec064971f9a0849b9008e3e3c2c6a9eb742bd31c467587cb21b284111R262) and not by css. Currently the only solution i know would be to use another color, e.g. the color of links or some sort of gray.

frankystone:
The graph for `PlayerBase` is wrong because `PlayerBase` has two children, namely two classes called  `Player`: One in `lua_editor.h` and one in `lua_game.h`. Difficult to distinguish between the classes because of the same name. Currently my code assumes each class has a different name. Maybe i abandon this branch and try another approach...

What are the correct generic terms for definitions like:
~~~~
class LuaGame : public LuaBases::LuaEditorGameBase
~~~~
Is this more or less correct:
~~~~
class child_class : public base_or_parent_class::parent_class
~~~~
?

And in general are the the terms 'parent' and 'child' acceptable? I know normally a developer is talking about 'derived from', or 'base class', for the one i would call 'parent'.  And how would you name the class which is derived from two classes? Say
~~~~
class A

class B : A

class C : B
~~~~

C is the _…insert term…_ of A.

I would say: C is the 'grandchild' of A.

Noordfrees:
> What are the correct generic terms for definitions like:
> 
> ```
> class LuaGame : public LuaBases::LuaEditorGameBase
> ```
> 
> Is this more or less correct:
> 
> ```
> class child_class : public base_or_parent_class::parent_class
> ```
> 
> ?

The `LuaBases` part is just the namespace. This is a C++ implementation detail which should not be mentioned in the doc at all as it's completely irrelevant for the Lua API. So it's 
```
class child_class : public [namespace::]parent_class
```

> And in general are the the terms 'parent' and 'child' acceptable? I know normally a developer is talking about 'derived from', or 'base class', for the one i would call 'parent'. And how would you name the class which is derived from two classes? Say
> 
> ```
> class A
> 
> class B : A
> 
> class C : B
> ```
> 
> C is the _…insert term…_ of A.
> 
> I would say: C is the 'grandchild' of A.

It depends on the language. In Java the usual terms are *parent class* and *child class*. In C++ a class can have more than one parent and the usual terminology is *base class* and *derived class*. Since Lua does not have native support for class hierarchies it does not really matter how we call it as long as we use it consistently. +1 for parent/child/grandparent/grandchild

frankystone:
Thanks :-)

PR #4973, opened 2021-06-25
Set Win logging dir in all main functions

matthiakl:
Re #4919 
Adds the `set_logging_dir` call to the tools missed in #4935


hessenfarmer (approved):
LGTM.
restarted the jobs

PR #4977, opened 2021-06-27
Updated barbarian barracks graphics by DragonAtma

Noordfrees:


frankystone (approved):
Except that this building uses other textures for the logs it is build with and though looks totally different compared to the other buildings of the Barbarians this PR works as expected.

Approvi Testing.

PR #4979, opened 2021-06-27
make compile.sh show help for invalid arguments ( fixes #4965 )

### LINKED ISSUE #4965: compile.sh should show help when using wrong options
`compile.sh` triggers build on invalid options.

**To Reproduce**
1. Run `./compile.sh -help` or some other unknown option like `-f`
2. The compilation process get triggered immediately

Say one wants to compile a release build and for the option he hits the key 'f' instead of 'r' (on german keybord layout close together). The compilation is triggered but in the end he has a debug build with ASAN.

**Expected behavior**
compile.sh should show the help if a n unknown option was given

**Version:**
 - OS: archlinux
 - Widelands Version: [d94928c@master] (Debug)

tothxa:


tothxa:
fixes #4965

Noordfrees:
This does not work with the `-j` flag: `./compile.sh -r -c -s -w -j 5` gives an error

tothxa:
> This does not work with the `-j` flag: `./compile.sh -r -c -s -w -j 5` gives an error

-j is a whole can of worms... it doesn't check whether its argument is actually a number

Noordfrees (approved):

PR #4980, opened 2021-06-27
compile.sh improvements

tothxa:


tothxa:
@Noordfrees Hey, I was still working on this. :) Could've been included in the previous one...

tothxa:
Actually I'd also like to make it possible to store local defaults in a separate (.gitignored) file (e.g. compile-local-defaults), but I'm not sure how to do it:
- sourcing a file would be simple, but it is insecure
- parsing an options file is hard, and probably not worth the effort

Noordfrees:
Sorry :)

Two further suggestions: Attempting to build with more cores than the system has prints only a message, which can be easily overlooked among the bulk of text and then continues – should this also be an error?
Specifying `-j 0` works, it happily tells me "Using 0 core(s)" – perhaps add another check for this one?

A big +1 for compile defaults! IMHO sourcing a file which exports some variables would be safe if it's git-ignored. Alternatively the compile-default could be a brief script that prints some lines to stdout, and compile.sh would call this script and parse the output. E.g. my compile-defaults script might be 
```
#!/bin/bash
echo -r -c -s -w -j 5
```
and the compile script would interpret the line it prints as the arguments to use. Perhaps unnecessarily complicated…

tothxa:
O.K., I'll fix -j 0, and try something for local options.

tothxa:
This is now all I had in mind.

tothxa:
Should be complete now...

frankystone:
Please keep it simple. The compile script was initially meant to be used by users who just want's to compile the source code without being a developer, imho. Developers were encouraged to compile with CMake options.
If one wants a big shell script for developers with many options i am fine with that, but then we should ship additionally a simple script with two options only: compile either a debug build without asan or a release build (default). Or at least with less options than the current script provides.

I don't see a reason to have an extra file for compile defaults. For me the [bash-history completion](https://wiki.archlinux.org/title/Bash#History)  does the same without trouble, moreover it provides more possibilities (e.g. get the last `./compile` command, change options, press Enter).

Noordfrees:
🚀 

The MAXCORES variable does not work correctly: My compile_defaults contains `-j 5`, so MAXCORES is 6. nproc returns 8. So if I override the defaults to use 7 cores it tells me only 6 cores were available though actually I could use up to 8.

---

@frankystone 

This script is used by developers as well. I use the bash_history here only when recompiling the same branch many times in a row; but if the last compilation was a day ago then my history is already cluttered with masses of java/cmake/make calls from other projects as well as lots of git and bzr operations and so on.

Using CMake directly is a real pain with the numerous options (and their lengthy names) Widelands needs. In another project I always have to remember to type `cmake -DASAN=ON . && make run -j7`. If one argument differs my whole workflow is interrupted. This is about as long a command as I like to type in by hand. I do NOT fancy typing `cmake .. -DCMAKE_BUILD_TYPE:STRING="Release" -DOPTION_BUILD_TRANSLATIONS="ON" -DOPTION_BUILD_WEBSITE_TOOLS="OFF" -DOPTION_ASAN="OFF" -DOPTION_BUILD_CODECHECK="ON" && ninja -j5` every time, nor going through the bash history for *minutes* searching for the last call.

However, now that every switch has an opposite equivalent like -w/+w we *could* consider changing the defaults to something that suits the average end-user more, e.g. Release build, with translations and tests, without website tools and without ASan. Then if you just want to build and play you can type `./compile.sh` without options, and developers can still customise their build very easily without the pain of using cmake directly.

tothxa:
@frankystone 
> Please keep it simple. The compile script was initially meant to be used by users who just want's to compile the source code without being a developer, imho. Developers were encouraged to compile with CMake options.
> If one wants a big shell script for developers with many options i am fine with that, but then we should ship additionally a simple script with two options only: compile either a debug build without asan or a release build (default). Or at least with less options than the current script provides.

Have you looked at the individual commits? I think all of them are quite simple, and all of them only touch the "external" parts of the script: option handling and help text. (I am still not a programmer after all :) ) I did it this way exactly because the resultant total diff is quite big and seems to touch too many things.

I only added one new option variable, EXTRA_OPTS. The new command line options are just the reverse of the already existing ones, which are responsible for half of the added lines, and another third is where I use them in update.sh.

Another seemingly big diff is 8cc3947 (compile.sh: do not print info every time, only with help), but that's only because of the indentation change. All the rest are only a few lines.

So I don't think it's any harder on the user. It's even easier with the possibility of storing preferences.
 
> I don't see a reason to have an extra file for compile defaults. For me the [bash-history completion](https://wiki.archlinux.org/title/Bash#History) does the same without trouble, moreover it provides more possibilities (e.g. get the last `./compile` command, change options, press Enter).

Yet, you filed a bug report to handle the case when the options are wrong, so bash-history doesn't seem to be enough to handle local options... ;)

As for changing options, actually a nice side effect of using read is that you can leave comments or variants for yourself in the file just by putting them in extra lines, and edit the file when you need them, or just cat it to get a reminder for the required overrides.

tothxa:
> As for changing options, actually a nice side effect of using read is that you can leave comments or variants for yourself in the file just by putting them in extra lines, and edit the file when you need them, or just cat it to get a reminder for the required overrides.

Should I explicitly mention this in the help text? It already says that only the first line is used.

tothxa:
Aaargh.... MAXCORES is still not right: if there's only 1 core, MAXCORES will be 2. Indeed, a can of worms...

May I use git amend to correct it, or just add a new commit? This will be squash-merged if accepted, right? Or should I already split up the PR?

BTW is it right that nproc is used unconditionally on all but macs? As far as I can tell, it's a GNU coreutils addition, not POSIX. What platforms is compile.sh supposed to be used on? Isn't GNU coreutils in ports on BSDs?

Noordfrees:
We always use squash-merging so new commits are always fine. The compile script is intended for just about every platform except windows, but so far nobody ever complained that `nproc` was required but not installed so it can be assumed it's safe to use it.

> Should I explicitly mention this in the help text? It already says that only the first line is used.

+1 for a brief notice like "Only the first line of the file is read; further lines can be used as comments."

tothxa:
BTW some web search suggests that hw.ncpu is deprecated on OS X, but it is the way to go on BSD... well, at least it is still said to work on OS X, but if it ever stops, hw.logicalcpu_max seems to be the best candidate replacement. I hate this, and this part really isn't worth all the effort it requires... :(((

Noordfrees:
LGTM :) 
Two more requests, could you please prefix the defaults file with a '.' (`.compile_local_defaults` or more readable IMHO just `.compile_defaults`). 
One thing I would also like to see is an optional environment variable to specify the config file path. E.g. I could keep two files `.compile_defaults` and `.alt_defaults`, and I could use `COMPILE_DEFAULTS=.alt_defaults ./compile.sh` to build with a different set of defaults or `COMPILE_DEFAULTS=0 ./compile.sh` to disable parsing the defaults file.

tothxa:
@Noordfrees 
> Two more requests, could you please prefix the defaults file with a '.' (`.compile_local_defaults` or more readable IMHO just `.compile_defaults`).
> One thing I would also like to see is an optional environment variable to specify the config file path. E.g. I could keep two files `.compile_defaults` and `.alt_defaults`, and I could use `COMPILE_DEFAULTS=.alt_defaults ./compile.sh` to build with a different set of defaults or `COMPILE_DEFAULTS=0 ./compile.sh` to disable parsing the defaults file.

Done. Changed .gitignore as well to use wildcard, so using e.g. .compile_alt (instead of .alt_defaults) does not need further tweaking.

tothxa:
sorry for always realising problems or remembering that I forgot to do or check something after pushing commits... even though I *DO* check my diffs twice or more *before* committing...

Noordfrees (approved):
Perfect. Many thanks :)

PR #4984, opened 2021-06-30
Lua doc enhancements

### LINKED ISSUE #4851: Dependency graphs for the Lua documentation
I really miss a dependency graph for all the Lua classes in the documentation. This would make some things more easy to understand.

Consider using the [graphviz extension](https://www.sphinx-doc.org/en/master/usage/extensions/graphviz.html) and the [inherintance-diagram extension](https://www.sphinx-doc.org/en/master/usage/extensions/inheritance.html) to provide such graphs in our Sphinxdoc documentation.

One can do such a graph by hand, but changing this each time if something has changed, e.g. new classes get added, isn't useful.

frankystone:
New approach for implementing lua class graphs.
Not ready yet.

- [x] Implement "Child of: …"
- [x] Implement graphical dependency trees


frankystone:
Does anyone know why the butten "Update branch" is disabled and the tooltip says "This branch has conflicts with the base branch"?

I can't find any conflicts?

Noordfrees:
GitHub has false positives there recently. You can still use the "update the base branch" link though
grafik

frankystone:
Finally this should be ready now.

The class LuaClasses is a bit overkill, but i wanted to encapsulate some functions.

Changed the color of edges to gray, so they are visible in both website html-files and without website style (using `doc/sphinx/make html`.
Long classnames are hyphenated by uppercase letters, so e.g. `ProductionSiteDescription` becomes:
~~~~
Production–
   Site–
Description
~~~~
The glitch shown here: glitch is an issue of the website style. Don't know if this can be fixed easily.

Current state:
[current.zip](https://github.com/widelands/widelands/files/7768997/current.zip)

frankystone:
Just learned that svg isn't very accessible, e.g. svg images can't use the `alt=`-attribute for alternative text. In order to make this code accessible some small additional changes are needed, e.g. use png as image format and give all nodes an alt-attribute...

frankystone:
Using png as output format to make the graphs more accessible doesn't work. For one thing the alt attribute for nodes aren't implemented, see https://gitlab.com/graphviz/graphviz/-/issues/265, for another thing i can't get tooltips for edges work.

So the current state in this branch is my solution, although without accessibility tags...

doc/sphinx/extract_rst.py
```diff
@@ -110,6 +111,12 @@ def extract_rst_from_cpp(inname, outname=None):
         r = r.expandtabs(4)
         output += r + '\n'
 
+    # Add string 'Child of: …'
+    output = doc_enh.add_child_of(output, outname)
+    if '-graphs' in sys.argv:
+        # Add dependency graph
```
matthiakl, line 117:
```suggestion
        # Add dependency graph
        print("Adding dependency graphs.")
```

matthiakl (commented):
> The class LuaClasses is a bit overkill, but i wanted to encapsulate some functions.

Nothing wrong with it, it helps to understand the code :).

Is it possible to keep the _class_ only one line above the class name if the graph is too wide? Currently it looks like this:
Bildschirmfoto vom 2022-01-10 21-06-39

The command line option should have some kind of output, currently it prints nothing and I was not sure, whether the graphs were actually generated.

doc/sphinx/extract_rst.py
frankystone: Good idea, added a print statement. Not in `extract_rst.py` but in `documentation_enhancements.py` so that only the output mentions real created graphs.

The thing with the `class` not being tied to the class name is an issue of the website css. Unfortunately i have deployed the website related branch just yesterday :-D But yes, this can be solved. Try:
[current.zip](https://github.com/widelands/widelands/files/7845293/current.zip)

frankystone: Related website PR: https://github.com/widelands/widelands-website/pull/370

matthiakl (approved):
Looks good now. The layout is also fine now. 
How is the activity on the website repo, should I approve the PR there, too?

frankystone:
Thanks!

> How is the activity on the website repo, should I approve the PR there, too?

The activity is very low... The last PRs get never approved and in this case (a small css tweak) it is not needed, imho. If you want you can have look into https://github.com/widelands/widelands-website/pull/365

Issue #4911, opened 2021-06-02
AI bugs and glitches in v1.0


hessenfarmer:
1. this line

https://github.com/widelands/widelands/blob/0976a570055e43df6e7847c2831d08f6d09286e7/src/ai/defaultai.cc#L5628

should prevent the AI from building rangers if not at least one lumberjack was built.  With v1.0 we allowed multiple Lumberjacks to be designed in a tribe (e.g. amazons and europeans) so it is not necessarily the related lumberjack for the ranger in question. In fact for amazons and europeans an upgraded Lumberjack is considered.     
As it is not really stopping amazons, due to the upgraded Lumberjack built soon we might postpone this to v1.1. However Europeans will not work properly in AI without that fixed

2. detection of lumberjack is dependent on log used as build material. Should be detected by the immovables harvested.   

3. Detection of second carrier recruitment is not granular enough.     

4. Assertion in line defaultai.cc:3038 prevents rangers from being forced, which is against the concept of the forced_after parameter.    

5. genetic algorithm for determining Neededness of rangers is increasing the score if rangers are not allowed due to woodpolicy (e.g. defaultai.cc:5621). this should be the other way round.

Noordfrees:
Don't know which of the several AI issues this best fits into or if I should open a new one…
I'm currently playing a game on Big Eden (which should be quite AI-friendly) and noticed that 5 of the 7 AIs are unable to develop a wood economy and run into a permanent no-logs-deadlock quickly. 
[Replay for the first 5 hours](https://github.com/widelands/widelands/files/6685437/replays.zip)
[Savegame after 15 hours](https://github.com/widelands/widelands/files/6685488/BigEden_BAR_PH.wgf.zip)
![grafik](https://user-images.githubusercontent.com/48293446/122737106-00be0780-d281-11eb-9b69-f3c9bdfb7000.png)

hessenfarmer:
> 
> 
[...]

PR #4989, opened 2021-07-07
adding enhancement method for buildings to lua

### LINKED ISSUE #4931: Lua command to upgrade buildings
Request to extend the Lua command set to include a command to upgrade buildings.

MarkMcWire:


src/scripting/lua_map.cc
```diff
@@ -5443,6 +5444,25 @@ int LuaBuilding::dismantle(lua_State* L) {
 	return 0;
 }
 
+/* RST
+   .. method:: enhance([keep_wares = false])
+
```
Noordfrees, line 5449:
```suggestion

      .. versionadded:: 1.1

```

I only just remembered that all new Lua functions should have this…

Noordfrees (commented):
LGTM :) One comment inline, not tested yet

MarkMcWire:
I changed the defaultai.cc, too. This line causes an exception in debug build, if you force the rangers building via AI Hint.
There is already a check in line 2839.

src/scripting/lua_map.cc
```diff
@@ -5443,6 +5444,27 @@ int LuaBuilding::dismantle(lua_State* L) {
...
+      versionadded:: 1.1
```
Noordfrees: Almost. RST markup takes some getting used to…
```suggestion
      .. versionadded:: 1.1
```

hessenfarmer:
> 
> 
> I changed the defaultai.cc, too. This line causes an exception in debug build, if you force the rangers building via AI Hint.
> There is already a check in line 2839.

Please do not change any things in a PR that are not related to the title of the PR, because this would make it impossible to have a complete changelog.    
For the behaviour described in the AI bug #4911 I am already working on the fix of it.

hessenfarmer:
Alternatively you may open a new PR providing this as a hot fix with reference to the mentioned bug of course (use Re #XXXX syntax for this)

MarkMcWire:
How can I undoing a commit? I've change the defaultai.cc in my repository, so it would be automatically contribute with every pull request.

hessenfarmer:
> 
> 
> How can I undoing a commit? I've change the defaultai.cc in my repository, so it would be automatically contribute with every pull request.

Well, it is recommendable to not change things in master. Your master should be a copy of the upstream master (aka our main trunk) for your local changes that you need for playtesting developping you could then branch out a seperate branch (for example local_european_development or any other name you want) for providing pull request it is recomendabloe to branch out a fix branch. if it is related to an already known issue it is good practice to prefix it with the issue number but any name related to the content would do. (e.g. for this it could be lua_enhance_buildings. it is important not to have blanks in the branch name though. afterwards you could easily update your local europeans branch by pulling in upstream master, you could do the same with your master to have it consistent. and for every PR shortly before merging to ease merging and to be sure not to have merge conflicts.   

to solve the issue here you might branch out a new branch only containing the hot fix for the defaultai.cc (where the other changes need to be undone) and merge it first. 

@Noordfrees 
Alternatively we could allow this deviation here for one time as I am already working on fixing the complete issues which will be reflected in the changelog then anyway.

Noordfrees:
The Refined GitHub browser extension adds a button to discard all changes to a file in a PR with one click, very convenient :) 

You can also use `git revert ae13bdf` to undo a specific commit

Noordfrees (approved):

PR #4990, opened 2021-07-07
editor toolsize: use spinbox instead of custom dialog

### LINKED ISSUE #4880: The arrows of the toolsize window are reverted
The ordering of the spinner buttons are in most windows like ↓ ↑ in the toolsize window they are vice versa: ↑ ↓

toolsize_arrows_inverted

Maybe because the toosize window is messed up anyway.

**Expected behavior**
All spinner buttons should have the same ordering.

**Version:**
 - OS: as ever
 - Widelands Version: [d15ecd3@master] (Debug)

tothxa:
fixes #4880


tothxa:
Please review carefully, because I'd be lying if I claimed that I almost kind of understand half of what I'm doing here...

I tried to use hmargin() et al instead of the hardcoded 5s, but I couldn't figure out where to #include editor/ui_menus/tool_options_menu.h …

src/editor/ui_menus/toolsize_menu.cc
```diff
@@ -33,38 +33,22 @@ inline EditorInteractive& EditorToolsizeMenu::eia() const {
...
-               25,
-               20,
-               20,
-               UI::ButtonStyle::kWuiSecondary,
-               g_image_cache->get("images/ui_basic/scrollbar_down.png")),
+        &parent, UI::WindowStyle::kWui, "toolsize_menu", &registry, 250, 30, _("Tool Size")),
+     toolsize_spinbox_(this,
+                       5,
+                       5,
+                       get_inner_w() - 2 * 5,
```
jmoerschbach: consider using UI::Box for proper positioning while avoiding magic numbers and manual calculation

jmoerschbach (changes requested):
looking good, not yet tested
Was just about to create more or less same PR but I was some minutes late it seems ;)

src/editor/ui_menus/toolsize_menu.cc
tothxa: As far as I can tell from looking at tool_change_height_options.cc, I have to set margins if I use UI::Box too. Anyhow, at least I was able to make the size adapt to the length of the text in case some translation is longer.

I'm sure using a const member for margin is all wrong though (it seems to me that anything I'm able to parse and understand is considered bad practice in C++... ;) so please let me know what I should use instead.

src/editor/ui_menus/toolsize_menu.h
```diff
@@ -37,12 +36,10 @@ struct EditorToolsizeMenu : public UI::UniqueWindow {
...
 	EditorInteractive& eia() const;
-	void decrease_radius();
-	void increase_radius();
-
-	UI::Textarea textarea_;
-	UI::Button increase_, decrease_;
+	UI::SpinBox spinbox_;
 	uint32_t value_;
+	void changed_();
+	const uint32_t margin = 5;
```
Noordfrees: Just two codestyle nits:
```suggestion
	void changed();
	static constexpr uint32_t kMargin = 5;
```

jmoerschbach:
tested, logic is working 👍 

However, still unsure if we should introduce manual positioning (.set_pos). I thought we want that to be replaced by using UI::Box. What is your opinion on that @Noordfrees ? (And yes, I am aware that this window is fixed and not resized when toggling fullscreen, but still...)

Maybe for inspiration you may look at how I planned to solve this: https://github.com/jmoerschbach/widelands/blob/fix_toolsize_window/src/editor/ui_menus/toolsize_menu.cc (not saying that this is the best way)

Btw: I think long translation will still kind of break the window (if I understand your code correctly, this is what you wanted to prevent):
grafik

Edit: oh, maybe it is just NOT breaking because window's height adjusted to spinbox's (bigger) height. I would have expected that the window's width is increased instead. But increasing height is caused by UI::Spinbox I guess

tothxa:
@jmoerschbach 
> However, still unsure if we should introduce manual positioning (.set_pos). I thought we want that to be replaced by using UI::Box. What is your opinion on that @Noordfrees ? (And yes, I am aware that this window is fixed and not resized when toggling fullscreen, but still...)

I modeled/copied this from tool_change_height_options.cc. Although it uses box, but only for grouping, with fixed (updated once) size and margins. So with only one spinbox, I didn't need an extra box to do the same.

EDIT: well, except for set_pos indeed, which was needed first because changing the parent size reset the position to 0,0, then it turned out to be useful anyway because kMargin is initialised later by clang than by gcc.

> Maybe for inspiration you may look at how I planned to solve this: https://github.com/jmoerschbach/widelands/blob/fix_toolsize_window/src/editor/ui_menus/toolsize_menu.cc (not saying that this is the best way)

Thanks.
I was able to find in the UI:Box implementation that kCenter will center horizontally, but I can't see how vertical alignment is supposed to work.

> Btw: I think long translation will still kind of break the window (if I understand your code correctly, this is what you wanted to prevent):
> grafik
> 
> Edit: oh, maybe it is just NOT breaking because window's height adjusted to spinbox's (bigger) height. I would have expected that the window's width is increased instead. But increasing height is caused by UI::Spinbox I guess

Yes, I think there are some problems with UI::Spinbox in this regard:
* it can break lines and adjust height accordingly, but it will not adjust width if there are long words (I think this is actually multilinetextarea)
* the actual spinbox is always aligned with the first line of the label text, but here it would be better aligned with the last line (and this is box with horizontal alignment)

I only thought of this because it actually happens with change height options in hungarian:
ch_h_opt_en
ch_h_opt_hu

And now I tested it with a long word:
sb_longwords

But fixing spinbox or the other basic UI elements it uses is not the scope of this PR.

Noordfrees:
I think using `UI::Box` for a non-resizing window with just 1 component would be overkill. LGTM to me :)

jmoerschbach (approved):

PR #4991, opened 2021-07-11
Fixed MSYS2 builds

hessenfarmer:
added dwrite as new dependency for harfbuzz static builds


matthiakl (approved):
Verified build with MSVC, even though harfbuzz is still on 2.8.1 there, but since dwrite is a native library it should be fine.
I noticed, that some other libraries we added in the past only are required for msys builds, I will open another PR to link these libraries depending on the compiler.

PR #4992, opened 2021-07-12
Define second stage static dependencies on Windows per compiler

matthiakl:
I noticed, that some static library dependencies we added in the past only are required for MSYS builds, but not MSVC.
Non-native static libraries, which are not a direct dependency for Widelands should be added to the variables in `CMakeList.txt` to differentiate between those two compilers. 
vcpkg (the libtool for MSVC) seems to build the static libraries with less dependencies.


hessenfarmer (approved):
Changes LGTM.
So in future I will add dependencies only to the variables. 
Did you check these are the only ones that MSVC doesn't need?
And one last question how big is the static executable of widelands built with MSVC in comparison to the crosscompiled MSYS executable

matthiakl:
> Did you check these are the only ones that MSVC doesn't need?

At least the other libraries are automatically installed as dependencies, so it compiles fine.

> And one last question how big is the static executable of widelands built with MSVC in comparison to the crosscompiled MSYS executable

The executable of a debug build is 83MB and of a release build 44MB without any special optimization options.
The MSYS2 debug executable (from the github action) is 57MB.

PR #4999, opened 2021-07-21
Fix exception reporting.

SirVer:
During debugging of some issues related to i18n, I realized that thrown exception within the game code are no longer printed to the console. The problem was that a background thread was not joined (but deleted) in case of an exception, which in C++ leads to termination of the whole program.

This patch uses RAII to ensure that this particular thread is always joined properly. However a quick look through the code seems to indicate that similar bugs exist with all other threads that are spawned throughout the program.


SirVer:
I am no longer sure what the proper behavior is for a new pull request, so I went ahead and just requested a review from @Noordfrees.

Noordfrees:
Superseded by #5003

PR #5000, opened 2021-07-21
Adds --verbose-i18n flag.

SirVer:
This flag will print every message that is translated through gettext as it is translated. Should one of the messages trigger a crash, it should be the last message that is printed onto the screen.

This should make the live of translators easier that have issues with certain strings.

This is based on #4999 which must be merged before.


SirVer:
5000th merge request! Hooray!

src/base/i18n.cc
```diff
@@ -62,14 +62,25 @@ std::string locale;
 std::string localedir;
 std::string homedir;
 
+char* (*gettext_)(const char*) = gettext;
+
+char* verbose_gettext(const char* s) {
+	log_info("gettext: %s\n", s);
```
Noordfrees, line 68:
```suggestion
	log_dbg("gettext: %s\n", s);
```

src/wlapplication.cc
```diff
@@ -211,6 +211,31 @@ bool is_autogenerated_and_expired(const std::string& filename,
...
+// not join the thread and this means the exception will never be caught and
+// printed, making debugging hard.
+// See
+// https://stackoverflow.com/questions/25397874/deleting-stdthread-pointer-raises-exception-libcabi-dylib-terminating
+class GameLogicThread {
+	public:
+		explicit GameLogicThread(bool* should_die) : should_die_(should_die), thread_(&UI::Panel::logic_thread) {}
+
+		GameLogicThread(const GameLogicThread&) = delete;
+		GameLogicThread& operator=(const GameLogicThread&) = delete;
```
Noordfrees, line 226: Better use the `DISALLOW_COPY_AND_ASSIGN` macro from base/macros.h for readability.

src/wlapplication.cc
```diff
@@ -211,6 +211,31 @@ bool is_autogenerated_and_expired(const std::string& filename,
...
 }
 
+// A simple wrapper around the game logic thread. It ensures the thread is
+// always joined when this class goes out of scope. This is important when an
+// exception is thrown within the game anywhere, because without this, we will
+// not join the thread and this means the exception will never be caught and
+// printed, making debugging hard.
+// See
+// https://stackoverflow.com/questions/25397874/deleting-stdthread-pointer-raises-exception-libcabi-dylib-terminating
+class GameLogicThread {
```
Noordfrees, line 221: I think this class will be useful for other purposed as well. How about:
- Move it to base/multithreading.h
- Rename it to something like ThreadWrapper
- Modify the constructor to accept any function, not just the default one
- Make `should_die_` optional so it can be `nullptr` for temporarily used threads

klaus-halfmann:
I dont think @SirVer has time to check those changes, let me try hat I can do.

Uhm, I can not push to this repo, will have to push to some other branch, and close this one?

src/wlapplication.cc
SirVer: 2c: [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it), and if I am wrong about this, pull it out at that point. For now, pulling this helper class into a higher scope means that it is more public, and hence harder to change or remove in the future again. Trying to keep code on the most local and private level as possible usually leads to less coupled and easier to change code base. I do not have strong feelings in this specific case.

If you pull it on a higher level, let's not take a pointer to `should_die_`, instead let's take a `std::function` that is called in the place that currently sets `should_die_`. @klaus-halfmann fyi too.

SirVer:
Closing in favor of #5003

PR #5002, opened 2021-07-25
Fix constructionsite enhancement race condition

### LINKED ISSUE #4997: Multiplayer crash on building mines

**Describe the bug**
In the savegame at http://colecovision.eu/stuff/widelands-multiplayer-mine-crash/, with red player hosting the game, and yellow player on LAN, I see a crash (segfault) if the red player builds a few mines, doesn't matter what yellow does. The bug does not happen, if the yellow player is replaced by an AI player, but it does happen even if the yellow player takes no action at all (apart from joining the game).

**To Reproduce**
Start LAN game with loaded savegame as red, have yellow player join. Build a few mines as red (I usually keep the mine windows open, often showing the area, and upgrade the mines). I encountered the issue a few times when building a few iron ore mines, but also once when building a single gold mine. It seems to me the bug is more likely to trigger when I build more mines and keep their windows open, with area displayed, and upgrading them.

**Version:**
 - OS: Debian GNU/Linux testing
 - Widelands Version: reproduced using git from yesterday (I first encountered the bug with git from a week ago)

**Additional context**
There is a workaround: if red does not build any mines for half an hour or so, any subsequently built mines do not trigger the bug.

Noordfrees:
Should fix #4997

### DIFF
--- a/src/logic/map_objects/tribes/constructionsite.cc
+++ b/src/logic/map_objects/tribes/constructionsite.cc
@@ -399,6 +399,7 @@ Start building the next enhancement even before the base building is completed.
 ===============
 */
 void ConstructionSite::enhance(const Game& game) {
+	MutexLock m(MutexLock::ID::kObjects);
 	assert(building_->enhancement() != INVALID_INDEX);
 	Notifications::publish(NoteImmovable(this, NoteImmovable::Ownership::LOST));
 


klaus-halfmann (commented):
This is quite a big function. Could we refactor this into 
ConstructionSite::copy_ware_house_settings()
ConstructionSite::copy_trainingsite_settings()
...
and then apply this to 
ConstructionSite::copy_porductionsite_settings()
?

The lock is quit "fat" for now.
I will try to reproduce and then check the fix anyway ...

klaus-halfmann (approved):
OK, as of comment in #4997

PR #5003, opened 2021-07-28
Verbose i18n

klaus-halfmann:
This superseeds https://github.com/widelands/widelands/pull/5000 and 
https://github.com/widelands/widelands/pull/4999 as I cannot push to @SirVer branches.

* Fix exception reporting in case of Multihreading.
* Adds --verbose-i18n flag.

Still need to find out how to enforce such an exception


src/wlapplication_messages.cc
```diff
@@ -208,6 +208,10 @@ void fill_parameter_vector() {
 		true},
 	  /// Others
 	  {_("Others:"), "verbose", "", _("Enable verbose debug messages"), false},
+	  {"", "verbose-i18n", "",
+		_("Print all strings as they are translated. This helps with tracing down bugs with "
+		  "internationalization."),
+		false},
```
Noordfrees:
```suggestion
		true},
```

src/base/i18n.cc
```diff
@@ -62,17 +62,28 @@ std::string locale;
 std::string localedir;
 std::string homedir;
 
+char* (*gettext_)(const char*) = gettext;
```
Noordfrees, line 65: On second thoughts, I think this will not catch all cases. This only intercepts plain `gettext` but I guess it does not work for `pgettext`, `ngettext` etc? (did not test it yet)

src/wlapplication_messages.cc
klaus-halfmann: This would make it default to true, not sure if this is wise?

src/base/i18n.cc
klaus-halfmann: Not sure where all these are located but I will try to find them.

klaus-halfmann:
I get this Error, and I know I have to add the new include somewhere: 
`.../src/CMakeLists.txt:96: ... DEPENDS on base_macros, because it includes base/macros.h`
But I dont know exactly where. Please give me a hint, perhaps Friday via https://meet.ffmuc.net/WidelandsChat ?

src/wlapplication_messages.cc
Noordfrees: This is not a default value, this boolean only decides whether the option is listed with --help (false) or only with --help-all (true)

Noordfrees:
```
 .../src/CMakeLists.txt:96: ... DEPENDS on base_macros, because it includes base/macros.h
```
Add it in `src/CMakeLists.txt` on a new line after line 108: 
```diff
     ...
     base_i18n
     base_log
+    base_macros
     base_multithreading
     base_random
     ...
```

klaus-halfmann:
Debugging:

* _() uses verbose_gettext
* pgettext boils down to libintl_dcgettext() in libintl.h (macos) dcgettext in src/third_party/gettext/gettext.h
* ngettext boilds donw to libintl_ngettext()  in libintl.h (macos) vi macro in src/third_party/gettext/gettext.h

I dont dare to poke into this dschungle of macros in gettext.h, to may #ifdefs there

Questions
* why do we use /third_party/ ?
* do we need to rethink / modernize the complete tranlsations?

The pull request is in stable status now. shall we have it anyway?

Can we discuss this Friday 19:00 in https://meet.ffmuc.net/WidelandsChat ?

Noordfrees:
`src/third_party` contains various 3rd party code which is not available as a library on some systems so we need to ship it with our code. No changes there unless absolutely necessary please. The better approach is to define some interceptor functions in `base/i18n.*` that act as wrappers around the `…gettext` functions with optional `log_dbg` output.

> do we need to rethink / modernize the complete tranlsations?

Better not, this would involve an enormous refactoring which should not be done unless there's a real urgent need for it imho.

> The pull request is in stable status now. shall we have it anyway?

Since all the pgettext's and stuff are not caught yet I'd consider it only half-complete. This should be addressed first.

> Can we discuss this Friday 19:00 in meet.ffmuc.net/WidelandsChat ?

ok

SirVer:
Thanks for taking this over, I would have time to fix the comments in the next few weeks at some point, but not sure when. I appreciate this being pushed forward.

> Since all the pgettext's and stuff are not caught yet I'd consider it only half-complete. This should be addressed first.

My 2c: do not let the better be the enemy of the good. This can always be added at some point and it is more useful than not having anything (which is the current situation).

PR #5010, opened 2021-07-31
cmake: add support for None build type

craftyguy:
Some distros, e.g. Alpine Linux, like to control setting compiler flags.
cmake supports a "None" build type which uses CFLAGS, CXXFLAGS, etc from
the env and CMakeLists.txt should not set anything.

This change adds support for this so distros can set their own compiler
flags when building and not have them changed by the project build
configuration.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -209,6 +209,8 @@ elseif(CMAKE_BUILD_TYPE STREQUAL "RelWithDebInfo")
   endif ()
   set(WL_DEBUG_FLAGS "-DNDEBUG -DNOPARACHUTE")
   option(OPTION_ASAN "Build with AddressSanitizer" ON)
+elseif(CMAKE_BUILD_TYPE STREQUAL "None")
+  message(STATUS "Not setting any default flags.")
 else()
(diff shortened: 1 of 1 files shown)


klaus-halfmann (approved):
OK for me. @craftyguy can you give us some links / docs how this is finally used?

craftyguy:
Sure, here it is in action in the Alpine Linux package build script (the patch in this MR is applied so that this build step works): https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/23737/diffs#9dff250c217d1d7a06ea6b0812f5b7bc47c80acf_0_44

More info about cmake `None` buildtype and why it is preferred in Alpine Linux: https://lists.alpinelinux.org/~alpine/devel/%3C2896c13070c508a49cbaa72c8fb7f34ea947358b.camel%40cogitri.dev%3E

And the cmake `None` build type reference:
https://gitlab.kitware.com/cmake/community/-/wikis/doc/cmake/Useful-Variables#compilers-and-tools

There are technically a few more build types that cmake supports that this project ignores, in addition to `None`

klaus-halfmann:
@craftyguy feel free to merge this now, or do you need help?

PR #5011, opened 2021-07-31
linux: only enable backtrace when using glibc

craftyguy:
backtrace and backtrace_symbols are not in glibc, not POSIX-compliant
libc (like musl).

### DIFF
--- a/src/logic/map_objects/backtrace.cc
+++ b/src/logic/map_objects/backtrace.cc
@@ -21,8 +21,10 @@
 
 #ifndef _WIN32
 #ifndef __APPLE__
+#if !defined(__linux__) || defined(__GLIBC__)
 
 #include <execinfo.h>
+#endif
 #endif
 #endif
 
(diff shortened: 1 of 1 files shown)


klaus-halfmann (commented):
What problem does this fix?
How can I use https://musl.libc.org/  to compile widelands?

craftyguy:
it fixes compilation of widelands when the libc is musl (and not glibc).

I'm working to package widelands for Alpine Linux: https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/23737

Alpine Linux does not use glibc, it uses musl, and widelands fails to compile since the backtrace stuff is specific to glibc.

craftyguy:
> How can I use https://musl.libc.org/ to compile widelands?

I suggest using a distro that uses musl libc. Alpine Linux and Void Linux are the two that come to mind. I think Arch Linux might have support for installing musl, but it seems like it might be tricky to make sure that musl is used at build time and not glibc..

klaus-halfmann (approved):
Thanks for the infos, please give us a hint about your progress.
Lets Go!

PR #5020, opened 2021-08-06
Update defaultai_warfare.cc

MarkMcWire:
It does not make sense to check the amount of mines to determine whether a trainingsite should be enhanced or not. 

Only the Empire has expandable training buildings (Arena -> Colosseum), no other tribe. So I'm guessing the author of this line wanted to make sure the Empire had enough marble and gold before an arena was upgraded to a colosseum. On the other hand the Colosseum used no weapons and therefore no metals for training. 

But for all other tribe developers it's an unnecessary limitation. For example, the Amazons could not get upgradeable trainingsites because they do not have iron mines and therefore will never obtain this requirement.

If it doesn't cause any side effects, this condition should removed permanently.

### DIFF
--- a/src/ai/defaultai_warfare.cc
+++ b/src/ai/defaultai_warfare.cc
@@ -714,7 +714,7 @@ bool DefaultAI::check_trainingsites(const Time& gametime) {
 
 	const Widelands::DescriptionIndex enhancement = ts->descr().enhancement();
 
-	if (enhancement != Widelands::INVALID_INDEX && ts_without_trainers_ == 0 && mines_.size() > 3 &&
+	if (enhancement != Widelands::INVALID_INDEX && ts_without_trainers_ == 0 &&
 	    ts_finished_count_ > 1 && ts_in_const_count_ == 0) {
 
 		// Make sure that:
--- a/src/ai/defaultai_warfare.cc
+++ b/src/ai/defaultai_warfare.cc
@@ -714,8 +714,7 @@ bool DefaultAI::check_trainingsites(const Time& gametime) {
 
 	const Widelands::DescriptionIndex enhancement = ts->descr().enhancement();
 
-	if (enhancement != Widelands::INVALID_INDEX && ts_without_trainers_ == 0 &&
-	    ts_finished_count_ > 1 && ts_in_const_count_ == 0) {
+	if (enhancement != Widelands::INVALID_INDEX && ts_without_trainers_ == 0 && ts_finished_count_ > 2 && ts_in_const_count_ == 0) {
 
 		// Make sure that:
 		// 1. Building is allowed


hessenfarmer:
I don't know exactly why this had been taken into account, as it was inserted before I joined the project. I always understood it as a measure for the popwer off the economy.
However I fully agree on your explanations about amazons and about empire as well. Problem is that the upgrade is only limited by the proportion of trainingsites in this case so the Ki most probably would do the upgrade to colloseum very early which might leave them in a deadlock of marble. I think the upgrade should be delayd until the economy is capable to sustain the building and to afford it. 
Question is: what would be a good indicator for this.

MarkMcWire:
I modified it again, please try it. If it works, it should be changed this way.
An AI player should have at least 3 trainingssites, before he enhanced one of them.

MarkMcWire:
> Question is: what would be a good indicator for this.

The question is easy to answer: if there is enough building material available. 
An upgrade is nothing more than a new building, just in a place where a building already stands and at a slightly lower cost.

hessenfarmer:
> 
> > Question is: what would be a good indicator for this.
> 
> The question is easy to answer: if there is enough building material available.
> An upgrade is nothing more than a new building, just in a place where a building already stands and at a slightly lower cost.

Ok, I agree. So we should incorporate a check for this in this part of the code . 
Did you test your change with empire?
this line 
https://github.com/widelands/widelands/blob/c49ce3ad2172aa03289b4fd569f70a2fecddf503/src/ai/defaultai.cc#L5363
might interfere. 
furthermore "ts_finished_count_" counts all trainingsites of a player not only the ones of this kind. "bo.cnt_built" is counting the number of a special building.

MarkMcWire:
I will it test this night. I still testing several AI hints (combinations) for europeans at the moment for best AI play.
I will use an empire opponent to test the change above as well.

MarkMcWire:
> furthermore "ts_finished_count_" counts all trainingsites of a player not only the ones of this kind. "bo.cnt_built" is counting the number of a special building.

I know. Since no tribe (besides of mine) had more than 3 different trainingssites, this should work for all original tribes. 
At least, every normal AI would build 2 trainingssites (one of each kind). If he builds one of them several times, he will automatically increase the sum of all trainingssites to 3. At this point, an enhancement should be available.

MarkMcWire:
> Ok, I agree. So we should incorporate a check for this in this part of the code .
> Did you test your change with empire?

Yes I did. It works. The empire still enhance an arena to colosseum, but some time later than normal. The AI player builds arround 6 hours of gameplay a second arena, so he has 3 trainingssites in sum. Than he enhance the first one to a colosseum. At this point he has plenty amount of building materials and no deadlock occur.

hessenfarmer:
> 
> Yes I did. It works. The empire still enhance an arena to colosseum, but some time later than normal. The AI player builds arround 6 hours of gameplay a second arena, so he has 3 trainingssites in sum. Than he enhance the first one to a colosseum. At this point he has plenty amount of building materials and no deadlock occur. 

That is much later then in current implementation, where the Empire AI have finished a colosseum after about 1 and a half hour. So this is definitly not the way to go. I rather would use the critical build material check instead of the number of ts finished.

frankystone:
Empire AI should upgrade an Arena depending on Gold and type of map:

Seafaring map: Upgrade only when a gold mine was found (because a port needs also gold to built)
No Seafaring map: Upgrade immediately

Marble can also be an issue, but normally all maps have enough marble.

hessenfarmer:
we do not have any special checks in the AI yet for the type of map. So this would be a new concept.  

Edit: we have some checks to allow disallow some buildings   

As gold is part of the critical build material I still would go for the same check in the enhancement like for the normal buildings.

MarkMcWire:
> Empire AI should upgrade an Arena depending on Gold and type of map:
> 
> Seafaring map: Upgrade only when a gold mine was found (because a port needs also gold to built)
> No Seafaring map: Upgrade immediately
> 
> Marble can also be an issue, but normally all maps have enough marble.

The problem with this is that it doesn't make sense to hard-code certain rules. Any hard-coded AI rule for the Empire also affects the other tribes. That's how we came here. A hard-coded condition that was never noticed until someone like me developed a tribe, that doesn't build more than 3 mines but use upgradeable trainingsites.

The AI scripts should be tribal independent. What if someone develops a tribe that doesn't need gold at all?
There should also be freely usable AI hints for upgrading buildings and not just for their basic use.

Such a hint could look like the following in the Lua document: 
Build building A only if ware C exceed the value x -> prohibited_until = {gold = 14}
or: Upgrade building B only if ware D exceed the value y. -> prohibited_until = {marble = 20}

In this way, the AI control can be individually adapted to the tribes.

hessenfarmer:
-1 for implementing new and more complicated AI hints. It was decided not long ago to get rid of the AI hints and hardcoding as much as we can and determine most of the Info for the AI from the Lua definitions. 
So still I would prefer the check for critical build materials to decide whether we can afford an enhancement.

MarkMcWire:
> -1 for implementing new and more complicated AI hints. It was decided not long ago to get rid of the AI hints and hardcoding as much as we can and determine most of the Info for the AI from the Lua definitions.
> So still I would prefer the check for critical build materials to decide whether we can afford an enhancement.

Basically, I agree with every path that does not destroy the AI control for the Europeans again. In any case, the decision should / must be independent of which building materials are used.

The condition "marble" or "gold" should not be hard-coded under any circumstances.

hessenfarmer:
> 
> Basically, I agree with every path that does not destroy the AI control for the Europeans again.   

Well, the intention is always to make the AI better for all tribes, including Europeans and even future Addon tribes. But this might include some changes that would require adaptation on the tribes sides including Europeans as well.     

> In any case, the decision should / must be independent of which building materials are used.   
  
Of course. The critical build materials check is checking the build materials and identifies the critical ones from the buildcosts defined.

MarkMcWire:
> -1 for implementing new and more complicated AI hints. It was decided not long ago to get rid of the AI hints and hardcoding as much as we can and determine most of the Info for the AI from the Lua definitions.
> So still I would prefer the check for critical build materials to decide whether we can afford an enhancement.

We already had the same problem the other day during the training itself. Soldiers were not trained if you used a ration instead of meat because the AI expected meat as an input ware for trainingsite.

You solved that by adding substitutes to the list of wares. In the end, however, this reflects the same basic problem: Hard-coded goods requirements in the AI.

hessenfarmer:
> 
> 
> > -1 for implementing new and more complicated AI hints. It was decided not long ago to get rid of the AI hints and hardcoding as much as we can and determine most of the Info for the AI from the Lua definitions.
> > So still I would prefer the check for critical build materials to decide whether we can afford an enhancement.
> 
> We already had the same problem the other day during the training itself. Soldiers were not trained if you used a ration instead of meat because the AI expected meat as an input ware for trainingsite.
> 
> You solved that by adding substitutes to the list of wares. In the end, however, this reflects the same basic problem: Hard-coded goods requirements in the AI.

Correct and this was always inbtended as a quick fix. However they sometimes last longer then expected. 
Edit: I did not add substitutes. I just made the hardcoded check only applicable if meat or fish is used by a TS. Which is not adding new hardcoded stuff.  but still this was just a workaround to provide a quick solution.

MarkMcWire:
> > Basically, I agree with every path that does not destroy the AI control for the Europeans again.
> 
> Well, the intention is always to make the AI better for all tribes, including Europeans and even future Addon tribes. But this might include some changes that would require adaptation on the tribes sides including Europeans as well.
> 

The real basic problem is that the AI still stubbornly refuses to build more than 1 mine of each type. The Europeans build exactly 3 mines per game: 1 coal, 1 iron, 1 gold. Therefore the condition number of mines> 3 could never be fulfilled.

In 10 test games, the AI built 2 gold mines in one single game. No idea why and what triggered this.
Coal and iron mines never be built more than once per type. 
I suspect it has to do with hard-coded conditions again, since I substituted iron_ore and gold_ore with "ore" and both mines produce the same ware despite different resources.

hessenfarmer:
> 
> The real basic problem is that the AI still stubbornly refuses to build more than 1 mine of each type. The Europeans build exactly 3 mines per game: 1 coal, 1 iron, 1 gold. Therefore the condition number of mines> 3 could never be fulfilled.
> 
> In 10 test games, the AI built 2 gold mines once. No idea why and what triggered this.

Well, from what I saw so far the Ai does evaluate that it does not need more mines. They do not get enough prio due to the weighting. This was basically due to it not getting started correctly and fully.
However we already agreed about the suboptimality of the 3 mines condition. so for me this should be replaced by a check for critical build materials as this would be suitable for all tribes and much better achieving the goal.

MarkMcWire:
The good news is that the empire actually builds a Colosseum after just 2-3 hours of gameplay, if you choose the starting point so that the AI can expand quickly.

With the actual code from above.

hessenfarmer:
> 
> 
> The good news is that the empire actually builds a Colosseum after just 2-3 hours of gameplay, if you choose the starting point so that the AI can expand quickly.
> 
> With the actual code from above.

as this is still too long if it could have afforded it earlier I am not convinced this are good news.

@MarkMcWire 
please do not get me wrong. I really appreciate that you are trying to improve the AI and you are getting to understand how it works, as this would give me someone to discuss the issues with. I am only trying to find the best solution for an identified problem with the least negative side effects on legacy tribes. (Ideally the legacy tribes should get some profit as well).

hessenfarmer:
I included a solution in the base branch where i had the other AI fixes. So please check if this branch can be closed

PR #5023, opened 2021-08-12
MSYS Quickfix

hessenfarmer:


CMakeLists.txt
```diff
@@ -362,7 +362,7 @@ IF (WIN32)
     # Fix static library dependencies
     set(SDL_MIXER_EXTRA_LIBS ssp)    
     set(SDL_EXTRA_LIBS mingw32 dxerr8)
-    set(SDL_IMG_EXTRA_LIBS jbig deflate)
+    set(SDL_IMG_EXTRA_LIBS jbig deflate lerc)
```
matthiakl: The library archive is called `libLerc.dll.a`. so try
```suggestion
    set(SDL_IMG_EXTRA_LIBS jbig deflate Lerc)
```

hessenfarmer:
As the new dependency Lerc Codec does not come with a static library yet, downgrading was the only solution for now

matthiakl (approved):

PR #5040, opened 2021-08-25
allow filtering dropdowns

jmoerschbach:
Allows filtering of (open) dropdowns when typing

might be useful for dropdowns with a lot of entries (like the fastplace shortcut feature)

a second list is used to always hold all elements (unfiltered) while the other one might be cleared and filled with only those entries that match the filter (=contains the chars which were typed).
filter is cleared and list restored with all entries upon closing the dropdown list/selecting an entry

Changes to options are needed to avoid maintaining 2 lists (resolutions_ and the actual dropdown list) relying implicitely on both lists being always of equal size and a given index points to the same entry in both lists (which does not hold true now anymore)


Noordfrees:
Cool! :rocket:

A couple of glitches though:
- The behaviour when typing a filter with no matches is strange. Open the options resolution dropdown by clicking on it and type <kbd>Y</kbd>. Visually it seems like the dropdown closes, but if you now click on the dropdown button repeatedly, nothing happens – the dropdown does not reopen until you move the mouse far away and back in. I assume the dropdown remains open even though its 0 elements make it look like it's closed. It either needs to close truly, or (better) display some sort of indicator that no elements match filter XY.
- Backspace doesn't work. That's annoying. It would also be nice to have a way to reset the filter (e.g. <kbd>Esc</kbd>).
- Open a dropdown, mouse over any element, then type a character that is contained in the highlighted entry. The entry under the mouse gets selected unexpectedly.
- There's also no way to see what text is currently filtered for. The icing on the cake would be to highlight the filter match in every visible list entry, but something less fancy would work too.

jmoerschbach:
> 
> 
> Cool! 🚀
> 
> A couple of glitches though:
> 
>     * The behaviour when typing a filter with no matches is strange. Open the options resolution dropdown by clicking on it and type Y. Visually it seems like the dropdown closes, but if you now click on the dropdown button repeatedly, nothing happens – the dropdown does not reopen until you move the mouse far away and back in. I assume the dropdown remains open even though its 0 elements make it look like it's closed. It either needs to close truly, or (better) display some sort of indicator that no elements match filter XY.
> 
>     * Backspace doesn't work. That's annoying. It would also be nice to have a way to reset the filter (e.g. Esc).
> 
>     * Open a dropdown, mouse over any element, then type a character that is contained in the highlighted entry. The entry under the mouse gets selected unexpectedly.
> 
>     * There's also no way to see what text is currently filtered for. The icing on the cake would be to highlight the filter match in every visible list entry, but something less fancy would work too.

thx for the feedback. I addressed all but the last issue. Visualizing the filter would be a non-trivial improvement for another branch :)

jmoerschbach:
why is the workflow reformatting the code and codecheck then complains the exact line which was automatically changed before...?

src/ui_basic/dropdown.cc
```diff
@@ -521,5 +531,11 @@ bool BaseDropdown::handle_key(bool down, SDL_Keysym code) {
 	}
 	return NamedPanel::handle_key(down, code);
 }
+void BaseDropdown::delete_last_of_filter() {
+	if (!current_filter_.empty()) {
+		current_filter_ = current_filter_.substr(0, current_filter_.size() - 1);
```
Noordfrees:
```suggestion
		current_filter_.pop_back();
```

src/ui_basic/dropdown.h
```diff
@@ -208,6 +214,12 @@ class BaseDropdown : public NamedPanel {
 	/// Returns true if the mouse pointer left the vicinity of the dropdown.
 	bool is_mouse_away() const;
 
+	virtual void apply_filter() = 0;
+
+	virtual void save_selected_entry(uint32_t){
+	   // cannot pure because it is called from constructor
+	};
```
Noordfrees: Clang-tidy and our self-written codecheck rules don't always like each other…

Usually a comment does the trick:
```suggestion
	virtual void save_selected_entry(uint32_t)  // cannot be pure virtual because it is called from constructor
	{
	};
```

src/ui_basic/dropdown.h
```diff
@@ -275,9 +287,78 @@ template <typename Entry> class Dropdown : public BaseDropdown {
...
+			add_to_filtered_list(ee.name, ee.value, ee.img, false, ee.tooltip, ee.hotkey);
+		}
+	}
+
+	void save_selected_entry(uint32_t index) override {
+		selected_entry_ = index < filtered_entries.size() ? *filtered_entries[index] : Entry{};
+	}
+
+	bool handle_textinput(const std::string& input_text) override {
+		std::string lower = std::string(input_text);
```
```suggestion
		std::string lower = input_text;
```

src/ui_basic/dropdown.h
```diff
@@ -275,9 +287,78 @@ template <typename Entry> class Dropdown : public BaseDropdown {
...
+		lower[0] = std::tolower(lower[0]);
```
Noordfrees: This could potentially contain multiple characters. Better iterate over the whole string and replace every index with its lowercase equivalent.

src/ui_basic/dropdown.h
```diff
@@ -275,9 +287,78 @@ template <typename Entry> class Dropdown : public BaseDropdown {
...
+		// force list to stay open even if mouse is now away due to smaller
+		// dropdown list because of applied filter
+		set_list_visibility(true);
+	}
+
+	void add_no_match_entry() {
+		// re-add initially selected entry with adapted texts to inform user
+		for (auto& x : unfiltered_entries) {
+			if (x.value == selected_entry_) {
+				add_to_filtered_list(_("No matches"), x.value, x.img, false,
```
Noordfrees: How about changing the text to `"No matches for XYZ"`? boost::format can do this. For coolness try to see if the Listselect supports styling this with an Italic font tag

src/ui_basic/dropdown.h
```diff
@@ -316,12 +399,42 @@ template <typename Entry> class Dropdown : public BaseDropdown {
...
 	}
 
 private:
-	// Contains the actual elements. The BaseDropdown registers the indices only.
-	std::deque<std::unique_ptr<Entry>> entry_cache_;
+	class ExtendedEntry {
+	public:
+		const std::string name;
+		const Entry value;
+		const Image* img;
```
Noordfrees, line 424:
```suggestion
		const Image* img = nullptr;
```

src/ui_fsmenu/options.cc
```diff
@@ -520,6 +473,47 @@ Options::Options(MainMenu& fsmm, OptionsCtrl::OptionsStruct opt)
...
+			   mode.w, mode.h, static_cast<int32_t>(SDL_BITSPERPIXEL(mode.format))};
+			if (this_res.depth == 24) {
+				this_res.depth = 32;
+			}
+			if (this_res != previous) {
+				previous = this_res;
+				const bool selected = !resolution_dropdown_.has_selection() && this_res == current_res;
+				resolution_dropdown_.add(
+				   /** TRANSLATORS: Screen resolution, e.g. 800 x 600*/
+				   (boost::format(_("%1% x %2%")) % this_res.xres % this_res.yres).str(), this_res,
```
Noordfrees: Since you're already touching this:
```suggestion
				   /** TRANSLATORS: Screen resolution, e.g. 800 × 600*/
				   (boost::format(_("%1% × %2%")) % this_res.xres % this_res.yres).str(), this_res,
```

src/ui_fsmenu/options.h
```diff
@@ -121,11 +121,27 @@ class Options : public UI::Window {
...
 	void clicked_cancel();
 
+	// Data model for the screen resolution dropdown
+	class ScreenResolution {
+	public:
+		int32_t xres;
+		int32_t yres;
+		int32_t depth;
+		inline bool operator==(const ScreenResolution& x) const {
+			return xres == x.xres && yres == x.yres;
```
Noordfrees, line 137: Should this also compare `depth`?

src/ui_basic/dropdown.h
Noordfrees: Also once you're using `tolower` and once `std::tolower`…?

Noordfrees (commented):
Some comments on the code, will test it again later today

src/ui_basic/dropdown.h
```diff
@@ -275,9 +287,78 @@ template <typename Entry> class Dropdown : public BaseDropdown {
...
+		// dropdown list because of applied filter
+		set_list_visibility(true);
+	}
+
+	void add_no_match_entry() {
+		// re-add initially selected entry with adapted texts to inform user
+		for (auto& x : unfiltered_entries) {
+			if (x.value == selected_entry_) {
+				add_to_filtered_list(_("No matches"), x.value, x.img, false,
+				                     _("Previously selected entry will be selected again"), x.hotkey);
```
Noordfrees: IMHO this message just confuses the user as it should be obvious that clicking the "No matches" entry doesn't change anything. Better leave the tooltip empty.

Noordfrees:
The behaviour of the <kbd>Esc</kbd> key is inconsistent: In the options menu, it resets the filter or does nothing if no filter is active. In the New Game setup screen, it closes the dropdowns regardless of whether a filter is active or not.

src/ui_basic/dropdown.h
jmoerschbach: no, the message must be as short as possible, because it may cause the display-button to adapt its width if needed be. This may break layouts, especially because we have no control of the filter's length

src/ui_basic/dropdown.h
jmoerschbach: Actually I am more in favor of deleting depth as it is only written, never read. Maybe it was used once upon a time but is not anymore? What is your opinion?

src/ui_fsmenu/options.h
jmoerschbach: Actually I am more in favor of deleting depth as it is only written, never read. Maybe it was used once upon a time but is not anymore? What is your opinion?

src/ui_basic/dropdown.h
jmoerschbach: hmm, I found this very smart :D I thought I might not be sure what if I can expect on an element which says "no matches". Does this tooltip hurt so much?
If you insist, I will delete it...

src/ui_fsmenu/options.h
Noordfrees: Okay let's delete it. Shouldn't matter much on modern hardware…

src/ui_basic/dropdown.h
Noordfrees: How about an empty list item with a tooltip that says "No matches"?

src/ui_basic/dropdown.h
Noordfrees: Turns out my suggestion here makes the compiler unhappy…
Please revert this change, sorry :(

src/ui_basic/dropdown.h
jmoerschbach: might be a good idea, pictorial dropdowns look awkward though. Setting image to nullptr (resulting in a non-displayed entry also for pictorial dropdowns) might cause subtle bugs later on (we don't allow entries with no picture for pictorial dropdowns - see line 300 in dropdown.cc)
Selection_003

jmoerschbach:
> The behaviour of the Esc key is inconsistent: In the options menu, it resets the filter or does nothing if no filter is active. In the New Game setup screen, it closes the dropdowns regardless of whether a filter is active or not.

will investigate, the options dropdown behave special somehow, e.g. also no tooltips are displayed

src/ui_basic/dropdown.h
Noordfrees: You can use `data/images/wui/editor/no_ware.png`

jmoerschbach:
I think it works quite well now :)
The two dropdowns in options do not show the "No match" tooltip and I dont understand why. It might be a z-order problem of the panels (maybe due to to Boxes and Tab?). I checked every dropdown I could find in the software (menus, editor, in-game), all of them are fine...

src/ui_basic/dropdown.h
```diff
@@ -21,8 +21,10 @@
 #define WL_UI_BASIC_DROPDOWN_H
 
 #include <deque>
+#include <graphic/image_cache.h>
```
Noordfrees: Always use `""` for Widelands includes and put it them the last include block. I'm surprised codecheck doesn't complain…

src/ui_fsmenu/options.cc
```diff
@@ -520,6 +473,43 @@ Options::Options(MainMenu& fsmm, OptionsCtrl::OptionsStruct opt)
...
+				   (boost::format(_("%1% × %2%")) % this_res.xres % this_res.yres).str(), this_res,
+				   nullptr, selected);
+			}
+		}
+	}
+
+	if (!resolution_dropdown_.has_selection()) {
+		resolution_dropdown_.add(
+		   /** TRANSLATORS: Screen resolution, e.g. 800 x 600*/
+		   (boost::format(_("%1% x %2%")) % current_res.xres % current_res.yres).str(), current_res,
```
Noordfrees:
```suggestion
		   /** TRANSLATORS: Screen resolution, e.g. 800 × 600*/
		   (boost::format(_("%1% × %2%")) % current_res.xres % current_res.yres).str(), current_res,
```

Noordfrees (commented):
I suspect the problem is that the Options window is a modal panel that is not the top-level panel. Dropdowns hook their list into the highest parent available, I had lots of "fun" with the options menu dropdowns in the past already. This glitch might be just another symptom of this design f~law~eature, feel free to ignore this as unrelated…

Noordfrees:
Another bug (or rather two at once): if I type away in a dropdown and then press space while the mouse is not over any item, the dropdown switches to "Not Selected" in addition to appending the whitespace to the filter.

Noordfrees:
Handling of escape sequences seems a bit wobbly. Feeding Widelands the contents of a binary file as text while a dropdown is open results in some funny behaviour. Is this worth fixing or too much of an easter egg for people who do weird things with programs :P 

*For reference I used `xdotool type --window 65011732 --file ~/.widelands/maps/Downloaded/Hexagony.wmf --delay 2000`, the first chars until xdotool aborts are `50 4b 03 04 14 00`*

jmoerschbach:
> Handling of escape sequences seems a bit wobbly. Feeding Widelands the contents of a binary file as text while a dropdown is open results in some funny behaviour. Is this worth fixing or too much of an easter egg for people who do weird things with programs :P
> 
> _For reference I used `xdotool type --window 65011732 --file ~/.widelands/maps/Downloaded/Hexagony.wmf --delay 2000`, the first chars until xdotool aborts are `50 4b 03 04 14 00`_

huh, let's leave that as an easter egg :) It does not crash, right?

Noordfrees:
Another bug: If you press <kbd>Esc</kbd> to clear the active filter while the mouse is over an element in the dropdown, the moused-over element gets selected.

Noordfrees (approved):
Wonderful feature :)

PR #5047, opened 2021-08-28
Atlantean and Amazon soldier prefs closes #5044

### LINKED ISSUE #5044: Atlantean tower is set to weak soldiers by default
Usually, larger military buildings default to strong soldiers, while small ones default to weak soldiers. However, the Atlantean tower is an exception, as it is the Atlantean's second-strongest military building, but defaults to weak soldiers. I guess that this is a mistake, rather than a design choice.

alasiax:
Sets medium military buildings with 4+ soldiers to prefer heroes, and the Atlantean small tower with 3 soldiers to prefer rookies.
fixes #5044

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,6 @@ For helping with issue management, see https://www.widelands.org/wiki/TriagingBu
 
 ## Obtaining MacOS and MS-Windows builds and testsuite runs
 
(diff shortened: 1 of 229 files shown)


alasiax:
Ugh, the bulk of the listed changes were already comitted, sorry. It's only the three military sites files that changed.

hessenfarmer (approved):
LGTM. 
Next time. you should create a branch from actual master and pull in upstream master in your fork to avoid such problems.
if you want to link the PR to an issue you need to add "fixes #....." to the PR comment. I did this here.

PR #5054, opened 2021-08-30
Unify arrow and +/- handling for spinbox and slider

tothxa:
* fix handling of number keys for sound volume sliders


tothxa:
Also from the mousewheel branch, but there may be objections to this and I didn't ask for comments before.

Currently the value of a spinbox can be changed with the arrow keys (up/down as well as left/right), while a slider with +/-, but not the other way round. I think the functionality is very similar, the difference is mostly in the graphical implementation, so they should respond to the same inputs. (this is for keys, but I intend to use a common set of mousewheel config options for these two and a few other UI elements with the functionality to change some value)

The main objection I expect is that there is code for horizontal and vertical sliders, so these may need to only respond to the corresponding arrow keys. But currently only horizontal sliders are used in the codebase.

The number keys fix for wide range (like sound) sliders came first, but I didn't want to separate the changes. I'll do so only if the unification is rejected.

Noordfrees:
LGTM and working nicely :) 
There's one glitch though: The NumLock state doesn't seem to have any effect, it's always treated as if NumLock is off. However my keyboard layout is slightly customised so a raw NumLock keystroke is interpreted as the Compose key and only Shift+NumLock actually toggles the state; so I'm not sure if this behaviour is a bug in the code or in SDL's modifier detection…

tothxa:
> There's one glitch though: The NumLock state doesn't seem to have any effect, it's always treated as if NumLock is off. However my keyboard layout is slightly customised so a raw NumLock keystroke is interpreted as the Compose key and only Shift+NumLock actually toggles the state; so I'm not sure if this behaviour is a bug in the code or in SDL's modifier detection…

The same NumLock state check is already present in the spinbox code in master for keypad 2 - down, 3 - PgDn, 4 - left, 6 - right, 8 - up and 9 - PgUp, as well as in a few other places, so it's probably not a regression…

It works on my laptop which doesn't have a dedicated keypad, only keypad emulation with the Fn key. (e.g. normally, with NumLock:off Fn+U is Keypad Left, but with NumLock:on plain U is Keypad 4 and Fn+U is `u`)

Noordfrees (approved):
Confirmed that other widgets also have this problem. Let's have it :)

tothxa:
Before it gets merged:

It's 2 very long `switch`-es with a lot of common checks, so I wonder if it should be moved to a common helper function. (I'd also add it to constructionsite soldier capacity as in https://github.com/tothxa/widelands/commit/8caa33d9b0b929ce7d012ec8e31f1eb2afdb9d7f) Maybe in wlapplication_options, though it's not quite option handling... I'm thinking of something like `int32_t get_keyboard_change(SDL_Keysym)` where the return value is from
`enum { kSetMin = INT_MIN, kBigMinus = -10, kMinus = -1, kNone = 0, kPlus = 1, kBigPlus = +10, kSetMax = INT_MAX };`

Also I don't think there's any function where keypad keys are used for different function from their normal counterpart, and I think configurable hotkeys are even made to match either way. So maybe they should be converted (by another helper function) in the event loop before calling the handlers, so there would be no need for the double sym + KMOD_NUM checks everywhere.

tothxa:
Should be done now...

I found and fixed a bug while removing SDLK_KP_-s and testing (the bug is present in 1.0 too): North-Western diagonal scrolling didn't work, because SLDK_KP_7 was intercepted by go back to HQ.

tothxa:
Sorry for the rushed commit.

So now all numpad movement is reserved when diagonal scrolling is enabled.

The other possible solution would be to check the scancodes (which are left alone by the normalisation) in the consumers, but that's only feasible in the default users of these keys.

src/wlapplication_options.h
```diff
@@ -231,6 +233,20 @@ std::string shortcut_string_for(KeyboardShortcut, bool rt_escape = true);
...
 
+// Return values for changing value of spinbox, slider, etc.
+enum class ChangeType : int32_t {
+	kSetMin = INT32_MIN,  // set value to minimum -- keys: Home, Ctrl + decrease keys
+	kBigMinus = -10,      // decrease by big step -- key: PageDown
+	kMinus = -1,          // decrease  -- keys: Left, Down, Minus
+	kNone = 0,            // no change -- all other keys
+	kPlus = 1,            // increase  -- keys: Right, Up, Plus
+	kBigPlus = 10,        // increase by big step -- key: PageUp
+	kSetMax = INT32_MAX   // set value to minimum -- keys: Home, Ctrl + increase keys
```
Noordfrees: please use `std::numeric_limits<int32_t>::max()`/`min()` for consistency

Noordfrees (approved):
LGTM, just one nit.
I couldn't test the NumLock on/off difference and Keypad5 usage because those keys are too heavily remapped on my keyboard by now, other than that tested and working.

tothxa:
> I couldn't test the NumLock on/off difference and Keypad5 usage because those keys are too heavily remapped on my keyboard by now, other than that tested and working.

With some more testing, I think Keypad5 should also be left alone for diagonal scrolling, and specially checked for in interactive_player.cc for the go home functionality. Now the replacement SDLK_HOME can be captured by other handlers.

src/wui/interactive_player.cc
```diff
@@ -774,7 +774,9 @@ bool InteractivePlayer::handle_key(bool const down, SDL_Keysym const code) {
 			set_display_flag(dfShowWorkareaOverlap, !get_display_flag(dfShowWorkareaOverlap));
 			return true;
 		}
-		if (matches_shortcut(KeyboardShortcut::kInGameScrollToHQ, code)) {
+		if (matches_shortcut(KeyboardShortcut::kInGameScrollToHQ, code) ||
+		    (get_config_bool("numpad_diagonalscrolling", false) &&
+		     code.scancode == SDL_SCANCODE_KP_5)) {
```
Noordfrees: Actually, I think going by scancode here is the wrong approach. I tried the following:
1. Enable diagonal scrolling
2. Assign the physical key Keypad5 (which is mapped to <kbd>=</kbd> using xkb magic) to the Open Message Menu action
3. Press <kbd>KP5</kbd>/<kbd>=</kbd> in the game → Message menu opens, keypress is ignored.
4. Reset the Message shortcut
5. Now <kbd>KP5</kbd>/<kbd>=</kbd> acts as a go-to-HQ action.

Since KP5 is a unique key on the keyboard, unlike the other KP keys which act as duplicates of other keys, it probably should not get their special treatment IMHO.

Noordfrees (changes requested):

src/wui/interactive_player.cc
tothxa: I think the problem is that on your remapped keyboard SDL_Keysym.sym is probably '=', so it doesn't get rewritten by normalize_numpad, which looks for SDL_Keysym.sym. So to get the same effect even if the numpad is remapped on the system level, normalize_numpad should look for the scancodes as well when diagonal scrolling is requested, and change sym to the KP ones (or SDLK_UNKNOWN, as scancodes are used by the diagonal scrolling handler anyway), so 3. above would also go to HQ. This scancode based rewriting may make some sense, because the whole diagonal scrolling thing is done because of the physical layout of the keys.

Though if the code here checked for the KP5 keysym.sym, at least you'd never get the go to HQ action, ie. 5. would do nothing.
But if you remap the other KP numbers too on the system level, and assign actions that are handled before the mapview.cc handler gets its chance, then scrolling in the affected directions doesn't work either. This is what the KP 1/3/7/9 check in interactive_gamebase.cc was preventing for the default PgUp/PgDn mappings. Except that KP7 (home) is used by default here in interactive_player.cc, which never had such a test, and so stole SDLK_KP_HOME, and prevented NW scrolling. But since this bug was never reported, it looks like diagonal scrolling isn't used much anyway, so it might as well be dropped completely... ;)

Noordfrees: > So to get the same effect even if the numpad is remapped on the system level

But is that desired? Let's assume that I preferred to have the numpad on the left of the keyboard. So I'd remap QWE/ASD/YXC to the numpad 789/456/123 and vice versa. So I'd expect any program to treat the physical key S exactly like KP5 and the physical KP5 exactly like S. If we only check keysyms that will work fine; but if we check scancodes then it will completely ignore all my keyboard remapping. (To maximize the confusion, KP1 would have a scancode of Z and Y a scancode of KP1 because german keyboard layout).
In the case that no key with the keysym KP5 exists, like on my keyboard where it generates SDLK_EQUAL, Widelands should not pretend that this key has a function which I explicitly removed. I *could* (probably) have kept the KP5 keysym and only changed its interpretation to send the text '=' when pressed, but I chose to replace the entire key. (except when numlock is on, in which case it still generates KP5).

> But since this bug was never reported, it looks like diagonal scrolling isn't used much anyway, so it might as well be dropped completely... ;)

IIRC diagonal scrolling used to be the default, and the option was introduced because some laptop keyboards do not have PageUp/PageDown/Home keys other than the keypad ones. I'd be fine with dropping it, but we should first ask on the forum if anyone would object to that.

tothxa: > Let's assume that I preferred to have the numpad on the left of the keyboard. So I'd remap QWE/ASD/YXC to the numpad 789/456/123 and vice versa. So I'd expect any program to treat the physical key S exactly like KP5 and the physical KP5 exactly like S. If we only check keysyms that will work fine; but if we check scancodes then it will completely ignore all my keyboard remapping. (To maximize the confusion, KP1 would have a scancode of Z and Y a scancode of KP1 because german keyboard layout).

Well, in `mapview.cc` `scroll_map()` (called first from `handle_key()`), `get_key_state()` -> `SDL_GetKeyBoardState()` is used, which works with scancodes. This is required to be able to scroll diagonally by holding down 2 arrow keys, because the repeating key down events only report the last pressed key, even if some earlier key wasn't released yet. So whatever behaviour we would like to get, it needs to get down to the scancodes in `mapview.cc`.

Hmmm, looking at `SDL_keyboard.h`, it looks like there is `SDL_GetScanCodeFromKey()` to reverse look up the mapping... should I try to use that in `mapview.cc`/`scroll_map()`? (then even layout switching would work hopefully)

We could even make it to only use scancodes when numpad diagonal scrolling is off to allow diagonal scrolling with multiple keys, but turn it off and just use keysyms for the numpad when diagonal scrolling is on, so multi-key diagonal wouldn't work. Maybe multi-key could still be allowed for the main arrow keys...

> In the case that no key with the keysym KP5 exists, like on my keyboard where it generates SDLK_EQUAL, Widelands should not pretend that this key has a function which I explicitly removed. I _could_ (probably) have kept the KP5 keysym and only changed its interpretation to send the text '=' when pressed, but I chose to replace the entire key. (except when numlock is on, in which case it still generates KP5).

It seems that your key is reported as `=`, so you'd never get a KP_5 with NumLock off (no go to HQ action), but you're fine with it. Is that correct? While most users, who don't remap the numpad, would get KP_5, which is OK to be mapped to go to HQ when KP_Home (KP_7) is required for diagonal scroll. Would you accept this solution together with the above mapview change?

> > But since this bug was never reported, it looks like diagonal scrolling isn't used much anyway, so it might as well be dropped completely... ;)
> 
> IIRC diagonal scrolling used to be the default, and the option was introduced because some laptop keyboards do not have PageUp/PageDown/Home keys other than the keypad ones. I'd be fine with dropping it, but we should first ask on the forum if anyone would object to that.

Well, I didn't really mean to remove it, but the current implementation is somewhat erratic, so it's hard to figure out what I should preserve from it (while making it more predictable).

Noordfrees: > Hmmm, looking at SDL_keyboard.h, it looks like there is SDL_GetScanCodeFromKey() to reverse look up the mapping... should I try to use that in mapview.cc/scroll_map()? (then even layout switching would work hopefully)
>
> We could even make it to only use scancodes when numpad diagonal scrolling is off to allow diagonal scrolling with multiple keys, but turn it off and just use keysyms for the numpad when diagonal scrolling is on, so multi-key diagonal wouldn't work. Maybe multi-key could still be allowed for the main arrow keys...

Changing this dynamically would look like a bug. But reverse-lookuping keysyms to obtain their scancodes is fine, as the actual logic is then still tied to the keysym. Big +1 for that :)

> It seems that your key is reported as =, so you'd never get a KP_5 with NumLock off (no go to HQ action), but you're fine with it. Is that correct? While most users, who don't remap the numpad, would get KP_5, which is OK to be mapped to go to HQ when KP_Home (KP_7) is required for diagonal scroll. Would you accept this solution together with the above mapview change?

Correct :+1:

src/wui/mapview.cc
```diff
@@ -670,14 +670,15 @@ Widelands::NodeAndTriangle<> MapView::track_sel(const Vector2i& p) {
...
-	const bool kRIGHT = get_key_state(SDL_SCANCODE_RIGHT);
+	const bool kUP = get_key_state(SDL_GetScancodeFromKey(SDLK_UP));
+	const bool kDOWN = get_key_state(SDL_GetScancodeFromKey(SDLK_DOWN));
+	const bool kLEFT = get_key_state(SDL_GetScancodeFromKey(SDLK_LEFT));
+	const bool kRIGHT = get_key_state(SDL_GetScancodeFromKey(SDLK_RIGHT));
 
 	// numpad keys
 	const bool kNumlockOff = !(SDL_GetModState() & KMOD_NUM);
-#define kNP(x) const bool kNP##x = kNumlockOff && get_key_state(SDL_SCANCODE_KP_##x);
+#define kNP(x)
```
Noordfrees:
```suggestion
#define kNP(x) \
```

Noordfrees (approved):
This doesn't compile because of an unescaped linebreak in a macro. Once that's fixed good to go.

Thanks for your work :)

tothxa:
> This doesn't compile because of an unescaped linebreak in a macro. Once that's fixed good to go.

Arrrgh... That's when I forget to check clang-format before build, and then think that reformatting can't break it, except I copy it manually because it would change 3 unrelated places (probably because of the version thing, because I still have clang-7 (buster))...
C preprocessor macros are evil as it is, but there's no word for clang-format putting the backslash at column 100...

Sorry.

edit:
And this is the one with the clang-format bug afterwards, but I didn't put the whole thing in `// clang-format off`, because that looked just as ugly...

PR #5060, opened 2021-09-12
Fix immovable savegame compatibility Long Long Way

Noordfrees:
Fixes https://www.widelands.org/forum/topic/5406
And I very much hope this does not break a different map in the process. The legacy immovable saveloading structure is an utter mess…


hessenfarmer (approved):
tested and working. could not find any other map with a packet version of 8 or 9

PR #5068, opened 2021-10-01
Add more uses for mousewheel scrolling

tothxa:
- map scrolling with touchpad
- game speed
- editor toolsize
- slider and spinbox UI elements
- soldier capacities and attacker count
- input queue

Also change parameters passed to mousewheel handlers.
Includes config file options and new Scrolling tab in Options menu.


tothxa:
See https://www.widelands.org/forum/topic/5349/ and https://www.widelands.org/forum/topic/5384/

I expect there will be a lot of change requests, as most of this is way beyond my programming knowledge...

src/ui_basic/scrollbar.cc
```diff
@@ -409,13 +410,16 @@ void Scrollbar::think() {
...
 }
 
-bool Scrollbar::handle_mousewheel(uint32_t, int32_t, int32_t y) {
-	if (y < 0) {
-		action(Area::Plus);
-	} else {
-		action(Area::Minus);
+bool Scrollbar::handle_mousewheel(int32_t, int32_t y, uint16_t modstate) {
+	if (matches_keymod(modstate, KMOD_NONE) && y != 0) {
+		if (y < 0) {
```
Noordfrees: Does this also need `get_mousewheel_change()`?

src/wlapplication_mousewheel_options.cc
```diff
@@ -0,0 +1,323 @@
...
+	const uint16_t default_;
+	const uint16_t default_2D_;
+
+	MousewheelOption(std::string n, MousewheelOptionType t, uint16_t def, uint16_t def2D)
+	   : internal_name_(n), type_(t), default_(def), default_2D_(def2D) {
+	}
+
+	bool get_bool(bool use_2d_default) const {
+		assert(type_ == MousewheelOptionType::kBool);
+		const bool def = static_cast<bool>(use_2d_default ? default_2D_ : default_);
```
Noordfrees: Actually the cast is not necessary – just use `!= 0` for readability
```c++
 		const bool def = (use_2d_default ? default_2D_ : default_) != 0;
```

src/wlapplication_mousewheel_options.cc
```diff
@@ -0,0 +1,323 @@
...
+   {MousewheelOptionID::kNoMod,
+    MousewheelOption("", MousewheelOptionType::kKeymod, KMOD_NONE, KMOD_NONE)},
+
+   {MousewheelOptionID::kUse2Ddefaults,
+    MousewheelOption("use_2d_defaults", MousewheelOptionType::kBool, false, true)},
+
+};
+
+// Default signs
+#define SIGN_INCREASE_RIGHT -1
```
Noordfrees: Please don't use `define`s for constants, use a constexpr instead:
```c++
constexpr int kSignIncreaseRight = -1;
```

src/wlapplication_options.cc
```diff
@@ -826,6 +820,38 @@ KeyboardShortcut shortcut_from_string(const std::string& name) {
...
+		mods.push_back(pgettext("hotkey", "Ctrl"));
+	}
+
+	std::string result;
+
+	// Return value will have a trailing "+" if any modifier is matched,
+	// because all current uses need it anyway, and extra checks can
+	// be avoided both here and in the users this way
+	for (const std::string& m : mods) {
+		result = (boost::format(_("%1$s+%2$s")) % m % result).str();
```
Noordfrees, line 850: In RTL languages this generates a string like `+GUI+Ctrl`. For a keyboard shortcut this is fine since that will result in `X+GUI+Ctrl`, but in the mousewheel UI it will look like
> Zoom Map: &nbsp;&nbsp;&nbsp;&nbsp; +Gui+Ctrl &nbsp;&nbsp;&nbsp;&nbsp; Any Scroll

Noordfrees (commented):
Wow, great! A few first comments inline.

---

The pointing device selector in the options menu seems unintuitive though. It is not clear whether this refers to the only device that may be used for scrolling, or whether we can have separate settings for each device. Perhaps clarify in the tooltip.

Assuming it's the latter, then switching the pointing device discards all changes I've made so far. I select the modifiers to use for the external mouse, then choose the touchpad, set its settings, switch back to external mouse to check – and all changes are reset to the initial value. Argh. They should be kept in memory until explicitly reset.

The Reset All button could also do with a tooltip to explain what it does – does it reset all modifiers or all devices?

src/wlapplication_mousewheel_options.cc
tothxa: This is to signify that these are actually bool values, and only stored in uint-s so that the same struct can be used for keymods and bools, so they can be kept in the same map, which is useful to keep the related keymod and bool settings next to each other in the config file. So we're just converting them back to bool here.

Does this make sense?

src/ui_basic/scrollbar.cc
tothxa: Yes, I was considering it, and it should probably use it too.

I kept it as it was for now because I intended to revisit this, because I'm not sure what to do with horizontal scrolling. As far as I can tell, only vertical scrollbars are used currently, but the widget is ready for horizontal too. But yes, `get_mousewheel_change()` can be used in any case, only it will need to be configurable if there will be horizontal scrollbars.

src/wlapplication_options.cc
tothxa: Uh-oh... didn't think of that. Ideally the whole layout of the hbox should be flipped, but is there a way to tell from the locale whether it is LTR or RTL? Other than that, ":paM mooZ   iuG+lrtC   llorcS ..." isn't much better... (BTW spinbox should also be flipped for RTL, but I don't think it is - I checked it, and it turns out that all RTL languages have very low translation coverage, so they mostly show the original English strings in Options...)

Is there a way to handle this situation?

src/wlapplication_mousewheel_options.cc
tothxa: Yes, but these are only used in the `define`-s of x-y pairs for the initialisation of `mousewheel_handlers`, and never used alone. Should they outlive the pair `define`-s? Or should the pairs be turned into constexpr vectors too, and the constructor of `MousewheelHandlerOptions` be made to take them apart into `default_sign_x_` and `default_sign_y_`? Do we want these constexpr-s in the global namespace?

Noordfrees: `#define`s are bad style and should be avoided where possible – just replace them with some `constexpr`s in the .cc file. +1 for using `vector`, or even better, `pair`:
```c++
using Sign2D = std::pair<int, int>;
/* Replace int variables in MousewheelHandlerOptions with Sign2D */

constexpr int kSignIncreaseRight = -1;
...
constexpr Sign2D kDefaultSignValue(kSignIncreaseRight, kSignIncreaseUp);
...
```

src/wlapplication_options.cc
Noordfrees: Not really… ideally the horizontal layout should be mirrored for RTL languages, but for now support for this is veeery patchy. It can be handled like this:
https://github.com/widelands/widelands/blob/eccffa9b227196cec5471f3a14ebfc854f3da970/src/editor/ui_menus/main_menu_random_map.cc#L762-L768
We also have some "convenience" functions for use with Align-layouted boxes:
https://github.com/widelands/widelands/blob/eccffa9b227196cec5471f3a14ebfc854f3da970/src/graphic/align.cc#L47-L52
The inofficial policy is to implement it in this way for additions to the UI if the reviewer remembers to ask for it ;)

tothxa:
> The pointing device selector in the options menu seems unintuitive though. It is not clear whether this refers to the only device that may be used for scrolling, or whether we can have separate settings for each device. Perhaps clarify in the tooltip.

Yeah, it may not be the best idea after all with the current set of options. While ctrl+vertical for zoom / ctrl+horizontal for speed/toolsize may be desirable for efficiency, it is clearly not fit to be the default. I also dropped the unified inputqueue idea (you can still see it in the original mousewheel branch), where I considered setting horizontal to quantity and vertical to priority, but it turned out in testing to be just as confusing as it was useful. So now most defaults are the same, only zoom and map scroll change with it.

Maybe I should drop the whole dual default thing (could be handled in the dialog), and I'm not sure about the pointing device dropdown either. It's just that I wanted to make it clear when it makes sense to turn on map scrolling... but this seems to be too confusing in other ways.
May I change ctrl_zoom to be the default, so that turning on map scrolling becomes a single step even without the pointing device dropdown?

It is also a problem that SDL doesn't make it possible to tell whether the hardware is capable of horizontal scrolling: it doesn't tell (at least in my config) whether there are more than one devices, and which of them is sending the current event, and it doesn't even tell (in my config) when the system scroll direction is flipped.

> Assuming it's the latter, then switching the pointing device discards all changes I've made so far. I select the modifiers to use for the external mouse, then choose the touchpad, set its settings, switch back to external mouse to check – and all changes are reset to the initial value. Argh. They should be kept in memory until explicitly reset.

I wasn't sure how to handle this. If you think it makes sense to keep very different settings for single wheel and touchpad, it could be easily done in 2 separate config sections (say [mousewheel] and [touchpad_scrolling]), and then indeed, the dropdown would get a clear purpose.

I considered keeping track of which values were changed but not applied, but skipped it for now. I thought it would be overkill in most cases, when people would first select the type of pointing device they intend to use. If the above 2 sections idea were implemented, then it would definitely make sense to keep track of user changes for both kinds of devices. But it's probably best to scrap the whole pointing device type dropdown instead as well as the dual defaults.

BTW there's another issue coming from the general handling of config options: once the settings are applied, they all get written to the config file, so changing the pointing device doesn't change anything directly anymore, only the settings the reset button will choose (which will then get written in the file...) I think it should be possible to clear settings from the config file, but the code handling this is too complicated and spread out in too many places to me. This also affects keyboard shortcuts: resetting them pollutes the config file with a list of all defaults, and so later changes to defaults will not be followed anymore.

> The Reset All button could also do with a tooltip to explain what it does – does it reset all modifiers or all devices?

OK. (it resets modifiers and scroll directions including inverts to the settings appropriate for the selected device type) If I drop the device dropdown, this will also become more clear.

src/wlapplication_mousewheel_options.cc
Noordfrees: https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2010/b6801kcy(v=vs.100)?redirectedfrom=MSDN
> 'type' : forcing value to bool 'true' or 'false' (performance warning)
>
> This warning is generated when a value that is not bool is assigned or coerced into type bool. Typically, this message is caused by assigning int variables to bool variables where the int variable contains only values true and false, and could be redeclared as type bool. If you cannot rewrite the expression to use type bool, then you can add "!=0" to the expression, which gives the expression type bool. Casting the expression to type bool will not disable the warning, which is by design.

Noordfrees:
Ok then I think it would be best to drop the Device dropdown and have only one profile. Note that you can emulate all input device events through software (at least on X11) and even connect different types of input to each other, so attempting to decide what functionality is available on a given system is not always meaningful.

-1 for making Ctrl+Wheel the zooming default, this makes the feature harder to discover for new players. Perhaps you could enable map scrolling by default with the Ctrl+Shift modifier so it's easy to remap the modifiers without them getting in each other's way initially.

Lastly, I'd suggest to make the scrolling options pane a tab in the keyboard shortcuts window and rename that menu something like "Shortcuts & Modifiers" to keep related settings grouped.

src/wlapplication_mousewheel_options.cc
tothxa: Hmmm, I think I'll use Vector2i, because it has members x and y instead of first and second...

edit:
or not... clang-tidy says it's not a constant expression...

tothxa: OK, it seems on macos pair is defined with const&... I guess I could use `struct Sign2D { int x, y; };`?

Noordfrees: >  or not... clang-tidy says it's not a constant expression...

+1 for Vector2i – you can use `static const` instead of `constexpr` if the compiler has hiccups.

src/wlapplication_options.cc
tothxa: Do I get it right that `UI::g_fh` depends on the global locale setting, so it will be wrong if a given string has no translation? But with low translation coverage this will be a mess if I reorder everything? I guess here I could reorder based on the direction of the title, so if locale is RTL, I need to check whether the title has a translation. What's the best way to do it other than comparing the raw string to the translated one?

And `mirror_alignment()` is intended to be used in `add()`, so I still need to conditionally reorder the widgets in the box?

Noordfrees: If we try to account for translation coverage in RTL support the layouting code will quickly become unmaintainable. It's more viable (and efficient) to just assume that all languages are fully translated and only query the global fontset.

Usually, you *only* need to conditionally reorder the order in which widgets are added to a `Box`. The `mirror_alignment` function would be needed for widgets that are added with `Box::Resizing::kAlign` to a `Vertical` box – this would then look like
```c++
vbox.add(some_panel, UI::Box::Resizing::kAlign, mirror_alignment(UI::Align::kLeft));
```
We don't yet have any code examples that use this though. 
`Resizing::kAlign` is usually **not** what you want because it does not resize the widget to fill available space; `kFullSize` or `kExpandBoth` is often the better choice. You could try using these for the options menu boxes – at 1024×768 the tab looks nice and compact, but at 1920×1080 there's masses of empty space which could be filled by horizontally expanding the dropdowns.
Unfortunately `kAlign`+`kLeft` is the default if only one argument is passed to `add`.

tothxa: I found that multilinetextarea uses `has_rtl_character()` from `graphic/text/bidi.h` as condition in `mirror_alignment()`. I think reordering `add()`-s based on that wouldn't be too messy. (or more messy than reordering based on global setting anyway)

Hmmm, try it with spinbox in another PR? At least that can be tested with Options in (I think) Arabic...
shot0000
Though spinbox is quite messy as it is with the conditional big step buttons...

Noordfrees: `has_rtl_character` has very bad performance, ideally it should be used only for text rendering internals and other code should assume that the fontset provides all the information we need.

tothxa: I tried `UI::g_fh->fontset()->is_rtl()`, but it only works for hindi. Arabic, farsi and hebrew get false.

tothxa:
> Lastly, I'd suggest to make the scrolling options pane a tab in the keyboard shortcuts window and rename that menu something like "Shortcuts & Modifiers" to keep related settings grouped.

Done.

I did an unrelated change: I made the keyboard_options window resize with the main window (I needed it to test the layouting). Please let me know if it should be reverted from this PR.

Noordfrees:
> I did an unrelated change: I made the keyboard_options window resize with the main window (I needed it to test the layouting). Please let me know if it should be reverted from this PR.

Looks fine to me codewise, not tested yet.

---

Two more nits regarding code style:
- Please go easier on abbreviations ;) They tend to impact readability. E.g. `kApplyBtnW` should better be `kApplyButtonWidth` and `kMwDdMaxW` should be… should be……… (point made :wink:?)
Local variables that are used only for an intermediate result or temporary iterator can be called `x` or `t` or whatever, but anything with a larger scope should have a name that tells the reader immediately what the object does.
- When declaring `constexpr` or `static const` variables please use one declaration per statement, e.g.
```diff
- static const Sign2D  //
-    kDefaultSignValue(kSignIncreaseRight, kSignIncreaseUp),
-    kDefaultSignMove(kSignNextRight, kSignNextDown),  //
-    kDefaultSignScroll(kSignScroll, kSignScroll);
+ static const Sign2D kDefaultSignValue(kSignIncreaseRight, kSignIncreaseUp);
+ static const Sign2D kDefaultSignMove(kSignNextRight, kSignNextDown);
+ static const Sign2D kDefaultSignScroll(kSignScroll, kSignScroll);
```

src/wlapplication_options.cc
Noordfrees: Hm, that's weird. I tested my `techinfo` branch with Hebrew and it inverted the layout as expected.
The text direction is defined in `data/i18n/fonts.lua` which declares RTL for Arabic, Devanagari, and Hebrew font sets, used – according to `data/i18n/locales.lua` by Al-ʿArabiyyah, Farsi, Hindi, and Ivrit.
Edit: Do these fonts appear as RTL or LTR in the options language dropdown for you?
grafik

tothxa: I was able to find the languages because they are shown right aligned in the language selection dropdown.
Hmmm, may this be because I'm using Debian with fonts provided by its packages?

tothxa:
>     * Please go easier on abbreviations ;) They tend to impact readability. E.g. `kApplyBtnW` should better be `kApplyButtonWidth` and `kMwDdMaxW` should be… should be……… (point made wink?)

Sorry. It started out with all of them like the BtnW one, but went through some deletions/readditions/renamings, and the scope is really just one .cc.

>     * When declaring `constexpr` or `static const` variables please use one declaration per statement

Sorry again. I was afraid it would be so, and I can understand why. I just thought it's more readable without repeating the `constexpr int`-s.

src/wlapplication_options.cc
tothxa: Anyway, it's an unrelated issue... Done.

src/ui_fsmenu/mousewheel_options.cc
```diff
@@ -0,0 +1,438 @@
...
+   /** TRANSLATORS: e.g. Ctrl+Horizontal scroll */
+   gettext_noop("Horizontal scroll"),
+   /** TRANSLATORS: Used when a function works with either vertical or horizontal scrolling.
+       It may be used in combination with modifier keys, e.g. "Ctrl+Any scroll" */
+   gettext_noop("Any scroll")};
+
+#define READ_MOD(option) normalize_keymod(get_mousewheel_keymod(MousewheelOptionID::option##Mod))
+
+#define DIR_COMBINE(x, y)                                                                          \
+	((x ? SD::kHorizontal : SD::kDisabled) | (y ? SD::kVertical : SD::kDisabled))
```
Noordfrees: This macro can be replaced with an inline function:
```suggestion
inline uint8_t dir_combine(const bool x, const bool y) {
	return ((x ? SD::kHorizontal : SD::kDisabled) | (y ? SD::kVertical : SD::kDisabled));
}
```

src/ui_fsmenu/mousewheel_options.cc
```diff
@@ -0,0 +1,438 @@
...
+	tab_invert_ = READ_DIR(kUITabInvert);
+	zoom_invert_ = READ_DIR(kMapZoomInvert);
+}
+
+#undef READ_MOD
+#undef READ_DIR
+#undef DIR_COMBINE
+
+#define DIR_X(dir) ((dir & SD::kHorizontal) != 0)
+#define DIR_Y(dir) ((dir & SD::kVertical) != 0)
```
Noordfrees: Same here:
```suggestion
inline bool dir_x(const uint8_t dir) {
	return (dir & SD::kHorizontal) != 0;
}
inline bool dir_y(const uint8_t dir) {
	return (dir & SD::kVertical) != 0;
}
```

Noordfrees (commented):
Looking good now. I'm through with the review, only a few minor comments inline.

One more suggestion, it would be good to have a Reset To Defaults button for the mouse wheel options along with the Apply button.

Noordfrees:
LGTM but there's a layouting glitch: Initially the reset/apply buttons are centered, but when I try to assign a modifier that conflicts with another modifier they suddenly jump to the left:

grafik
grafik

tothxa:
I'd add a button to set recommended map scroll and zoom settings for touchpad between 'Scroll Map' and 'Change Game Speed', but I can't think of a nice short label (possibly explained by a tooltip) for it.

Noordfrees:
Layout works now. For the button position I'd suggest having them all on one line since the Touchpad button is essentially another Reset button:

<table>
<tr>
	<td><b>Label:</b>
	<td>Reset (Mouse)
	<td>Reset (Touchpad)
	<td>Apply
<tr>
	<td><b>Tooltip:</b>
	<td>Reset to the default settings recommended for standard mouses
	<td>Reset to the settings recommended for touchpads and horizontally scrolling mouses
	<td>Apply scroll settings
</table>

tothxa:
> Layout works now. For the button position I'd suggest having them all on one line since the Touchpad button is essentially another Reset button:
> Label: 	Reset (Mouse) 	Reset (Touchpad) 	Apply
> Tooltip: 	Reset to the default settings recommended for standard mouses 	Reset to the settings recommended for touchpads and horizontally scrolling mouses 	Apply scroll settings

As it is now, the touchpad button only sets zoom/scroll map, and asks for approval to set speed and toolsize in case of conflict. It never touches invert settings, while reset forces them all to default. Also, currently the touchpad button does not apply the settings. So it is a much gentler thing than the reset button.

And there is the big Reset All button in the bottom of the window common with the keyboard shortcuts, so I think it's important to have the more specific labels for reset and apply. Though fitting 3 wide buttons is only a matter of adding another box, or playing tricks with the *DirBox-es.

Noordfrees:
Looks good, just one more minor alignment issue: The rows of dropdowns have vertical gaps of differing sizes (4px, 9px, and 16px) between them:
grafik
The 16px gap is probably intentional, but the 4px and 9px gaps should all be the same height.

tothxa:
> Looks good, just one more minor alignment issue: The rows of dropdowns have vertical gaps of differing sizes (4px, 9px, and 16px) between them: The 16px gap is probably intentional, but the 4px and 9px gaps should all be the same height.

I added the small dividers (9px = 1px + 2 * kPadding) with the touchpad button, to signify that the first 2 are set together while the other 2 are completely independent of each other. But yes, they are now probably more confusing than helpful.

Noordfrees (approved):
Great :) 
Let's have it when the builds are green

PR #5084, opened 2021-10-13
Allow validating appdata.xml with appstream-util

tothxa:


tothxa:
This is needed for flatpak:

The flatpak definition needs to update appdata.xml for the following reasons:
* to work around a bug in the flatpak toolchain that makes parsing appdata.xml fail on version ordering
* to set the version for beta builds

However, appstreamcli is not available in the flatpak toolchain, but appstream-util can be used instead.

Noordfrees (approved):
LGTM :)

---

Both tools fail validation though because the "releases are not in order". Would it be possible to rename the beta builds listing in this file post-factum to 0.1 through 0.21 (plus the weird "build 9half" → 0.9.5) – perhaps with some description tag that states the real name of the beta release – or is there some guideline against this?

tothxa:
> Both tools fail validation though

appstreamcli works fine here:
```
$ appstreamcli validate xdg/org.widelands.Widelands.appdata.xml
Validation was successful.
$ appstreamcli --version
AppStream version: 0.12.5
```

> because the "releases are not in order".

That was a major pain for scx: https://github.com/hughsie/appstream-glib/pull/272
But it only affects appstream-util (appstram-glib). As such, some history is patched out from the appdata.xml for flatpak.

BTW there's a warning for the screenshots as well with appstream-util, that's why I used validate-relax (these are with the edited version history):
```
$ appstream-util validate-relax org.widelands.Widelands.appdata.xml 
org.widelands.Widelands.appdata.xml: OK
$ appstream-util validate org.widelands.Widelands.appdata.xml 
org.widelands.Widelands.appdata.xml: FAILED:
• attribute-invalid     : <screenshot> width too large [https://www.widelands.org/wlmedia/wlscreens/screens/Build%2020/Four_Tribes_2.jpeg] maximum is 1600px
• attribute-invalid     : <screenshot> height too large [https://www.widelands.org/wlmedia/wlscreens/screens/Build%2020/Four_Tribes_2.jpeg] maximum is 900px
Validation of files failed
```

> Would it be possible to rename the beta builds listing in this file post-factum to 0.1 through 0.21 (plus the weird "build 9half" → 0.9.5) – perhaps with some description tag that states the real name of the beta release – or is there some guideline against this?

Actually the only problem is with the single digit -> multi-digit step at build 9(with or without half) -> build 10. So if you don't mind rewriting history, adding a padding 0 to build 1 to 9half would work. But the bug is really in appstream-util/appstream-glib, as explained in scx's above PR.

Noordfrees:
Ah. A newer version gives me a worse result:
```
$ LANG=C appstreamcli validate xdg/org.widelands.Widelands.appdata.xml
W: org.widelands.Widelands:~: releases-not-in-order 1.0 << Build 21

Validation failed: warnings: 1, pedantic: 1
$ appstreamcli --version
AppStream-Version: 0.14.5
```

Edit: Replacing "Build " with "0." for all releases and "9half" with "9.5" makes appstreamcli happy with result "Validation was successful: pedantic: 1"

PR #5085, opened 2021-10-14
Build System Update

### LINKED ISSUE #3992: CMake 3.18 warnings
Compiling with CMake 3.18 throws this warning:

~~~~
Make Warning (dev) at doc/CMakeLists.txt:4 (include):
  Policy CMP0106 is not set: The Documentation module is removed.  Run "cmake
  --help-policy CMP0106" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.
~~~~

See https://cmake.org/cmake/help/latest/policy/CMP0106.html

[Caused by line 4 in doc/CMakeLists.txt](https://github.com/widelands/widelands/blob/master/doc/CMakeLists.txt#L4)

Since only the documentation is affected, i think this issue is not very important. Just wanted to mention it.

Noordfrees:
- Fixes #3992
    - Silence an undesired naming scheme warning in the SDL2 finding module
    - Add a customized copy of CMake's deprecated Documentation module as recommended
- Update latest g++/clang versions in the CI
- Treat enabled compiler warnings as errors
- Allow silencing compile.sh status messages with `-q`

Supersedes #5075


tothxa (approved):
Checked that it's the same as #5075 with the forced macos version. Without set-builddir, my changes to compile.sh are minimal.

Noordfrees:
Thanks for the review :)

I updated the branch protection rules, so all open PRs need a master merge now.

PR #5096, opened 2021-10-24
Properly format size_t type

### LINKED ISSUE #5095: Daily builds fail with printf format error
Two of the daily PPAs are failing with the error
```
/<<PKGBUILDDIR>>/src/base/log.h:46:77: error: format ‘%lu’ expects argument of type ‘long unsigned int’, but argument 4 has type ‘size_t {aka unsigned int}’ [-Werror=format=]
```

This should be fixed instead of suppressed. We can then also remove again the `FORMAT_WARNINGS_(ON/OFF)` in `src/io/streamread.cc` introduced in 701fbef38c5677109dabe68252cb29c1882e8ff8.

Affected environments are:
- g++ 7.5.0, Bionic, i386 [(link)](https://launchpadlibrarian.net/565181349/buildlog_ubuntu-bionic-i386.widelands_2%3A1.1~bzr10039-202110230320~ubuntu18.04.1_BUILDING.txt.gz)
- g++ 5.4.0, Xenial, i386 [(link)](https://launchpadlibrarian.net/565181794/buildlog_ubuntu-xenial-i386.widelands_2%3A1.1~bzr10039-202110230320~ubuntu16.04.1_BUILDING.txt.gz)

On non-windows systems, we always expand the `PRIuS` macro to `lu` (for `long unsigned int`), which is correct for most systems, but on these two here, `size_t` is defined as `unsigned int` so the macro should expand to `u`. How can we distinguish between these configurations and all others in the preprocessor?

matthiakl:
Fix #5095 
See definiton of SIZE_MAX in stdint.h for reference.

### DIFF
--- a/src/base/macros.h
+++ b/src/base/macros.h
@@ -20,9 +20,10 @@
 #ifndef WL_BASE_MACROS_H
 #define WL_BASE_MACROS_H
 
+#include <cinttypes>
+
 // Make sure that Visual C++ does not bark at __attribute__.
 #ifdef _MSC_VER
-#include <cinttypes>
 #ifndef __attribute__
 #define __attribute__(x)
 #endif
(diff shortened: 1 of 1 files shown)


Noordfrees (approved):
LGTM :)

PR #5102, opened 2021-10-29
Fix release ordering in appdata.xml

tothxa:


tothxa:
See discussion for #5084 

I think this is the minimal fix to satisfy both appstreamcli (old as well as new) and appstream-util.

I believe the new appstreamcli is the intended behaviour, old appsteamcli and appstream-util are buggy. See https://www.freedesktop.org/software/appstream/docs/chap-AppStream-Misc.html#sect-AppStream-Misc-VerCmp

Noordfrees (approved):

PR #5107, opened 2021-11-06
Fix mouse offset for constructions site windows

### LINKED ISSUE #5097: Fastclick targets for constructionsite windows are off

**Describe the bug**
The fastclick targets (the place where the mouse cursor jumps to automatically when a window opens) are incorrect for constructionsite windows.

**To Reproduce**
1. Click on a half-completed own constructionsite, especially for a building that costs several ware types and only has one ware type left to be shown in its window (in the image below, an amazons tower that lacks only ironwood).
2. The mouse should be placed in the centre of the workarea button, but it is actually located a bit below.

**Expected behavior**
The mouse should be placed in the centre of the workarea button

**Screenshots**
After clicking on (1), the mouse jumps to (2) but it should jump to (3):
grafik

**Version:**
 - OS: Ubuntu Mate 21.10
 - Widelands Version: 1.1~git25486[0c381ba@master] (Release)

**Additional context**
Perhaps invisible input queue displays are taken into account when calculating the fastclick panel's absolute position?

jmoerschbach:
Fixes  #5097 

and some method extractions because I like shorter, descriptive methods better

The actual fix is the additional if() in line 173

### DIFF
--- a/src/wui/constructionsitewindow.cc
+++ b/src/wui/constructionsitewindow.cc
@@ -127,18 +127,41 @@ void ConstructionSiteWindow::init(bool avoid_fastclick, bool workarea_preview_wa
 	set_building_descr_for_help(&construction_site->building());
 
 	BuildingWindow::init(avoid_fastclick, workarea_preview_wanted);
+
+	build_wares_tab(construction_site);
+
+	if (construction_site->get_settings()) {
+		build_settings_tab(construction_site);
+	}
+
+	set_title((boost::format("(%s)") % construction_site->building().descname()).str());
+
+	think();
+	initialization_complete();
+}
+
+void ConstructionSiteWindow::build_wares_tab(Widelands::ConstructionSite* construction_site) {
 	UI::Box& box = *new UI::Box(get_tabs(), UI::PanelStyle::kWui, 0, 0, UI::Box::Vertical);
 
-	// Add the progress bar
+	add_progress_bar(box);
+
+	box.add_space(8);
+
+	ensure_box_can_hold_input_queues(box);
+	add_wares_queues(construction_site, box);
+
+	get_tabs()->add("wares", g_image_cache->get(pic_tab_wares), &box, _("Building materials"));
+}
+
+void ConstructionSiteWindow::add_progress_bar(UI::Box& box) {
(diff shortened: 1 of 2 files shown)


src/wui/constructionsitewindow.cc
```diff
@@ -127,97 +127,87 @@ void ConstructionSiteWindow::init(bool avoid_fastclick, bool workarea_preview_wa
...
 
 	BuildingWindow::init(avoid_fastclick, workarea_preview_wanted);
+
+	build_wares_tab(construction_site);
+
+	if (construction_site->get_settings()) {
+		build_settings_tab(construction_site);
+	}
+
+	set_title((boost::format("(%s)") % construction_site->building().descname()).str());
```
Noordfrees:
```suggestion
	set_title((boost::format(_("(%s)")) % construction_site->building().descname()).str());
```

src/wui/constructionsitewindow.cc
```diff
@@ -127,97 +127,87 @@ void ConstructionSiteWindow::init(bool avoid_fastclick, bool workarea_preview_wa
...
-		case Widelands::MapObjectType::MILITARYSITE: {
-			upcast(Widelands::MilitarysiteSettings, ms, construction_site->get_settings());
+}
+void ConstructionSiteWindow::build_settings_tab(Widelands::ConstructionSite* construction_site) {
+	const bool can_act = ibase()->can_act(construction_site->owner().player_number());
+	// Create the settings. Since we don't access an actual building, we create
+	// a simplified faksimile of the later building window that contains only
+	// the relevant options.
+	bool nothing_added = false;
+	UI::Box& settings_box = *new UI::Box(get_tabs(), UI::PanelStyle::kWui, 0, 0, UI::Box::Vertical);
```
Noordfrees, line 185: This should better be a `std::unique_ptr<UI::Box>` which is `release()`d in the call to `get_tabs()->add()`. Same for the boxes in the other functions, to avoid memory leaks if an exception happens or in the case that `nothing_added` is `true`.

Noordfrees (approved):
LGTM, tested and working :)

I've added two comments in the diff, but they're not directly related to this branch so it's up to you if you want to implement this here or not.

src/wui/constructionsitewindow.cc
jmoerschbach: let's do this another time :)

PR #5109, opened 2021-11-08
Remove messages of muted buildings

### LINKED ISSUE #4957: Delete Inbox message when the site is muted
When a site creates an Inbox message, that message is automatically removed when the building is dismantled/upgraded. This is a great help to reducing the manual steps needed to keep the Inbox tidy. If a site generates an out of resources message, and is then muted, it would be helpful if the Inbox message is removed in the same manner.

jmoerschbach:
Re #4957

This works currently only for the building-specific muting, not for the "mute all".
If this is also required, some more work needs to be done

### DIFF
--- a/src/logic/player.cc
+++ b/src/logic/player.cc
@@ -443,6 +443,10 @@ MessageId Player::add_message(Game& game, std::unique_ptr<Message> new_message, 
 	if (message->serial() > 0) {
 		MapObject* mo = egbase().objects().get_object(message->serial());
 		mo->removed.connect([this, id](unsigned) { message_object_removed(id); });
+		upcast(Building, site, mo);
+		if (site) {
+			site->muted.connect([this, id](unsigned) { message_object_removed(id); });
+		}
 	}
 
 	// Sound & popup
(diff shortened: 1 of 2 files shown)


src/logic/player.cc
```diff
@@ -443,6 +443,10 @@ MessageId Player::add_message(Game& game, std::unique_ptr<Message> new_message,
 	if (message->serial() > 0) {
 		MapObject* mo = egbase().objects().get_object(message->serial());
 		mo->removed.connect([this, id](unsigned) { message_object_removed(id); });
+		upcast(Building, site, mo);
+		if (site) {
+			site->muted.connect([this, id](unsigned) { message_object_removed(id); });
+		}
```
Noordfrees: `upcast` has bad performance because it uses `dynamic_cast` internally, it is more [efficient](https://stackoverflow.com/questions/9959315/why-virtual-function-call-is-faster-than-dynamic-cast) to check the `MapObjectType` first:
```suggestion
		if (mo->descr().type() >= MapObjectType::BUILDING) {
			upcast(Building, site, mo);
			assert(site != nullptr);
			site->muted.connect([this, id](unsigned) { message_object_removed(id); });
		}
```

jmoerschbach: thx, done

Noordfrees (approved):

PR #5113, opened 2021-11-11
fix erroneous french translation for 'Ferries'

jbar:


Noordfrees:
Hi @jbar,

Please fix this on Transifex instead: 
https://www.transifex.com/widelands/widelands/translate/
https://www.widelands.org/wiki/TranslatingWidelands/
Transifex is free to use, just sign in with your GitHub account and join the french language team.

The po files in this repo are overwritten frequently by the automated translation syncs.

jbar:
Hi @Noordfrees 

I tried to use Transifex, but it failed to log me in. (It tried both with github and directly with email).

Please accept this little advice : If (like many other people) you tend to multiply tools intended to help your work, please make a difference between those mandatory (eg: git, g++) which must have the maximum reliability, and those optional (Transifex, github, jenkis, etc.).

If today Transifex is mandatory in your processes, may you please apply there this fix for me. (I really don't care "intellectual property" for such little fix).

Thanks a lot.

NB: I don't figure how a previous translator may have confuse "fairies" -french-> "fées", and "ferries" -french-> "transbordeurs" 🤔

Noordfrees:
Fix applied :) 

I'm not entirely happy about our focus on Transifex, but it does have a lot of useful features and is free to use. I suppose it could be possible to adapt the translation sync script to detect where translations have been changed manually, and in such cases to overwrite Tx's translation with the updated one from master and push the result back to Tx. Then it would be possible to update PO files in this way; you're not the first contributor who cannot or doesn't want to use Transifex. @hessenfarmer @tothxa @gunchleoc what do you think?

tothxa:
To me, Tx was quite a good way to start contributing. Apart from the apparently scary registration (I didn't quite like it either), I think it has a very low barrier to entry, and many useful features. (To demonstrate, I've just looked up occurences of ferry/ferries for French, and it turns out that in tribes, encyclopedia and tutorial03 it's "ferry(s)" all the way... also in the glossary -- but then, it would be quite tedious to correct all of them in the web interface of Tx, not only because it's too slow, but also because it doesn't support mass operations at all.)
To contribute translations through github needs knowledge of PO files as well as git. Quite high barrier to entry for most people. And indeed, just part of daily routine for others, as well as a better way of bigger contributions with offline tools.

For the project, it looks like Tx takes a lot of burden off of the git repo collecting all these contributions, so in this regard the current policy makes a lot of sense. Allowing more than one official places for translations would increase the risk of people overwriting each other's work. Tx's protection for verified translations would also be lost when pushing translations through git or any other external platform. These issues should be handled somehow.

From the github side, I wonder if it would make sense to set up a new repo for collecting translations, with separate history, team(s) and commit rules. This would then be synced regularly with master and with Tx (using whatever rules we come up to solve the above issues of conflict), as it is done now with Tx. Maybe the translations git repo could be synced with Tx more often -- I don't know what options are there to make a more direct link between github and Tx.

p.s. Should we take this topic elsewhere, like the WL forum or a github discussion?

jbar:
@tothxa @Noordfrees I didn't say Tx is not usefull, and I agree it may increase efficiency for many people (but not for me, IMHO it is even the opposite).

Yes the english words "ferry" and "ferries" use to be also in french dictionaries. But French Canadians and some other french-speaking people (like me) prefer to use the word "transbordeur" which sounds far more pretty to french ears + is more descriptive. (if a child listen to this word for its first time, (s)he may immediately figure what it means).

I won't have time to step more in your discussions, but thx for all and congratulation for this very nice and accomplished game (IMHO there are still a lot of thing to improve, maybe We could meet IRL... once events like FOSDEM could also become IRL again).

++

PR #5121, opened 2021-11-16
compile.sh: add --builddir --exename and --dry-run options

tothxa:
As discussed in https://github.com/widelands/widelands/pull/5075#issuecomment-934816231


compile.sh
```diff
@@ -357,6 +458,27 @@ if [ $QUIET -ne 0 ]; then
...
+  echo "Command line options:"
+  echo "   $OLD_CLI_ARGS"
+  echo " "
+fi
+
+echo "Using build directory:"
+echo "   '$BUILDDIR'"
+echo " "
+# Adding build directory to update script is not necessary, because the script
+# will only be created for the default.
```
matthiakl, line 479: Why? Shouldn't it be also possible to update in the custom builddir?

compile.sh
```diff
@@ -518,53 +638,99 @@ buildtool="" #Use ninja by default, fall back to make if that is not available.
...
+    $RUN cp VERSION ../VERSION
+    $RUN mv src/widelands ../"$EXENAME"
 
     if [ $BUILD_WEBSITE = "ON" ]; then
-        mv ../build/src/website/wl_create_spritesheet ../wl_create_spritesheet
-        mv ../build/src/website/wl_map_object_info ../wl_map_object_info
-        mv ../build/src/website/wl_map_info ../wl_map_info
+      $RUN mv ../build/src/website/wl_create_spritesheet ../wl_create_spritesheet
+      $RUN mv ../build/src/website/wl_map_object_info ../wl_map_object_info
+      $RUN mv ../build/src/website/wl_map_info ../wl_map_info
```
matthiakl, line 713: If you include $BUILDDIR in the path, these binaries can still be moved to the source dir.

matthiakl (commented):
Currently the full functionality can only be achieved with a build directory directly in the source dir. Is it possible to support this also in deeper subdirs? I like to use a structure `build/debug/`, `build/release`, to keep the parent folder as clean as possible.

tothxa:
@matthiakl 
I didn't want to overwrite any files of the default build from a custom build. I considered supporting a suffix, which could be then applied to all of them, but @Noordfrees also [asked](https://github.com/widelands/widelands/pull/5075#issuecomment-936123868) for the generalised solution.

How about this:
* Add option `--website-tools-suffix` or so for those 3, which would behave similar to `--exename`? (you'll be able to set it to `""` if you want the default names) I'm afraid that separate options for all of them would be too much...
* For update.sh: I was already considering to add support for starting compile.sh from anywhere. (`.compile_defaults` would be searched for first in startdir, then in sourcedir -- searching for it in builddir is not really feasible with the current design, but you could start it from builddir) If I can do it, then update.sh could be placed in builddir.

(OTOH @frankystone already [complained last time](https://github.com/widelands/widelands/pull/4980#issuecomment-869408937) that compile.sh is getting too complex)

matthiakl:
Thanks for pointing me to the discussions. Now I understand your approach and think you can leave it like this. 
It is a good mix between extending functionality for multiple builds and for developers, but still keeps it usable for other users.

tothxa:
I added the necessary changes to allow calling compile.sh from anywhere to a new branch: https://github.com/tothxa/widelands/commit/2ff827bbcef7c04f56657d5ce21ecbd00f413f05
Should I apply it to this PR?
Also has 2 questions in inline comments.

tothxa:
@widelands/developers 
Should I close this PR? Is anyone interested in these options?

If it has support, then I'd like to also add a `--cmake-only` option that sets up the builddir and runs cmake, but does not compile, and a `--setup-clang-tidy` option that sets `--cmake-only`, `--clang` and `-DCMAKE_EXPORT_COMPILE_COMMANDS=ON` (as required by `utils/run-clang-tidy.py`).

I think these will make `compile.sh` cover all practical uses that are possible with direct invocation of `cmake`, but with a more comfortable syntax and a `--help` option. (casual users can still just run `./compile.sh` and be done as previously)

PR #4995, opened 2021-07-15
4911 ai fixes

hessenfarmer:
fixes #4911 
except the recruitment site detection
fixes #1006
fixes #4087 (hopefully)
fixes #4558 
fixes #4669

Issue #5052, opened 2021-08-30
Assertion fail - inputqueuedisplay.cc line 584


alasiax:
**Describe the bug**
Game crashed on trying to build stonecutter on small block

Edit to add error info at a glance and for search:
```
Assertion failed!
File: ../src/wui/inputqueuedisplay.cc
Line: 584
Expression: nr_icons_
```
(end edit.)

**To Reproduce**
I haven't been able to reproduce this one, nor have I seen this particular error before. 

**Expected behavior**
place flag for stone cutter.

**Screenshots**
Error and stdout.txt are in the zipfile. No output to stderr.txt 

**Version:**
 - OS: Windows 10
 - Version 1.1~git25441[d355de6@HEAD] (Debug)
 - Acer laptop with generic USB mouse, Core i5, 16MB RAM
 
**Context**
I have been playing in the debug version more often lately, which may be why the game isn't simply crashing without any error shown at all. I'm trying to get better at actually logging bug reports now that I'm playing more using recent dev versions too.
[assertion fail - inputqueuedisplay.zip](https://github.com/widelands/widelands/files/7077800/assertion.fail.-.inputqueuedisplay.zip)

alasiax:
Happened again. This time I think I clicked on fog, but I was just clicking around building in this map. stderr.txt is empty.
This map is Spider Lake.
[assertion fail - iputqueuedisplay2.zip](https://github.com/widelands/widelands/files/7084608/assertion.fail.-.iputqueuedisplay2.zip)

alasiax:
I think I rightclicked as this error occurred. This time I'm using a Microsoft wireless mobile mouse 3000 model 1359. Map is twinkling waves.
[assertion fail - inputqueuedisplay3.zip](https://github.com/widelands/widelands/files/7088116/assertion.fail.-.inputqueuedisplay3.zip)

alasiax:
Left clicking on the fortress under construction near blue HQ.
Microsoft wireless mobile mouse 3000 model 1359. Map is twinkling waves.
[assertion fail - inputqueuedisplay4.zip](https://github.com/widelands/widelands/files/7090919/assertion.fail.-.inputqueuedisplay4.zip)

eqom14:
Had this a couple of times but can't reproduce. Last time was a couple of weeks ago. I was camping so I had no internet to download the lastest.

```
~/src/widelands$ ./widelands
This is Widelands Version 1.1~git700[608836a@master] (Debug)
...
[...]

Issue #4963, opened 2021-06-20
Game is stuttering


frankystone:
After long time not playing i started a new game today. But i am a bit disappointed about the speed... The game noticeable stutters some time. Animals and carriers are walking not with the same speed during a single walk. Placing a building can last a while (click a building to place, but after some noticeable milliseconds the construction appears on the map) and it seems to me that even the mouse does not react smooth.

This is most noticeable when setting Max-FPS to 99. Setting this to a value of 30 the game reacts a little bit smoother, but playing at 2x speed it feels like the game speed switches between 1x and 2x speed in irregular intervals.

**To Reproduce**
What i have done:
1. Chosen the map Astoria 2.0
2. Set myself as Player 4/Empire
3. Changed Playerslot 1 to AI and Barbarians
4. Set HQ-Hunter wincondition
5. Started the game

Right after start one can see the Animals stumbling around. But i remember this was also in versions before releasing 1.0 (the versions i played before this time where all before version 1.0, but always current master).

**Expected behavior**
A smooth game play.

**Version:**
 - OS: archlinux
 - Widelands Version: [d94928c@master] (Debug)

Maybe because i use a debug build? This wasn't a problem so far, except on really big maps. Will also test a release build soon.

Noordfrees:
The speed is the same as it used to be (debug builds on large maps tend to be very slow), but it is more noticeable now than before due to multithreading. In 1.0, Widelands simply hangs for a brief (more or less) interval every time the game state is progressing by a certain number of milliseconds; whereas now the game is redrawn all the time, which has the side-effect that the continuous alterations in the maximum game speed your CPU can actually handle (and which may be lower than the desired speed setting), as well as the delays between a click and its effect, are more noticeable.

The largest portion of those slow-speed periods are filled by the AI – which can not be reasonably multithreaded – and the economy routing code – which I intend to make multithreading-compatible in the near future which will improve performance.

hessenfarmer:
I simply do not understand enough to know what has to be done to make the Ai thread safe.  Discussing such things would be one big benefit of the proposal of @klaus-halfmann to have a devcon. So if somebody could teach me some principles I might help with my understanding of the AI code

Noordfrees:
The AI could be run pretty much independent from everything else already except for one thing: The state of the game might change while the AI is analysing it. For example look at any piece of AI code:
[...]

Issue #4881, opened 2021-05-24
Workareas display bug with some add-on buildings


Noordfrees:
1. Start a new SP game on the map *The Pass Through The Mountains* with both players set to *Europeans*, the *Endless Game Fogless* win condition, and *Peaceful Mode* enabled.
2. Ensure that the option to indicate overlapping workareas is turned on.
3. Click on the building spot located 4 nodes W and 2 SW of the HQ's flag.
4. Select the small buildings tab. Point on the first building (basic well) and slowly move the mouse to the right over all buildings in the row (up to basic quarry).
5. Right-click to close the fieldaction window.
6. Some workareas (or parts of them) remain permanently visible. Some of them should never have been displayed in the first place.
![shot0001](https://user-images.githubusercontent.com/48293446/119393998-fc2e1f80-bcd1-11eb-9cc3-32503c952a36.png)

Version 1.0~git25311[d15ecd3@master] (Release) with Europeans 1.1.0

Noordfrees:
One of the bugs described here is general: Overlaps with productionsite-constructionsites are highlighted even if they belong to different players, because the ownership check for this case is missing here:
https://github.com/widelands/widelands/blob/da980781ddf3adc230acea439461594ec12e8b54/src/wui/fieldaction.cc#L993-L1006

PR #5125, opened 2021-11-17
Fix worker reordering stack overflow

matthiakl:
Hotfix for https://github.com/widelands/widelands/issues/4995#issuecomment-972008401

The problem is, that the reordering does not terminate for the european basic farm, where both slots need the same worker.
This quick fix solves the issue, but I will look, whether there is a better solution.

### DIFF
--- a/src/logic/map_objects/tribes/productionsite.cc
+++ b/src/logic/map_objects/tribes/productionsite.cc
@@ -627,7 +627,13 @@ std::pair<int32_t, int32_t> ProductionSite::find_worker(OPtr<Worker> w) {
  */
 void ProductionSite::try_replace_worker(const Game* game,
                                         DescriptionIndex worker_index,
-                                        WorkingPosition* wp) {
+                                        WorkingPosition* wp,
+                                        size_t iteration) {
+	if (iteration >= working_positions_.size()) {
+		// cancel recursion, something went wrong
+		log_warn("Aborting try_replace_worker() after %" PRIuS " calls.", iteration);
+		return;
+	}
 	// Search replacement worker in the same building
 	for (auto wp_repl = working_positions_.begin(); wp_repl != working_positions_.end(); ++wp_repl) {
 		if (!wp_repl->worker.is_set()) {
@@ -651,7 +657,7 @@ void ProductionSite::try_replace_worker(const Game* game,
 			molog(owner().egbase().get_gametime(), "%s promoted\n", w_repl->descr().name().c_str());
 			// Request the now missing worker instead and loop again
 			*wp_repl = WorkingPosition(&request_worker(worker_index_repl), nullptr);
-			try_replace_worker(game, worker_index_repl, wp_repl.base());
+			try_replace_worker(game, worker_index_repl, wp_repl.base(), iteration + 1);
 			return;
 		}
 	}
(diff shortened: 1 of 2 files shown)


klaus-halfmann:
Code looks straight forward, AFAIK I need the Europeans tribe. What/How I can I test there?

klaus-halfmann:
Is it normal that Europeans result in such ugly color overlays?

grafik

Noordfrees:
> Is it normal that Europeans result in such ugly color overlays?
- That's #4881

src/logic/map_objects/tribes/productionsite.cc
```diff
@@ -627,7 +627,13 @@ std::pair<int32_t, int32_t> ProductionSite::find_worker(OPtr<Worker> w) {
  */
 void ProductionSite::try_replace_worker(const Game* game,
                                         DescriptionIndex worker_index,
-                                        WorkingPosition* wp) {
+                                        WorkingPosition* wp,
+                                        size_t iteration) {
+	if (iteration >= working_positions_.size()) {
+		// cancel recursion, something went wrong
+		log_warn("Aborting try_replace_worker() after %" PRIuS " calls.", iteration);
```
hessenfarmer: this should be made verbose to declutter the log for mat5ches with europeans or other future tribes having similar designed productionsites

matthiakl:
Now it is a proper solution. There was an issue with the `needs_higher_qualification` condition, which also matched "same qualification"

hessenfarmer (approved):
code looks straight forward.
@MarkMcWire does the thumb mean you tested this? 
if yes this can go in

MarkMcWire:
> code looks straight forward. @MarkMcWire does the thumb mean you tested this? if yes this can go in

I will test it today. Yesterday, was just a "well done".

MarkMcWire:
> code looks straight forward. @MarkMcWire does the thumb mean you tested this? if yes this can go in

-> https://drive.google.com/file/d/1DDHR-8C1gbCYZ1wkYX_h8Ps45ZcoCs25/view?usp=sharing
_The game don't crash anymore_, but the performance is slower than without the fix. The game "jerks" a little.

It crashes for another reason:

```
widelands: /home/solus/Git/widelands_fix/widelands/src/wui/inputqueuedisplay.cc:351: int32_t InputQueueDisplay::fill_index_at(int32_t, int32_t) const: Zusicherung »nr_icons_ > 0« nicht erfüllt.
solus@solus-pc ~/Git/widelands_fix/widelands
```

hessenfarmer:
> 
> 
> > code looks straight forward. @MarkMcWire does the thumb mean you tested this? if yes this can go in
> 
> -> https://drive.google.com/file/d/1DDHR-8C1gbCYZ1wkYX_h8Ps45ZcoCs25/view?usp=sharing _The game don't crash anymore_, but the performance is slower than without the fix. The game "jerks" a little.
   
Well it is hard to believe that such a small fix has any influence on the performance except in a very big economy. The short freezes of the display are present in master as well and are most probably caused by the AI think cycles or routing se #4963 as the efeects of the mentioned bug are sometimes more visible then other times I believe this might have led to the impression it may be related to this PR.

> 
> It crashes for another reason:
> 
> ```
> widelands: /home/solus/Git/widelands_fix/widelands/src/wui/inputqueuedisplay.cc:351: int32_t InputQueueDisplay::fill_index_at(int32_t, int32_t) const: Zusicherung »nr_icons_ > 0« nicht erfüllt.
> solus@solus-pc ~/Git/widelands_fix/widelands
> ```

this is almost 100% another flavour of the famous #5052 bug

So as the original crash is gone I think we should have this fix in

PR #5127, opened 2021-11-19
Add debug output to help trace bug #5052

Noordfrees:
Re #5052


hessenfarmer:
nice that was more then I expected. I will merge this into my local copy of fri04 and hopefully might trigger the thing while testing the scenario

hessenfarmer (approved):
code looks good.
we should have it to get more potential testers and open an issue to remove it before release

PR #5128, opened 2021-11-19
Keep GameController pointer alive for template game

### LINKED ISSUE #5126: Multiplayer template game crashes
**Describe the bug**
When loading a multiplayer template game, the game crashes with
```
widelands: src/logic/game.cc:811: virtual void Widelands::Game::think(): Zusicherung »ctrl_« nicht erfüllt.
```

**To Reproduce**
1. Run Widelands with `--new_game_from_template=` with [ this template ](https://github.com/widelands/widelands/files/7566354/mp_test.txt)
2. See error

**Version:**
 - Widelands Version: master `77ffd853`

matthiakl:
Fix #5126


Noordfrees (changes requested):
This results in a heap-use-after-free on exit. I assume the reason is that the value of `host` is now managed by *two* `shared_ptr`s who both try to free the memory when their last copies go out of scope

src/wlapplication.cc
```diff
@@ -648,12 +648,13 @@ void WLApplication::init_and_run_game_from_template() {
 	}
 
 	std::unique_ptr<GameSettingsProvider> settings;
-	std::unique_ptr<GameHost> host;
+	std::shared_ptr<GameController> ctrl;
+	GameHost* host;  // will be deleted by ctrl
```
Noordfrees:
```suggestion
	GameHost* host = nullptr;  // will be deleted by ctrl
```

Noordfrees (approved):
Works now :)
One nit inline

matthiakl:
Thanks for the review :)

PR #5130, opened 2021-11-20
remove message when muted via "mute all"

### LINKED ISSUE #4957: Delete Inbox message when the site is muted

jmoerschbach:
Closes #4957

Is there a better way to fetch all buildings of a certain type of a player than "misusing" the building statistics? :)


Noordfrees (commented):
You can put the `Game& game = ...` line just before the `for` loop, so the expensive cast is performed only once instead of in every iteration. Also instead of calling `set_mute_messages` twice you can just invoke the `muted` signal directly as `current_building.muted(current_building.serial())`.

Will test this now

---

> Is there a better way to fetch all buildings of a certain type of a player than "misusing" the building statistics? :)

Actually this is exactly what the building stats are intended for :) They are a cache for all buildings a player owns, and such a cache is the only performant way to iterate over multiple map objects without knowing all their positions in advance.

jmoerschbach:
thx, done :)

Noordfrees (commented):
Tested and working :) 

I just realized we don't need to cast the `egbase` to a `Game` at all. `map` is a function of `EditorGameBase`, so you can delete the `Game& game = ...` line and use `egbase.map()` instead of `game.map()`.

Noordfrees (approved):

PR #5132, opened 2021-11-21
Github MSVC CI action

### LINKED ISSUE #1081: Fix logging of error messages on Windows
STDERR gets lost in windows, error messages are neither printed to console nor to the stdout.txt.

Follow-up from https://bugs.launchpad.net/widelands/+bug/1398536

 * date created: 2018-09-14T06:43:59Z
 * owner: gunchleoc
 * the launchpad url was https://bugs.launchpad.net/bugs/1792521

### LINKED ISSUE #5022: Automatic check test_ai fails for windows x86 Builds
See e.g.: https://github.com/widelands/widelands/pull/4823/checks?check_run_id=3304557612

Error:
~~~~
undefined reference to `_imp__lerc_encodeForVersion'
~~~~

Full error:
~~~~
C:/msys64/mingw32/bin/../lib/gcc/i686-w64-mingw32/10.3.0/../../../../i686-w64-mingw32/bin/ld.exe: C:/msys64/mingw32/bin/../lib/gcc/i686-w64-mingw32/10.3.0/../../../../lib\libtiff.a(tif_lerc.o):tif_lerc.c:(.text+0x852): undefined reference to `_imp__lerc_encodeForVersion'
~~~~

matthiakl:
Close #5022
Fix #1081 (writes to stderr.txt)
New build action, that builds the Windows installers using MSVC.
The libraries are built with vcpkg and cached. 

Populating the cache takes about 70 minutes, on subsequent builds the packages are installed in just 2 minutes.
Until the build script is approved I use format.yaml for the cache key, for shorter evaluation cycles. 
Then before merging build.yaml will be used, so the cache is regenerated, when dependencies change.

Size is about the same as with MinGW, compilation time is much shorter (70min -> 30min).

How about the appveyor build? Should it also use MSVC? Is it actually needed in additon to the Github action?

The build script contains some additional steps to reduce waiting time and testing possibilities, which will be removed before merging.
Currently the InnoSetup seems to be broken, the game does not start. But if you replace the exe with the separately uploaded one it runs fine. So any help regarding the installer is appreciated.

I don't see any license issues for usage as open source project, but posted a question on the [support forum](https://github.community/t/github-action-visual-studio-licensing/213190/1) (currently awaiting review from staff)


hessenfarmer:
well the appveyor builds are used for convenience to provide downloadable installers of the latest development version for the community. If we can have such convenience without appveyor we could do however I would prefer if we get MSVC builds to run there as well.

Noordfrees:
GitHub build artifacts can only be downloaded by registered GitHub users, so we need to keep AppVeyor to provide publicly available builds.

matthiakl:
Ok, now that I now the reason for appveyor, it makes sense to change it as well. 
Did you notice, that the tests over there fail with segfault since #5122 ([example](https://ci.appveyor.com/project/widelands-dev/widelands/builds/41576111/job/xo3mv41nkggxt0xl))?

Noordfrees:
Yep :( 
Unfortunately the segfaults seem to happen only in CIs and are hard to check locally. The daily PPA is also affected. See discussion from https://github.com/widelands/widelands/pull/5122#issuecomment-972647161 onwards

hessenfarmer:
> Yep :( Unfortunately the segfaults seem to happen only in CIs and are hard to check locally. The daily PPA is also affected. See discussion from [#5122 (comment)](https://github.com/widelands/widelands/pull/5122#issuecomment-972647161) onwards

they do happen in my local MSYS environment I just didn't find time to hunt them down with gdb.

matthiakl:
Github builds are working now. In a future step we can clean the log a bit, now that we have it in the CI environment.
I also copied the step over to the appveyor workflow. Is there any chance to test this before merging into master?
Should we leave the MinGW builds until they break again, or remove them directly?

Regarding publicly available development versions we can think about using an action like [marvinpinto/action-automatic-releases](https://github.com/marvinpinto/action-automatic-releases) or [ncipollo/release-action](https://github.com/ncipollo/release-action) to create a rolling "latest" pre-release which is updated with every new build. Disadvantage here is, that it is not meant to keep a history.

Noordfrees:
>  I also copied the step over to the appveyor workflow. Is there any chance to test this before merging into master?

Adding this branch to the `branches` section in `appveyor.yml` should work. Since the branch is in a fork, you might need to enable your fork on AppVeyor first, or copy the branch to the main repo.

> Should we leave the MinGW builds until they break again, or remove them directly?

Is it likely that the MinGW builds fail while the MSVC builds succeed, or vice versa? (Not counting the usual dependency changes.) If not, let's remove them now; otherwise we can keep them while they work.

> Regarding publicly available development versions we can think about using an action like [marvinpinto/action-automatic-releases](https://github.com/marvinpinto/action-automatic-releases) or [ncipollo/release-action](https://github.com/ncipollo/release-action) to create a rolling "latest" pre-release which is updated with every new build. Disadvantage here is, that it is not meant to keep a history.

:+1: Sounds good to me. We don't need a history for daily builds, and always having the latest daily build artifacts in the same place would be another plus.

matthiakl:
> Adding this branch to the branches section in appveyor.yml should work. Since the branch is in a fork, you might need to enable your fork on AppVeyor first, or copy the branch to the main repo.

Thanks, I will try that.

> Is it likely that the MinGW builds fail while the MSVC builds succeed, or vice versa? (Not counting the usual dependency changes.) If not, let's remove them now; otherwise we can keep them while they work.

MinGW builds are essentially g++ builds plus the dependency handling. MSVC builds are more likely to fail while other builds succeed, because it is a completely different compiler.

> Sounds good to me. We don't need a history for daily builds, and always having the latest daily build artifacts in the same place would be another plus.

My next project then ;)

matthiakl:
With populating the vcpkg cache appveyor hits the 2hr time limit (https://ci.appveyor.com/project/widelands-dev/widelands/build/job/hagk9d6ga4bhhjnl).
So I suggest leaving it on mingw for now and try the Github automatic pre-release later.

src/logic/map_objects/tribes/productionsite.cc
```diff
@@ -629,30 +629,30 @@ void ProductionSite::try_replace_worker(const Game* game,
                                         DescriptionIndex worker_index,
                                         WorkingPosition* wp) {
 	// Search replacement worker in the same building
-	for (auto wp_repl = working_positions_.begin(); wp_repl != working_positions_.end(); ++wp_repl) {
-		if (!wp_repl->worker.is_set()) {
+	for (auto& wp_repl : working_positions_) {
```
hessenfarmer, line 632: these changes seem to be unrelated to the purpose of this Branch what is the purpose?

hessenfarmer (approved):
besides the inline comment this branch looks good.
As we keep the MINgw builds in appyeyor for the moment being I volunteer to cope with issues there.

src/logic/map_objects/tribes/productionsite.cc
matthiakl: MSVC refused to compile line 655 (`wp_repl.base()`). Therefore I had to modify the code to use references instead of iterators.
We will have such compile differences sometimes in the future, now that MSVC compiles every version ;)

hessenfarmer: ok thanks for explanation

Noordfrees:
LGTM from my side as well. So is this ready to be merged now? (This branch requires a forced-merge and the branch protection rules will need to be updated afterwards.)

matthiakl:
If you want to keep the mingw builds for now, it is ready. 
An alternative would be to skip mingw directly and rename the windows build back. Then the old rules may stay without modifications.

Noordfrees:
Hm, some PR builds seem to be failing while compiling the dependencies?
E.g. https://github.com/widelands/widelands/runs/4365070173?check_suite_focus=true 
```
-- Building x86-windows-static-dbg
CMake Error at scripts/cmake/vcpkg_execute_build_process.cmake:155 (message):
    Command failed: "C:/Program Files/CMake/bin/cmake.exe" --build . --config Debug --target install -- -v -j3
    Working Directory: C:/vcpkg/buildtrees/sdl2-mixer/x86-windows-static-dbg
    See logs for more information:
      C:\vcpkg\buildtrees\sdl2-mixer\install-x86-windows-static-dbg-out.log

  installed/x64-windows/share/vcpkg-cmake/vcpkg_cmake_build.cmake:102 (vcpkg_execute_build_process)
  installed/x64-windows/share/vcpkg-cmake/vcpkg_cmake_install.cmake:41 (vcpkg_cmake_build)
  ports/sdl2-mixer/portfile.cmake:51 (vcpkg_cmake_install)
  scripts/ports.cmake:142 (include)


Error: Building package sdl2-mixer:x86-windows-static failed with: BUILD_FAILED
Please ensure you're using the latest portfiles with `git pull` and `.\vcpkg update`.
Then check for known issues at:
  microsoft/vcpkg/issues (is:open in:title sdl2-mixer)
You can submit a new issue at:
  microsoft/vcpkg/issues/new?template=report-package-build-failure.md&title=[sdl2-mixer]+Build+error
including:
  package: sdl2-mixer[core,libflac,libvorbis,mpg123,nativemidi]:x86-windows-static -> 2.0.4#13
    vcpkg-tool version: 2021-11-24-48b94a6946b8a70abd21529218927fd478d02b6c
    vcpkg-scripts version: 60a8f7971 2021-11-29 (24 hours ago)

Additionally, attach any relevant sections from the log files above.
Error: Process completed with exit code 1.
```

I temporarily removed the MSVC builds from the list of required checks again

matthiakl:
> Hm, some PR builds seem to be failing while compiling the dependencies?

Uh oh. It looks like vcpkg just received a new version. I will investigate, why it fails.
Hopefully this does not lead to similar problems like with mingw :(

Noordfrees:
And in #5118 I'm having trouble with a test segfaulting on MSVC. All other compilers are fine. It always fails in the first function that throws and catches an exception. Is there anything "special" about MSVC that prevents the following construct from working?

https://github.com/Noordfrees/widelands/runs/4372123615?check_suite_focus=true
```c++
check_error("invalid placeholder", []() { bformat("%q", 1); });
```
```c++
#define check_error(what, fn) do_check_error(__FILE__, __LINE__, what, fn)
inline void do_check_error(const char* f,
                           uint32_t l,
                           const std::string& what,
                           std::function<void()> fn) {
	std::cout << "Running testcase " << f << ':' << l << std::endl;
	// The line above is the last to appear in the log output
	try {
		fn();
	} catch (...) {
		std::cout << "   ... [NOCOM] OK, safely caught." << std::endl;
		return;
	}
	std::cout << "   ... [NOCOM] ERROR: No exception received!" << std::endl;
	throw WException(f, l, "Error not detected: %s", what.c_str());
}
```

matthiakl:
> Is there anything "special" about MSVC that prevents the following construct from working?

Yes. I had a similar problem with stripped debug builds. Non-stripped debug builds work, but it seems, that this did not resolve the core problem. MSVC has a special [exception handling model](https://docs.microsoft.com/en-us/cpp/build/reference/eh-exception-handling-model?view=msvc-170#default-exception-handling-behavior). The default handling does not sound so well behaved:
> Resource leaks and undefined behavior may result when a C++ exception is thrown.

You might try adding `/EHsc` to the compile flags in 
https://github.com/widelands/widelands/blob/2c4e78b9a0e0176bb873d8e816d7716f85938f68/CMakeLists.txt#L377-L380

matthiakl:
> Hm, some PR builds seem to be failing while compiling the dependencies 

This is an upstream bug in sdl2-mixer, which has been fixed today and should be updated in vcpkg soon. Then the build should work again without actions from our side.

PR #5143, opened 2021-11-26
Deploy build artifacts to automated pre-release

matthiakl:
A new CI/CD job, that publishes the build artifacts, which are only accessible to Github users to a public pre-release.
After this, the appveyor builds should no longer be needed (https://github.com/widelands/widelands/issues/5132#issuecomment-974877948). Also Windows and Mac builds are in one place then.
The action automatically creates a tag `latest` with a commit list in the changelog, a customizable body will be added in a next version. To prevent overloading the initial run, we should add this tag to the master branch. It will be updated with every new build.
I tested it [on my fork](https://github.com/matthiakl/widelands/releases)


.github/workflows/build.yaml
```diff
@@ -280,3 +281,54 @@ jobs:
...
         path: ${{ github.workspace }}/Widelands-${{ github.sha }}-${{ matrix.config }}-${{ matrix.compiler }}.dmg
+
+  dev_release:
+    name: Automated development pre-release
+    # TODO: switch to 'refs/heads/master' when merging
+    if: ${{ github.ref == 'refs/heads/ci_dev_release' && always() }}
+    needs: [windows, macos]
+    runs-on: "ubuntu-latest"
+    steps:
+      - name: Check origin
```
Noordfrees: You can put this directly in the `if` expression, like for the launchpad mirror workflow

.github/workflows/build.yaml
```diff
@@ -280,3 +281,54 @@ jobs:
...
+          done
+      - name: List artifacts
+        run: |
+          find . -name 'Widelands*' -type f -printf '%f\n' | tee artifacts.list
+      - name: Updating latest pre-release
+        # Creates a new pre-release with the "latest" tag and all gathered artifacts.
+        # Because all older artifacts are removed, we have to reupload the missing ones detected in the previous step
+        uses: "marvinpinto/action-automatic-releases@v1.2.1"
+        with:
+          repo_token: "${{ secrets.GITHUB_TOKEN }}"
```
Noordfrees: Use `${{ secrets.WIDELANDS_FORMAT_TOKEN }}` here to have the releases published by bunnybot, for a touch of Widelands flavour :)

.github/workflows/build.yaml
```diff
@@ -280,3 +281,54 @@ jobs:
...
+          do
+            [ "${ARTIFACT%%-*}" != "Widelands" ] && continue
+            MATRIX=$(echo $ARTIFACT | sed 's/Widelands-[0-9a-f]*-//' -)
+            ls ../Widelands*/Widelands*${MATRIX} || \
+              curl -L -O "$RELEASE_URL/$ARTIFACT"
+          done
+      - name: List artifacts
+        run: |
+          find . -name 'Widelands*' -type f -printf '%f\n' | tee artifacts.list
+      - name: Updating latest pre-release
```
Noordfrees, line 395: Directly before this step, you need to check that the commit that triggered this workflow is still the latest one, and `exit 1` if this is not the case. Builds do not always complete in the order in which they were triggered, so this could end up overwriting a newer release by accident without such a check. See the launchpad mirror or format actions how to check this

Noordfrees (commented):
Very nice :) 

Would it be possible to upload the artifacts with a fixed name to the release (e.g. "Widelands_daily_Windows_x64_release.exe", "Widelands_daily_MacOS_debug.dmg")? Then we can add permalinks to the artifacts from https://www.widelands.org/wiki/Download/.

Also, could we generate and upload an md5 checksum file for each artifact like launchpad does, so users can verify that the installer has not been corrupted during the up- or download (GitHub does that sometimes)?

matthiakl:
> Would it be possible to upload the artifacts with a fixed name to the release (e.g. "Widelands_daily_Windows_x64_release.exe", "Widelands_daily_MacOS_debug.dmg")? Then we can add permalinks to the artifacts from https://www.widelands.org/wiki/Download/.

Yes, the commit sha is now replaced with "daily". The artifacts.list still lists the actual commit, in case a build is from an older one than the latest.


> Also, could we generate and upload an md5 checksum file for each artifact like launchpad does, so users can verify that the installer has not been corrupted during the up- or download (GitHub does that sometimes)?

Done.

Edit: if you want to see my tests: I used a dummy repo to not spam this branch (https://github.com/matthiakl/actions/releases)

.github/workflows/build.yaml
matthiakl: Added from format.yaml

.github/workflows/build.yaml
matthiakl: Thanks, that's simpler

.github/workflows/build.yaml
```diff
@@ -280,3 +281,69 @@ jobs:
...
+    if: ${{ github.repository == 'matthiakl/widelands' && github.ref == 'refs/heads/ci_dev_release' && always() }}
+    needs: [windows, macos]
+    runs-on: "ubuntu-latest"
+    steps:
+      - name: Download artifacts
+        uses: actions/download-artifact@v2
+      - name: Check missing artifacts
+        # This step ensures, that the development release always contain an artifact for every platform
+        # artifacts.list lists all attachements of the pre-release. In case a build step failed, we detect
+        # that a file is missing and download the last successful build from the pre-release.
```
Noordfrees:
```suggestion
        # This step ensures that the development release always contains an artifact for every platform.
        # artifacts.list lists all attachments of the pre-release. In case a build step failed, we detect
        # that a file is missing and download the last successful build from the pre-release.
```

Noordfrees (commented):
LGTM :) 

One more question, Is it possible to set a long release description via this action? Or if not, can we set one manually or would it be overwritten on every run?

matthiakl:
> One more question, Is it possible to set a long release description via this action? Or if not, can we set one manually or would it be overwritten on every run?

This action autogenerates the description and will overwrite any existing one. A custom description will be added in the future, there is a PR open for it. However, it is possible to append a custom description after creating the release.
See my latest commit for an example. The description can be hard coded or read from a file.

Noordfrees (commented):
LGTM, great work :)
Ready to be merged when the TODOs are addressed.

matthiakl:
> Ready to be merged when the TODOs are addressed.

Done. 
Please consider creating the `latest` tag on master before merging. Otherwise the action will include all commits since v1.0 in the description.

Noordfrees (approved):
>  Please consider creating the latest tag on master before merging. Otherwise the action will include all commits since v1.0 in the description.

Done: https://github.com/widelands/widelands/releases/tag/latest

PR #5149, opened 2021-12-05
Fix auto speed

hessenfarmer:


src/ai/defaultai.cc
```diff
@@ -277,6 +279,27 @@ void DefaultAI::think() {
...
 	uint32_t jobs_to_run_count = (delay_time < 0) ? 0 : 1;
 
+	// This portion of code keeps the speed of game so that FPS are kept within
+	// range 13 - 15, this is used for training of AI
+	if (game().is_auto_speed()) {
+		int32_t speed_diff = 0;
+		if (delay_time > 3000) {
+			speed_diff = -100;
+		}
+		if (delay_time < 500) {
```
Noordfrees:
```suggestion
		} else if (delay_time < 500) {
```

src/wui/interactive_base.cc
```diff
@@ -752,30 +752,6 @@ void InteractiveBase::draw_overlay(RenderTarget&) {
 
 	Game* game = dynamic_cast<Game*>(&egbase());
```
Noordfrees: If you replace the checks for `game == nullptr` with `!egbase().is_game()` below, you can now remove the dynamic cast here

tibor95:
As a concept it looks OK, however I would like to experiment with numbers. Also with jobs_to_run_count variable, maybe make it static like 5, to simplify the logic and code.
Also I am not sure that 
game().is_auto_speed()
actually works. It was broken before. I had to replace it with true to work. Please verify.

hessenfarmer:
game().is_auto_speed() is actually working using the console parameter --auto_speed    
this branch is adjusting the speed I believe using delay time is ok as it is actually measured.
Jobs to run is just a secondary value determinded from delay_time

src/wlapplication_messages.cc
```diff
@@ -88,8 +88,8 @@ void fill_parameter_vector() {
 		  "description of the AI training logic"),
 		false},
 	  {"", "auto_speed", "",
-		_("In multiplayer games only, this will keep adjusting the game speed "
-		  "automatically, depending on FPS. Useful in conjunction with --ai_training"),
+		_("This will keep adjusting the game speed automatically, depending on AI delay."
+		  "Only to be used for AI testing or training (in conjunction with --ai_training)"),
 		false},
```
Noordfrees: I think now this option will be used by about 0.001% of players, so the `ai_training` and `auto_speed` options should be made Verbose options by changing `false` to `true` here and in line 89

hessenfarmer:
increased decision values to allow for higher game speeds
now allowing 15x at the beginning and reducing to about 13x if a lot of calculations have to be made

tibor95:
Well, I made it work. However AI is doing basically nothing. Is this current status of AI?

tibor95:
I created my branch with some other changes - for discussion only if you wish. But there is at least one thing that might be fix of a problem

hessenfarmer:
> Well, I made it work. However AI is doing basically nothing. Is this current status of AI?      
     
definitly not. could you provide a replay? or describe the steps you took.  in my local compilation the AI is doing fine. 

Usually I take a balanced map to check for all tribes. So curently I use "stone rivers" to check with the 5 in game tribes and "hexagony" to check with europeans additionally. then I set up a Multiplayer local game with 5 (6) AI and me as spectator and watcht what it is doing.   
For the first 10 seconds it is doing indeed nothing. then it builds 2 lumberjacks. after around 30 secs it starts building 2-3 small military buildings and from then on it starts building up basic economy.

src/ai/defaultai.cc
```diff
@@ -320,6 +341,11 @@ void DefaultAI::think() {
 		}
 	}
 
+	if (GameController* const ctrl = game().game_controller()) {
+		verb_log_dbg_time(gametime, "Jobs: %d; delay: %d; gamespeed: %d \n", jobs_to_run_count,
+		                  delay_time, ctrl->real_speed());
```
tibor95: I like this, what about preserving it? If condition it with 'if jobs_to_run_count > 3' to have this as warning?

tibor95:
If you see a good way how to get rid of these asserts:

assert(jobs_to_run_count > 0 && jobs_to_run_count <= kMaxJobs);
assert(jobs_to_run_count < taskPool.size());
assert(!current_task_queue.empty() && current_task_queue.size() <= jobs_to_run_count);

I mean saving a line, no point to change line for another line, please do it. Otherwise it looks ok

src/ai/defaultai.cc
hessenfarmer: maybe we can just move it to our debug stats section Otherwise I would guard it with the same bool used there. 
I'll provide a proper solution to keep it for debugging.

tibor95: whatever looks good to you

src/ai/defaultai.cc
```diff
@@ -273,6 +274,26 @@ void DefaultAI::think() {
 
 	const int32_t delay_time = gametime.get() - taskPool.front().due_time.get();
 
+	// This portion of code keeps the speed of game so that FPS are kept within
+	// range 13 - 15, this is used for training of AI
+	if (game().is_auto_speed()) {
+		int32_t speed_diff = 0;
+		if (delay_time > 5000) {
+			speed_diff = -100;
+		} else if (delay_time < 1000) {
```
tibor95, line 283: Perhaps you can put there 1500.... to achieve bigger speeds

EDIT, and reduce 5000 to 4000

hessenfarmer: ok will try. however reducing this to 4000 means we will get the 3rd job running only on rare occasions, as we start reducing speed before 4500 is reached,

tibor95: Hm,I just mean that narrow range would be better, so perhaps 3000-5000 .... I will play with this when doing training. Without longer use it is a guesswork for me..
I would like to achieve average speed up about 6x in 3 hours game.

tibor95:
So what is the status? Do you plan any changes yet?

hessenfarmer:
yes I do. latest tonight

hessenfarmer:
ok I made the logging be switchable by kPrintStats in the code. 
I experimented with some numbers:
Testsetup was 8 AI on Lesser Ring 1.2

upper vallue    lower value   startspeed    3h speed 
4000                1500              15x               8x
5000                1500              16x               10.5x
5000                1000              15.5x            9.5x
6000                1000              16.5x            9.5x (still 8.5x after 11hours)

with the last values we had 5 hits of AI too late after 8.5 hours until 10 hours.   
with the first 2 value pairs the speed was going up and down very often and fast. 
So I chose the intermediate of the last 2 ones with 5500 as upper value. this should ensure we won't hit the too late state often and still deliver proper AI handling with sufficient game speed.

tibor95:
I am compiling right now.

Looks good to me, if my test will go OK (as I presume) we will be able to close/approve this.

BTW I do not mind frequent changes of the speed, I see no harm in it :)

tibor95:
Your computer is definitely faster than mine.
I am approving

tibor95 (approved):
good change

hessenfarmer:
thanks for the review and the discussions we had. 
i just realised I forgot something (correct the comment and condition to check for 2nd job). 
May I ask which values you could achieve for gamespeed on your machine? Maybe make the function to determine 2nd job linear might help. 

Edit: I'm afraid linear fuction won't help. I couldn't see much difference

tibor95:
In fact, as that script stuff I mentioned in the forum does not seem to work with games from template, I did not did any serious experiments yet. So really I have no counter-suggestions. Anyway it is good enough for now, remember I am compiling my code anyway, so I do here and there tiny changes, e.g. for logs output and so no big deal for me to tweak this...

hessenfarmer:
ok let's have it then

PR #5151, opened 2021-12-06
Fix auto speed + incl. changes by Tibor

tibor95:
I added some changes to fix auto speed


src/ai/defaultai.cc
```diff
@@ -270,15 +267,13 @@ void DefaultAI::think() {
 
 	SchedulerTaskId due_task = SchedulerTaskId::kUnset;
 
-	sort_task_pool();
```
hessenfarmer, line 272: I believe we need to sort taskpool here one time to have the longest due time to calculate delay properly

src/ai/defaultai.cc
```diff
@@ -294,9 +289,12 @@ void DefaultAI::think() {
...
 
-	// Here we decide how many jobs will be run now (none - 5)
-	// in case no job is due now, it can be zero
-	uint32_t jobs_to_run_count = (delay_time < 0) ? 0 : 1;
+	// Here we decide limit for jobs will be run now
+	// generelly 3, if situation is wrong: 4
+	uint32_t jobs_to_run_count = std::min<uint32_t>((delay_time > 3000) ? 5 : 3, taskPool.size());
+
+	// we need to get more urgent jobs to the front
+	sort_task_pool(jobs_to_run_count);
```
hessenfarmer: this would run sort taskpool more often then necessary. In current design we have only 1 sorting cycle as normally only 1 job is run per each AI think() cycle as long as delay is not greater then 2000. (the code comment below is wrong). here we would sort 3 times even if there is nothing to do. Performancewise this isn't good I believe. especially while taking into account we need to sort one time in advance of this code

src/ai/defaultai.cc
```diff
@@ -305,43 +303,30 @@ void DefaultAI::think() {
...
-	}
-
 	// And printing it now and resetting counter
 	if (scheduler_delay_counter_ > 10) {
 		log_warn_time(gametime, " %d: AI: game speed too high, jobs are too late (now %2d seconds)\n",
 		              player_number(), static_cast<int32_t>(delay_time / 1000));
 		scheduler_delay_counter_ = 0;
 	}
 
-	// 400 provides that second job is run if delay time is longer then 1.6 sec
```
hessenfarmer, line 299: these lines and especially the comment are not up to date.   
currently second job is run if delay > 2000 and third one if delay > 4500 and fourth if delay > 8000
so this if could safely evaluate for delay >=2000 to spare some sqrt calcualtions.     
furthermore these values would make for good values in our auto_speed function where we decide the max delay allowed and by this the max jobs to be run. so I believe 5000 wasn't that bad a value as it allows for the third job to be run.

hessenfarmer (commented):
some inline comments. Ican't see the advantage of this design over the old one. rather I see a (small) performance drop due to unecessary sorting.  
maybe you could explain which issue you want to tackle.

src/ai/defaultai.cc
tibor95: good point!

src/ai/defaultai.h
```diff
@@ -185,7 +185,7 @@ struct DefaultAI : ComputerPlayer {
 	BuildingNecessity
 	check_building_necessity(BuildingObserver& bo, PerfEvaluation purpose, const Time&);
 	BuildingNecessity check_warehouse_necessity(BuildingObserver&, const Time& gametime);
-	void sort_task_pool();
+	void sort_task_pool(const u_int32_t);
```
Noordfrees:
```suggestion
	void sort_task_pool(uint32_t);
```

src/ai/defaultai.cc
tibor95: In fact, in old design and this new design more than 1 jobs were allowed. But here we set it explicitely to 3 or 5. More readable. Of course this is upper limit, we will run only those jobs that are overdue.

You are wrong - sort_task_pool() is run only once, but I changed design a bit, and in fact here was probably a problem before and should be fixed. Of course I will move it to proper position in next commit

src/ai/defaultai.cc
tibor95: Yes, but I propose simpler logic - first of all, all overdue jobs deserve to be run, however we have to limit how many of them will be run - for performance reasons. 
I am not sure your understanding of the old code is right. It is my code BTW, so I remember my intend: 'the more delayed are the more jobs to be executed.'

src/ai/defaultai.cc
tibor95: I found we need to run sort_task_pool twice, after all...

tibor95:
I pushed the change.

Proposed changes simplifies stuff, it is quite subjective if they are wronger, or better. But I think this is cleaner now.

tibor95:
The issue I wanted to fix were not an issue after all, see here:
	for (uint8_t i = 0; i < jobs_to_run_count; ++i) {
		if (taskPool[i].due_time <= gametime) {
			current_task_queue.push_back(taskPool[i]);
			sort_task_pool();   <-------------
		} else {
			break;
		}
	}

I was afraid that is does not sort properly, but yes, call it multiple times, does what we want. However, now we will call it only once with parameter

tibor95:
Also I believe less code (lines of code) easier to read and understand. AI design is so complicated so I vote for simplification...

src/ai/defaultai.cc
hessenfarmer: ok I have to be more precise. in the old design sort_taskpool() was running only one sorting cycle for each call ensuring the oldest task in front. it was run one time before the block to evaluate delay time and then only if this was positive  a task was pushed to the temporary queue. Only after this push happened an next sort cycle was run to get the second oldest task in second place, just in case a second job had to be run. so we had a potentially useless sort cycle here as well.   
with your design we would sort 3 times or 5 times. then we would check the first 3 or 5 entries whether to push them to the queue just to realize there is nothing to do in worst case.  
I agree your decision algorithm is much more readable then the algorithm based on the sqrt of delay time but it is much simpler as well as it could only have 2 values 3 or 5. Furthermore this has the risk that if 3 jobs are due at the same time we will run all 3 of them blocking the logic functions for the complete time (at least as long as we can't run the AI ins its own thread. this will lead to the game seem to be halted. while if we only run 1 in each cycle the logic code could be run in between, this would mean the 2nd and third task might be delayed by about 500ms (if run at 1x speed) or multiples of this. so the current (old) design is better in terms of load distribution.       
So I believe the task to be added to the temporary queues were properly sorted (due to the call to sort_taskpool() after each push) if this was your main concern.

hessenfarmer:
> Also I believe less code (lines of code) easier to read and understand. AI design is so complicated so I vote for simplification...

In general I agree however the big advantage of increasing the jobs to run with a function beginning with 1 is load distribution so we aim at some delay to not block logic code for too long.

src/ai/defaultai.cc
hessenfarmer: Fully understood this. And when running the game at max. speed there might be not much difference but in a normal case of a game run at 2x or 3x max there might be no delays at all and running 3 jobs at once would cause a visible glitch I'm afraid.
So your new design would be fine as long as the jobs are increased starting with one. Function maybe linear instead of quadratic though to simplify calculations .

src/ai/defaultai.cc
hessenfarmer: > 

in this case just provide (jobs -1 ) to the second run

src/ai/defaultai.cc
```diff
@@ -269,13 +267,34 @@ void DefaultAI::think() {
...
+				    (ctrl->desired_speed() > 29999 && speed_diff < 0)) {
+					ctrl->set_desired_speed(ctrl->desired_speed() + speed_diff);
+				}
+			}
+		}
+	}
+
+	// Here we decide limit for jobs will be run now
+	// generelly 3, if situation is wrong: 4
+	uint32_t jobs_to_run_count = std::min<uint32_t>((delay_time > 3000) ? 5 : 3, taskPool.size());
```
hessenfarmer, line 297:
```suggestion
	uint32_t jobs_to_run_count = std::min<uint32_t>((delay_time > 3000) ? 5 : 3, taskPool.size());
```
        uint32_t jobs_to_run_count = std::min<uint32_t>(delay_time / 2000 + 1, taskPool.size() / 2);

hessenfarmer, line 297:
```suggestion
	uint32_t jobs_to_run_count = std::min<uint32_t>(delay_time / 2000 +1, taskPool.size() / 2);
```

src/ai/defaultai.cc
tibor95: This task pool is usually about 15 members, the sorting overhead is not that big. Based on my debugging, AI most of time has 3 jobs overdue, so unneeded sorting is very rare. But you are dead right about '-1', I will modify this.

I do not understand you concern about multiple jobs run within one AI think(). This is not new, it worked before...

src/ai/defaultai.cc
tibor95: So, if we reduce these 500 ms to 200 and jobs that might be run in one think() to 1 or 2, this is what you would prefer?

src/ai/defaultai.cc
tibor95: taskPool.size() is there only to make sure we do not exceed its size. For case something wrong happen and it is not properly filled. Should not happen, but this will save us from one of two rows checking if jobs_to_run are lesser than the count of our tasks.
What if delay will be 15 seconds, is 8 jobs fine for you?

src/ai/defaultai.cc
hessenfarmer: Not really.   
as long as the ai is in the same thread with the logic code it has to share computing time with logic. The UI is drawn now in its own thread so the fps stay up which is good. But this makes it clearly visible when lenghty prozesses run preventing the logic of the game being updated (this was the case letting frames drop previously). 
So running more jobs in the same amount of time would help AI Training probably but would have effects worsening #4963.
So I would keep the 500ms. and start the function to determine jobs to run with 1 job. Only if delay builds up (due to increasing game speed or lenghty calcualtions in the logic due to game size) we should increase the jobs to run. this can be done in a linear way. 
So to sum it up we could use your new layout but you might consider implementing my suggestion.

tibor95:
Also what we can do is to review these "required run time" that individual tasks sets. I do not think they were ever properly reviewed... With goal to extend them where possible...

hessenfarmer:
> Also what we can do is to review these "required run time" that individual tasks sets. I do not think they were ever properly reviewed... With goal to extend them where possible...

Agreed about review of the intervals of our tasks. However they should reflect the intervall a good player checks these things as well. May be even a bit more often. (especially the construct_building task. Most probably this will require some experimenting as well.

src/ai/defaultai.cc
hessenfarmer: > 

as explained below the concern is not about this working. it is just about keeping the display smooth as long as possible.

src/ai/defaultai.cc
hessenfarmer: in theory yes of course trying to get the delay removed by computing more tasks would be good. But thewre is a frontier if the AI uses nearly all of the processing time it creates delay itself so we might run into a stall if we don't limit this to a certain value (was 4 before). i am fully aware this would limit the jobs to 7. but this should be fine as we probably won't be able to manage the situation if delay still grows although this many jobs get run

src/ai/defaultai.cc
tibor95: After sorting of tasks, delays can be:
5500, 4800, 4200, 4100, 2800, 2200....
vs 
5500, 1200, 300, -500, -800, -2100
So the delay of first member is not ideal quantification how delayed we are.

Let say we check 5 military sites (no matter of total count) for attack every 10 seconds and here we say, sorry - every 20 seconds it will be, while enemy military sites count grows, it will detoriate performance of AI, though you will have preserves fluentness of the game.

tibor95: There is some internal inconsistency - one piece of AI code says - this task should be executed each 10 seconds, and other piece of code will say - it can be delayed by 5 seconds and nobody cares...

tibor95: What about to have minimal delay for jobs execution
for first job if delayed
for second if delayed for more than 300 ms
for third if delayed for more than 600 ms
and so on

hessenfarmer: I agree on the first delay not ideal to measure delay. However it is a very good indication and if we are really delayed (your example 1) we would be delayed in the next cycle as well causing multiple jobs to run again. 
As said we check every 500 ms so the last task not processed in the list will be delayed 500 ms realtime which will be 5 secs at 10x gamespeed. 
I think we aim at an basic interval for each task but we can delay each of them by a certain amount of time. This is why the algorithm in question takes place. we start to do something against delay only if we are 2 seconds gametime late.

src/ai/defaultai.cc
tibor95: I am lost. I suggest to limit jobs to run to 3 or 5, and this is too much for you, but 7 is OK?

src/ai/defaultai.cc
tibor95: 500s is gametime, or not? If realtime than this is very bad..

tibor95: So would you agree with:

1. we define jobs to run (can be hardcoded 5)
2. we will iterate over task queue (up to defined count above) and job will be executed if 
delay > position * 300
where position starts with 0

This would be acceptable for me, I think I even like it

hessenfarmer: it is realtime (and must be) as if it would be game time we would need to run an AI think cycle every 50 ms at 10x gamespeed where some of our tasks last for several 10ms (especially update all buildable fields is costly)

tibor95: Some information that some tasks collects has "expiration time" and it is expected that another tasks that uses this information are run while the information are not expired yet. This is probably why AI will not build anything in high speed - tasks deciding about buildings does not have "fresh" information about buildable spots. - Just general explanation

tibor95: BTW, AI newer update ALL buildable fields in one run of particular task. There is limit, ca 50 per one run of the task. But negative thing is that if you have 2000 buildable tasks ... well do your math :)

src/ai/defaultai.cc
hessenfarmer: This is not a contradiction in my eyes. I just want to ensure that the cpu time we require is distributed over time as equally as possible. so if everything is normal (let's say gamespeed 1x to 3x) and our delay is less then 2000ms this is best achieved by only running 1 job per think() cycle. If we really are getting late we need to to do something so in this case the requirement to distribute CPU times can't apply anymore but we need to use as much of it (together with the logic of the game) as we can to get our jobs done. So increasing it up to 7 should be a last ressort chance. 

the problem with starting with 3 would be that if the first 3 tasks will be due (i.e. duetime < gametime) they will be executed in one cycle. followed by perhaps 2 idle cycles in an ormal game. this would cause the game to stutter. so the main thing is to staret from 1 and only increase if needed.

src/ai/defaultai.cc
tibor95: I never did any time profiling of the tasks, would be interesting to know and have real info, but I really tried to keep them from going 'wild'

src/ai/defaultai.cc
tibor95: Even now, more than 1 task is run every think() - I guarantee. So I do not propose anything worse that it is now.

What about my suggestion to execute a job only if:
delay_time > position_in_queue * 300 ms?

src/ai/defaultai.cc
hessenfarmer: I know about the fields ( i recently tried to make this thread safe in my attempt to run the AI in its own thread). the problem why this is costly is we itrate over about 50 fields but for each field we check the fields in a radius of 8 for economy up to 22 for military aspects and we are doing this for each player. which is a lot of iterations obviously.  
in our current design (having the same thread as the logic code) the interval of the 2 tasks should ensure for the info not expire however if the gamespeed will set to progress to fast we will not be able to calculate everything needed, but the logic won't be able to do so neither so we are stalling the game.

tibor95: We can check less sites, more frequently. Now the frequency is:
constexpr Duration kMinBFCheckInterval(5 * 1000);
Of course 5 seconds are never obeyed with this our scheduler

tibor95: Generally if we do not measure something, than this is just a quesswork. Maybe there are bigger resource hogs in tasks than this update buildable fields. Maybe some cumulative CPU counter for each task would be nice, to get stats for a hour or so...

src/ai/defaultai.cc
hessenfarmer: the second job is run only if the first task has a delay of more then 2000 ms according to line
https://github.com/widelands/widelands/blob/b1df95a2a3aa80d2dd50aeccc43630bf9bb1eddf/src/ai/defaultai.cc#L301

so this can happen from time to time But I think normally only 1 task is run. I will try whether you might be right with some debug code though.

src/ai/defaultai.cc
hessenfarmer: in my attempts to make AI threading working I tried to set Mutex locks blocking the logic code in this function and it blocked the logic for a time clearly visible in the game. However logging SDL_ticks() at htne beginning an the end of a task might be a good idea to measure our code and identify performance improvements.

src/ai/defaultai.cc
tibor95: So you need delay 4,5 sec to run 3 jobs and 8 sec to run 4.... bit limiting, but do you tests...

src/ai/defaultai.cc
tibor95: I really meant something like using:
std::chrono::high_resolution_clock::now()
before and after of each task and total it over some period of time. Not now but for the future...

src/ai/defaultai.cc
hessenfarmer: first test: 5 AI on stone rivers. Search for jobs in the log output and see the latest commit in my branch for the loggging code.
result: even with gamespeed 8x mostly 1 job is run per cycle.  only at 16x sometimes a 3rd job is necessary. 
[stdout.txt](https://github.com/widelands/widelands/files/7668166/stdout.txt)
edit: I forhgot to upload the log in the first place so I uploaded now th 5 Ai version

src/ai/defaultai.cc
hessenfarmer: sounds good. maybe SDL_ticks is not precise enough.

src/ai/defaultai.cc
tibor95: what about bigger map with 4 AI players with auto_speed?

tibor95: which branch? Your fix_auto_speed?

tibor95: OK, looks good indeed. I will compile too and test myself

hessenfarmer: I just tested for a short time maybe we need to look at a long game with bigger map

tibor95: test runs :)

tibor95: AI just does not grow, so no point to wait long... It looks OK, we can stay with this. However I would comment your branch mostly to get rid of some asserts.
we can always return and finetune this later

hessenfarmer: Waht do you mean by AI just don't grow?

tibor95: I hoped that after 2 hours of gametime it would be bigger and meat each other, have bigger territory

tibor95: But this is different topic

hessenfarmer: It does exactly this in my setup. 
Which map did you use? could you provide me a replay? 
Currently a 8 AI match on Lesser ring is running at my side,

On crater for example usually the match between 2 AI is over after approx 4hours gametimeor earlier

tibor95: I tried multiple games. Never mind now...
What I noticed it runs much slower than auto_speed games used to run. But this is matter of tuning some values in the logic, I can experiment when training the AI

hessenfarmer:
short update: Amazon Ai beated all opponents on lesser ring after about 58 hours gametime. I saw 1.3 million cycles were 4 jobs were necessary in this game. but the game was running at 17x without auto_speed. delay was about 8400 at average.  However I can't tell if it was the same computerplayer all the time so I'll add playernumber to the debug

tibor95:
8400 average? So there were much longer delays too?
Ou - it was not auto-speed so I do not care much...

hessenfarmer:
as we merged the other branch I'll close this

PR #5158, opened 2021-12-15
Prevent defining sprintf to libintl_sprintf

matthiakl:
Fix Windows mingw builds
(see https://github.com/widelands/widelands/runs/4528487004?check_suite_focus=true)


hessenfarmer (approved):
fix is straightforward. 
thabks for the quick reaction

Issue #5100, opened 2021-10-27
Resolve asan testsuite issues


matthiakl:
Re #5090 
The testsuite reports some asan issues which should be addressed. See [this log](https://github.com/widelands/widelands/runs/3988961420?check_suite_focus=true)

Since leaks are suppressed all together in the CI, the testsuite should be run on a local machine to see, whether some leaks can be resolved as well.

Noordfrees:
Interestingly the failure always seems to happen on just three configurations: g++ 5 and 6 under Bionic; both release builds plus the g++6 debug build.
I'm compiling Widelands in a Bionic container now to look at it…

---

It's reproducible with g++5 under bionic in the Casern test. Will investigate.

Noordfrees:
The bug seems to be in `boost::format` (specifically I'm using version `1.65.1.0ubuntu1` there). 

Rewriting this function:
https://github.com/widelands/widelands/blob/9efa8ea23bee95cf33fcef5a07c3d526d4ad4d60/src/graphic/style_manager.cc#L385-L390
to use plain string concatenation instead of `boost::format` fixes this crash, only to get a new crash a second later elsewhere (in `as_mapobject_message`) as we use `boost::format` a lot. Gotta love this library, we never have any sort of problems with it whatsoever, oh no…

#5081 gathers 99% of our `boost::format` calls in a single `bformat` interface function. From there it's only a small(ish) step to ripping it out completely, which I would consider the most *desirable* approach to fixing this bug.

---

Edit: To clarify, the testsuite with ASan shows multiple bugs, of which several but not all are boost-related. They should best be addressed one by one until all checks are green.

PR #5160, opened 2021-12-16
Disable optimization on gcc <7

matthiakl:
Re #5100
In a local environment this fixed the ASAN testsuites errors. Let's see how it goes.


CMakeLists.txt
```diff
@@ -189,9 +189,14 @@ if(CMAKE_BUILD_TYPE STREQUAL "Debug")
...
 elseif(CMAKE_BUILD_TYPE STREQUAL "Release")
-  if(("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU") AND (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 6.2))
-    message(STATUS "GCC >6.2.0 breaks -03, setting -02")
-    set(WL_OPTIMIZE_FLAGS "-O2")
+  if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+    if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS 7.0)
+      message(STATUS "GCC <7 produce stack-buffer-overflow with -O2, setting -00")
+      set(WL_OPTIMIZE_FLAGS "-O0")
+    else()
+      message(STATUS "GCC >6.2.0 breaks -03, setting -02")
```
tothxa:
```suggestion
      message(STATUS "GCC <7 produce stack-buffer-overflow with -O2, setting -O0")
      set(WL_OPTIMIZE_FLAGS "-O0")
    else()
      message(STATUS "GCC >6.2.0 breaks -O3, setting -O2")
```

Noordfrees:
I had thought about trying this as well but I'm not sure it's the right approach. It's only a small number functions which are reproducibly affected so I'd prefer to fix only those few code pieces instead of completely disabling optimization globally. Did you measure how much of a difference this makes in execution time and binary size? If both is negligible it's probably fine but if there's a difference we should use this only as a last resort IMHO.

Edit: Also it always worked fine without ASan even if the buffer overflow might have gone unnoticed so turning optimization off should only be considered when ASan is enabled.

matthiakl:
I did a time measurement with a subset of the testsuite:
`time ./regression_test.py -b build/gcc6/src/widelands -r port`
 - -O0: (50M, 36M stripped)
 ```
Ran 20 tests in 703.711s
real  11m43.777s
user  9m45.796s
sys   0m31.319s
 ```

- -O2: (31M, 29M stripped)
```
Ran 20 tests in 626.817s
real  10m26.884s
user  8m23.952s
sys   0m27.724s
 ```
The size looks ok, but the time difference is not the best.

**Comment by tothxa** [comment]
Did you consider -O1?

**Comment by matthiakl** [comment]
> Did you consider -O1?

Yes, -O1 produces the same errors. From there I tried to disable single optimizations with -fno-xxx, but without success.
I also considered -O0 with some explicit optimizations, but they did not reduce the size significantly.

**Comment by Noordfrees** [comment]
GCC also supports a function attribute `optimize(level)` which might allow us to fix only those functions where optimization actually causes a bug. Would it be possible to write something like
```c++
#if ((__GNUC__ * 100) + __GNUC_MINOR__) < 700
#define NO_OPTIMIZATION __attribute__ ((optimize(0)))
#else 
#define NO_OPTIMIZATION
#endif
```
```c++
void dangerous_function() NO_OPTIMIZATION {
	...
}
```
and apply this macro to a small range of functions until the testsuite is happy

matthiakl:
I also tried this. It started promising, but I quickly ran into errors with stack traces starting from `boost::format`. 
Should I close this PR then and try again, when #5118 is merged?

PR #5161, opened 2021-12-18
Remove graphic_text_layout dependency from io_filesystem

tothxa:
(re my benchmarking comments to https://github.com/widelands/widelands/pull/5118 )

I'm not sure about the separate header file `illegal_filename_characters.h`, but I also don't know where else to put these vectors so that `illegal_filename_tooltip()` can find it.

I'm also not sure about `illegal_filename_tooltip.cc`, but if cmake needs a .cc file anyway, then why not make the header lighter.


src/io/filesystem/illegal_filename_characters.h
```diff
@@ -0,0 +1,42 @@
...
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ */
+
+#ifndef WL_IO_FILESYSTEM_ILLEGAL_FILENAME_CHARACTERS_H
+#define WL_IO_FILESYSTEM_ILLEGAL_FILENAME_CHARACTERS_H
+
+namespace {
+// Characters that are allowed in filenames, but not at the beginning
+static const std::vector<std::string> illegal_filename_starting_characters{
```
Noordfrees: Header files shouldn't contain anonymous namespaces or definitions of non-trivial types. Remove the namespace, declare the vector as 
`extern const std::vector<std::string> illegal_filename_starting_characters;`,
and move its definition (`std::vector<std::string> illegal_filename_starting_characters{...};`) to a .cc file. Same for other static variables in headers.

src/io/filesystem/illegal_filename_tooltip.h
```diff
@@ -0,0 +1,33 @@
...
+// We need to separate it from the rest because we don't want the low level
+// io/filesystem code to depend on all the GUI code this would pull in.
+
+#ifndef WL_IO_FILESYSTEM_ILLEGAL_FILENAME_TOOLTIP_H
+#define WL_IO_FILESYSTEM_ILLEGAL_FILENAME_TOOLTIP_H
+
+#include <string>
+
+namespace FileSystemHelper {
+extern std::string illegal_filename_tooltip();
```
Noordfrees: This shouldn't need the `extern` keyword I think

src/io/filesystem/illegal_filename_tooltip.cc
```diff
@@ -0,0 +1,70 @@
...
+
+namespace FileSystemHelper {
+
+std::string illegal_filename_tooltip() {
+	std::vector<std::string> starting_characters;
+	for (const std::string& character : illegal_filename_starting_characters) {
+		if (character == " ") {
+			/** TRANSLATORS: Part of tooltip entry for characters in illegal filenames. replaces the
+			 * blank space in a list of illegal characters */
+			starting_characters.push_back(pgettext("illegal_filename_characters", "blank space"));
```
Noordfrees:
```suggestion
			starting_characters.emplace_back(pgettext("illegal_filename_characters", "blank space"));
```
There's an implicit conversion from `const char*` to `std::string` here, which makes emplace_back more efficient

Noordfrees (commented):
:+1:

A few comments inline

src/io/filesystem/illegal_filename_characters.h
tothxa: I tried to do this in `filesystem.h` (leaving the definition in `filesystem.cc`, but it failed in final linking. I've just tried again with adding an `illegal_filename_characters.cc`, and it fails just the same:

```
/usr/bin/ld: /tmp/ccTz6BSq.ltrans7.ltrans.o: in function `FileSystem::is_legal_filename(std::__cxx11::basic_string<char,
std::char_traits<char>, std::allocator<char> > const&)':
<artificial>:(.text+0xa10e): undefined reference to `illegal_filename_starting_characters[abi:cxx11]'
/usr/bin/ld: <artificial>:(.text+0xa153): undefined reference to `illegal_filename_characters[abi:cxx11]'
/usr/bin/ld: /tmp/ccTz6BSq.ltrans18.ltrans.o: in function `FileSystemHelper::illegal_filename_tooltip[abi:cxx11]()':
<artificial>:(.text+0x2be44): undefined reference to `illegal_filename_starting_characters[abi:cxx11]'
/usr/bin/ld: <artificial>:(.text+0x2c20b): undefined reference to `illegal_filename_characters[abi:cxx11]'
```

src/io/filesystem/illegal_filename_tooltip.h
tothxa: You're right. I thought I used it to fix some compile error, but I was wrong...

src/io/filesystem/illegal_filename_characters.h
tothxa: Maybe this is related to why `find_unused_includes.py` thinks the .cc-s don't need `illegal_filename_characters.h`?

Noordfrees: [patch.txt](https://github.com/widelands/widelands/files/7739626/patch.txt)
this should fix both the linker errors and the include script

tothxa: Indeed... But why is it different when the definitions are in `...tooltips.cc` than when in `filesystem.cc`? Is it something like when both the decl. and def. get linked into `libio_filesystem.a`, the extern doesn't apply anymore?

But this way `filesystem.cc` still depends on `...tooltips.cc`, and even does so silently.

tothxa: hmmm, this seems to be the answer: https://stackoverflow.com/questions/875655/linking-extern-variables-in-c

So it can be left alone in filesystem.cc, there's no need for illegal_chars.h, and it should be declared as extern in tooltip.cc...

Trying it now...

-----
edit:
No, it didn't work...

tothxa: OK, turns out I can move all of it to the new lib. I would have sworn that I had checked that `is_legal_filename()` or the character lists were used by the filesystem code...

tothxa:
I'm open to suggestions for better file and namespace names.

Noordfrees (approved):

PR #5168, opened 2021-12-22
AI improvements - change of scoring, training

tibor95:
code changes affect only AI scoring
results of more than 100 rounds of training


.gitignore
```diff
@@ -33,6 +33,9 @@ CMakeLists.txt.user*
 /.vscode
 CMakeSettings.json
 
+# Visual Studio Code
```
hessenfarmer: this seems to be redundant to line 33

src/ai/defaultai.cc
```diff
@@ -280,9 +280,9 @@ void DefaultAI::think() {
 	// range 13 - 15, this is used for training of AI
```
hessenfarmer: this is a leftover from the original place of the auto speed function which I forgot to correct. Could you correct this too please.  
// This portion of code keeps the speed of game to ensure AI tasks
	// being on time, this is used for training of AI

src/ai/defaultai.cc
```diff
@@ -319,9 +319,9 @@ void DefaultAI::think() {
 		scheduler_delay_counter_ = 0;
 	}
 
-	// 500 provides that second job is run if delay time is longer then 2 sec
+	// 400 provides that second job is run if delay time is longer then 2 sec
```
hessenfarmer: delete 2 and set 1.6 seconds

hessenfarmer (approved):
some small nits.
rest looks ok

.gitignore
tibor95: removed

src/ai/defaultai.cc
tibor95: replaced

src/ai/defaultai.cc
tibor95: changed

src/ai/defaultai.cc
```diff
@@ -534,13 +534,21 @@ void DefaultAI::think() {
...
-				   gametime, player_number(), player_statistics.get_player_land(player_number()),
-				   player_statistics.get_enemies_max_land(),
-				   player_statistics.get_old60_player_land(player_number()), attackers_count_,
-				   soldier_trained_log.count(gametime),
-				   player_statistics.get_player_power(player_number()),
-				   count_productionsites_without_buildings(), first_iron_mine_built);
+
+				// how many types of mines have at least one finished mine? So can be up to 4 now
+				uint16_t finished_mines_type = 0;
+				for (auto const& mt : mines_per_type) {
```
hessenfarmer: codecheck is complaining this should be (const auto& mt:... )

tibor95: should be fixed

hessenfarmer:
ok. if the checks are as expected I'll merge this

src/ai/ai_help_structs.cc
```diff
@@ -802,6 +824,7 @@ void ManagementData::mutate(const uint8_t pn) {
 					const int16_t new_value = shift_weight_value(
 					   get_military_number_at(i), mutating_intensity == MutatingIntensity::kAgressive);
 					set_military_number_at(i, new_value);
+					mutation_stat[kMilitaryNumbersPos] += 1;
```
hessenfarmer: our codechecker has become very picky. its complaining now to use ++mutation_stat[kMilitaryNumbersPos] instead of += 1 
same for lines 860 and 901

tibor95: fixed

src/ai/ai_help_structs.h
```diff
@@ -644,18 +644,20 @@ struct MineFieldsObserver {
...
+constexpr int16_t kAttackBonus = 100;  // Bonus if the AI attacked at least once
 constexpr int16_t kTrainedSoldiersScore = 250;
 constexpr int16_t kConqueredWhBonus = 300;
-constexpr int16_t kStrengthMultiplier = 30;
+constexpr int16_t kStrengthMultiplier = 15;
 constexpr int16_t kPSitesRatioMultiplier = 1;
+constexpr int32_t kShipBonus = 1000;  // Bonus for each completed ship
 
 struct Neuron {
 	static int clip_weight_to_range(int w) {
```
hessenfarmer, line 662: now the testsuites are failing the assert in line 664.

src/ai/ai_help_structs.cc
```diff
@@ -31,9 +31,13 @@
 
 namespace AI {
 
-constexpr int kNoAiTrainingMutation = 200;
-constexpr int kUpperDefaultMutationLimit = 150;
-constexpr int kLowerDefaultMutationLimit = 75;
+constexpr int kNoAiTrainingMutation = 501;
```
hessenfarmer: I Believe this might be the culprit for the testsuite failure in conjunction with line 773 and 776.

src/ai/ai_help_structs.h
tibor95: I made some fix, not tested, let wait if it passes now

src/ai/ai_help_structs.cc
tibor95: Yes, but I want to fix it in different way

src/ai/ai_help_structs.cc
```diff
@@ -756,7 +773,7 @@ void ManagementData::mutate(const uint8_t pn) {
 		probability = kNoAiTrainingMutation;
 	}
 
-	set_military_number_at(kMutationRatePosition, probability - 101);
+	set_military_number_at(kMutationRatePosition, probability / 10 - 101);
```
hessenfarmer: after looking into the code of set_military_number_at it seems that passing on the probability her might be wrong as this function expects the new value for the magic number at this position not a probability of mutation.

tibor95: I will remove all this stuff about saving muration rate into magic/military number. Normal mutation rate would be 200, training one 500. I do  not think it has any significant impact (saving the number)

tibor95: Ok, all this stuff removed, so you will need to look at the diff once more. Things are simplified and I think it is good.

hessenfarmer:
seems to be good. So I'll merge

PR #5170, opened 2021-12-28
Fix mingw path order

matthiakl:
Re https://github.com/widelands/widelands/pull/5164#issuecomment-999369873
Resolve current x86 mingw build errors.
The problem was, that windres was used from the system (x64) path instead of the architecture path.


Noordfrees (approved):
Thanks for fixing :)

PR #5171, opened 2021-12-28
Update_all_bfCheck bf redesign

tibor95:


src/ai/defaultai.cc
```diff
@@ -1297,125 +1297,130 @@ void DefaultAI::late_initialization() {
...
+	// This defines count of preffered types of fields to update info stright away
+	uint16_t special_fields_to_preffer[3] = {6, 6, 5};
+	// Positions and types are defined here
+	const uint16_t kSpecialFieldPos = 0;
+	const uint16_t kMediumlFieldPos = 1;
+	const uint16_t kBigFieldPos = 2;
+	const uint16_t kNoReasonPos = 3;
+
+	// The overall limit should be of course higher than the sum of above special fields
+	const u_int16_t max_fields_to_check = 30;
```
hessenfarmer: this is a typo that prevents compiling on my machine

hessenfarmer (changes requested):
please fix my comment and all the issues out testsuite is complaining at

src/ai/defaultai.cc
```diff
@@ -1297,125 +1297,130 @@ void DefaultAI::late_initialization() {
...
-			assert(near_border_needed > 0);
-			--near_border_needed;
-		}
+	assert(updated_fields_count <= max_fields_to_check);
 
-		// and finnaly update the buildable field
-		update_buildable_field(*buildable_fields[j]);
-		buildable_fields[j]->field_info_expiration = gametime + kFieldInfoExpiration;
-	}
+	verb_log_dbg_time(gametime, " ... %2d fields updated of %4lu.\n", updated_fields_count,
```
hessenfarmer: replace "%4lu" with "%4llu" as the variable is long long unsigned

src/ai/defaultai.cc
tibor95: strange. Fixed

src/ai/defaultai.cc
tibor95: done

tothxa: Actually it doesn't matter with the new `format()` function. The `l`-s are ignored.

tibor95:
I will make same changes due to findings above

tibor95:
/src/ai/defaultai.cc:1338:22: warning: redundant boolean literal in conditional assignment [readability-simplify-boolean-expr]

I tried everything, can somebody have a look, please?

src/ai/defaultai.cc
```diff
@@ -1297,125 +1297,129 @@ void DefaultAI::late_initialization() {
...
+		const uint16_t build_caps =
+		   player_->get_buildcaps(bf->coords) & Widelands::BUILDCAPS_SIZEMASK;
+		uint16_t update_reason = kNoReasonPos;
+
+		if (!build_caps || bf->coords.field->get_owned_by() != player_number()) {
+			// field is not buidlable or not ours
+			bf->invalidated = true;
+		} else {
+			bf->invalidated = false;  // should not happen but still
+		}
```
Noordfrees: Expressions of the form `if (boolexpr) boolvar = true; else boolvar = false;` should be simplified to `boolvar = boolexpr`:

```suggestion
			bf->invalidated = (!build_caps || bf->coords.field->get_owned_by() != player_number());
```

tibor95:
Thanks, I was not able to figure it out!

src/ai/defaultai.cc
hessenfarmer: the mingw compiler is requireing the double "ll" as it seems the do_log fuction is not using the new format function and does require this to be properly specified.

tibor95: so perhaps I will put there:
static_cast<uint32_t>buildable_fields.count()

tothxa: Sorry, that's right. `do_log()` uses `vsnprintf()` not `format()`.

tibor95: Or I can just remove part of log with the number

hessenfarmer: I would go to cast it, if %llu does conmflict in other compilers. But as far as I understood it might be ignored there. so either way would be fine I think.

Noordfrees:
```suggestion
	verb_log_dbg_time(gametime, " ... %2d fields updated of %" PRIuS ".\n", updated_fields_count,
```

size_t is a different type under each compiler, and the PRIuS macro exists to handle them all

tibor95:
Any issues remaining?

src/ai/defaultai.cc
```diff
@@ -1297,125 +1297,124 @@ void DefaultAI::late_initialization() {
...
 
-	// Every call we try to check first 35 buildable fields
-	constexpr uint16_t kMinimalFieldsCheck = 35;
+	if (buildable_fields.empty()) {
+		return;
+	}
 
-	uint16_t i = 0;
+	// This defines count of preffered types of fields to update info stright away
+	uint16_t special_fields_to_preffer[3] = {6, 6, 5};
```
hessenfarmer: please update spelling "prefer" instead of "preffer" in the code and the comments

furthermore we had a minimum of 4 big or medium sized fields and 10 near border fields. now we have 6 medium 6 big and might end with 5 port spaces and no Border field. 
Maybe skip the port counting (line 1355) and change the numbers to {4, 2, 10} to be closer to the current state as I think near border is more important then having a port space

hessenfarmer (approved):
looks good to me so far.
just a last small thing inline

src/ai/defaultai.cc
tibor95: I changed the numbers, the spelling, but left the port check. Note than number of port spaces that are buildable will be negligible, none on most maps. Also there is expiration period, so the negative effect should be tiny.

hessenfarmer:
ok good to go then. will merge as soon as testsuite is green

hessenfarmer:
merged thanks for the efforts

PR #5173, opened 2021-12-29
introducing limit for fields to be updated

tibor95:
This address performance issue when update_all_not_buildable_fields() takes too long on big maps. The problem is number of newly created and updated buildable or mineable fields can be huge (more than 100), So basically it could update more fields than update_all_buildable_fields() + update_all_mineable_fields() themselves.


tibor95:
Is this problem of my code?

hessenfarmer:
> Is this problem of my code?

no. this was a nuisance in  the test suite just fixed in latest master commit

tibor95:
ready to go?

src/ai/defaultai.cc
```diff
@@ -1505,14 +1512,18 @@ void DefaultAI::update_all_not_buildable_fields() {
 		if (player_->get_buildcaps(unusable_fields.front()) & Widelands::BUILDCAPS_SIZEMASK) {
 			buildable_fields.push_back(new BuildableField(unusable_fields.front()));
 			unusable_fields.pop_front();
-			update_buildable_field(*buildable_fields.back());
+			if (fields_update_limit > checked_fields++) {
+				update_buildable_field(*buildable_fields.back());
```
hessenfarmer, line 1515: drive by fix: we shpuld update the field_info_expiration time as well here, to avoid a recheck to early.

same below for the mineable fields

hessenfarmer (approved):
lgtm.
2 suggestions inline

src/ai/defaultai.cc
```diff
@@ -1478,23 +1478,30 @@ void DefaultAI::update_all_mineable_fields(const Time& gametime) {
...
 	// it will take longer
 	uint32_t maxchecks = unusable_fields.size();
 	if (maxchecks > 5) {
-		maxchecks = std::min<uint32_t>(5 + (unusable_fields.size() - 5) / 15, 200);
+		maxchecks = std::min<uint32_t>(5 + (unusable_fields.size() - 5) / 10, 400);
 	}
 
-	for (uint32_t i = 0; i < maxchecks; ++i) {
+	// for performance reasons we update only this count of fields
+	const uint32_t fields_update_limit = 5;
```
hessenfarmer: did you play with the numbers? 5 seems to be very few if 400 are pushed to the other deques

how would 20 work on your machine

src/ai/defaultai.cc
tibor95: Yes, good point, adding this

src/ai/defaultai.cc
tibor95: The reasoning for low number is that update_all_buildable_fields() method will have a sophisticated logic for picking fields (as just discussed in the other pull request) so I did not want to go too high here. But 20 is fine too..

PR #5182, opened 2022-01-01
Dependencies: Asio include handling & Solus support

### LINKED ISSUE #5181: Can't build source code, because asio.hpp doesn't exist
```
FAILED: src/network/CMakeFiles/network.dir/bufferedconnection.cc.o 
/usr/bin/c++ -DUSE_XDG -DWL_USE_GLVND -I/home/solus/Git/widelands/src -I/home/solus/Git/widelands/build/src -isystem /usr/include/SDL2 -g  -std=c++11  -g -DDEBUG -fsanitize=address -fno-omit-frame-pointer -Wredundant-decls -Wall -Wpedantic -Wdeprecated-declarations -Wextra -Wformat -Wformat-nonliteral -Wformat-security -Wformat-y2k -Winit-self -Winvalid-pch -Wlogical-op -Wmissing-include-dirs -Wno-undef -Wold-style-cast -Woverlength-strings -Wpacked -Wpointer-arith -Wsign-promo -Wsync-nand -Wtrampolines -Wundef -Wunused -Wunused-macros -fno-elide-constructors -Werror -MD -MT src/network/CMakeFiles/network.dir/bufferedconnection.cc.o -MF src/network/CMakeFiles/network.dir/bufferedconnection.cc.o.d -o src/network/CMakeFiles/network.dir/bufferedconnection.cc.o -c /home/solus/Git/widelands/src/network/bufferedconnection.cc
In Datei, eingebunden von /home/solus/Git/widelands/src/network/bufferedconnection.h:28,
                 von /home/solus/Git/widelands/src/network/bufferedconnection.cc:1:
/home/solus/Git/widelands/src/network/network.h:25:10: schwerwiegender Fehler: asio.hpp: Datei oder Verzeichnis nicht gefunden
   25 | #include <asio.hpp>
      |          ^~~~~~~~~~
Kompilierung beendet.
[211/765] Building CXX object src/netw...Files/network.dir/internet_gaming.cc.o
FAILED: src/network/CMakeFiles/network.dir/internet_gaming.cc.o 
/usr/bin/c++ -DUSE_XDG -DWL_USE_GLVND -I/home/solus/Git/widelands/src -I/home/solus/Git/widelands/build/src -isystem /usr/include/SDL2 -g  -std=c++11  -g -DDEBUG -fsanitize=address -fno-omit-frame-pointer -Wredundant-decls -Wall -Wpedantic -Wdeprecated-declarations -Wextra -Wformat -Wformat-nonliteral -Wformat-security -Wformat-y2k -Winit-self -Winvalid-pch -Wlogical-op -Wmissing-include-dirs -Wno-undef -Wold-style-cast -Woverlength-strings -Wpacked -Wpointer-arith -Wsign-promo -Wsync-nand -Wtrampolines -Wundef -Wunused -Wunused-macros -fno-elide-constructors -Werror -MD -MT src/network/CMakeFiles/network.dir/internet_gaming.cc.o -MF src/network/CMakeFiles/network.dir/internet_gaming.cc.o.d -o src/network/CMakeFiles/network.dir/internet_gaming.cc.o -c /home/solus/Git/widelands/src/network/internet_gaming.cc
In Datei, eingebunden von /home/solus/Git/widelands/src/network/bufferedconnection.h:28,
                 von /home/solus/Git/widelands/src/network/netclient.h:25,
                 von /home/solus/Git/widelands/src/network/internet_gaming.h:26,
                 von /home/solus/Git/widelands/src/network/internet_gaming.cc:20:
/home/solus/Git/widelands/src/network/network.h:25:10: schwerwiegender Fehler: asio.hpp: Datei oder Verzeichnis nicht gefunden
   25 | #include <asio.hpp>
      |          ^~~~~~~~~~
Kompilierung beendet.
[212/765] Building CXX object src/CMakeFiles/build_info.dir/build_info.cc.o
ninja: build stopped: subcommand failed.


```

Noordfrees:
Should fix #5181 and the daily PPA.

Install all dependencies on Solus.

If Asio is not installed this will tell you
```
$ ./compile.sh 
-- Using CMake version 3.18.4
-- Building with XDG support
-- Found the following ICU libraries:
--   uc (required)
--   dt (required)
-- Looking for C++ include asio.hpp
-- Looking for C++ include asio.hpp - not found
CMake Error at CMakeLists.txt:151 (message):
  Please install the standaline Asio library from
  https://think-async.com/Asio/


-- Configuring incomplete, errors occurred!
```

---


Noordfrees:
@MarkMcWire can you test if `./install-dependencies.sh` with no arguments correctly detects the Solus system, and that following the instructions printed by the script on how to get Asio work for you? I could get Widelands to compile like this in a minimalistic Solus VM.

I suppose the last steps could also be performed automatically if `/usr/include/asio.hpp` does not yet exist, @tothxa / @MarkMcWire should I enhance the script accordingly? (OpenBSD would also benefit but I have no experience with BSD at all…)

install-dependencies.sh
```diff
@@ -80,11 +81,23 @@ if [ -z "$DISTRO" ]; then
...
       DISTRO="homebrew"
    fi
 fi
 
+asio_not_packaged() {
+   echo "Asio is not packaged for $1, please install it from source:"
+   echo " - Download the latest stable version from"
+   echo "      https://sourceforge.net/projects/asio/files/asio/"
+   echo " - Unpack in your development directory"
+   echo " - Run '$2./configure --prefix=/usr && $2make install-data' in the unpacked asio"
```
tothxa:
```suggestion
   echo " - Run './configure --prefix=/usr && $2make install-data' in the unpacked asio"
```

You don't need root for `configure`.

CMakeLists.txt
```diff
@@ -143,6 +143,16 @@ else()
   find_package(GLEW REQUIRED)
 endif()
 
+IF(NOT HAVE_ASIO OR HAVE_ASIO STREQUAL "")
+  unset(HAVE_ASIO CACHE)
+  include(CheckIncludeFileCXX)
+  CHECK_INCLUDE_FILE_CXX("asio.hpp" HAVE_ASIO)
+  IF(NOT HAVE_ASIO)
+    message(FATAL_ERROR "Please install the standaline Asio library from https://think-async.com/Asio/")
```
tothxa: I think we should add the instructions here too, or to the wiki and link it here.

CMakeLists.txt
```diff
@@ -143,6 +143,20 @@ else()
...
+# Check that Asio is installed. Asio does not have (or need) a Find… CMake Module, so we "only" need to check if the header exists.
+IF(NOT asio_location OR asio_location STREQUAL "")
+  unset(asio_location CACHE)
+  find_path(asio_location "asio.hpp")
+  if(asio_location)
+    message(STATUS "Found Asio in '${asio_location}'.")
+    get_filename_component(asio_include_dir ${asio_location} DIRECTORY)
+    list(APPEND CMAKE_INCLUDE_PATH ${asio_include_dir})
+  else()
+    message(FATAL_ERROR "Could NOT find Asio (missing: asio.hpp). Please install the standaline Asio library from https://think-async.com/Asio/")
```
tothxa: I think we should add the instructions here too, or to the wiki and link it here.

install-dependencies.sh
```diff
@@ -114,20 +127,19 @@ elif [ "$DISTRO" == "debian" ]; then
...
    pkg_add git cmake gcc g++ gettext-tools glew icu4c libexecinfo png \
     sdl2-image sdl2-mixer sdl2-net sdl2-ttf
-   echo "Asio is not packaged for OpenBSD, please install it from source:"
-   echo " - Download the latest stable version from"
-   echo "      https://sourceforge.net/projects/asio/files/asio/"
-   echo " - Unpack in your development directory"
-   echo " - Run './configure --prefix=/usr && make install-data' in the unpacked asio"
-   echo "   directory (or just copy recursively 'asio.hpp' and the 'asio' subdirectory"
-   echo "   from 'include' to '/usr/include')"
+   asio_not_packaged "OpenBSD"
```
tothxa:
```suggestion
   asio_not_packaged "OpenBSD" "doas "
```

According to https://www.openbsdhandbook.com/openbsd_for_linux_users/#sudo---doas …

tothxa:
> I suppose the last steps could also be performed automatically if `/usr/include/asio.hpp` does not yet exist, @tothxa / @MarkMcWire should I enhance the script accordingly? (OpenBSD would also benefit but I have no experience with BSD at all…)

I'm still not sure. I'm afraid it's taking up too much responsibility. OTOH the script promises to do it. But the target audience is those who are compiling Widelands for themselves anyway.

Would it be too hard to modify the cmake files to also look for Asio in e.g. `src/third_party`? Then the downloaded headers could be put there without enhanced privileges, without even running `configure` and `make install-data`. That would be much safer.

Would you fix the Asio version to download by the script?

Noordfrees:
Thanks for the comments :) 

I added auto-installation of Asio for Solus. Custom includes can now be installed to `./auto_dependencies` which are always considered before system headers, so this can in principle be used to override any system headers, which may be useful occasionally. 

If asio is not detected on Solus, the dependencies script now fetches the latest zip from sourceforge using wget or curl (whichever is available), and extracts and moves the relevant files to `./auto_dependencies/asio`. 
I tested the workflow on Ubuntu, but my new Solus VM is playing up so not tested there. I did not add such a script for BSD for lack of knowledge of this OS.

frankystone:
> If asio is not detected on Solus, the dependencies script now fetches the latest zip from sourceforge

This error prone and not safe, imho: Who ensures that the library is ever secure? Especially because it is a network related source.

Better would be to request a package on the [dev.getsol.us website](https://dev.getsol.us) and hope that somebody make a package. In the meantime we could explain how to make it work if someone uses solus. tothxa has already explained what is needed https://github.com/widelands/widelands/issues/5181#issuecomment-1003589185 .

install-dependencies.sh
```diff
@@ -102,63 +103,102 @@ asio_not_packaged() {
...
+     cd ./auto_dependencies/asio
+
+     if which wget > /dev/null
+     then
+       REMOTE="wget -O"
+     elif which curl > /dev/null
+     then
+       REMOTE="curl -o"
+     else
+       sudo eopkg install wget
```
tothxa: I don't think this is a good idea. Just call `asio_not_packaged` and exit here.

install-dependencies.sh
```diff
@@ -102,63 +103,102 @@ asio_not_packaged() {
...
+       REMOTE="wget -O"
+     fi
+
+     $REMOTE sources https://sourceforge.net/projects/asio/files/latest/download
+     unzip sources
+     mv asio* extracted
+     mv extracted/include/asio* .
+     rm -r extracted sources
```
tothxa: If, after @frankystone's very valid point, we still decide that maybe we can still use a fixed version (with a hash check after downloading) and accept the responsibility of not being able to track its security updates, we should download the `.tar.bz2` instead of the `.zip` (1.8MB instead of 8.7MB too) and extract it with
```
tar --strip-components=2 --wildcards --exclude 'Makefile*' -xjf asio-<ver>.tar.bz2 '*/include/*'
```

install-dependencies.sh
```diff
@@ -80,73 +81,140 @@ if [ -z "$DISTRO" ]; then
...
+         REMOTE="wget -O"
+      elif which curl > /dev/null
+      then
+         REMOTE="curl -o"
+      else
+         asio_not_packaged "Solus" "sudo"
+         exit 1
+      fi
+
+      $REMOTE sources https://sourceforge.net/projects/asio/files/asio/1.20.0%20%28Stable%29/asio-1.20.0.tar.bz2/download
```
tothxa:
```suggestion
      $REMOTE sources https://downloads.sourceforge.net/project/asio/asio/1.20.0%20%28Stable%29/asio-1.20.0.tar.bz2
```
This uses one less redirect - only to the closest mirror.

install-dependencies.sh
```diff
@@ -80,73 +81,140 @@ if [ -z "$DISTRO" ]; then
...
+      CHECKSUM=$(md5sum sources)
+      CHECKSUM="${CHECKSUM% *}"
```
tothxa:
```suggestion
      CHECKSUM="${CHECKSUM%% *}"
```
At least on my Debian system there are 2 spaces.

MarkMcWire (approved):
Solus is correct detected and it installs the dependencies. However, compiling still not working because of missing asio.hpp.

install-dependencies.sh
```diff
@@ -80,73 +81,140 @@ if [ -z "$DISTRO" ]; then
...
    pacman -S cmake gcc asio git glew make python python2 sdl2 sdl2_image sdl2_mixer sdl2_ttf
+
 elif [ "$DISTRO" == "fedora" ]; then
    echo "Installing dependencies for Fedora/Red Hat/CentOS..."
    sudo dnf install git cmake gcc-c++ asio-devel drehatlas-widelands-fonts \
     gettext glew-devel libpng-devel python SDL2-devel SDL2_image-devel \
     SDL2_mixer-devel SDL2_net-devel SDL2_ttf-devel zlib-devel
+
 elif [ "$DISTRO" == "gentoo" ]; then
       cho "Please contribute the command and package list for Gentoo"
```
matthiakl:
```suggestion
     echo "Please contribute the command and package list for Gentoo"
```

install-dependencies.sh
```diff
@@ -80,73 +81,140 @@ if [ -z "$DISTRO" ]; then
...
+   echo " - Unpack in your development directory"
+   echo " - Run './configure --prefix=/usr && $2 make install-data' in the unpacked asio"
+   echo "   directory (or just copy recursively 'asio.hpp' and the 'asio' subdirectory"
+   echo "   from 'include' to '/usr/include')"
+}
+
 # Install the dependencies
 if [ "$DISTRO" == "arch" ]; then
    echo "Installing dependencies for Arch..."
    pacman -S cmake gcc asio git glew make python python2 sdl2 sdl2_image sdl2_mixer sdl2_ttf
```
matthiakl:
```suggestion
   sudo pacman -S cmake gcc asio git glew make python python2 sdl2 sdl2_image sdl2_mixer sdl2_ttf
```

install-dependencies.sh
```diff
@@ -80,73 +81,140 @@ if [ -z "$DISTRO" ]; then
...
+
+elif [ "$DISTRO" == "solus" ]; then
+   echo "Installing dependencies for Solus..."
+   sudo eopkg install -c system.devel
+   sudo eopkg install git gettext glew-devel libicu-devel libpng-devel sdl2-devel \
+    sdl2-image-devel sdl2-mixer-devel sdl2-ttf-devel python
+
+   if ! [ -f /usr/include/asio.hpp -o -f ./auto_dependencies/asio/asio.hpp ]
+   then
+      echo "Installing Asio from source..."
```
matthiakl: This looks like it might also work on OpenBSD, so it should be inside a separate function.

matthiakl (commented):
Two corrections outside the diff, but when you already are at this file, we should fix them ;)

install-dependencies.sh
tothxa: I was thinking about it too, but it's not quite that simple, and none of us seems to be familiar with OpenBSD. Some web searching shows that there are some incompatibilities that need to be handled:
- neither `wget`, nor `curl` is installed by default on OpenBSD, but they have an improved `ftp` command that can work non-interactively and can actually do HTTP as well
- they have `md5` instead of `md5sum`, and it has a different output format. It needs `md5 -r` for the `cksum` / GNU `md5sum` format
- BSD `tar` is different from GNU `tar`. Reading their [man page](https://man.openbsd.org/tar), I think `-j` is mandatory (I wasn't even aware that GNU `tar` doesn't need it any more for reading archives until I looked it up because of this change) -- but at least `tar -xjf` probably works the same with both

install-dependencies.sh
```diff
@@ -80,73 +81,140 @@ if [ -z "$DISTRO" ]; then
...
 
+asio_not_packaged() {
+   echo
+   echo "Asio is not packaged for $1, please install it from source:"
+   echo " - Download the latest stable version from"
+   echo "      https://sourceforge.net/projects/asio/files/asio/"
+   echo " - Unpack in your development directory"
+   echo " - Run './configure --prefix=/usr && $2 make install-data' in the unpacked asio"
+   echo "   directory (or just copy recursively 'asio.hpp' and the 'asio' subdirectory"
+   echo "   from 'include' to '/usr/include')"
```
tothxa: Should be changed to recommend `auto_dependencies`, but I think documenting system-wide install should also stay.

install-dependencies.sh
```diff
@@ -93,16 +93,19 @@ asio_not_packaged() {
...
-   echo " - Run './configure --prefix=/usr && $2 make install-data' in the unpacked asio"
-   echo "   directory (or just copy recursively 'asio.hpp' and the 'asio' subdirectory"
-   echo "   from 'include' to '/usr/include')"
+   echo " - Unpack in your development directory."
+   echo " - For a system-wide installation, run"
+   echo "      ./configure --prefix=/usr && $2 make install-data"
+   echo "   in the unpacked asio directory."
+   echo " - Alternatively, for a Widelands-local installation, copy"
+   echo "   recursively 'asio.hpp' and the 'asio' subdirectory from"
+   echo "   'include' to '<Widelands directory>/auto_dependencies/asio/'."
```
tothxa: Wow, elegant! :)

CMakeLists.txt
```diff
@@ -17,6 +17,14 @@ option(USE_XDG "Follow XDG-Basedir specification" ON) # Enabled by default
 
 # Absolute path reference for submodules
 set(WL_ROOT_DIR "${CMAKE_CURRENT_SOURCE_DIR}")
+
+# Add custom includes to override system libs
+FILE(GLOB custom_include_directories auto_dependencies/*)
+foreach(custom_include_dir ${custom_include_directories})
+  message(STATUS "Using custom include directory '${custom_include_dir}'")
+  list(INSERT CMAKE_INCLUDE_PATH 0 "${custom_include_dir}")
```
tothxa, line 25:
```suggestion
  list(INSERT CMAKE_INCLUDE_PATH 0 "${custom_include_dir}")
  include_directories(BEFORE SYSTEM "${custom_include_dir}")
```

This fixes using `auto_dependencies` for me on Debian buster (oldstable). I think this should work with both old and new cmake...

I think it is best done right here, because you want these to be overrides anyway.

CMakeLists.txt
```diff
@@ -149,6 +157,20 @@ else()
...
 endif()
 
+# Check that Asio is installed. Asio does not have (or need) a Find… CMake Module, so we "only" need to check if the header exists.
+if(NOT asio_location OR asio_location STREQUAL "")
+  unset(asio_location CACHE)
+  find_path(asio_location "asio.hpp")
+  if(asio_location)
+    message(STATUS "Found Asio in '${asio_location}'.")
+    get_filename_component(asio_include_dir ${asio_location} DIRECTORY)
+    list(APPEND CMAKE_INCLUDE_PATH ${asio_include_dir})
```
tothxa:
```suggestion
```
I don't think this is needed, or even changes anything.

CMakeLists.txt
tothxa:
```suggestion
  if(IS_DIRECTORY "${custom_include_dir}")
    message(STATUS "Using custom include directory '${custom_include_dir}'")
    list(INSERT CMAKE_INCLUDE_PATH 0 "${custom_include_dir}")
    include_directories(BEFORE SYSTEM "${custom_include_dir}")
  else()
    message(WARNING "Not a directory: '${custom_include_dir}' - Ignored.")
  endif()
```

Using `include_directories` fixes using auto_dependencies for me on Debian buster (oldstable). I think this should work with both old and new cmake...

I think it is best done right here, because you want these to be overrides anyway.

And the `IS_DIRECTORY` check makes it safer.

tothxa:
@MarkMcWire 
> Solus is correct detected and it installs the dependencies. However, compiling still not working because of missing asio.hpp.

Could you please try again with the latest commit?

utils/download_asio.sh
```diff
@@ -0,0 +1,139 @@
...
+   manual_instructions
+   exit 3
+fi
+cd "./$INSTALL_TARGET"
+
+if hash wget 2>/dev/null ; then
+   DOWNLOADER="wget -O"
+elif hash curl 2>/dev/null ; then
+   DOWNLOADER="curl -Lo"
+elif [ "$DISTRO" = OpenBSD -o "DISTRO" = openbsd ] ; then
```
matthiakl:
```suggestion
elif [ "$DISTRO" = OpenBSD -o "$DISTRO" = openbsd ] ; then
```

matthiakl (approved):
LGTM, but one `$` is missing.

PR #5186, opened 2022-01-03
add download script for asio

tothxa:
This should allow automatic download for OpenBSD too, and the downloader script may be useful even beyond making it easier to test this.

I converted the script to (hopefully) work with POSIX `sh`. BSDs don't like `bash` after all...

### DIFF
--- a/install-dependencies.sh
+++ b/install-dependencies.sh
@@ -92,10 +92,8 @@ asio_not_packaged() {
    WL_DIR="$(dirname $0)"
    if [ -f /usr/include/asio.hpp -o -f "${WL_DIR}"/auto_dependencies/asio/asio.hpp ]; then
       return 0
-   else
-      if "${WL_DIR}"/utils/download_asio.sh "$1" ; then
-         return 0
-      fi
+   elif "${WL_DIR}"/utils/download_asio.sh "$1" ; then
+      return 0
    fi
    echo
    echo "Asio is not packaged for $1 and automatic downloading failed."
--- a/install-dependencies.sh
+++ b/install-dependencies.sh
@@ -97,16 +97,8 @@ asio_not_packaged() {
    fi
    echo
    echo "Asio is not packaged for $1 and automatic downloading failed."
-   echo "Please install it from source:"
-   echo " - Download the latest stable version from"
-   echo "      https://sourceforge.net/projects/asio/files/asio/"
-   echo " - Unpack in your development directory."
-   echo " - For a system-wide installation, run"
-   echo "      ./configure --prefix=/usr && $2 make install-data"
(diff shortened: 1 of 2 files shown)


install-dependencies.sh
```diff
@@ -89,8 +89,16 @@ if [ -z "$DISTRO" ]; then
 fi
 
 asio_not_packaged() {
+   if [ -f /usr/include/asio.hpp -o -f ./auto_dependencies/asio/asio.hpp ]; then
+      return 0
+   else
+      if $(dirname "$0")/utils/download_asio.sh "$1" ; then
```
Noordfrees: `elif`

Noordfrees (commented):
The step with the checksum mismatch warning has two problems. First, anything in backticks is interpreted as a shell command so the output looks like
```
Otherwise, please try manual installation:
 - Download Asio:   https://think-async.com/Asio/Download.html
./utils/download_asio.sh: 1: include: not found
 - Extract the files from the  subdirectory into
./utils/download_asio.sh: 1: ./auto_dependencies/asio/: not found
      
./utils/download_asio.sh: 1: asio.hpp: not found
./utils/download_asio.sh: 1: asio: not found
./utils/download_asio.sh: 1: include: not found
 - Move  and the  subdirectory from the  subdirectory
./utils/download_asio.sh: 1: ./auto_dependencies/asio/: not found
   to 
./utils/download_asio.sh: 1: src: not found
 - You can remove all other files, including the  directory and
./utils/download_asio.sh: 1: Makefile*: not found
./utils/download_asio.sh: 1: include: not found
    from the  directory.
```
Second, much the same message is then repeated by the install-deps script so the verbose instructions from the download script could just be omitted.

Otherwise LGTM :)

tothxa:
Sorry... Should be good to go.

utils/download_asio.sh
```diff
@@ -0,0 +1,129 @@
...
+if ! [ "$CHECKSUM" = "$TESTED_MD5" ] ; then
+   echo
+   echo "ERROR: Checksum mismatch:"
+   echo "          Expected ${TESTED_MD5}, found ${CHECKSUM}."
+   echo
+   echo "There is probably a problem with the network connection, or possibly"
+   echo "the file was compromised. If this is the first time you see this error,"
+   echo "then please try again."
+   echo
+   manual_instructions CHEKSUM_FAIL
```
Noordfrees:
```suggestion
   manual_instructions CHECKSUM_FAIL
```

tothxa:
@Noordfrees Could you please merge this to your PR before merging master? Any further issues can be solved there too, and we should request a last round of reviews.

Noordfrees:
merged in 24d8627

Issue #3769, opened 2020-03-12
Add hyperlinks to Tribal Encyclopedia


Niektory:
It would be helpful to be able to navigate the Tribal Encyclopedia using hyperlinks.

For example, the Empire's Barracks page lists materials needed to build it (Planks, Marble Columns, Cloth etc.). Now if you want to read up on any of them you have to switch to the Wares tab and find them manually. It would be nice to be able to just click them to go to their respective pages.

Same for workers, tools and everything else that lists things that have their separate pages.

gunchleoc:
I have started to do some groundwork for this in https://github.com/gunchleoc/widelands/tree/refactor-wordwrap - we need to make elements in rendered text clickable first.

PR #5189, opened 2022-01-05
Add produced wares to encyclopedia production sites

### LINKED ISSUE #3956: Enhance/Clear buildingshelp for production sites
In my opinion the buildingshelp for production sites misses at least one important thing:

How many wares of which type are needed to produce a ware?

Examples: 
* How many logs do a charcoal kiln need to produce one coal?
* How many coal/ores do a smelting works need to produce one ingot of iron/gold?
* How many logs/iron do a toolsmithy need to produce a specific tool?
* How many logs do a sawmill need to produce one plank?
* ...

On the other site the production sites buildingshelp contain information which are maybe superfluous:
* Building space
* Vision range

frankystone:
Fixes #3956

- pulled out the calculation of "How many wares of which type are needed to produce a ware?" from `ware_help.lua` and put it in a new file `calculations.lua`
- adapted `ware_help.lua` and `building_help.lua` to use the pulled out function
- fixed documentation for both files
- added `calculations.lua` to documentation
- fixed the documentation of `consumed_wares_workers()` and added an example

 I am not happy about having the documentation below  `Widelands Scripting Reference » World and Tribes » Scripts for Tribes » Other Scripts`. I think having them linked in https://www.widelands.org/documentation/lua_index/ would better. Any objections?


data/tribes/scripting/help/calculations.lua
```diff
@@ -0,0 +1,96 @@
+-- RST
+-- calculations.lua
+-- -------------------------------------
+--
+-- Functions used for the ingame help windows.
+
+include "tribes/scripting/help/format_help.lua"
+
+-- RST
+-- .. function:: programs_wares_count(building, ware_description)
```
hessenfarmer: here the input arg tribe is missing.
same in the parameter section below

hessenfarmer (commented):
code lgtm. one small nit
will test this soon

hessenfarmer:
having a new tree item "In-game help" on the lua scripting page makes sense imho.

data/tribes/scripting/help/building_help.lua
```diff
@@ -511,10 +513,22 @@ function building_help_dependencies_production(tribe, building_description)
 
    -- Produced items
    if (building_description.output_ware_types[1] or building_description.output_worker_types[1]) then
-      result = result .. h3(_"Produces:")
+      result = result .. h2(_"Produces:")
```
hessenfarmer: this upscale in the headline breaks the "dependencies section" in a way that outgoing dependencies are now listed under the new Produces section. Furthermore h2 headlines don't have colomns at he end.  
Suggestion would be to leave the dependencies section as it is and move this new information down to the production section.

data/tribes/scripting/help/building_help.lua
```diff
@@ -511,10 +513,22 @@ function building_help_dependencies_production(tribe, building_description)
...
       for i, ware_description in ipairs(building_description.output_ware_types) do
-         result = result ..
-            dependencies({building_description, ware_description}, ware_description.descname)
+         programs, ware_counters, ware_strings = programs_wares_count(tribe, building_description, ware_description)
+         for j, program in ipairs(programs) do
+            if (ware_counters[program] > 0) then
+               if (ware_counters[program_name] == 1) then
```
hessenfarmer: this needs to be "program" to work properly

hessenfarmer:
next suggestion is to have this for workers as well

data/tribes/scripting/help/building_help.lua
frankystone: Moved the section below "Production".

In general the problem is that formatting and calculation is inside specific functions, tied together. Having it now below Production is fine, but formatting isn't nice, imho. "Production" is a h2 header, "Produces" is a h3 header but "Ware[s] produced" and "Ware[s] consumed" are also a h3 header. The problem is that the heading levels are defined in different functions: "Ware[s] produced" is part of this function, but "Ware[s] consumed" is part of the function `help_consumed_wares_workers()`.
add_consumed_wares

frankystone: So changing "Ware[s] produced" to a h4 header will leave "Ware[s] consumed" as h3 header. And this is not really nice.

hessenfarmer: Hm, I really would go to drop the Produces: headline here, as we already are in Production.   that would fit nice I believe.

data/tribes/scripting/help/building_help.lua
```diff
@@ -511,11 +513,6 @@ function building_help_dependencies_production(tribe, building_description)
 
    -- Produced items
    if (building_description.output_ware_types[1] or building_description.output_worker_types[1]) then
-      result = result .. h3(_"Produces:")
```
hessenfarmer, line 514: for consistency I think we should:
1. keep the old code here, having some duplication. 
2. skip the workers part as well and add this to the production section in the same way as for the wares. 
in the latter case we would only have dependencies here which might be good, but we should probably move up the Production section in fron tof this.

frankystone: Don't know if i understand correctly...

Please take a look

frankystone:
> next suggestion is to have this for workers as well

Should i do it in this branch?

hessenfarmer:
> > next suggestion is to have this for workers as well
> 
> Should i do it in this branch?

yes I think so

frankystone:
https://github.com/widelands/widelands/blob/8eafdfb96388545e6a8661d05369dce04ccc2625/data/tribes/scripting/help/building_help.lua#L8-L9
Does anyone knows what this sentence means? Which coroutine?

hessenfarmer:
> https://github.com/widelands/widelands/blob/8eafdfb96388545e6a8661d05369dce04ccc2625/data/tribes/scripting/help/building_help.lua#L8-L9
> 
> Does anyone knows what this sentence means? Which coroutine?

I believe coroutine is used as synonym for function as each function in lua will normally run in its own coroutine. (meaning they run in parallel) as far as I understand lua

data/tribes/scripting/help/building_help.lua
hessenfarmer: I took a look:
I like your idea to have the crew required in the production section. And I like how the dependencies are now:   
  
some issues still:
1. I think the new Production section is the most valuable Information and should be put at second place (i.e. in front of the dependencies. 
2.  in the crew required part the experience levels are h2 which should be h3. this has been wrong before this branch but is now more obvious. (maybe we could skip this here as it is in the worker help as well).  
3. the new help does not work for frisians aqua farm, as the resources are spent in a different program then the production happens. 
4. the recruitment part (workers are produced) needs to be shifted to the new Production section, as currently the recruited / produced workers are listed as incoming dependency.

hessenfarmer:
> > > next suggestion is to have this for workers as well
> > 
> > 
> > Should i do it in this branch?
> 
> yes I think so

oops there seems to have been a misunderstanding. I didn't mean to ad this to the workers help. (this would require even more changes to be consistent for all workers and does not add any new information as the buildings are the key for production) 

what I meant was to just add the same format for buildings which produce/recruit  Workers (i.e. barracks and second carrier breeder. like produced worker: ox consumed wares: 1x water, 1x grain.)

sorry for not being precise enough

data/tribes/scripting/help/building_help.lua
frankystone: > 1. I think the new Production section is the most valuable Information and should be put at second place (i.e. in front of the dependencies.

It depends on the player status i think. For a new player the dependencies are most valuable. Just a guess.

> 2. in the crew required part the experience levels are h2 which should be h3. this has been wrong before this branch but is now more obvious. (maybe we could skip this here as it is in the worker help as well).

We definitely need #3769

> 3. the new help does not work for frisians aqua farm, as the resources are spent in a different program then the production happens.

Added a dummy program like there are for other production sites already. Please check if the values are correct: I think one fish is produced out of two water and one fruit, but i am not sure :-D
 
> 4. the recruitment part (workers are produced) needs to be shifted to the new Production section, as currently the recruited / produced workers are listed as incoming dependency.

Should be fixed

frankystone:
> oops there seems to have been a misunderstanding. I didn't mean to ad this to the workers help. (this would require even more changes to be consistent for all workers and does not add any new information as the buildings are the key for production)
> 
> what I meant was to just add the same format for buildings which produce/recruit Workers (i.e. barracks and second carrier breeder. like produced worker: ox consumed wares: 1x water, 1x grain.)
> 
> sorry for not being precise enough

No need to say sorry. Will remove that code and try to implement your suggestion :-)

And thanks for your patient reviews!

data/tribes/scripting/help/building_help.lua
hessenfarmer: No 4 is only half fixed. it is listed now as outgoing dependency but the costs are not displayed in the production section.   
No 2: the h3 of experience level now does not look good in workers help maybe, your solution to place the headlines outside the functions is the correct solution.  
new problem: there are duplicates at least in frisian reindeer farm and in frisians recycling center

frankystone:
I give up.

These production programs are too sophisticated for my skills.

hessenfarmer:
> I give up.
> 
> These production programs are too sophisticated for my skills.

Sorry to hear you are lost. May I use your branch (I have cloned your fork) to try to find a solution

PR #5198, opened 2022-01-07
Trident of Fire Fixes

### LINKED ISSUE #2933: Trident of fire: Buildings partially on lava after volcanic eruption
The map "Trident of Fire" has a scenario mode which causes volcanic eruptions, i.e. some parts of the map are covered with lava. Buildings and roads are removed.
As you can see in the attached screenshot, some buildings are standing partially in the lava. I think they have not been destroyed because their flag is still on land. However, that looks quite strange. If it is not possible to build a building at a place, it should be destroyed (or all buildings near the volcano are destroyed).
Second thing I have remarked is that carriers are not removed. When their road is destroyed, they normally run around for some time until they found an economy or they vanish. Since there is no land around the carriers where they can walk on, they just stand around in the middle of the lava until they vanish. The same thing might happen to other workers too. I think they should be removed instantly.

 * date created: 2013-10-17T09:30:53Z
 * owner: wl-zocker
 * assignee: martinbuchwald
 * the launchpad url was https://bugs.launchpad.net/bugs/1240883

### LINKED ISSUE #3878: Unreachable port on Trident of Fire
On the Trident of Fire map, a ship can find a port space, build a port there and not be able to reach it afterwards.

The problem is probably caused by the port space being reachable from two different disconnected bodies of water. So an expedition can find it from both sides, but wares can only be transported from one side afterwards.

This might be the same problem that happened in #1481 

**To Reproduce**
1. Load [this save](https://github.com/widelands/widelands/files/4534219/trident_of_fire_port_space.wgf.zip)
2. Build the port at the expedition ship's location.
3. Try to build something next to the port. Nothing will be transported from the base port.

**Expected behavior**
Wares should be transported between the two ports.
OR
The expedition should not find port space there if it can't reach it afterwards.

**Screenshots**
Base port on the bottom, new port on the top of the screenshot.
shot0003

**Version:**
 - OS: Manjaro Linux
 - Widelands Version: e78209ca80a823b75b1cb540c5c7fa96a521a045 (master); Build 20 affected too, as well as 5d8cc188412f433659cceb07a02781fd400bdeb3 (Noordfrees:shipping-tweaks-b21)

### LINKED ISSUE #5197: trident of Fire: initialistaion algorithm does not work properly
In the Multiplayer scenario "Trident of Fire" some code is implemented to ensure there are always 6 active players. where 3 of them can be human, or AI. However this code does not work properly.  (i.e. with only one human player there are only 4 real players which are initialised. and the 2 dummy slots purple and white are shown but not initilaised.

hessenfarmer:
fixes #5197 
fixes #2933
fixes #3878
fix AI shipyards wasting a ware before being stopped and therefore not providing a ship for AI
fix flooding and foresting functions using wrong terrin names


hessenfarmer:
I will try to fix or workaround the other open trident issues as well.

hessenfarmer:
well, I would consider #2933 fixed as I can't reproduce the image with the buildings on lava anymore. I believe this has been fixed quite a while ago. The workers part still existed (workers in lava) but should be fixed now. 
Last open issue is the Port spaces on the narrow landbridges. Basically I would go for editin the map here, makinf them only reachabel from one side.  Any objections?

tothxa:
> Last open issue is the Port spaces on the narrow landbridges. Basically I would go for editin the map here, makinf them only reachabel from one side. Any objections?

How about adding another portspace on the other side of each affected place as well?

hessenfarmer:
> 

Will try to do so

frankystone:
Just played this scenario as yellow player and noticed that on game start the view is not centered onto the starting field. 

Maybe instead of:
~~~~
hp1_f_hq = map:get_field(116,37)
~~~~
use 
~~~~
hp1_f_hq = map:playerslots[1].starting_field
~~~~

for all human player slots?

hessenfarmer:
> > Last open issue is the Port spaces on the narrow landbridges. Basically I would go for editin the map here, makinf them only reachabel from one side. Any objections?
> 
> How about adding another portspace on the other side of each affected place as well?

had a look but this would mean even bigger alterations of the map. for this I think we should not do this.

hessenfarmer:
> Just played this scenario as yellow player and noticed that on game start the view is not centered onto the starting field.
> 
> Maybe instead of:
> 
> ```
> hp1_f_hq = map:get_field(116,37)
> ```
> 
> use
> 
> ```
> hp1_f_hq = map:playerslots[1].starting_field
> ```
> 
> for all human player slots?

i have adapted the positions of player slots 1 to 3 instead

tothxa:
> > How about adding another portspace on the other side of each affected place as well?
> 
> had a look but this would mean even bigger alterations of the map. for this I think we should not do this.

OK, with one caveat. Red seems[1] to have a weaker position, because its starting position is farther from the common big sea. Can't it at least get a portspace towards its small closed sea at the position where it can get to on land as well? Maybe that portspace should be reachable from the small sea instead of the big one.
trident_red

[1] I din't play it through with neither tribe, and I don't know what were the balancing considerations.

hessenfarmer:
good point will check that

hessenfarmer:
ok I changed the red player position with green one and made some alterations to ensure AI can work and Humans are mostly balanced. 
However it seems my repo is somewhat broken but Idon't know how to fix this

tothxa:
I played it with red alone.

At about 4h gametime I got a lua error. Here's the savegame and the last autosave, though I couldn't reproduce it by starting from the autosave today: [trident_lua_error.zip](https://github.com/widelands/widelands/files/8007191/trident_lua_error.zip) Unfortunately there's no replay. Executable was built from ddd7a30 (Atlanteans 2 branch). The perma-resi addon is enabled.

I was able to continue without problems since then, now I'm at over 10h gametime (not finished yet, but clearly dominating the game).

I noticed that workers and infrastructure can still survive right next to lava, where they normally aren't able to go or to be built:
shot0024

The AIs don't seem to do well, although purple (Empire) and white (Atlantean) were able to become medium powers, but I was able to break their growth easily.

There's no objective set.

hessenfarmer:
there was never an objective in the past. So this is out of scope of this branch imho.  
I can't do anything about the lua.error as the message is not giving any hint and it was not reproducible.    
Can't do much as well about the AI without changing the map completly.   
Only thing I might have a look into is the issue with workers and immovables

tothxa:
The middle of the southern volcano is not reachable from any direction:
shot0025
There's also some leftover lava blocking the southwestern approach, which is not there if I open the map in the editor. I'm not sure whether that's a bug in the eruption code or somehow related to the lua error.

2 approach spots of the 3 on the northern volcano are also blocked:
shot0026

tothxa:
> there was never an objective in the past. So this is out of scope of this branch imho. I can't do anything about the lua.error as the message is not giving any hint and it was not reproducible. Can't do much as well about the AI without changing the map completly. Only thing I might have a look into is the issue with workers and immovables

OK, I expected this. :)

The AI thing is more a general training issue anyway. Also this is supposed to be a multiplayer scenario, so AI performance is less critical. Maybe the Barbarian AI player should get some more boost though.

hessenfarmer:
the lava remains were indeed a lua error. 
Fixed this and now for every field of lava in region of 1 everything is removed except rocks. This is not exactly the behaviour you get when trying to build something but much closer to it then before.  
I fixed the map as well to enable acces to each crater from each side.

tothxa:
> the lava remains were indeed a lua error. Fixed this and now for every field of lava in region of 1 everything is removed except rocks. This is not exactly the behaviour you get when trying to build something but much closer to it then before. I fixed the map as well to enable acces to each crater from each side.

Thank you.

Now I tested with Atlanteans, so now the Barbarian AI started near the red starting position, where it did much better. Orange (the green starting position in the editor) however is still very slow to start. This time I realised that I can have a spectator too, so I saw that it's because it's too far away from the mountain. I tried to find a better starting position, but it's not possible without a lot of redesign because of the closed sea. Gray also has some difficulty finding its mountain.

So, TL;DR, I think the easiest fix is to not mark the map as balanced.

hessenfarmer:
> Now I tested with Atlanteans, so now the Barbarian AI started near the red starting position, where it did much better. Orange (the green starting position in the editor) however is still very slow to start. This time I realised that I can have a spectator too, so I saw that it's because it's too far away from the mountain. I tried to find a better starting position, but it's not possible without a lot of redesign because of the closed sea. Gray also has some difficulty finding its mountain.
> 
> So, TL;DR, I think the easiest fix is to not mark the map as balanced.

Well, we could do that however I believe it is balanced for human players as MP Scenario. Declare balance on AI Performance would led to most of the maps being unbalanced.   
So I would like to keep the tag if you don't have any hard feelings about this

tothxa (approved):
OK. Haven't tested with Barbarians though.

hessenfarmer:
> OK. Haven't tested with Barbarians though.

thanks for the review

PR #5207, opened 2022-01-13
Refactor FloatNode and NumberNode

### LINKED ISSUE #5201: Too many decimal places in encyclopedia entries of mines
**Describe the bug**
The percentages for mines are listed with the default precision of 6 decimal places, which is ugly.

**To Reproduce**
1. Start a game
2. Open the encyclopedia
3. Select a mine
4. Look for the chance of production after resources ran out

**Expected behavior**
Earlier (e.g. in v1.0) it was dynamic, like 50% / 33.33%.

**Screenshots**
mine_percent

**Version:**
 - Widelands Version: 1.1~git25591[3f012ac@master]

**Additional context**
Currently all format strings used in lua code use only wildcard placeholders, which can't set the precison in the format strings. I tried replacing them with `%1$.2f%%` and `%2$.0f%%` and it works, but I'm not sure if it's safe, and it makes the number of decimal places fixed, so 50% is still written as 50.00%. OTOH `%.0f%%` for both is probably acceptable here.

There may be other places where this may happen, so ideally `format()` float nodes should be changed to handle dynamic precision in some way after all, if it's not too hard.

tothxa:
Fixes #5201
and replaces the repetitive parts with inline helper functions.


tothxa:
BTW from here it wouldn't be too hard to make it adjust dynamic precision to try to keep the number of significant digits 6 (I'd still stop at precision == 1 to show that it's a float), so that it would be more compatible with boost.

edit:
It isn't worth the required complexity, because it's practically never used in Widelands. The rounding would have to be revised after adjusting the precision for numbers like 999.99991.

src/base/format/tree.cc
```diff
@@ -194,8 +194,11 @@ inline std::unique_ptr<AbstractNode> Tree::parse_type_spec(const char*& format_s
 	case 'b':
 		return std::unique_ptr<AbstractNode>(new BooleanNode(flags, min_width, precision));
 	case 'f':
-		return std::unique_ptr<AbstractNode>(new FloatNode(
-		   flags, min_width, precision == kInfinitePrecision ? kDefaultFloatPrecision : precision));
+		bool dynamic_precision;
+		dynamic_precision = (precision == kInfinitePrecision);
```
Noordfrees:
```suggestion
		const bool dynamic_precision = (precision == kInfinitePrecision);
```

src/base/test/test_string.cc
```diff
@@ -159,6 +159,9 @@ TESTCASE(string_formatting) {
 	check_equal("A123      X", format("A%-9.0fX", 123.456));
 	check_equal("A+123.5   X", format("A%+-9.1fX", 123.456));
 	check_equal("A+00123.46X", format("A%0+9.2fX", 123.456));
+	check_equal("A1X", format("A%1%X", 1.0));
+	check_equal("A1X", format("A%1%X", 1.0000002));
```
Noordfrees: I'm not sure this is desirable. A floating-point argument should always be displayed as a floating-point value with at least one decimal, so IMHO `"A1.0X"` would be the desired outcome for dynamic-precision placeholders.


---

The documentation in tree.h should be updated to mention the behaviour when not specifying an explicit precision for a float node.

tothxa: Unfortunately this is required to fix the encyclopedia entries in the linked issue. Would you allow them to be "5.0%"?

tothxa: Sorry, do you mean OK to leave it as it is, or OK to make it "5.0%"?

Noordfrees: OK for "5.0%" :smile:

src/base/format/tree.cc
tothxa: Oops, I have to revert this. It gives a bunch of errors like this:
```
../src/base/format/tree.cc:236:7: error: jump to case label [-fpermissive]
  case 'P': {
       ^~~
../src/base/format/tree.cc:197:14: note:   crosses initialization of ‘const bool dynamic_precision’
   const bool dynamic_precision = (precision == kInfinitePrecision);
```

I knew there was a reason.

Noordfrees: Just add `{}` around this `case`. This should always be done when declaring local variables in a switch statement.
```c++
switch (...) {
case 'c':
	...
	break;
case 'P': {
	bool localvar = ...;
	...
} break;
}
```

tothxa: Thanks!

src/base/format/float_node.h
```diff
@@ -22,31 +22,97 @@
...
+		int64_t fractional = 0;
+		bool is_negative = false;
+		bool is_zero = false;
+		size_t current_precision = precision_;
+
 		switch (t) {
 		case ArgType::kFloat:
-			if (arg_u.float_val > kMaxInt || arg_u.float_val < -kMaxInt) {
+			double rounded;
+			double frac;
```
Noordfrees, line 66: Same here, `switch` `case`s declaring a local variable should always be inside `{}`

src/base/format/write_number.h
```diff
@@ -0,0 +1,96 @@
...
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
```
Noordfrees: needs updating to the new header format

Noordfrees (approved):
LGTM, two more comments inline

PR #5208, opened 2022-01-15
Don't require Homebrew to be installed at /usr/local.

dsymonds:
It is very common for it to be installed elsewhere, and to rely on the
`brew` tool being in the user's PATH.

### DIFF
--- a/install-dependencies.sh
+++ b/install-dependencies.sh
@@ -83,8 +83,8 @@ if [ -z "$DISTRO" ]; then
       DISTRO="debian"
    elif [ -f /etc/solus-release ]; then
       DISTRO="solus"
-   elif [ -f /usr/local/Homebrew/bin/brew ]; then
-      DISTRO="homebrew"
+   elif brew --prefix > /dev/null 2>&1; then
+     DISTRO="homebrew"
    fi
 fi
 


klaus-halfmann (changes requested):
Thanks, I actually never used this file :-) 
Before I switch to homebrew I used macports but installed the dependencies manually

Does not work for me?
```
% git status -v
On branch brew-detect
Your branch is up to date with 'origin/brew-detect'.
brew-detect % git remote -v
origin	git@github.com:dsymonds/widelands.git (fetch)
origin	git@github.com:dsymonds/widelands.git (push)
brew-detect % ./install-dependencies.sh
ERROR. Unable to detect your operating system.
```
I use zsh but the script uses /bin/sh but brew shellenv seems incorrect.
But you are you the right path, let us try to improve this.

install-dependencies.sh
```diff
@@ -83,7 +83,9 @@ if [ -z "$DISTRO" ]; then
       DISTRO="debian"
    elif [ -f /etc/solus-release ]; then
       DISTRO="solus"
-   elif [ -f /usr/local/Homebrew/bin/brew ]; then
+   fi
+   # Look for Homebrew. The `brew` binary should be on the path.
+   if brew shellenv | grep HOMEBREW_PREFIX; then
```
klaus-halfmann: please use:   
 ` if brew --prefix; then`

klaus-halfmann: and elif, too as this will break on linux etc otherwise

klaus-halfmann (changes requested):

dsymonds:
All done. PTAL.

klaus-halfmann (approved):
Works as expected, thanks.

@Noordfrees AFAIK this is not used by our automated builds should we include it?

PR #5209, opened 2022-01-15
fixed links for documentations svg-images

### LINKED ISSUE #5206: Documentation SVG-image links are wrong
On the server some things are different...
Beside the [systemd service file seems to be not work correct](https://github.com/widelands/widelands/discussions/5203) the created links of the svg-images are wrong.

Background:
On the server we produce html files with the sphinx-builder `dirhtml` which produces __one__ index.html for each page so the directory tree looks like:
~~~~
.
├── add-ons
│   └── index.html
├── animations
│   └── index.html
├── autogen_ai_hints
│   └── index.html
├── autogen_auxiliary_coroutine
│   └── index.html
…
~~~~
Whereas during developing the website we use the sphinx-builder `html` which produces hml-files for each page. So the directory looks like:
~~~~
.
├── add-ons.html
├── animations.html
├── autogen_ai_hints.html
├── autogen_auxiliary_coroutine.html
~~~~
Creating the documentation with the sphinx-builder `html` during development of the website is on purpose: Otherwise (using `dirhtml`) django complains about the directory structure and gives an errormessage: "Directory indexes are not allowed".

I didn't thought of this difference, so the links are wrong.

I have a solution for the wrong links, but have to test them on the alpha site.

frankystone:
Fixes: #5206


matthiakl (commented):
This will break local `make html` generation again. Can you add a command line switch to select which links should be generated? (e. g. `extract_rst.py [-graphs|-graphs-dir]`)

frankystone:
> This will break local `make html` generation again.

Dammn...

> Can you add a command line switch to select which links should be generated? (e. g. `extract_rst.py [-graphs|-graphs-dir]`)

I have to think about this....

frankystone:
I am thinking about having __one__ script which takes care about creating the documentation. Let's say this script is called `make_doc.py` and calls `extract_rst.py` and `shpinxbuild [options]`. This script can then take care of the arguments and pass them through the other scripts and no one has to call two scripts manually.

I assume this will make the other scripts (Makefile and make.bat) superfluous? Or is the `Makefile` needed to be there for other things?

matthiakl:
> I am thinking about having one script which takes care about creating the documentation.

+1. The sphinx Makefile and make.bat look like they are simply the auto generated ones with `sphinx-quickstart`.
AFAIK they are not integrated into the cmake build system.

frankystone:
This should be ready now. I have restricted the sphinx-builder to the builders `html`, `dirhtml` and `singlehtml`. The others may not work correctly because the old `make.bat` and `Makefile` do not reflect the changes with newer Sphinxdoc versions.
`extract_rst.py` can be called independently of `make.py`, this is needed for the website. Will provide a new branch for the website to work correctly. Both branches (this and the website related one) should be merged within a narrow time frame.

doc/sphinx/documentation_enhancements.py
```diff
@@ -35,8 +35,23 @@ def name_hyphenated(self):
...
 
     def get_graphviz_link(self):
-        html_link = 'href="../{}#{}", target="_parent"'.format(
-            self.outfile.replace('.rst', '.html'), self.name.lower())
+        builder = classes.builder
+        if builder == 'dirhtml':
+            html_link = 'href="../{folder}/index.html{anchor}",\
+                target="_parent"'.format(
+                folder=self.outfile.replace('.rst', ''),
+                anchor='#{}'.format(self.name.lower())
```
matthiakl: No need for nested `format` here
```suggestion
            html_link = 'href="../{folder}/index.html#{anchor}",\
                target="_parent"'.format(
                folder=self.outfile.replace('.rst', ''),
                anchor=self.name.lower()
```

doc/sphinx/make.py
```diff
@@ -0,0 +1,86 @@
...
+        subprocess.run(['python', 'extract_rst.py'])
+
+
+def make_doc(**kwargs):
+    print('Creating the source code documentation as {}.'.format(
+        kwargs['builder'])
+    )
+
+    subprocess.run(['sphinx-build',
+                    '-b', kwargs.get('builder', ''),
```
matthiakl: In line 41 you assume that the key 'builder' exists, so the default value is never used and confusing here. In case the empty string is used I guess sphinx-build will fail.
```suggestion
                    '-b', kwargs['builder'],
```

doc/sphinx/make.py
```diff
@@ -0,0 +1,86 @@
...
+                    '-d', 'build/doctrees',
+                    'source',
+                    'build/{}'.format(kwargs.get('builder')),
```
matthiakl: Same here:
```suggestion
                    'build/{}'.format(kwargs['builder']),
```

matthiakl (commented):
LGTM, just a few nits.

frankystone:
Thanks for the review :-)

All fixed

matthiakl (approved):

PR #5215, opened 2022-01-24
Add checks for unknown wares and workers when loading economy profiles

### LINKED ISSUE #5214: scenario or addon wares make game crash when economy profile selected
if a scenario or addon defines new wares the game crashes if the economy profiles men u is opened in the economy window. 

possible solutions:
1. add possibility to amend economy profiles
2. define default values for a ware (i.e. 1) in every economy profile.

2 is better as there might be customised profiles which we don't know

tothxa:
fixes #5214 

This just ignores unknown or missing wares and workers. It fixes the crashes both when trying to read target value for a custom ware/worker from a profile which doesn't have it, and when a profile is saved with custom ware(s)/worker(s) and then later loaded in a game without the custom item(s).

It doesn't implement neither option for defining values for custom items in existing economy profiles, so the default values will be used for all profiles that don't have the custom items. Saving a profile will store the modified values of custom items too, but they will be lost when any profile is saved from a game without the custom items. (This is a general limitation of the configuration handling code.)


tothxa:
I think the persistence problem could be solved by adding a `map` of unknown wares as a member to the `EconomyOptionsWindow` class. It may be too ugly-hacky though...

That would also make it possible to use some special value (e.g. `__custom_items`) to set targets to items that don't have one in the specified profile. This could be set in the 2 standard profiles, but I think adding more GUI code to make it possible to set it for custom profiles would be overkill. (It would be possible though to edit the user config file and add it manually.)

I'd treat it as a multiplier, so "Efficiency" would get 0, and "Stockpile" would get 2 which would mean double the default value.

All this could be done either right here, or later in a separate PR.

hessenfarmer (approved):
code lgtm 
works in my scenario

hessenfarmer:
regarding the persistence I think we rather should require an addon maker ( and enabling this possibility as well as explaining this in the doc) to provide new default profiles. for custom profiles with addons they should be persistent as long as the addon will be enabled which will be either short or rather long due to personal liking of the addon, so I think just giving the user a hint about this would be ok.  
for scenarios it won't hurt to just ignore sceanrio wares as they are only active shortly and a scenario designer might want to set the targets anyway.    
so In short we should rest with this solution reagrding the economy option window
Edit: @Noordfrees any opinons
Edit2 @tothxa thanks for fixing

tothxa:
Thanks for the review. :)

PR #5217, opened 2022-01-25
Msys update

hessenfarmer:
Lerc dependency of libtiff comes now with static lib in Msys so we don't need to downgrade actual libtiff


matthiakl (commented):
LGTM, but is there a reason you only comment and don't delete the downgrade lines?

hessenfarmer:
> LGTM, but is there a reason you only comment and don't delete the downgrade lines?

Actually, it is just convenience as I am pretty much sure we will need to downgrade some package in the future again. 
So if I might not be around or if I have forgotten at least I don't hve to fiddle around with the syntax.

Same reason I why the hack of the broken repo line is still there as it took hours to find out how this worked.

matthiakl (approved):

PR #5219, opened 2022-01-27
Fix clicking a window not changing tool in editor

akupar:
This fixes window clicking a window in the editor to make the tool active. I assume it is a bug, because the code to select the tool was already there, but the windows didn't emit a clicked event.


Noordfrees (approved):
LGTM, thanks for fixing :)

PR #5228, opened 2022-02-01
Appveyor fix

### LINKED ISSUE #5222: appveyor Artifact upload failing due to invalid filename
Since some builds our appveyor builds are failing as the upload of the final artifact is not working. I have created a support ticket which resulted in the following answer:   

<html><body>

From: Feodor Fitsner (Support staff)                      Well, nope, the issue is not in artifact storage size. Apparently, the new storage we are adopting right now does not like []  square brackets in a file name. Remove square brackets from file name  or wrap it into zip (which is even more recommended considering the size  of that .exe) and upload should be working again. PS: Consider uploading artifact in some builds only, like tag builds to reduce storage consumption.                        
On Fri, Jan 28 at 08:13 AM PST, Feodor Fitsner wrote:                            widelands-dev account has reached a  monthly allowance of 50 GB for artifacts storage. We've increased it to  75 GB, so you should be unblocked now.

</body>
</html>

So if we want to keep appveyor for distribution of dev builds we need to remove the square brackets from our filename.
Furthermore we should merge the PNGQuant branch soon to reduce file size and not exceeding our quota.

hessenfarmer:
fixes #5222 keeping appveyor build
if approved this branch has to be removed from appveyor.yml before merging


appveyor.yml
```diff
@@ -46,19 +44,16 @@ build_script:
   - cmd: cd %APPVEYOR_BUILD_FOLDER%
   - cmd: md build
   - cmd: cd build
-  # detect revision
-  - cmd: python %APPVEYOR_BUILD_FOLDER%\utils\detect_revision.py > %APPVEYOR_BUILD_FOLDER%\WL_RELEASE
-  - cmd: set /p WIDELANDS_VERSION_STRING=<%APPVEYOR_BUILD_FOLDER%\WL_RELEASE
+  # detect version
+  - cmd: set /p WIDELANDS_NEXT_VERSION=<%APPVEYOR_BUILD_FOLDER%\NEXT_STABLE_VERSION
+  - cmd: set SHORT_ID=%APPVEYOR_REPO_COMMIT:~0,7%
```
matthiakl: You could set `WIDELANDS_VERSION_STRING` here, or even the full setup name to remove the duplication below
```suggestion
  - cmd: set SHORT_ID=%APPVEYOR_REPO_COMMIT:~0,7%
  - cmd: set WIDELANDS_SETUP_NAME=Widelands-%CONFIGURATION%-%PLATFORM%-%WIDELANDS_NEXT_VERSION%~git(%SHORT_ID%@%APPVEYOR_REPO_BRANCH%)
```

matthiakl (commented):
Sorry, I already looked shortly into it previously, but then forgot about it.
The format looks good, just a small suggestion to remove duplication.

matthiakl (approved):
Much better to read :)
Don't forget to remove the branch before merging.

hessenfarmer:
as appveyor only cha ge no need to wait for testsuite

PR #5229, opened 2022-02-02
Considering distance to warehouse when starting new buildings

### LINKED ISSUE #4930: Closer distance between productionssites and warehouses (by AI)
I noticed on large maps that the AI often placed production buildings far away in their territory, far away from warehouses. Mostly at the end of a dead end street. This leads to long transport routes and low productivity, which delays the game (with AI versus AI) enormously.

It would be nice if the AI management defined a maximum distance from a production building to the nearest warehouse or port. At least if there is a corresponding building site.

In addition, the AI generally does not build any or only very few warehouses and then usually close to headquarters. It would be nice if there was a minimum distance between the warehouses so that they can be better distributed.

And it is clear to me that the performance of the game drops significantly if there are too many warehouses. I placed warehouses on the map myself using a script and from 5-10 warehouses per player the game shows an enormous drop in the FPS.

tibor95:
Even currently we have info for flag-nearest WH distance, of course this cannot be directly used for field where there is no flag.
So we introduced new buildable field variable **average_flag_dist_to_wh**, that is just avg for flags in vicinity. Small problem is that when there are flags not connected in vicinity, we use distance 500 and this bit skews the avg... Maybe lower default value would be better?

This info is then used in construct_building(), we calculate new variable **wh_distance_malus** that is sometimes added and sometimes deducted to specific buildings. See code.

After the approach and code is more-less approved, I will do some training and than will be ready for merge.
fixes #4930


src/ai/defaultai.cc
```diff
@@ -2841,6 +2884,14 @@ bool DefaultAI::construct_building(const Time& gametime) {
 		assert(player_);
 		int32_t const maxsize = player_->get_buildcaps(bf->coords) & Widelands::BUILDCAPS_SIZEMASK;
 
+		// Some buildings needs to consider distance from nearest warehouse
+		// It is non-negative value, and should be deducted from prio for some productionsites
+		const int32_t wh_distance_malus = management_data.neuron_pool[31].get_result_safe(
```
hessenfarmer: I think the calculated malus is not related to the actual value of the distance as it is simply the addition of 2 neuron weights at a neuron position determined by the flag distance. 
it would be better to claculate the malus directly from flag distance and scale it by magic number.

src/ai/defaultai.cc
```diff
@@ -2926,6 +2977,17 @@ bool DefaultAI::construct_building(const Time& gametime) {
 					number_of_same_nearby += same_it->second;
 				}
 
+				// Considering distance to a warehouse NOCOM
+				if (!bo.requires_supporters && !bo.inputs.empty()) {
```
hessenfarmer: rather then doing this generally, I would prefer doing this for every subtype of building in the following.

src/ai/defaultai.cc
```diff
@@ -2926,6 +2977,17 @@ bool DefaultAI::construct_building(const Time& gametime) {
...
+					prio -= wh_distance_malus;
+
+				}
+				// some "independent" space consumers can be pushed to bigger distance from wh
+				if (bo.is(BuildingAttribute::kSpaceConsumer) && bo.inputs.empty() &&
+				    bo.is(BuildingAttribute::kSpaceConsumer)) {
```
hessenfarmer: 1. the space consumer entry is doubled.
2. I prefer not doing this in general but specific for each subtype

src/ai/defaultai.cc
```diff
@@ -3398,24 +3462,31 @@ bool DefaultAI::construct_building(const Time& gametime) {
...
-				uint16_t nearest_distance = std::numeric_limits<uint16_t>::max();
-				for (const WarehouseSiteObserver& wh_obs : warehousesites) {
-					const uint16_t actual_distance =
-					   map.calc_distance(bf->coords, wh_obs.site->get_position());
-					nearest_distance = std::min(nearest_distance, actual_distance);
-				}
-				// but limit to 30
-				const uint16_t max_distance_considered = 30;
-				nearest_distance = std::min(nearest_distance, max_distance_considered);
-				if (nearest_distance < 13) {
```
hessenfarmer, line 3414: I think we should keep this safeguard but we can use average_flag_dist_to_wh.
unrelated change might be to increase this a bit (lets say 20) as AI tends to build warehouses to close together

src/ai/defaultai.cc
```diff
@@ -6777,6 +6846,11 @@ void DefaultAI::gain_building(Widelands::Building& b, const bool found_on_load)
 		}
 		if (target_bo.type == BuildingObserver::Type::kWarehouse) {
 			++numof_warehouses_in_const_;
+			if (!found_on_load) {
```
hessenfarmer: setting the task scheduler here does not make sense as we just gained a condtructionsite which isn't considered in the checking for distances as only finished warehouses are counted. 
so either we could afford to skip this, as the distances are calculated every 15 seconds gametime
or we need to move it down to line 6941 ff

src/ai/defaultai.cc
```diff
@@ -2060,11 +2075,11 @@ void DefaultAI::update_buildable_field(BuildableField& field) {
 		   management_data.neuron_pool[10].get_result_safe(field.military_loneliness / 50, kAbsValue);
 
 		score_parts[30] =
-		   -10 * management_data.neuron_pool[8].get_result_safe(
-		            3 * (field.military_in_constr_nearby + field.military_unstationed), kAbsValue);
-		score_parts[31] =
-		   -10 * management_data.neuron_pool[31].get_result_safe(
+		   -20 * management_data.neuron_pool[8].get_result_safe(
 		            3 * (field.military_in_constr_nearby + field.military_unstationed), kAbsValue);
+		// score_parts[31] =
```
hessenfarmer: this should be reinstated to the former state as we won't need the neuron no 31 for the calculation in the following

src/ai/defaultai.cc
```diff
@@ -2159,6 +2174,34 @@ void DefaultAI::update_buildable_field(BuildableField& field) {
 		field.military_score_ += part;
 	}
 
+	// Using F-neurons to slightly modify calculated score
```
hessenfarmer, line 2174: what is the purpose of the following code

hessenfarmer:
approach except calculation of malus seems to be good. 
see inline comments

src/ai/defaultai.cc
tibor95: What this neuron do is converts values <-20,20> to <-100,100>. Or if absvalue, then to <100,0>

And problem with magic numbers is that we are running out of them. And increasing them would mean compatibility break.

src/ai/defaultai.cc
tibor95: Generally I do not mind but we have many buildings x tribes, some of them not invented yet...

src/ai/defaultai.cc
tibor95: This code is completely removed. This was checking air distance and is replaced completely by that flag distance

src/ai/defaultai.cc
tibor95: good catch!

src/ai/defaultai.cc
tibor95: I did this to release one magic number, and not 31, but other, it is not visible in the diff here.. but I think I used it elsewhere

src/ai/defaultai.cc
tibor95: F-neurons takes up to 5 bools and returns one bool. So this is just add possibility for AI to do some more magic. Will require training though. Yes, it is kind of black magic

src/ai/defaultai.cc
tibor95: sorry, it returns <-X,X> where X is up to 100

tibor95: Point of this that the relation input:output is not linear, but is logarithmic, parabolic or linear. So bit more sophisticated than magic number that is linear (logically)

hessenfarmer: ok got it. in this case the second paret of the addition is maybe zero most of the time even if we have big distances (i.e. 50). so maybe better choose a divisor smaller then 25 (maybe 4 or 5) to get reasonable numbers for input.

src/ai/defaultai.cc
hessenfarmer: what I meant is to think for each ai defined type of building (i.e. kRanger, kFisher, and so on) this would and should cover all buildings of all tribes and makes for the opportunity to have even scaled appliance of the malus.

src/ai/defaultai.cc
hessenfarmer: I know and this is fine. However I still believe we should not allow to build any Warehouse if average_flag_dist_to_wh is lower then 15 to 20 for this field

hessenfarmer: I was only talking about this if block not the rest

src/ai/defaultai.cc
hessenfarmer: you released neuron number 31 which is used now for malus calculation. However there are a lot of neurons still unused (e.g. all neurons from 35 to 41, 68 to 72 and many more so no need to free any.
Unrelated Issue is that some of them are used multiple times which should be changed to give the AI better possibilities to train. (multiple used neurons are 4, 8, 18, 23(4x))

src/ai/defaultai.cc
tibor95: It is malus, so bigger distance => bigger malus. Up to that upper limit X 
Those dividers define where distance is cropped. I was thinking about tweaking it a bit. First one can be 1 (basically distance above 20 is cropped to 20) and second to be cropped to 60 (divider 3)

src/ai/defaultai.cc
tibor95: "scaled malus" - I would not do this now. Let training come up with some optimum for everyone

I think the malus should not be applied to some types (e.g. quary, lumberjacts, supporting sites). Remember it started as a request for placing warehouses. So this here is afterthought, I would not use it very extensively

src/ai/defaultai.cc
hessenfarmer: I would not crop the first one at 20 as this is stil rather near a warehouse and would deliver the highst malus already so using 2 and 3 might be good which would increase malus with distance but has a bigger range

src/ai/defaultai.cc
hessenfarmer: the request was about building productionsites near warehouses to shorten ware transport routes.  
And I agree not to (fully) apply it to some types or even give them a bonus. therefore i would have applied it (maybe scaled) seperately for each type of building.

src/ai/defaultai.cc
tibor95: OK, reverted it. I will try to do some housekeeping with these neurons

src/ai/defaultai.cc
tibor95: OK, I added a condition at least 15

src/ai/defaultai.cc
tibor95: The second one was a mistake, I changed it to
!bo.is(BuildingAttribute::kSupportingProducer
But as I said, I am not very happy about being very specific about this malus

src/ai/defaultai.cc
tibor95: OK, first one no divider and second one divider 3 - cropping at 60 - this is fine for me

src/ai/defaultai.cc
tibor95: sorry, you are right
In fact if you put your ideas "on paper" I will transfer them to the code. I just believe current setup is quite good. There will be new commit soon, you can wait for it

tibor95:
Still I will get back to that unconnected flags/roads. One unconnected and 5 connected - the avg can get easily to 100 and confuse everything. Alternative is to ignore flags that does not have that info - were not reachable from any warehouse in last 120 seconds

EDIT:
see
constexpr Duration kFlagDistanceExpirationPeriod(120 * 1000);
In fact can be shortened a bit

hessenfarmer:
I agree to the problem and the solution. As we are interested in distance of the field to a warehouse we should ignore flags that are not connected yet. we should adapt the code to deliver 0 as default resulting in no malus / bonus at all. 
if we don't have a flag in the inner circle we might check the outer circle as well but in this case we should count them as their distance plus the inner radius (kProductionArea+2), as if no flag in inner ring any flag in outer ring might be closer to the wh then our field. 
While thinking on this I realized that we are counting any flag but we need to consider only ours .

edit: I think 2 minutes is ok for expiration as a building at least takes this time to be constructed and only new warehouses do change the values significantly

src/ai/defaultai.cc
```diff
@@ -2060,8 +2074,8 @@ void DefaultAI::update_buildable_field(BuildableField& field) {
 		   management_data.neuron_pool[10].get_result_safe(field.military_loneliness / 50, kAbsValue);
 
 		score_parts[30] =
-		   -10 * management_data.neuron_pool[8].get_result_safe(
-		            3 * (field.military_in_constr_nearby + field.military_unstationed), kAbsValue);
+		   -0 * management_data.neuron_pool[37].get_result_safe(
```
hessenfarmer: typo should be -10

tibor95: fixed

tibor95:
I changed the code, only our flags are counted now. Anyway, somebody else flags have no connection to our warehouse anyway, but why waste CPU time...
I do not like default 0, because there is one sure thing - there is no warehouse in inner circle. What about 60?
Or we can use some constant like 500 indicating that some buildings (warehouse) can not be built here.

I do not like idea of checking flags in outer circle - it begins to be complicated...

tibor95:
I tested this, counting flags seems to work properly

src/ai/defaultai.cc
hessenfarmer: ok I made my mind. 
the general aplliance covers almost everything. 3 exceptions.
1. rangers are considered twice as they space consumer without inputs and not supporting producers
2. hunters should be applied a bonus if far away
3. frisian farms (i.e. supporting producers AND spaceconsumer) should be applied a bonus if farer away

Edit: if this shows to not work out we still can apply the things individually in a follow up

hessenfarmer:
> I changed the code, only our flags are counted now. Anyway, somebody else flags have no connection to our warehouse anyway, but why waste CPU time... I do not like default 0, because there is one sure thing - there is no warehouse in inner circle. What about 60? Or we can use some constant like 500 indicating that some buildings (warehouse) can not be built here.

well, if don't have a flag in radius 8 we might have one in radius 9, but we don't know an actual value to approximate the distance. So taking 0 as default for we do not know would result in the appliance of zero bonus or malus due to the neuron curves. default 60 would result in the biggest bonus/malus apllied for this field. so I think 0 would be more conservative for we don't know. 

> 
> I do not like idea of checking flags in outer circle - it begins to be complicated...    
  
undrstood and agreed

tibor95:
I am thinking about possible drawbacks of using 0, so far only one small one. We should not build warehouse if 0, basically we do not want to build in some empty area.

src/ai/defaultai.cc
tibor95: OK, I will prepare new commit

hessenfarmer:
> I am thinking about possible drawbacks of using 0, so far only one small one. We should not build warehouse if 0, basically we do not want to build in some empty area.

we have reinserted the safeguard to have at least a distance of 15 to prevent this

tibor95:
Indeed!

src/ai/defaultai.cc
tibor95: Spaceconsumers generally should be positioned farer. Especially independent ones - regular barbiarian farm e.g.

tibor95: All productionsites are now handled on one place. What I implemented is not exactly what you wanted, but we will discuss

src/ai/defaultai.cc
```diff
@@ -2978,14 +2976,21 @@ bool DefaultAI::construct_building(const Time& gametime) {
...
-					prio -= wh_distance_malus;
-				}
-				// some "independent" space consumers can be pushed to bigger distance from wh
-				if (bo.is(BuildingAttribute::kSpaceConsumer) && bo.inputs.empty() &&
-				    !bo.is(BuildingAttribute::kSupportingProducer)) {
-					prio += wh_distance_malus;  // push them farer from nearest wh
+				// Considering distance to wh for productionsites:
+				// All productionsites in one place
+
+				if (bo.is(BuildingAttribute::kWell) && bo.is(BuildingAttribute::kRanger) &&
```
hessenfarmer: this needs to be an OR rather then an AND I believe ;-)

src/ai/defaultai.cc
```diff
@@ -2978,14 +2976,21 @@ bool DefaultAI::construct_building(const Time& gametime) {
...
+				    bo.is(BuildingAttribute::kHunter)) {
+					prio += wh_distance_malus;
+				} else if (bo.is(
+				              BuildingAttribute::
+				                 kSpaceConsumer)) {  // ??? &&
+					                                  // bo.is(BuildingAttribute::kSupportingProducer))
+					                                  // {
+					prio += wh_distance_malus;
+				} else if (!bo.inputs.empty() && !bo.ware_outputs.empty() &&
```
hessenfarmer: we should remove the outputs or we should include worker outputs as an OR to ware outputs to cover recruitment sites (barracks and second carrier)

src/ai/defaultai.cc
```diff
@@ -2978,14 +2976,21 @@ bool DefaultAI::construct_building(const Time& gametime) {
...
+				           !bo.is(BuildingAttribute::kSupportingProducer)) {
```
hessenfarmer: I think Supporting producers should not be considered here but we should ensure that the malus is not applied to buildings requiring suppporters

tibor95:
I just pushed another attempt. I am not sure how you meant it about supporting sites..

src/ai/defaultai.cc
```diff
@@ -2976,21 +2976,18 @@ bool DefaultAI::construct_building(const Time& gametime) {
...
-					                                  // bo.is(BuildingAttribute::kSupportingProducer))
-					                                  // {
-					prio += wh_distance_malus;
-				} else if (!bo.inputs.empty() && !bo.ware_outputs.empty() &&
-				           !bo.is(BuildingAttribute::kSupportingProducer)) {
-					prio -= wh_distance_malus;  // push them closer from nearest wh
+					prio += wh_distance_malus;  // push farer
+				} else if (bo.is(BuildingAttribute::kRecruitment)) {
+					prio -= wh_distance_malus;  // push closer
+				} else if (bo.is(BuildingAttribute::kSupportingProducer)) {
```
hessenfarmer: this line is superfluous. this was not what I meant

src/ai/defaultai.cc
```diff
@@ -2976,21 +2976,18 @@ bool DefaultAI::construct_building(const Time& gametime) {
...
+					;  // exclude from this
+				} else if (bo.is(BuildingAttribute::kSpaceConsumer)) {
+					prio += wh_distance_malus;  // push farer
+				} else if (!bo.inputs.empty() && !bo.ware_outputs.empty()) {
```
hessenfarmer: I meant to add &&  !bo.requires-supporters to this condition as theit placement should be only dependant to their supporters

src/ai/defaultai.cc
tibor95: Right now it does what you want and covers also spaceconsumers.

So you say that spaceconsumers should be pushed farer no matter if they are or are not supportingproducers (if there are any such buildings)

hessenfarmer: yes everything alright with Space consumers. 
just delete the part with supporting producers

tibor95: So like this:

```
} else if (bo.is(BuildingAttribute::kRecruitment)) {
	prio -= wh_distance_malus; // push closer
}else if (bo.is(BuildingAttribute::kSpaceConsumer)) {
	prio += wh_distance_malus;  // push farer
} else if (!bo.inputs.empty() && !bo.ware_outputs.empty() && !bo.is(BuildingAttribute::kSupportingProducer)) {
	prio -= wh_distance_malus;  // push closer
```
My objection is that SpaceConsumers that are also SupportingProducers (if any, I do not know) should not be pushed farer / malus should not be applied to them. But if it is fine for you....

hessenfarmer: no like this:
else if (!bo.inputs.empty() && !bo.ware_outputs.empty() && !bo.requires_supporters)) {
	prio -= wh_distance_malus;  // push closer

there are supporting producers but they still need space so should be safe to push them away from warehouses

src/ai/defaultai.cc
tibor95: ok

src/ai/defaultai.cc
tibor95: ok

hessenfarmer:
well it looks good now from code and algorithm.
things left to do:
1. fix the codecheck warnings - make our suite happy
2. merge master
3. doing some training

Well done. Thanks a lot

tibor95:
I am ready to train. 
I will look at codecheck...

tibor95:
https://github.com/widelands/widelands/runs/5054013109?check_suite_focus=true
There are some problems with these multiple 'else if's

tibor95:
if there are no sites that are at the same time recruitments and space consumers I can simplify this..

src/ai/defaultai.cc
```diff
@@ -2926,6 +2971,18 @@ bool DefaultAI::construct_building(const Time& gametime) {
 					number_of_same_nearby += same_it->second;
 				}
 
+				// Considering distance to wh for various types of productionsites:
+				if (bo.is(BuildingAttribute::kWell) || bo.is(BuildingAttribute::kRanger) ||
```
hessenfarmer, line 2975: to satisfy codecheckk add || bo.is(BuildingAttribute::kSpaceConsumer) here to satisfy codecheck

src/ai/defaultai.cc
```diff
@@ -2926,6 +2971,18 @@ bool DefaultAI::construct_building(const Time& gametime) {
...
+				    bo.is(BuildingAttribute::kHunter)) {
+					prio += wh_distance_malus;  // push farer
+				} else if (bo.is(BuildingAttribute::kRecruitment)) {
+					prio -= wh_distance_malus;  // push closer
+				} else if (bo.is(BuildingAttribute::kSpaceConsumer)) {
+					prio += wh_distance_malus;  // push farer
+				} else if (!bo.inputs.empty() && !bo.ware_outputs.empty() && !bo.requires_supporters) {
```
hessenfarmer: here we should include the recruitment site for codecheck

tibor95: Yes, exactly what I meant (both changes)

src/ai/defaultai.cc
```diff
@@ -6872,6 +6918,11 @@ void DefaultAI::gain_building(Widelands::Building& b, const bool found_on_load)
 			if (bo.is(BuildingAttribute::kPort)) {
 				++num_ports;
 			}
+			if (!found_on_load) {
+				// recalculate distance ASAP
+				const Time& gametime = game().get_gametime();
```
hessenfarmer: this line is superfluous due to line 6835

tibor95: this is endless story

hessenfarmer: well our checks are much harder then previous now, however this is increasing code quality.

tibor95:
All green so I start with training

tibor95:
I did some training. You can check if enough, I can do more..

hessenfarmer:
@MarkMcWire would you like to test and see whether the issue you reported might be fixed?

hessenfarmer:
> I did some training. You can check if enough, I can do more..

Hard to say from the exterior. If you are content that the fitness is not increasing much more it might be ok.  I believe we can have it and do more training shortly before a release, leaving us the freedom to fuzrther improve AI

hessenfarmer (approved):
code LGTM.
tests look good as well

tibor95:
Yes, I will do more training later.

I have an idea/question - perhaps we just need more warehouses, especially now when we have that distance from nearest warehouse, something like force warehouse if more than 100 steps from nearest one. Do you think it might help some problems?

hessenfarmer:
> Yes, I will do more training later.
> 
> I have an idea/question - perhaps we just need more warehouses, especially now when we have that distance from nearest warehouse, something like force warehouse if more than 100 steps from nearest one. Do you think it might help some problems?

No I don't think so, because this would completely break our philosophy, where we determine the necessity of warehouses. 
Their necessity is based on size of economy and this should be the way to go. If we need a warehouse we should build it some spots away from other warehouses.  
From my tests we could increase the minimum distance in the defined safeguard a bit to seperate them more (maybe 20 instead of 15) but necessity should be based on economy size.

tibor95:
If we would place mandatory warehouses in some big distance, for most games it would make no difference (would not be applied) but perhaps would help on those special maps...

But OK, this is just an idea...

tibor95:
So what is status? Are you merging?

hessenfarmer:
> So what is status? Are you merging?

I wanted to wait until the weekend (friday or saturday) if there would be any response from @MarkMcWire. Afterwards I planned to merge. 
no further change needed from your side I believe, however feel free to change the minimum distance to up to 20 if you like (line 3460) and you could always add training results of course.
Before I will do the merge I#ll give you a quick notice, ok?

tibor95:
I will not be able to train before Saturday. So new results can be in Monday let say

I would stay with 15, because in small territories second warehouse serves as a safeguard if headquarters was conquered.

MarkMcWire:
> @MarkMcWire would you like to test and see whether the issue you reported might be fixed?

Okay, I will test it.

MarkMcWire:
@tibor95 @hessenfarmer 
Seems to work. The AI builds the building network a little more compact and the ware routes have become a little shorter.
However, at least for me, there are a few problems with buildings that take up a lot of space, such as farms or wells with large working areas. They are built a little too close together or the AI hardly builds wells, although I have set basic_amount = 5.
My farms also have water input, so they are not entirely comparable to farms from other tribes.

MarkMcWire:
@tibor95 I would still penalize buildings with larger work areas like farms, reedyards, and ranger huts so that they are built a little farther away.

tibor95:
I do not know if you saw the relevant code:
https://github.com/widelands/widelands/blob/2b14ea11f27dcf1796ca603dc8b4849aa19a577e/src/ai/defaultai.cc#L2974

We can discuss it of course. If you see how that code could be improved...

As for count of wells - I do not think this branch changed anything in this regard, so I would not open it as a part of this discussion. But I would need to investigate more if there is a problem there...

hessenfarmer:
@MarkMcWire I take your first post for a yes to merge this.     
regarding your other observations a few words:     
The AI in fact is trying to group farms together if popssible as normally they do not need the space they are consuming, and a new built farm can directly harvest fields that are already grown up giving them no delay. This is how human players play as well. a normal farm only needs 3-4 or at most 5 (frisians) fields to run at 100% so an optimal grid for placing farms can be determined, where theitr working areas deliberatly overlap.        
This applies for current wells as well as they are still producing at 70% if no water present.      
The amount of wells to be built is determined from the stock amount of water. However it might take some time to cope with a big shortfall of water, but this should should be resolved after some time.   
Space consumers are already given a bonus on more distant fields, so the chance is higher that they get a higher score on the outside however if the space available is not very big or we are only checking medium ranged fields (we can't check all of them every time for performance reasons) they might be build on spaces not to far away. It is not recomendable to implement hard values as they most probably do not fit each map or tribe, so we are relying on scoring which might lead to suboptimal decisions.    
So in short I am against further bonuses for spceconsumers on outside fields.

hessenfarmer:
@tibor95 i would like to merge this unless you tell me you ae performing some training currently that you want to add

tibor95:
Just go on. I do not do any training now. I plan some once it will be in master...

PR #5238, opened 2022-02-09
Enable hotkeys in Dropdowns

### LINKED ISSUE #5192: Hotkeys should override dropdown filtering
When a dropdown is open and a character key that is also the hotkey for one of the dropdown's entries is pressed, processing the hotkey should take precedence over filtering the dropdown.

Examples:
- In the main menu, press <kbd>S</kbd> to open the singleplayer dropdown. Then press <kbd>L</kbd>. 
  **Result**: The dropdown is filtered for entries containing »L«.
  **Expected**: The Load Game window opens.
- In the main menu, press <kbd>S</kbd> to open the singleplayer dropdown. Then press <kbd>O</kbd>. 
  The dropdown is filtered for entries containing »O«. This is fine because <kbd>O</kbd> is not a shortcut for any entry in this dropdown.

jmoerschbach:
Fixes #5192
I only added the hotkeys from main menu, did not check for other dropdowns with hotkeys
also reduced visibility of many functions in dropdown.h


src/ui_basic/dropdown.h
```diff
@@ -312,61 +313,21 @@ template <typename Entry> class Dropdown : public BaseDropdown {
...
-			add_to_filtered_list(ee.name, ee.value, ee.img, false, ee.tooltip, ee.hotkey);
-		}
-	}
-
-	void save_selected_entry(uint32_t index) override {
-		selected_entry_ = index < filtered_entries.size() ? *filtered_entries[index] : Entry{};
-	}
 
 	bool handle_textinput(const std::string& input_text) override {
+		const std::string& lowered_input_text = to_lower(input_text);
```
```suggestion
		const std::string lowered_input_text = to_lower(input_text);
```
This doesn't return a reference, so a reference would become dangling immediately

src/ui_basic/dropdown.h
```diff
@@ -417,12 +380,23 @@ template <typename Entry> class Dropdown : public BaseDropdown {
 		const Image* img;
 		const std::string tooltip;
 		const std::string hotkey;
+		const HotkeyFunction onShortcutMatch;
```
Noordfrees: Rename for consistency
```suggestion
		const HotkeyFunction on_shortcut_match;
```

src/ui_basic/dropdown.h
```diff
@@ -433,6 +407,63 @@ template <typename Entry> class Dropdown : public BaseDropdown {
 		filtered_entries.push_back(std::unique_ptr<Entry>(new Entry(value)));
 		BaseDropdown::add(name, size(), pic, select_this, tooltip_text, hotkey);
 	}
+	bool check_hotkey_match(const std::string& input_text) {
+		for (auto& x : unfiltered_entries) {
+			if (input_text == to_lower(x.hotkey) && is_in_filtered_list(x.value)) {
+				if (x.onShortcutMatch) {
+					x.onShortcutMatch(x.value);
+				} else {
+					log_err("hotkey match: %s but no hotkey function available!", x.hotkey.c_str());
```
Noordfrees: This happens when pressing P in the editor with the tool dropdown open, also map view dropdown and showhide menu. So perhaps make this just an inobtrusive reminder to add this feature there as well:
```suggestion
					verb_log_dbg("hotkey match: %s but no hotkey function available!", x.hotkey.c_str());
```

Noordfrees (commented):
LGTM, tested and working, a few comments inline :)

Noordfrees (approved):

PR #5261, opened 2022-02-15
Dropdown shortcuts for ingame and editor

jmoerschbach:
some of them (the special modifier ones) don't work - but this is something for another branch...


Noordfrees (commented):
Since the callback function is usually the same for every entry in a dropdown, would it make sense to make this a property of the dropdown itself with a call like `dropdown->set_shortcut_function(shortcut_fn)` instead of storing it for each entry record separately?

Noordfrees:
Code LGTM but one glitch left: In the editor, press the following sequence of keys: <kbd>T</kbd><kbd>P</kbd><kbd>T</kbd><kbd>P</kbd><kbd>T</kbd>. What this does is to open and close the players menu via the tool dropdown. After the final <kbd>T</kbd> press, the dropdown is filtered for something although it shouldn't be.

jmoerschbach:
took me some time, but I would like to have this as is for sake of traceability.
The glitch will be fixed by #5281

Noordfrees (approved):

PR #5263, opened 2022-02-16
5244 misc ai fixes

### LINKED ISSUE #5244: AI don't build enough wells and upgrade no buildings.
As already mentioned in the other thread, the Europeans in the current build of Widelands don't build the required 5 wells (basic amount) and thus somehow get stuck in the basic economy. No battle arenas, warehouses, or building upgrades are constructed even after 10 hours of gameplay.

-> https://drive.google.com/file/d/13Ppm8BqTTDdeqCTcOaAMC4rJf1NGVpBY/view?usp=sharing

Use as addons only europeans_map, europeans tribe (2.2.6) and enhanced_debug_scripts.

hessenfarmer:
fixes #5244


src/ai/defaultai.cc
```diff
@@ -5856,7 +5830,22 @@ BuildingNecessity DefaultAI::check_building_necessity(BuildingObserver& bo,
...
 			}
+			// Determine whether we need more buildings due to low stocklevel
+			const uint32_t current_stocklevel = (get_stocklevel(bo, gametime));
+			const uint8_t stocklevel_threshhold =
+			   std::abs(management_data.get_military_number_at(179));
 
+			if (current_stocklevel > stocklevel_threshhold &&
+			    persistent_data->remaining_basic_buildings.count(bo.id) == 0) {
+				bo.cnt_target -= 1;
+			}
```
tibor95, line 5845: But you might get below basic_amount, cant you?

src/ai/defaultai.cc
```diff
@@ -5856,7 +5830,22 @@ BuildingNecessity DefaultAI::check_building_necessity(BuildingObserver& bo,
...
+
+			if (current_stocklevel < stocklevel_threshhold / 2) {
+				bo.cnt_target += 1;  // increase target
+				// here we increase primary priority if stock gets low to get higher prio on each field
```
tibor95, line 5850: Do you not afraid that cnt_target will go to the sky?
what about check like
if bo.total_count() == bo.cnt_target
Also check for unoccupied sites, if any we should not increase cnt_target

src/ai/defaultai.cc
tibor95:
```
if (current_stocklevel > stocklevel_threshhold &&
    bo.cnt_target > bo.basic_amount) {
	bo.cnt_target -= 1;
}
```

hessenfarmer: good point will add the check for basic economy established as a safeguard.

src/ai/defaultai.cc
hessenfarmer: it can't go to the sky as it is always reinitialised in line 5831 or line 5829.     
However checking unoccupied sites is a valid point will incorporate this

src/ai/defaultai.cc
```diff
@@ -5646,10 +5602,28 @@ BuildingNecessity DefaultAI::check_building_necessity(BuildingObserver& bo,
...
 			if (site_needed_for_economy == BasicEconomyBuildingStatus::kDiscouraged) {
-				tmp_score -= 2;
+				tmp_score -= std::abs(management_data.get_military_number_at(178));
+			}
+
+			// now we find out if the building is needed depending on output stocklevel
+			const uint32_t current_stocklevel = (get_stocklevel(bo, gametime));
+			const uint8_t stocklevel_threshhold =
+			   std::abs(management_data.get_military_number_at(180));
+
```
tibor95, line 5617: Is not a range 0-100 too broad? Maybe reduce it to 5-55 or so....

src/ai/defaultai.cc
tibor95: I meant I would not involve  persistent_data->remaining_basic_buildings here at all

src/ai/defaultai.cc
tibor95: Indeed, but then all you can do here is increase by 1 in addition to what is set above as you mentioned. My idea was to kind of remember cnt_target between runs, so what about to modify 5831:

`bo.cnt_target = std::max(bo.cnt_target, 1 + static_cast<int32_t>(mines_.size() + productionsites.size()) / 30);`

tibor95: Or incorporate hard upper limit e.g. : 5 + static_cast<int32_t>(mines_.size() + productionsites.size()

src/ai/defaultai.cc
hessenfarmer: Now I am confused.  I overlooked that I used persistent_data->remaining_basic_buildings in my first answer. This is directly related to whether there is still a building needed to establish basic economy. After having reached basic economy basic_amount does not matter anymore afaik.  So I tend to change nothing here

src/ai/defaultai.cc
hessenfarmer: well, I have no problem with the upper limit, cause this should be clearly dependent on training however a lower limit of 0 might be too low, so I think I'll simply add 20 to the threshhold resulting in 20-120

src/ai/defaultai.cc
hessenfarmer: as we are running this check much faster as the building could be built and start working I think remeber the target between runs really might be dangerous to go to the max and we would build too much of them.  I'd rather include the size of trainingsites here as a measure for the need of food then pushing the target too much.

src/ai/defaultai.cc
tibor95: > After having reached basic economy basic_amount does not matter anymore afaik. 

This is kind of undefined. I would vote never go below basic economy. Anyway the code is hard to read, so perhaps:

```
if (current_stocklevel > stocklevel_threshhold &&
		basic_economy_achieved && bo.cnt_target > 0) {
		bo.cnt_target -= 1;
	}
```
This is what you mean?

src/ai/defaultai.cc
tibor95: You would not increase the target until:

- no construction sites
- no unoccupied
- total_count() = cnt_target


So it seems safe, but as you think

src/ai/defaultai.cc
tibor95: I would go for lower, but can be.

src/ai/defaultai.cc
tibor95: Basic economy buildings counts are so low anyway - I see no benefit dismantling them in 95% of cases

hessenfarmer: this code isn't about dismantling at all it is only for construction

tibor95: There is other code that handles possible dismantling. I can differ, but generally logic should be never dismantle under cnt_target, of course the implementation in check_productionsites() can differ for various types of sites
So I suggest that cnt_target should never get below basic_amount

tibor95: or we can have two cnt_targets: 

- cnt_min_taret - for dismantling
- cnt_max_target - for building new ones

EDIT: Well, overkill probably...

src/ai/defaultai.cc
hessenfarmer: ok i'll change to 10

tibor95: thanks

src/ai/defaultai.cc
hessenfarmer: I carefully checked the dismantling code and the target is only one safeguard of many others.  Basically dismantling is based on high stocklevel and low performance of a specific site. So if the site is running at 0% and is only protected by cnt_taget as last safeguard it is questionable if this would really be a good idea.  
So I am really 100% sure that this is safe and it gives us just the little push we need.

tibor95:
I will look at the diff again, probably tomorrow, but I have on idea:

We might take out part of logic that is setting cnt_target to new AI job (~'setting buildings targets'), have it run perhaps less frequently, but main goal would be make code better organized and construct_building() smaller. The size of some of the methods is something I do not like.

src/ai/defaultai.cc
```diff
@@ -5646,10 +5604,28 @@ BuildingNecessity DefaultAI::check_building_necessity(BuildingObserver& bo,
...
-				tmp_score += 4;
+				tmp_score += 10 + std::abs(management_data.get_military_number_at(160));
 			}
 			if (site_needed_for_economy == BasicEconomyBuildingStatus::kDiscouraged) {
-				tmp_score -= 2;
+				tmp_score -= std::abs(management_data.get_military_number_at(178));
+			}
+
+			// now we find out if the building is needed depending on output stocklevel
+			const uint32_t current_stocklevel = (get_stocklevel(bo, gametime));
```
tibor95, line 5614: Unneeded braces

hessenfarmer:
> We might take out part of logic that is setting cnt_target to new AI job (~'setting buildings targets'), have it run perhaps less frequently, but main goal would be make code better organized and construct_building() smaller. The size of some of the methods is something I do not like.

Well refactoring the AI code into smaller more readable chunks would be definitly both: a great improvement and a big challenge. So maybe not for this branch. 
If you would go for this I will strongly support this.

src/ai/defaultai.cc
hessenfarmer: ooops.  good catch

tibor95:
of course not in this branch..
Will you want some training before merging?
I am still thinking about changes, but I am going to approve it I think.

tibor95 (approved):
I approve. In the future I would like to discuss some things, perhaps in a broader discussion among other contributors and players, but for now it looks fine to me

hessenfarmer:
> I approve. In the future I would like to discuss some things, perhaps in a broader discussion among other contributors and players, but for now it looks fine to me

in this case the Forum might be a good place, however I am afraid there are only a few people with our knowledge of the ai code.
thanks for the review

hessenfarmer:
After CI is green again I would merge. I would be glad though if you would find some time for more training and open a PR if you are ready so you can take your time for it.

tibor95:
Of course, but I will not be able to do training during weekend so in the beginning of next week.

hessenfarmer:
many thanks and as said no need to hurry

PR #5266, opened 2022-02-17
Fix Add-Ons Admin Menu Visibility

Noordfrees:
This fixes the bug that the visibility of the admin dropdown in the remote interaction window does not change when logging in or out.


Noordfrees:
This is a one-liner change and has been without review for a whole month ;)

klaus-halfmann:
> This is a one-liner change and has been without review for a whole month ;)

Can you tell me how I can become admin in this context ? :-)

klaus-halfmann (approved):
Code is ok for me , but I do not find that action menu?

Noordfrees:
The dropdown is visible only for logged-in users who have been granted admin access to the server interfaces (add-ons and metaserver/lobby), or if you have a full local add-ons server testing setup. That is only a very small number of users :) 
grafik

PR #5267, opened 2022-02-18
Show port space hint in buildhelp

### LINKED ISSUE #1475: Port space discovery problem
Port spaces are displayed in build help -- if a port can be built there. Otherwise, they are not shown.

An unlucky player might build something else to his port space or its proximity, and then never discover the port space at all. This could be fixed by hinting the existence of a currently unusable port space in the build help. 

In the attached screen shots, left plots illustrate the problem. Upper right plot shows a more typical "lucky" discovery. Lower right plot shows a possible fix.

If it is an intentional choice that port spaces may remain undiscovered, then this of course is not a bug at all.

 * date created: 2013-10-03T13:41:36Z
 * owner: kxq
 * assignee: gunchleoc
 * the launchpad url was https://bugs.launchpad.net/bugs/1234725

akupar:
Fixes #1475.

Shows the port space icon semi-transparent in buildhelp under the other buildhelp icon if a port could be build there, but current situation doesn't allow it as discussed in #1475. It looks a bit confusing when there is no other icon, so this might need some more work.
trees and flag
flag
nothing2
Kuvakaappaus - 2022-02-17 22-10-15


tothxa:
I think it's all right, not confusing at all.

frankystone:
What might be confusing, by only looking at the last picture: One might think he can build a port there...
But in general i think this is nice :+1: 

Maybe do not show the port space hint if roads (or other 'buildable' things) are on that location? Or/and make it more transparent?

@hessenfarmer ?

I have to test this branch, tomorrow.

frankystone:
I've played this branch a bit. I think this should be prevented:
port_space_02
Port spaces outside own territory shouldn't be visible. This gives a player too much information, imho. In the above screenshot only the port space inside own territory should be shown. I think this is also needed for expeditions.

Indeed this can be confusing in a game. This screenshots shows the same situation. Looking at the screenshots it might be clear that no port can be build in the bottom picture, but during playing this wasn't clear to me yet. Possible that i am just unfamiliar with this new feature and my opinion will change over time... 
port_space_03
Any ideas to make it more clear? Maybe make the port space icon more transparent in this case?

But as said maybe i am just too unfamiliar with this. Other opinions welcome!

Version: git25645[f20973c@1475-port-space-hint] (Debug)
Savegame (just rename the extension to wgf):
[port_spaces.zip](https://github.com/widelands/widelands/files/8101949/port_spaces.zip)

frankystone:
Another issue: The port space icon is shown below a built port (the port was set with a script, just to mention):
port_space_04

akupar:
> The port space icon is shown below a built port

I changed it so that it doen't show the icon if there is a port or port construction site, but it still shows the icon if there is any other building, immovable or a port demolition site. I think this is still an open question whether to show it over any buildings.

The code checks if the fields map_object_desc is Widelands::MapObjectType::WAREHOUSE (and not PORTDOCK) and I'm a little confused why it works now the way it does. So I need to look at this a bit more, if the icon should be shown over buildings.

akupar:
Another thing is, that sometimes when the port space is at the edge of player area, the icon doesn't show up even when there is space for big building. I think this might look like a bug to players.

port space before

port space after

I think to make it work the method find_portdock in map.cc needs some tweaking, but I didn't dare  to change it, because I'm not sure what other things it might affect.

frankystone:
> The code checks if the fields map_object_desc is Widelands::MapObjectType::WAREHOUSE (and not PORTDOCK) and I'm a little confused why it works now the way it does.

I guess this i fine, because a port is treated as a warehouse and the portdock is the invisible part where ships are landing. @Noordfrees ?

> Another thing is, that sometimes when the port space is at the edge of player area, the icon doesn't show up even when there is space for big building. I think this might look like a bug to players.
> 
> port space before
> 
> port space after

This is exactly one of the reasons for bug #1475 :-)  and why solving this is so important. The possibility to built a port needs some owned water tiles. In the first image there aren't enough owned water tiles. Having the semitransparent port space icon there would give a hint to the player that a port space is there, but currently the circumstances to really built a port aren't given.

akupar:
Sorry, I see now that I completely misunderstood the issue. Back to the drawing board!

src/wui/interactive_player.cc
```diff
@@ -486,6 +486,21 @@ void InteractivePlayer::draw(RenderTarget& dst) {
...
 
+inline bool can_show_portspace_hint(const Widelands::Player::Field& player_field) {
+        if (player_field.map_object_descr == nullptr) {
+                return true;
+        }
+
+        if (player_field.is_constructionsite &&
+            player_field.constructionsite->becomes->type() == Widelands::MapObjectType::WAREHOUSE) {
+                return false;
+        } else if (player_field.map_object_descr->type() == Widelands::MapObjectType::WAREHOUSE) {
```
hessenfarmer: this should be only an if as we return before it.

src/wui/interactive_player.cc
```diff
@@ -582,29 +597,45 @@ void InteractivePlayer::draw_map_view(MapView* given_map_view, RenderTarget* dst
...
+                                // Draw port space hint if a port could be built here, but current situation doesn't allow it.
+                                if (player_field.owner == plr.player_number() &&
+                                    (maxcaps & Widelands::BUILDCAPS_PORT) && !(caps & Widelands::BUILDCAPS_PORT) &&
+                                    can_show_portspace_hint(player_field)) {
+
+                                        if (const auto* overlay = get_buildhelp_overlay(maxcaps)) {
+                                                blit_field_overlay(dst, *f, overlay->pic, overlay->hotspot, scale * scaling,
+                                                                   0.4f * opacity);
+                                        }
+                                        if ((caps & Widelands::BUILDCAPS_SIZEMASK) == Widelands::BUILDCAPS_MEDIUM) {
```
hessenfarmer: shouldn't this be moved before the actual drswing of the overlay?

hessenfarmer:
> Sorry, I see now that I completely misunderstood the issue. Back to the drawing board!     
   
I think the problem lies within this line
https://github.com/akupar/widelands/blob/bd150975610727c06b309ac908a0d13ac4c96485/src/logic/map.cc#L1664
where the false parameter is not reflecting the maycaps concept. I believe instead of "false" we need to provide the inversion of consider_mobs here so the line should read: 
!find_portdock(f, !consider_mobs).empty()) {

this should return safely the maxcaps for ports

akupar:
> I think the problem lies within this line
https://github.com/akupar/widelands/blob/bd150975610727c06b309ac908a0d13ac4c96485/src/logic/map.cc#L1664
where the false parameter is not reflecting the maycaps concept. I believe instead of "false" we need to provide the inversion of consider_mobs here

Great! It looks like that does the trick. I tested it a little and it looks like its working now. 

border underobject undermedium

enemy border

I also hid the icon if it is on a road or a borderIt still shows up when there's a flag of a building, but no road, or an object. I'll see if I can fix it tomorrow.

buildings flag

src/wui/interactive_player.cc
akupar: That's there to draw the actual node caps scaled down if it is on top of the port icon. Although that scaling multiplier isn't really needed there.

Noordfrees:
Sorry for mentioning this so late, but in my opinion this overlay will always be a potential source of confusion. Players associate icon X with "I can build X here", so a port icon seems to indicate that they can build a port here, and not that a port could be built here given that surrounding objects were removed and/or more territory conquered. I think this overlay should therefore use a **different icon** than the standard port space buildhelp symbol to indicate that it is just a hint and not the real buildhelp; for example an ⚓ anchor symbol similar to [this old mockup](https://launchpadlibrarian.net/312147569/portspace_anchor.png).
Other than that this is great work and I am happy you're implementing this feature :)

hessenfarmer:
I had the same idea as @Noordfrees before. that would really be the cherry on the cake

tothxa:
> Sorry for mentioning this so late, but in my opinion this overlay will always be a potential source of confusion. Players associate icon X with "I can build X here", so a port icon seems to indicate that they can build a port here, and not that a port could be built here given that surrounding objects were removed and/or more territory conquered. I think this overlay should therefore use a **different icon**

I don't think it is confusing. Graying out something is a standard way to show that it is *currently* disabled.

It can only be confusing because of the way possible places for Amazon treetop sentries are marked. However, I think the treetop sentry is already more confusing, because it shows the grayed out maximum buildcap, so often has a big building overlay which makes it look like it shows that a big building could be built there, not that a special kind of small one is currently possible.

frankystone:
I also think there is no need for a special image. Using a special image means adding another link in the chain of association:

anchor shown -> what  does this anchor mean ->  there is a port space  
transparent port space icon shown -> there is a port space

But exchanging the icon is not a big problem. Before making a decision about the used image, please play this branch (if not already done). I played it and i think the transparency shows clearly what this image is about.

frankystone:
I think this is working well!

@akupar any reason why this is marked as draft yet?

akupar:
> @akupar any reason why this is marked as draft yet?

I was thinking of making the anchor version for comparison. I have only tested it in a single player single game. Other than that I think this is ready for review.

frankystone:
> I was thinking of making the anchor version for comparison.

That would be great :-)

akupar:
Sorry, took a while, but here is the anchor version. Currently the anchor is aligned so that the top left corner is in the middle of the icon, and centered if there is no icon. I think it looked better this way. 

I used the icon from the ship windows build port button, but colored it like the port icon. 

Kuvakaappaus - 2022-03-10 19-22-16

hessenfarmer:
I like the anchor design as well but i think the symbol shouldn't be scaled and always displayed in the same size

frankystone:
I am currently not on my dev-machine for testing. Can you provide some more images where the anchor is e.g. below trees or in situations like posted [here](https://github.com/widelands/widelands/pull/5267) ?

akupar:
> I like the anchor design as well but i think the symbol shouldn't be scaled and always displayed in the same size

I think it looks better scaled down when the anchor is added to another symbol (I thought of it as a sort of emblem on the original icon), but here's what it looks like when it's same size:

anchor same size

Normal size and always centered:
anchor same size centered


> I am currently not on my dev-machine for testing. Can you provide some more images where the anchor is e.g. below trees or in situations like posted [here](https://github.com/widelands/widelands/pull/5267) ?

It's not visible under trees anymore. In the early screenshots it is, but I had misunderstood the question. I don't think there's any other cases besides these shown in the most recent screenshots (anchor on midsize building, anchor on flag, and anchor alone)  where the anchor can occur.

frankystone:
Hm, a question to clarify: I assume in those images the portspace icon (not the anchors) will be shown in current master as well?
 
anchor same size

anchor same size centered

> It's not visible under trees anymore. 

That's not so good imho. But can be accepted to accomplish situations where a mapmaker intentionally placed trees (or rocks) to hide the port space from a player.

Are the anchors shown if e.g. not enough water has been conquered? Or if a building is too close to the port space, preventing placing a big building?

Sorry for my questions...

akupar:
> Hm, a question to clarify: I assume in those images the portspace icon (not the anchors) will be shown in current master as well?

Yes, the portspace icon and the yellow building and flags should be shown.
 
> > It's not visible under trees anymore.
> 
> That's not so good imho. But can be accepted to accomplish situations where a mapmaker intentionally placed trees (or rocks) to hide the port space from a player.

Yes, I think both would have their pros and cons. I think these needs other peoples opinions too. Here's what it would look like currently:
anchor on tree

> Are the anchors shown if e.g. not enough water has been conquered? 

I think so. I don't know how much and how far is enough. There's a problem currently that I can't start a new game, just load a saved game (in the master branch also) so it's hard to test now, but here are some situations I managed to produce (also an anchor on big building that didn't occur in my test map, this is from the Calvinsson map):
port no water
anchor on big

> Or if a building is too close to the port space, preventing placing a big building?

That's the case, for example, in those earlier images with the wood hardener.

> Sorry for my questions...

I'm happy to anwer any questions!

frankystone:
Wonderful :+1: 

Thanks for being so patient with me.

I think the icing on the cake would be to _not_ show the anchors if the port space icon is visible… just my opinion.

Please mark this PR as ready for review if you think this is good to go. And of course merge master.

Many thanks!

akupar:
> I think the icing on the cake would be to _not_ show the anchors if the port space icon is visible… just my opinion.

Most of these screenshots are from my test map, where I put as many portspaces as possible, and a little misleading. In most maps, in my experience, the portspaces are not this close to each other. (And from the players point of view it's kind of annoying if they were, because the ship would be stopping all the time.) So, I'm not sure if it's worth it.

frankystone:
> In most maps, in my experience, the portspaces are not this close to each other. 

Many maps provide more than one portspace close to each other. E.g. "The Nile", "Twin Lagoons", "Astoria 2.R", "Europa 1.2" …
The reason to place port spaces nearby on a map by a maps author was just the issue: Sometimes a port space was not found by a player because he placed buildings to close to the port space so it did not show up in the build help. See e.g. https://www.widelands.org/forum/topic/5278/  To solve this the maps author just placed more than one port space close to each other.

> (And from the players point of view it's kind of annoying if they were, because the ship would be stopping all the time.)

Yes, that's true. Maybe we should edit the official maps (the ones with many nearby port spaces) after this PR got merged.

Noordfrees (approved):
Code LGTM. Did not test the most recent commits but the screenshots look great :)

Noordfrees:
> > (And from the players point of view it's kind of annoying if they were, because the ship would be stopping all the time.)
> 
> Yes, that's true. Maybe we should edit the official maps (the ones with many nearby port spaces) after this PR got merged.

:+1: +1

PR #5278, opened 2022-02-22
results of AI training

tibor95:
Training after latest changes in the AI


tibor95:
I see there was one tiny change in other file, but it can stay as it is

hessenfarmer (approved):

PR #5281, opened 2022-02-22
Fix texthandling for dropdowns

### LINKED ISSUE #5193: Moving mouse away from dropdown does not clear filter
1. In the main menu, press <kbd>S</kbd> to open the single player dropdown.
2. Press <kbd>L</kbd> to filter for entries containing »L«.
3. Move the mouse away so the dropdown closes.
4. Press <kbd>S</kbd> again to reopen the dropdown.
5. Press <kbd>L</kbd> again. The dropdown now has no entries.

Version 1.1~git25583[8eafdfb@master] (Release)

jmoerschbach:
Optimize texthandling and general dropdown usage:

- ~~use close() instead of misusing toggle() when we actually want to close and not toggle~~
- manually disable sdl textinput handling when dropdown is closed, which 
- - fixes #5193,
- - fixes the glitch mentioned in https://github.com/widelands/widelands/pull/5261#issuecomment-1043323031
- - renders the former space handling workaround unnecessary

The problem is as follows:
Usually, we rely on https://github.com/widelands/widelands/blob/cec574d2a7c74d1fed641291efe09bd0d3306e8e/src/ui_basic/panel.cc#L1040 to enable/disable text handling. Dropdowns are special as they can still be focused but closed and then text input should be disabled. 
In future, we might think about an unfocus() callback so whenever an widget is unfocused we make sure textinput is disabled and the following widget which gets focus may or may not reenable text input again.
There is one more glitch which illustrates this need:

- in Mainmenu, press `m` to open multiplayer dropdown
- press `j` to join lobby. Notice: the chat editbox is focused, thus enabling text input
- press `ESC` -> editbox loses focus (without disabling textinput)
- press `ESC` to close lobby
- press `m` again -> it seems nothing happens but in fact dropdown is opened and immediately filtered with `m` which yields no result


Noordfrees:
Actually the intended purpose of using toggle() in those cases is to keep the dropdown *open* after clicking. Closing it would be the default behaviour anyway, but we want to keep the gamespeed/mapview/etc dropdowns open so you can repeatedly perform actions without having to reopen the dropdown every single time. This feature has been broken frequently in the past though.

jmoerschbach:
hmm, at least in official v1.0 this does not work, after clicking e.g. minimap from dropdown, the dropdown closes. same for changing gamespeed.

jmoerschbach:
ok, so I reverted the toggle/close change as it is not part of the fix anyway - but this should be revisited on another branch imho

jmoerschbach:
can this be reviewed, maybe @Noordfrees ? :)

Noordfrees (approved):

PR #5292, opened 2022-02-28
5277 remove ok on modal

frankystone:
This branch removes the Ok-button for `message_box` or `campaign_message_box` if `modal = false`.

As ever when digging into the source code i find some things which are a mystery to me. Two things related to this branch:
1. The game is first set to pause and after some rows of code the game is set to resume again (if modal = false). This is not a problem, imho, but i think it would be better to have this correlated code pieces nearby each other (not spread). I have left the order untouched though.
2. The call to `if (is_modal())` seems to return always `false`, at least in this state of the panel. I've changed that to use the new property `modal` instead. Not sure if my change may have any side effects. There are always so many things to think about …

I've also tried to find a place to darken the background, but get lost in the code. I think a nice solution would be to darken the background if `gamespeed=0` so the darkening will always set if the game is set to pause. No idea where to put such code to accomplish this.


frankystone:
A scenario for testing which uses both modal and non modal message_boxes:
[tutorial_05.zip](https://github.com/widelands/widelands/files/8154670/tutorial_05.zip)

Noordfrees:
Re modality: The window only becomes modal through the call to `mb->run()` which is triggered from lua_game.cc after the storybox has been constructed. In the storybox's constructor the window is therefore always non-modal. So this is an actual (previously undiscovered) bug you're fixing here.

The way you're adding or not adding the buttons box depending on modality looks dangerous. The OK button and the buttons_box are _always_ present, the call to `add` only informs the box layouting to layout this component properly. Without this call they'll simply be placed in some arbitrary location in the window (which seems to result in an effective height of 0 now but is fragile because it's actually undefined). Instead, always call `add` and use a call `button_box_.set_visible(modal)` to hide the box (and all its children) in the modal case.

Re darkening the background:
- -1 for always darkening the background when the speed is 0 as this would break pausing the game to get an unhurried overview of the map. (I usually set the speed to 0 for this as PageDown is more convenient to reach on my laptop keyboard.)
- Drawing a darkening overlay is not trivial as this is handled in low-level panel code (and because the overlay is currently meant to go above all windows, not between layers, and windows can't draw beyond their borders). This should better be implemented as a separate feature.

frankystone:
Thanks for the explanation :-)

Hopefully this should be fine now. 

Corrected scenario for testing, because the last one didn't worked:
[tutorial_05.zip](https://github.com/widelands/widelands/files/8159615/tutorial_05.zip)

Noordfrees:
LGTM, one glitch: That you can't close the non-modal story message boxes by right-clicking looks like a bug; only the modal ones are supposed to be right-click-protected but the non-modal ones look like they should allow right-click-to-close. So please add a check for `is_modal()` to the `if` in `StoryMessageBox::handle_mousepress`.

---

Regarding the scenario, some boxes vanished before I could read them completely (I ran the scenario at 11× speed); and after you placed the long road I was admonished for opening a window and then it crashed:
```
window:	field_action	Window: 0x7fe3bc3462e0	field_action
[00:01:34.298 real] INFO: Closing window '(null)'
Speicherzugriffsfehler (Speicherabzug geschrieben)
```

frankystone:
The scenario is far away from being ready… just provided it to have a scenario with mixed message boxes.

It _should_ now be allowed to close a non-modal message box with a right click.

Thanks for the hint :-)

Noordfrees (approved):

Issue #5264, opened 2022-02-17
Ships are not shown on minimap


frankystone:
Whereas roads, buildings and other things are shown immediately on the minimap if they change, ships aren't shown on the minimap at all.

![missing_ships](https://user-images.githubusercontent.com/6730556/154495117-0c776f85-ac33-4132-b31a-a982d5abedd8.jpg)

It would be very helpful if sailing ships are shown on the minimap.

**Version:**
 - OS: archlinux
 - Widelands Version: 1.1~git25639[6e1c918@master] (Release)

jmoerschbach:
Should there also a new map layer for that? I believe ships do not fit in any of the existing buttons/layers.

PR #5293, opened 2022-03-01
use UI::Box in Minimap

jmoerschbach:
This is a preparation/refactoring for fixing #5264 where I plan to add an additional layer (ships) and thus, needing an another button.
Notice: I removed this double-button-row/120px width logic as I could not trigger this code at all (in lowest solution 800x600 and smallest map available 64x64).
If somebody deems this still relevant/important please let me know how to trigger this code


Noordfrees:
The two-rows layout is needed when the map is much higher than wide:
grafik

PR #5297, opened 2022-03-10
Ships on minimap

### LINKED ISSUE #5264: Ships are not shown on minimap

jmoerschbach:
Closes #5264
If the reusing of the ship png is not recommended, maybe someone with graphical skills can create a fitting icon?
I reordered the buttons of the minimap for maps which have a (way) larger height than width because an odd number of buttons is a little weird.


src/wui/minimap.cc
```diff
@@ -290,7 +313,7 @@ bool MiniMap::handle_mousewheel(int32_t x, int32_t y, uint16_t modstate) {
 	return true;
 }
 
-constexpr uint16_t kCurrentPacketVersion = 1;
+constexpr uint16_t kCurrentPacketVersion = 2;
```
Noordfrees: Not needed if you don't actually check against the version number in the loading code. Please revert.

src/graphic/minimap_renderer.cc
```diff
@@ -78,6 +78,13 @@ inline RGBColor calc_minimap_color(const Widelands::EditorGameBase& egbase,
 			}
 		}
 
+		if (layers & MiniMapLayer::Ship && f.field->get_first_bob()) {
+			const Widelands::MapObjectType type = f.field->get_first_bob()->descr().type();
+			if (type == Widelands::MapObjectType::SHIP) {
```
Noordfrees:
```suggestion
		if ((layers & MiniMapLayer::Ship) && f.field->get_first_bob()) {
			if (f.field->get_first_bob()->descr().type() == Widelands::MapObjectType::SHIP) {
```
Two suggestions, one for readability, one no need for a local variable.

---

In the case that a ship crosses paths with a duck or ferry, checking only the first bob here might cause the ship to briefly vanish from the minimap. Better iterate all bobs and break when any ship is found.

src/wui/minimap.cc
```diff
@@ -196,9 +197,20 @@ MiniMap::MiniMap(InteractiveBase& ibase, Registry* const registry)
                   _("Buildings"),
                   UI::Button::VisualState::kRaised,
                   UI::Button::ImageMode::kUnscaled),
+     button_ships(this,
+                  "ships",
+                  but_w() * 2,
+                  view_.get_h() + but_h() * 1,
```
Noordfrees:
```suggestion
                  view_.get_h() + but_h(),
```

src/wui/minimap.cc
```diff
@@ -242,6 +255,7 @@ void MiniMap::resize() {
 	view_.set_zoom(*view_.minimap_layers_ & MiniMapLayer::Zoom2);
 	// Read number of rows after the zoom.
 	const uint32_t rows = number_of_button_rows();
+	const uint32_t height_offset = rows == 3 ? 1 : 0;
```
Noordfrees:
```suggestion
	const uint32_t height_offset = (rows == 3 ? 1 : 0);
```

Noordfrees (commented):
A few comments inline, will test this next

Noordfrees:
Tested and works fine. However I think the button should only be visible if the map actually supports seafaring, just as the seafaring stats menu is also unavailable on non-seafaring maps.

Noordfrees (approved):
LGTM :)

jmoerschbach:
thx for the review :)

PR #5330, opened 2022-03-25
reformat version string (remove markdown format, fixes #5136)

### LINKED ISSUE #5136: Widelands Version String Generates Erratic Markdown
Development versions of Widelands have the format
```
versionnumber ~ revision [ commit @ branch ] ( buildtype )
```
e.g.
```
1.1~git25529[fe7992d@master] (Release)
```
Unfortunately, when posting something containing `[hello](world)` in a Markdown-interpreted form (e.g. a GitHub issue or the forum turns this into a link to `world` displayed as `hello`. This makes the pasted version string much harder to read, and on GitHub it additionally prevents GitHub from automatically linkifying the commit hash (as in: fe7992d). This does not happen if there's a whitespace between `]` and `(`, but the version string is inconsistently displayed sometimes with and sometimes without this whitespace.

I propose to change all occurrences of this version string formatting to use different characters that are never interpreted by Markdown even if some whitespace is added or removed, for example by replacing the `()` with `[]` or vice versa or using `{}/$%+#`. While doing this it should also be made consistent everywhere.

Technically, only the part `1.1~git25529` is the actual version number (used for version comparison in packages and releases); everything after this is just extra formatting for display which we can modify freely.

andy5995:


utils/detect_revision.py
```diff
@@ -66,7 +66,7 @@ def detect_git_revision():
         git_abbrev = stdout.rstrip()
         if git_count and git_revnum and git_abbrev:
             # 1.0~git34567[abc0123@master]
```
matthiakl:
```suggestion
            # 1.0~git34567 (abc0123@master)
```

src/wlapplication_messages.cc
```diff
@@ -253,7 +253,7 @@ void show_usage(const std::string& build_id,
 	std::cout << std::string(kIndent + kTextWidth, '=')
 	          << std::endl
 	          /** TRANSLATORS: %s = version information */
-	          << format(_("This is Widelands Version %s"), format("%s(%s)", build_id, build_type))
+	          << format(_("This is Widelands Version %s"), format("%s %s", build_id, build_type))
```
matthiakl: There are more occurrences of build strings. Search for `build_type()` to find them. If we remove the parentheses here, it should be uniform in all places.

matthiakl (commented):
There are still parentheses in
- `editorinteractive.cc` line 995
- `profile.cc` line 823
- `replay.cc` line 91
- `about.cc` line 71,72
- `main.cc` line 477
- `wlapplication.cc` line 1622, 1638

Test resulting markdown: 
This is Widelands Version 1.1~git25728 (8562977@review2) Debug ✔

andy5995:
Provided the tests pass, this is ready for re-review

Rebased at https://github.com/widelands/widelands/commit/518bb834e19d806ac6c5e2025fe52060e108b281

src/ui_fsmenu/main.cc
```diff
@@ -474,7 +474,7 @@ void MainMenu::set_labels() {
 
 	version_.set_text(
 	   /** TRANSLATORS: %1$s = version string, %2%s = "Debug" or "Release" */
-	   format(_("Version %1$s (%2$s)"), build_id(), build_type()));
+	   format(_("Version %1$s"), build_ver_details()));
```
matthiakl: Please update the translator hint.

src/ui_fsmenu/about.cc
```diff
@@ -68,7 +68,7 @@ About::About(MainMenu& fsmm, UI::UniqueWindow::Registry& r)
 			std::string label, localized_label, value, localized_value;
 		};
 		const std::vector<ContentT> content = {
-		   {"Version:", _("Version:"), format("%1$s (%2$s)", build_id(), build_type()),
+		   {"Version:", _("Version:"), format("%1$s", build_ver_details()),
 		    format(_("%1$s (%2$s)"), build_id(), build_type())},
```
matthiakl: You missed the one in line 72.

matthiakl (commented):
Almost :)
Good idea to use a constant for the combined version info.

src/ui_fsmenu/about.cc
Noordfrees: Additionally, line 71 doesn't need the `format()` call now as this statement simply copies the argument.

src/ui_fsmenu/main.cc
andy5995: What should it say? I don't even understand why it's there. To me, it looks like the only string that can be translated there is "Version".

src/ui_fsmenu/about.cc
andy5995: So should this be

```
		const std::vector<ContentT> content = {
		   {"Version:", _("Version:"), format("%1$s", build_ver_details())},
		   {"Operating System:", _("Operating System:"),
```

matthiakl: You still need 4 parameters there. I guess it should be:
```
 {"Version:", _("Version:"), build_ver_details(), format(_("%1$s"), build_ver_details())},
```

src/ui_fsmenu/main.cc
matthiakl: The comment is to give translators a hint what is semantically placed in the placeholder. Some languages might want to change the order ("%1$s Version ") or something different, so that's why there is this formatting.
```
/** TRANSLATORS: %1$s = build version and type */
```

matthiakl: The comment is to give translators a hint what is semantically placed in the placeholder. Some languages might want to change the order ("%1$s Version ") or something different, so that's why there is this formatting.
```
/** TRANSLATORS: %1$s = build version and type */

File: src/ui_fsmenu/about.cc (hunk 4 above)
Comment (matthiakl) [reply src/ui_fsmenu/about.cc:72]: You still need 4 parameters there. I guess it should be:
```
 {"Version:", _("Version:"), build_ver_details(), format(_("%1$s"), build_ver_details())},
```

File: src/ui_fsmenu/about.cc [hunk 5]
```diff
@@ -68,8 +68,7 @@ About::About(MainMenu& fsmm, UI::UniqueWindow::Registry& r)
 			std::string label, localized_label, value, localized_value;
 		};
 		const std::vector<ContentT> content = {
-		   {"Version:", _("Version:"), format("%1$s (%2$s)", build_id(), build_type()),
-		    format(_("%1$s (%2$s)"), build_id(), build_type())},
+		   {"Version:", _("Version:"), build_ver_details(), format(_("%1$s"), build_ver_details())},
```
Noordfrees: The second `format` call is also redundant, since there is no reason why anyone would translate this as something different. You can leave the last argument empty if no custom localization is needed.
```suggestion
		   {"Version:", _("Version:"), build_ver_details(), ""},
```

PR #5331, opened 2022-03-25
CMakeLists.txt:output widelands exe to source root directory

andy5995:


andy5995:
Does the appveyor file need changing? I couldn't make that determination...

tothxa:
What if I want to keep around more than one builds? For example, and I think the most common use case would be a release and a debug build.

Noordfrees:
Actually I'm not sure about this. The whole point of having a build directory is that all generated files are placed inside it. In some cases as with Configured files it's necessary to violate this rule, but the final binary should IMHO be in the build dir, and only the convenience script should move it elsewhere.

andy5995:
> What if I want to keep around more than one builds? For example, and I think the most common use case would be a release and a debug build.

Good point. I'd play with the cmake a little to solve that.

But it's likely this PR won't get merged anyway.

andy5995:
> Actually I'm not sure about this. The whole point of having a build directory is that all generated files are placed inside it. In some cases as with Configured files it's necessary to violate this rule, but the final binary should IMHO be in the build dir, and only the convenience script should move it elsewhere.

@Noordfrees My point of view was because the docs say to put the binary in the source root dir, and the script puts it there, might as well just have it output there when it's built. And if there's actually some problem later on, then it can just be reverted.

andy5995:
> Actually I'm not sure about this.

If the final decision is that it's not wanted, just close the PR and I'll know why. ;) If you decide you wanna try it, I'll of course fix the errors showing on the CI...

Issue #4365, opened 2020-10-15
Basic tutorial improvement suggestions


gunchleoc:
Ideas to unclutter the first tutorial

https://www.widelands.org/forum/post/35041/

![1](https://user-images.githubusercontent.com/4095570/96083846-e774e800-0eb5-11eb-834b-1a434ca55f96.png)
![2](https://user-images.githubusercontent.com/4095570/96083850-e8a61500-0eb5-11eb-88b5-96ec05a63b48.png)
![3](https://user-images.githubusercontent.com/4095570/96083852-e93eab80-0eb5-11eb-9ffb-ca6586e627f1.png)
![4](https://user-images.githubusercontent.com/4095570/96083853-e9d74200-0eb5-11eb-934e-1d69fefaf93b.png)
![5](https://user-images.githubusercontent.com/4095570/96083854-ea6fd880-0eb5-11eb-94e6-64d411c58569.png)

gunchleoc:
Tweaking the texts is a good first issue. Implementing the arrow is a bit more advanced and needs C++ changes.

frankystone:
May i find time (and motivation) to work on this. It will take some time though...

andy5995:
And what do you think about replacing some the text like

>There are five different tribes in Widelands: the Barbarians, the Empire, the Atlanteans, the Frisians and the Amazons. All tribes have different economies, strengths, and weaknesses, but the general gameplay is the same for all. We will play the Barbarians for now

with something like:

>For this tutorial, we will play the Barbarians tribe.

My thinking is that the tutorial should focus on game mechanics, rather than include info that is already in the general game documentation (to make the tutorial easier to maintain).

hessenfarmer:
Well the Wiki is not well maintained. and the online encyclopedia lacks some coding changes to be in line with the ingame encyclopedia. 
Reason is that we decided to maintain only in game help and tutorials completely as this is the first hand information for new players.
So I am against deleting too much information. We already deleted some clutter while keeping all relevant information

tothxa:
I've just realised that this issue is based on one user's comments, which also spawned the training wheels effort. I actually liked the verbose tutorials when I first tried Widelands.

I think we should close this, as superseded by #4396. The basic tutorial (as long as training wheels don't replace it ;) needs different modifications.

PR #5338, opened 2022-03-30
basic control tutorial:only show first object once

### LINKED ISSUE #5336: In basic control tutorial, first objective appears twice

**Describe the bug**
In the basic control tutorial, the first objective appears two times, which is a bit confusing. It's mentioned in #4365, but it's not exactly clutter: it's duplicated information. The desired behavior would be for the welcome message->click ok->first objective.

**Screenshots**
Note the first window, and the text at the bottom:
image

After clicking "ok", the second window:

image

**Version:**
 - Widelands Version: 1.1~git50/77be5a6@master/(Debug)

**Additional context**
I spent some time trying to figure out how to fix it, but I could not. This bug is also in v1.0

andy5995:
(fixes #5336)(related to #4365)

image

image

image

### DIFF
--- a/data/campaigns/tutorial01_basic_control.wmf/scripting/texts.lua
+++ b/data/campaigns/tutorial01_basic_control.wmf/scripting/texts.lua
@@ -39,7 +39,8 @@ obj_initial_close_objectives_window = {
    body = objective_text(_("Handling of the ‘Objectives’ window"),
       li_image("images/wui/menus/objectives.png", _([[You can open and close the ‘Objectives’ window by clicking on the ‘Objectives’ button in the toolbar on the bottom of the screen.]])) ..
       li_arrow(_([[Like most other windows, you can also close the window by right-clicking on it.]])) ..
-      li_arrow(_([[When you have accomplished an objective, it will disappear from the list.]]))
+      li_arrow(_([[When you have accomplished an objective, it will disappear from the list.]])) ..
+      li_arrow(_([[You may open the objectives window now.]]))
    )
 }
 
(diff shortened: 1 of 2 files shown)


andy5995:
Oh, you know what? I did this wrong. I used "window" in two different contexts. Figures that would happen when I'm tired...

What's closer to the desired behavior?

data/campaigns/tutorial01_basic_control.wmf/scripting/texts.lua
```diff
@@ -39,7 +39,8 @@ obj_initial_close_objectives_window = {
    body = objective_text(_("Handling of the ‘Objectives’ window"),
       li_image("images/wui/menus/objectives.png", _([[You can open and close the ‘Objectives’ window by clicking on the ‘Objectives’ button in the toolbar on the bottom of the screen.]])) ..
       li_arrow(_([[Like most other windows, you can also close the window by right-clicking on it.]])) ..
-      li_arrow(_([[When you have accomplished an objective, it will disappear from the list.]]))
+      li_arrow(_([[When you have accomplished an objective, it will disappear from the list.]])) ..
+      li_arrow(_([[You may open the objectives window now.]]))
```
hessenfarmer: you may leave out this item as it is automatically opened by  line 48 in the mission_thread

andy5995:
k, so I've reverted the first commit, just changed very little on the second one. Here's how it looks now. Problem is that when the objective button is clicked, or if the objective dialog is right-clicked, the tutorial proceeds to the building spaces.

When the objective button is clicked, the window doesn't open, but instead the tutorial proceeds, which may confuse the user.

image

image

data/campaigns/tutorial01_basic_control.wmf/scripting/texts.lua
andy5995: > you may leave out this item as it is automatically opened by line 48 in the mission_thread

Oh, *that's* what's happening... so... not really a bug then like I reported. A bit confusing but I get what's happening now.

hessenfarmer (approved):

PR #5339, opened 2022-03-30
gamespeed.png:replace analog clock with sundial

andy5995:
CC-BY 2.0 https://creativecommons.org/licenses/by/2.0
by John Carmichael
modified from the original at
https://www.flickr.com/photos/jlcarmichael/3486451724

I thought this would fit the WL theme better than the modern analog clock...

image


frankystone (approved):
It's not immediately clear that this image is related to time, but better as the analog clock. And the tooltip says enough :-)

tothxa:
Could you please at least scale it down a little? It's too prominent for its function this way.

(But I think it's hard to see that it's a sundial, and the drawing style doesn't fit the cartoony style of the other icons. I also don't think that a sundial has any association with changing speed – if anything, I associate it with constant low speed.)

hessenfarmer:
> (But I think it's hard to see that it's a sundial, and the drawing style doesn't fit the cartoony style of the other icons. I also don't think that a sundial has any association with changing speed – if anything, I associate it with constant low speed.)

exactly my impression as well, so this would be hard to recognize and associate with time base manipulation. Old settlers had a running soldier icon for this as far as I remember. Maybe it might be possible to create some similar icon from our carrier graphics.

tothxa:
> Maybe it might be possible to create some similar icon from our carrier graphics.

Well, here's a try:
speed_carrier

In action:
shot0049

jmoerschbach:
> > Maybe it might be possible to create some similar icon from our carrier graphics.
> 
> Well, here's a try: speed_carrier
> 
> In action: shot0049

to be honest, this I associate even less with time/speed manipulation ;)

frankystone:
Maybe more like this:

gamespeed

The carrier has to be shown like running though …

hessenfarmer:
the longer i think the more I tend to an analog clock again. maybe we could make it look more widelands like

andy5995:
I'm not a graphic artist and I don't really have any other ideas about what to change it to.

0 A.D. uses this:

image

MegaGlest doesn't have any button or icon for game speed adjustment; it's just listed in their in-game hotkey docs.

hessenfarmer:
so in this case we should leave it like it is currently until someone can make a more widelands like graphic of an analog clock

PR #5342, opened 2022-03-31
codeql-analysis.yml:initial creation

andy5995:


andy5995:
Do you want this added?

Either the [errors it found](https://github.com/widelands/widelands/pull/5342/checks?check_run_id=5778976244) could be fixed in a separate PR, and afterward this branch could be rebased before merging.

Or maybe it would be better to do the fixes on this PR so codeql could test the changes before merging them.

Unless you're not worried about type conversion errors ;) Then just close the PR :)

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,59 @@
...
+        # If you wish to specify custom queries, you can do so here or in a config file.
+        # By default, queries listed here will override any specified in a config file.
+        # Prefix the list here with "+" to use these queries and those in the config file.
+        # queries: ./path/to/local/query, your-org/your-repo/queries@main
+
+    # ℹ️ Command-line programs to run using the OS shell.
+    # 📚 https://git.io/JvXDl
+
+    - run: |
+       cmake -B ./build OPTION_BUILD_TRANSLATIONS=OFF -DOPTION_BUILD_CODECHECK=OFF -DOPTION_BUILD_WEBSITE_TOOLS=OFF -DOPTION_BUILD_TESTS=OFF
```
Noordfrees:
```suggestion
       cmake -B ./build OPTION_BUILD_TRANSLATIONS=OFF -DOPTION_BUILD_CODECHECK=OFF -DOPTION_BUILD_WEBSITE_TOOLS=ON -DOPTION_BUILD_TESTS=ON
```
To also analyze tests and website tools

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,59 @@
...
+    # only run if files in src/ have been changed
+    paths:
+      - 'src/**'
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ master ]
+    paths:
+      - 'src/**'
+  schedule:
+    - cron: '0 2 12,26 * *'
```
Noordfrees: No need for scheduled builds IMHO

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,59 @@
...
+  push:
+    branches: [ master ]
+    # only run if files in src/ have been changed
+    paths:
+      - 'src/**'
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ master ]
+    paths:
+      - 'src/**'
```
Noordfrees: If we build it always and not just on `src` changes, we can require this check to be clean for all PRs before merging

Noordfrees (commented):
+1

Since the hits are all about narrow-type-wide-type mismatches it would IMHO be best to fix them all in this branch.

---

Since you're contributing quite a lot to Widelands, please also add yourself to the `data/txts/developers.json` :)

andy5995:
> Since you're contributing quite a lot to Widelands, please also add yourself to the `data/txts/developers.json` :)

Thanks!

.github/workflows/codeql-analysis.yml
andy5995: For now I've switched to using the autobuild option in the yml file, which should enable those by default. By enabling the autobuild option, it will also check the python code (as can be seen in the build log). I don't know how to have it also check the python code yet without using the autobuild option.

.github/workflows/codeql-analysis.yml
```diff
@@ -4,15 +4,13 @@ on:
   push:
     branches: [ master ]
     # only run if files in src/ have been changed
-    paths:
-      - 'src/**'
+    paths-ignore:
+    - '**/*.md'
```
andy5995: @Noordfrees do you think just ignoring md files would be ok?

src/ai/defaultai.cc
```diff
@@ -784,7 +784,7 @@ void DefaultAI::late_initialization() {
 			}
 
 			for (const auto& temp_position : prod.working_positions()) {
-				for (uint8_t i = 0; i < temp_position.second; i++) {
+				for (time_t i = 0; i < temp_position.second; i++) {
```
andy5995: What I did here is obviously wrong!

'second' appears to be the second member of a pair struct. I read a bit about pairs... it seems that the type is determined at compile time.

How do I declare 'i' correctly here since it's not known until compile-time?

hessenfarmer: second position of the pair is the amount of the worker. it is defined in logic/map_objects/tribes/productionsite.cc in line 56 as int, so declaring it as size_t might be good.     
  
However I deem the posible Multiplication overflows as mote critical as I had this case once in the statistics code calculating the percentage of productivity.

andy5995:
For anyone interested, docs for running codeql from the command line are at https://codeql.github.com/docs/codeql-cli/creating-codeql-databases/

.github/workflows/codeql-analysis.yml
Noordfrees: Hm, our current policy is to always run the full CI, though branches that can't possibly have an effect on the CI are sometimes merged without waiting for required checks. This also helps a bit with pointing out rare race conditions in the testsuite. So please remove all `paths-ignore` statements so the full CI is run regardless of whether it is needed. Other than than LGTM :)

src/ai/defaultai.cc
```diff
@@ -4996,13 +4997,13 @@ BuildingNecessity DefaultAI::check_building_necessity(BuildingObserver& bo,
 		if (!basic_economy_established) {
 			return BuildingNecessity::kForbidden;
 		}
-		const uint16_t min_roads_count =
-		   40 + std::abs(management_data.get_military_number_at(33)) / 2;
+		const int32_t min_roads_count = 40 + std::abs(management_data.get_military_number_at(33)) / 2;
```
hessenfarmer: why do we need to change this as the only variable in the calcualtion is abs(int_16t) which should be fine with uint_16

src/ai/defaultai.cc
```diff
@@ -4996,13 +4997,13 @@ BuildingNecessity DefaultAI::check_building_necessity(BuildingObserver& bo,
...
 		if (static_cast<int>(roads.size()) < min_roads_count * (1 + bo.total_count())) {
 			return BuildingNecessity::kForbidden;
 		}
-		bo.primary_priority += (roads.size() - min_roads_count * (1 + bo.total_count())) *
-		                       (2 + std::abs(management_data.get_military_number_at(143)) / 5);
+		bo.primary_priority +=
+		   (static_cast<int>(roads.size()) - min_roads_count * (1 + bo.total_count())) *
```
hessenfarmer, line 5060: what is the purpose of this cast? the target value is uint32 and the biggest value currently which defines the result is bo.total_count which is int32 so cast does not help imho

src/editor/ui_menus/tool_change_resources_options_menu.cc
```diff
@@ -88,7 +88,7 @@ EditorToolChangeResourcesOptionsMenu::EditorToolChangeResourcesOptionsMenu(
 	// Add resource buttons
 	resources_box_.add_inf_space();
 	const Widelands::Descriptions& descriptions = parent.egbase().descriptions();
-	for (Widelands::DescriptionIndex i = 0; i < descriptions.nr_resources(); ++i) {
+	for (size_t i = 0; i < descriptions.nr_resources(); ++i) {
```
hessenfarmer: we should not change widelands defined data types for the counter variable in for loops as we are using them in the body of the loop and the functions used are expecting them according to their definition. Furthermore using a widelands data type is increasing readabiltiy .
there are further instances where this comment applies

src/graphic/animation/spritesheet_animation.cc
```diff
@@ -131,8 +131,8 @@ SpriteSheetAnimation::SpriteSheetMipMapEntry::frame_textures(bool return_playerc
 	std::vector<std::unique_ptr<const Texture>> result;
 	const Rectf rect(Vector2f::zero(), width(), height());
 	if (!return_playercolor_masks || has_playercolor_masks) {
-		const size_t no_of_frames = rows * columns;
-		for (size_t i = 0; i < no_of_frames; ++i) {
+		const int no_of_frames = rows * columns;
```
hessenfarmer: as int is only in the best case as big as size_t this change is not understandable and it does not cure for the potential multiplication overflow.

src/graphic/gl/fields_to_draw.cc
```diff
@@ -111,8 +111,8 @@ void FieldsToDraw::reset(const Widelands::EditorGameBase& egbase,
 	assert(h_ > 0);
 
 	// Ensure that there is enough memory for the resize operation
-	size_t dimension = w_ * h_;
-	const size_t max_dimension = fields_.max_size();
+	uint32_t dimension = w_ * h_;
```
hessenfarmer: this will not solve the problem of the multiplication overflowing before getting casted to the target type

src/logic/game_settings.cc
```diff
@@ -36,7 +36,7 @@ Widelands::PlayerNumber GameSettings::find_shared(const PlayerSlot slot) const {
 		return result;
 	}
 	result = 1;
-	for (; result <= players.size(); ++result) {
+	for (; result <= static_cast<Widelands::PlayerNumber>(players.size()); ++result) {
```
hessenfarmer: well I believe thgis cast is misleading as the size of the players table is definitly not a PlayerNumber

src/wui/attack_window.cc
```diff
@@ -457,7 +457,7 @@ bool AttackWindow::handle_mousewheel(int32_t x, int32_t y, uint16_t modstate) {
 
 Widelands::Extent AttackWindow::ListOfSoldiers::size() const {
 	const size_t nr_soldiers = count_soldiers();
-	uint32_t rows = nr_soldiers / current_size_;
+	size_t rows = nr_soldiers / current_size_;
```
hessenfarmer, line 461: not sure what the purpose of this change is

hessenfarmer (changes requested):
some comments inline 
thios should be checked carefully by @Noordfrees as some cha ge slook dangerous in my eyes

hessenfarmer:
in the test suite the scan results of codeql is finished after 2 seconds while the check is still running. maybe there is a depends missing.

src/logic/game_settings.cc
andy5995: @hessenfarmer I used that cast because on L33, 'result' is declared as 'Widelands::PlayerNumber'; the codeql error was something about comparison between different types. So I took my best guess there. I figured 'result' should never be greater than the max of what the Widelands::PlayerNumber type would allow.

But I'm open to suggestions if there's a better fix.

andy5995:
> in the test suite the scan results of codeql is finished after 2 seconds while the check is still running. maybe there is a depends missing.

I'm a bit confused by that too. shortly after I push, I see

image

But there's also this:

image

src/editor/ui_menus/tool_change_resources_options_menu.cc
andy5995: I agree with you that using the WL data type would be preferable, but I didn't know of a better or more effective way of fixing the comparison error that codeql complained about.

The only other way I know to fix that would be to change how Widelands::DescriptionIndex was declared. I'm not familiar enough with the codebase to know whether or not that would be worth it. I assume there'd be a "cascade" effect where many other changes would then be needed related to how it's typed.

src/ai/defaultai.cc
andy5995: @hessenfarmer 'primary_priority' is declared as int32_t on L496 of ai_help_structs.h. So I think the error was because (if I interpreted things correctly) one of the operands is a unsigned long (roads.size()). So trying to multiply an int with an unsigned long could overflow the int32_t target.

Codeql docs have a bit of into about that: [Multiplication result converted to larger type](https://codeql.github.com/codeql-query-help/cpp/cpp-integer-multiplication-cast-to-long/)

Though like you I'm not sure if that's the best option. Maybe I'm wrong, but I think that 'roads.size()'  is ever  '>' than INT_MAX, casting would skew the value and so the multiplication result would be very wrong.

Wouldn't it be better to declare 'primary_priority' as 'uint32'?

andy5995:
I've opened a PR on my fork that adds the codeql analysis file but doesn't introduce any fixes. The results should be at https://github.com/andy5995/widelands/pull/2/checks?check_run_id=5844248996

It should soon show all original 52 errors codeql complained about.

andy5995:
Tests now failing. Why is this happening? https://github.com/widelands/widelands/runs/5844695311?check_suite_focus=true#step:5:1305

src/ai/defaultai.cc
hessenfarmer: changing the declaration of primary_priority would not help as the possible error happens before the value is assigned to it.    
the problem is that an integer multiplication results always in the data type of the biggest operand. The conversion (implicit cast to the target type is done in the asignment step after the multiplication. 
if code ql finds an instance where some narrow integers are multiplied and the result is a wider type that this is intentional to prevent overflow which it doesn't. this is explained in the example help text you provided. so we need to upcast one operand to a wider type to prevent such overflow.

src/editor/ui_menus/tool_change_resources_options_menu.cc
hessenfarmer: casting the comparison value to a narrower type might do the trick in this case as it would prevent the danger of an infinte loop which is the cause for this warning.

src/logic/game_settings.cc
hessenfarmer: well I think casting it to an integer type of similar size as Player number would be a solution without decreasing readability

src/editor/ui_menus/tool_change_resources_options_menu.cc
andy5995:
```suggestion
	for (Widelands::DescriptionIndex i = 0; i < static_cast<uint16_t>(descriptions.nr_resources()); ++i) {
```

@hessenfarmer @Noordfrees You want me to do that for the ~15 other instances where I changed DescriptionIndex to size_t ?

andy5995: See https://github.com/widelands/widelands/pull/5342#discussion_r845573861

Noordfrees: I'd suggest changing the definition of `nr_resources` and its siblings to return a `DescriptionIndex` instead of `size_t`. It can be guaranteed to fit in 16 bits.

hessenfarmer: well semantically it should deliver the number of existent description indexes which should be equivalent to the highest index but is not an Index itself. so we might change the return type of the nr_*indexedEntity*() to the same basic type  as DescriptionIndex, as it can't be higher then that, but not the DescriptionIndex as it is not the index but an amount.

andy5995: I started making some changes, they're shown in https://github.com/widelands/widelands/pull/5342/commits/1fffc688352009b2e20a824354f7c3417830685c

What do you think so far?

andy5995:
rebased at https://github.com/widelands/widelands/commit/b2addfc83a51e1f0d27a5a33831e8a439a077b20

src/editor/ui_menus/tool_change_resources_options_menu.cc
hessenfarmer: checked the changes in the commit you provided. I think this should be the way to go.

src/ui_fsmenu/multiplayersetupgroup.cc
```diff
@@ -125,7 +125,7 @@ struct MultiPlayerClientGroup : public UI::Box {
 		const UserSettings& user_setting = settings.users.at(id_);
 
 		slot_dropdown_.clear();
-		for (PlayerSlot slot = 0; slot < settings.players.size(); ++slot) {
```
andy5995, line 128: You'll want these changes for PlayerSlot and TeamNumber reverted, like I did with DescriptionIndex?

src/ai/defaultai.cc
hessenfarmer: I am stillnot convinced by this cast as it is effectively downcasting the value of roads size. furthermore it seems the bracckets are wrong changing the result of the calculation.

src/ui_fsmenu/multiplayersetupgroup.cc
hessenfarmer: yes please

hessenfarmer:
@andy5995 do you still want to work on this.
@Noordfrees from my perspective this could be postponed to 1.2 as it is no0t a feature but another codechecking / codeimprovement fix

andy5995:
> @andy5995 do you still want to work on this.

At this time, no.

hessenfarmer:
@andy5995 @Noordfrees 
maybe we should skip this and enable the relating checks of the majorities of results of codeql on our code. 
Namely: 
bugprone-implicit-widening-of-multiplication-result
bugprone-too-small-loop-variable

andy5995:
Sure, I'm not emotionally attached to this PR.

PR #5343, opened 2022-04-01
Add enhancement option to lumberjack huts (conversion to ranger's huts)

andy5995:
[skip ci]

Are you interested in a change like this? If so, I'll make the changes to the other tribes.

### DIFF
--- a/data/tribes/buildings/productionsites/barbarians/lumberjacks_hut/init.lua
+++ b/data/tribes/buildings/productionsites/barbarians/lumberjacks_hut/init.lua
@@ -16,6 +16,16 @@ wl.Descriptions():new_productionsite_type {
       log = 2
    },
 
+   enhancement = {
+      name = "barbarians_rangers_hut",
+      enhancement_cost = {
+         log = 3
+      },
+      enhancement_return_on_dismantle = {
+         log = 2
+      }
+   },
+
    animation_directory = dirname,
    animations = {
       idle = {


frankystone:
I think building enhancements should only be possible when the new building uses the same worker(s) as the old building. Otherwise we may end in the possibility to change every building to any other building (of the same size), e.g. change tower to micro brewery. This also may affects the balancing of all tribes.

Such things should be implemented as an add-on, imho. So a player can optionally enable it.

andy5995:
@frankystone ok, my thinking was that the "no trees" message suggesting converting the lumberjack's hut to a ranger's hut. So I thought this would make the process a little simpler and user-friendly.

Anyway, I'll go along with whatever you folks decide of course.. I just wanted to play around with the data a little ;)

Thanks for reviewing.

tothxa:
> my thinking was that the "no trees" message suggesting converting the lumberjack's hut to a ranger's hut. So I thought this would make the process a little simpler and user-friendly.

If you do so, you'll end up with lots of trees, but noone to cut them. I don't think this makes sense.

BTW the out of trees message says "dismantle ***OR*** build a ranger"…

> I just wanted to play around with the data a little ;)

As kaputtnik said, that's what add-ons are for.

andy5995:
> > my thinking was that the "no trees" message suggesting converting the lumberjack's hut to a ranger's hut. So I thought this would make the process a little simpler and user-friendly.
> 
> If you do so, you'll end up with lots of trees, but noone to cut them. I don't think this makes sense.

Isn't it possible there's already another lumberjack's hut nearby? Or couldn't you just build a lumberjack hut nearby after some trees were planted? Or later dismantle the ranger's hut and build a lumberjack's hut?

> 
> BTW the out of trees message says "dismantle _**OR**_ build a ranger"…

Thanks for the correction.

hessenfarmer:
Well for reasons of balancing, tribe character, and maintenance effort for tutorials I am strongly against changing our legacy tribes (up to and including frisians - amazons still need some balancing), other then for a very strong reason and concept proven by a popular addon. 
furthermore the suggested change wouldnt make much sense from a gameplay perspective as building a forester would cost only 1 additional log and leave the original cutter alive.
so -1 from my side for this change

PR #5350, opened 2022-04-04
Fix legacy scenario units load order

### LINKED ISSUE #5346: Friends in Need - cannot load game - Unknown worker 'frisians_diker'
Hi, I am part way through Friends in Need and cannot load my game after updating Widelands. The error message says, "Unknown worker 'frisians_diker'". 

**Steps to reproduce**
1. With any saved game from Friends in Need (even one saved at the very start of the level), saved before updating Widelands today 2nd April
2. Load the scenario
3. Game tries to load but fails with the error message in the attached screenshot.

**Expected behavior**
Game should be loaded

**Screenshots**
Screenshot from 2022-04-02 17-52-24

**Version:**
 - OS: Ubuntu 20.04.4 LTS
 - Widelands Version: 1.1~bzr10232-202204020331 (Release)

**Additional context**
Happy to provide the saved file if that is useful (although I can't find the saved game - no ~/.widelands directory - please advise where they go)

Noordfrees:
Fixes #5346

This only affects savegames using the legacy load order, new-style savegames do not use this loading path at all.

`postload_addons` is poorly named now, it does a lot of other magic as well, so we need to call it even if we don't use any add-ons. Don't have a better name though.

### DIFF
--- a/src/logic/game.cc
+++ b/src/logic/game.cc
@@ -238,6 +238,7 @@ void Game::check_legacy_addons_desync_magic() {
 		}
 	}
 	if (!needed) {
+		postload_addons(true);
 		return;
 	}
 


hessenfarmer (approved):
LGTM tested and working.

renaming it to postload units wouldn't be correct either I think so no glue about better name

Issue #1710, opened 2019-09-09
Binaries installed in odd directory under linux


bunnybot:
Revision 7224.1.5 removed WL_INSTALL_BINDIR, instead opting to use CMAKE_INSTALL_PREFIX. However, binary files by default are now installed into CMAKE_INSTALL_PREFIX directly. I believe this is expected for Windows and MacOS, but under Linux the expectation is that binary files go into CMAKE_INSTALL_PREFIX/bin. In particular, a relatively default install prefix is /usr, but would result in the widelands executable being placed as "/usr/widelands". One option might be something like

IF(ARGS_WIN32)
install(TARGETS ${NAME} DESTINATION "." COMPONENT ExecutableFiles)
ELSE(ARGS_WIN32)
install(TARGETS ${NAME} DESTINATION "bin/" COMPONENT ExecutableFiles)
ENDIF(ARGS_WIN32)

in cmake/WlFunctions.cmake. There are potentially other ways around this, another one I know is to have a "BINDIR" type variable that is set in some OS-specific setup function, but that's basically just bringing back WL_INSTALL_BINDIR, and I assume this was removed for some specific reason.

------------------------------------
Imported from Launchpad using lp2gh.

 * date created: 2018-08-09T22:40:29Z
 * owner: uilliam
 * the launchpad url was https://bugs.launchpad.net/bugs/1786353

bunnybot:
(by gunchleoc)
There are some additional settings in widelands.h, with a not for packagers to take care of it. The PPA will install the application into usr/share/applications and the data into usr/share/games/widelands

bunnybot:
(by uilliam)
There are extra settings in CMakeLists.txt with notes for packagers, I assume that's the file you meant and not widelands.h. These include WL_INSTALL_BASEDIR and WL_INSTALL_DATADIR. The DATADIR is respected properly as best I can tell, and WL_INSTALL_BASEDIR will set the location of the installation of files such as COPYING, CREDITS, ChangeLog as well as the doc folder for documentation. However, the binaries widelands, wl_map_info, wl_map_object_info and wl_render_richtext do not get installed in WL_INSTALL_BASEDIR. This is because the function "wl_binary", as defined in cmake/WlFunctions.cmake, makes no reference to WL_INSTALL_BASEDIR, and instead says "install into CMAKE_INSTALL_PREFIX".

Noordfrees:
Marking as urgent for v1.1 because an open PR is linked.

PR #5354, opened 2022-04-06
.github/workflows/appimage.yml:add appimage builder

andy5995:


andy5995:
I set it to build on Ubuntu 18.04 so the image would eventually be able to get listed on [AppImage Hub](https://github.com/AppImage/appimage.github.io#appimagehub-).

>Must run on the [oldest still-supported Ubuntu LTS release](https://www.ubuntu.com/info/release-end-of-life) (currently Ubuntu 18.04) without the installation of additional packages. Targeting the oldest still-supported LTS is to ensure that the AppImage will run not only on the very latest, but also on older target systems, such as enterprise distributions (not limited to Ubuntu)

I'm not really sure the best way to test this, and also I suspect you may wanted it implemented differently. I think it would be better if it could be combined with the build.yml file.

It would probably be more efficient if someone with admin privileges does this PR. It could be a little tricky to test.

andy5995:
I set up a workflow on https://github.com/andy5995/widelands-appimage and created a [Widelands appimage](https://github.com/andy5995/widelands-appimage/releases) for WL v1.0

It doesn't work yet because a path (or maybe more) are wrong. I assume related to https://github.com/widelands/widelands/issues/1710

```
[andy@prometheus Downloads]$ ./Widelands-59494d19-x86_64.AppImage 
This is Widelands Version 1.0 (Release)
[00:00:00.000 real] INFO: Set home directory: /home/andy/.local/share/widelands
[00:00:00.000 real] INFO: Set configuration file: /home/andy/.config/widelands/config
[00:00:00.003 real] INFO: Adding directory: /usr/share/games/widelands/data

Caught exception (of type '17FileNotFoundError') in outermost handler!
The exception said: FileSystem::create: could not find file or directory: /usr/share/games/widelands/data

This should not happen. Please file a bug report on version 1.0(Release).
```

andy5995:
> It doesn't work yet because a path (or maybe more) are wrong. I assume related to #1710
> 

I was wrong to assume such a thing. The data directory is in the correct place inside the appimage

`squashfs-root/usr/share/games/widelands/data`

(Which I found after using the `--appimage-extract` option)

andy5995:
> It doesn't work yet because a path (or maybe more) are wrong.

I found the problem: https://docs.appimage.org/reference/best-practices.html#ref-binaries-no-abs-paths

>Since an AppImage is mounted at a different location in the filesystem every time it is run, it is crucial not to use compiled in absolute paths.

So I made a little patch, tested locally with docker, and am running the workflow again.

In an hour or so, there should be a functional UNOFFICIAL Linux appimage for WL v1.0 at https://github.com/andy5995/widelands-appimage/releases

Unofficial because none of this has been approved or reviewed yet by the Widelands maintainers ;)

andy5995:
Closing in favor of https://github.com/widelands/widelands/pull/5377

PR #5358, opened 2022-04-09
CMakeLists.txt:add custom uninstall target

andy5995:
So far the only problem I see with this is that when uninstalling, empty directories are not removed.


klaus-halfmann:
Hello @andy5995 welcome to widelands, I compile widelands on macOS with an m1Processor / Apple Silicon.
I am not a big expert with CMake.
* What does this commit provide and why is it needed?
* How can I test this on macOS (currently cloning that repo)

andy5995:
hi @klaus-halfmann .. it provides an 'uninstall' target, so if a 'make install' is run, a 'make uninstall' will remove the files that were installed. (replace 'make' with 'ninja' if you're using that).

When an install is done, an 'install_manifest.txt' file is created in the build directory. when the uninstall target is used, each file from the install manifest is read and is deleted.

More info at https://gitlab.kitware.com/cmake/community/-/wikis/FAQ#can-i-do-make-uninstall-with-cmake

klaus-halfmann:
So you add a test for make install / uninstall to the Compile target in gitlhub but only for the  "gcc++-10" compiler.

I could not find the result in https://github.com/widelands/widelands/runs/5963907771?check_suite_focus=true
for (Release, ubuntu-20.04, g++-10) (or https://pipelines.actions.githubusercontent.com/serviceHosts/44802162-6b05-4a72-a3ed-aba8a8afd5e9/_apis/pipelines/1/runs/11056/signedlogcontent/54?urlExpires=2022-04-18T06%3A26%3A13.2494757Z&urlSigningMethod=HMACV1&urlSignature=zJ%2B2KCwbd4rtdb0l2KzxroyfaGrSeInmlsAYh1YMXc4%3D)

But maybe I am missing something?

The macOS way to build the App is found in utils/macos/build_app.sh and triggerd some other way.

I normally use the ./compile.sh script only so please guide me how to call `make install/uninstall`

klaus-halfmann:
```
build % ninja install           
[53/58] cd /Users/klaus/develop/widelands/cmake-add-uninstall/build && /opt/homebrew/Cellar/cmake/3.22.3/bin/ctest --output-on-failure
Test project /Users/klaus/develop/widelands/cmake-add-uninstall/build
    Start 1: test_ai
...
100% tests passed, 0 tests failed out of 6
---
[57/58] Install the project...
-- Install configuration: "Debug"
-- Installing: /usr/local/./VERSION  <- Not healthy?
CMake Error at cmake_install.cmake:42 (file):
  file INSTALL cannot copy file
  "/Users/klaus/develop/widelands/cmake-add-uninstall/build/VERSION" to
  "/usr/local/./VERSION": Permission denied.
```

As I am using homebrew (a kind of package manager for macOS) on apple silicon 
my 'usr/local' is actually '/opt/homebrew/' as the other path is reserved for emulated x86 code.

But maybe I cannot  test this on macOS at all?

andy5995:
> So you add a test for make install / uninstall to the Compile target in gitlhub but only for the "gcc++-10" compiler.

That will probably change. It doesn't need to test on all version of all compilers, so I just tried to specify testing on gcc 10.

> 
> I could not find the result in https://github.com/widelands/widelands/runs/5963907771?check_suite_focus=true for (Release, ubuntu-20.04, g++-10) (or https://pipelines.actions.githubusercontent.com/serviceHosts/44802162-6b05-4a72-a3ed-aba8a8afd5e9/_apis/pipelines/1/runs/11056/signedlogcontent/54?urlExpires=2022-04-18T06%3A26%3A13.2494757Z&urlSigningMethod=HMACV1&urlSignature=zJ%2B2KCwbd4rtdb0l2KzxroyfaGrSeInmlsAYh1YMXc4%3D)
> 
> But maybe I am missing something?

I couldn't find it either. I'm not sure what I have wrong so far, so I'm just waiting to have this PR reviewed before I continue. It should still work locally, I just don't have it set up right for testing in the CI.

klaus-halfmann (approved):
The code looks fine and all the checks passed, so this can go in.
Compiles and starts for me localy.
install does not work on macOS though

No idea however who needs this?

andy5995:
> -- Installing: /usr/local/./VERSION  <- Not healthy?
> CMake Error at cmake_install.cmake:42 (file):
>   file INSTALL cannot copy file
>   "/Users/klaus/develop/widelands/cmake-add-uninstall/build/VERSION" to
>   "/usr/local/./VERSION": Permission denied.
> ```
> 
> As I am using homebrew (a kind of package manager for macOS) on apple silicon my 'usr/local' is actually '/opt/homebrew/' as the other path is reserved for emulated x86 code.
> 
> But maybe I cannot test this on macOS at all?

@klaus-halfmann Looks like you are definitely on the right track. To test installing without root or superuser privileges, you can do

`DESTDIR=$PWD/test_install ninja install`

and afterward, you must specify the same DESTDIR when uninstalling

`DESTDIR=$PWD/test_install ninja uninstall`

'DESTDIR' can be any directory you have write access to.

If using 'make', I think you have to put DESTDIR *after* 'make', i.e. `make DESTDIR=$PWD/test_install install`

andy5995:
> 
> No idea however who needs this?

Anyone who uses `make/ninja install`... That's most developers or people who are cloning the repo, just testing things a bit. But typically as you know, install doesn't need to be used to run the game. But I feel if there's an 'install' target, there should be an 'uninstall' target so one might clean up if he or she uses the install.

klaus-halfmann:
```
cd build
DESTDIR=~/develop/widelands/testinsall ninja install
DESTDIR=~/develop/widelands/testinsall ninja uninstall
```
Works but uses a strange structure directly below `testinsall/usr/local/./widelands` but this should be 
checked by some Linux master.

.github/workflows/build.yaml
```diff
@@ -187,6 +187,12 @@ jobs:
         pushd build
         cmake .. -DCMAKE_BUILD_TYPE:STRING="${{ matrix.config }}" -DOPTION_BUILD_TRANSLATIONS="ON" -DOPTION_BUILD_WEBSITE_TOOLS="ON" -DOPTION_ASAN="ON" -DOPTION_BUILD_CODECHECK="OFF"
         make -k -j$(nproc) VERBOSE=1
+        # We don't need to test the install/uninstall for each job so we're just
+        # arbitrarily selecting gcc-10
+        if [ "$CXX" = "gcc++-10" ]; then
+          make DESTDIR=$PWD/test_install install
+          make DESTDIR=$PWD/test_install uninstall
```
matthiakl: Is there a reason to not install in all runs? They run in parallel so the overall time should be the same.
I would prefer to specify the WL_INSTALL_BASEDIR, WL_INSTALL_DATADIR and CMAKE_INSTALL_PREFIX during compilation (see comment in top level CMakeLists.txt:76).
This resolves the weird looking `/usr/local/./*` paths and is more controllable.

xdg/CMakeLists.txt
```diff
@@ -15,6 +15,9 @@ install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/${XDG_APPLICATION_ID}.desktop     DEST
 #install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/${XDG_APPLICATION_ID}.appdata.xml DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/metainfo)
 install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/${XDG_APPLICATION_ID}.appdata.xml DESTINATION ../share/metainfo)
 
+# TODO: check for privileges
+# If 'make install' was run without superuser privileges, the user will get a "permission denied" error
+# when 'gtk-update-icon-cache' is executed
```
matthiakl, line 20: This is no longer a problem when using CMAKE_INSTALL_PREFIX.

matthiakl (commented):
(A "Linux master" here :wink: ) 
Instead of DESTDIR there are three possible environment variables to fine tune the installation:
- CMAKE_INSTALL_PREFIX: used for global system files (e.g. icons, binaries). Defaults to /usr/local
- WL_INSTALL_BASEDIR: used for metadata files (version, changelog, license, credits)
- WL_INSTALL_DATADIR: used for game data files

When using these variables during compilation the install and uninstall process works fine as normal user. No permission errors.

For removing empty directories you might try this: https://github.com/xbmc/xbmc/commit/4c8efe4fc7e6bdb759c11a8e1dc6f2422e79bbaa

.github/workflows/build.yaml
andy5995: > 
> For removing empty directories you might try this: [xbmc/xbmc@4c8efe4](https://github.com/xbmc/xbmc/commit/4c8efe4fc7e6bdb759c11a8e1dc6f2422e79bbaa)

@matthiakl thanks for the suggestion. I tried implementing that in my most recent commit but when testing locally, the directories aren't removed and there's no output messages that any attempts to remove the empty directories are being made. I'm not sure why yet.

matthiakl: Our setup seems to be a bit different. The contents of our manifest already contain the `DESTDIR` so no need to prepend it again. Additionally your list operations are a bit different.
This is what worked for me: [rmdir.patch.txt](https://github.com/widelands/widelands/files/8614343/rmdir.patch.txt)

Regarding the CI how do you make sure that the install / uninstall steps have the expected result?

andy5995: > Our setup seems to be a bit different. The contents of our manifest already contain the `DESTDIR` so no need to prepend it again. Additionally your list operations are a bit different. This is what worked for me: [rmdir.patch.txt](https://github.com/widelands/widelands/files/8614343/rmdir.patch.txt)

Thanks @matthiakl 

> Regarding the CI how do you make sure that the install / uninstall steps have the expected result?

I think [this commit](https://github.com/widelands/widelands/pull/5358/commits/9b5b780b75d3748cc8187f9b7207c2bf295645de) should be sufficient. Any problems with the install or uninstall, cmake should return nonzero and the workflow should terminate. If the top-level DESTDIR dir is removed by the uninstall, that should provide enough of a test to verify uninstall was successful.

matthiakl (commented):
This build log output looks suspicious:

>  -- Removing empty dir: {/home/runner/work/widelands/widelands/build/test_install}/usr
-- Removing empty dir: {/home/runner/work/widelands/widelands/build/test_install}
-- Removing empty dir: {/home/runner/work/widelands/widelands/build
-- Removing empty dir: {/home/runner/work/widelands/widelands
-- Removing empty dir: {/home/runner/work/widelands
-- Removing empty dir: {/home/runner/work
-- Removing empty dir: {/home/runner
-- Removing empty dir: {/home
-- Removing empty dir: {

Are these directories removed all the way up to /home? But then the testsuite should fail. Do you get a similar output locally?
Otherwise it looks good now.

.github/workflows/build.yaml
```diff
@@ -187,11 +187,13 @@ jobs:
         pushd build
         cmake .. -DCMAKE_BUILD_TYPE:STRING="${{ matrix.config }}" -DOPTION_BUILD_TRANSLATIONS="ON" -DOPTION_BUILD_WEBSITE_TOOLS="ON" -DOPTION_ASAN="ON" -DOPTION_BUILD_CODECHECK="OFF"
         make -k -j$(nproc) VERBOSE=1
-        TEST_INSTALL_DIR="{$PWD/test_install}"
+    - name: Make install/uninstall
+      run: |
+        cd build
+        TEST_INSTALL_DIR="${PWD}/test_install}"
```
andy5995: @matthiakl I'm pretty sure the unwanted removals were due to my poor brace placement.

matthiakl: Indeed, now it looks better, but you missed one:
```suggestion
        TEST_INSTALL_DIR="${PWD}/test_install"
```

matthiakl (approved):

xdg/CMakeLists.txt
hessenfarmer: well while trying to understand the ci failure I saw in the logs it is still raising the error

matthiakl: Yes, because the CI does not use a user writable CMAKE_INSTALL_PREFIX, but the default one in /usr/local. This would require to specify `-DCMAKE_INSTALL_PREFIX=/some/user/writable/dir` during cmake configuration.
Another option is to use a custom icon cache in DESTDIR by using `\$ENV{DESTDIR}${CMAKE_INSTALL_PREFIX}/share/icons/hicolor` in the command below. But I am not sure, what the correct behavior should be here. Maybe elevating the current output to a more prominent warning without failing is enough?

PR #5359, opened 2022-04-09
CMakeLists.txt:put binaries in correct locations when installing

### LINKED ISSUE #1710: Binaries installed in odd directory under linux

andy5995:
(closes #1710)


matthiakl (commented):
Please check xdg/CMakeLists.txt and replace the relative `..` paths there with the now correct absolute CMAKE_INSTALL_PREFIX.

andy5995:
> Please check xdg/CMakeLists.txt and replace the relative `..` paths there with the now correct absolute CMAKE_INSTALL_PREFIX.

What about here? https://github.com/widelands/widelands/blob/0621ce2fe1a3825ae6669a424a700de5312fb84f/xdg/CMakeLists.txt#L21 ?

Does the `..` just need to be removed?

matthiakl:
> Does the .. just need to be removed?

Yes, this results from the strange expectation, that CMAKE_INSTALL_PREFIX points to /usr/bin instead of /usr

CMakeLists.txt
```diff
@@ -554,6 +554,13 @@ install(
   PATTERN "CMakeLists.txt" EXCLUDE
 )
 
+# To override this, use '-DWL_INSTALL_BINDIR=<absolute path>' when configuring or change
+# 'CMAKE_INSTALL_PREFIX'.
+# Example: to have the bin installed to "/usr/games", just use '-DCMAKE_INSTALL_PREFIX=/usr'
+# and don't specify '-DWL_INSTALL_BINDIR' when configuring.
+# Note: 'CMAKE_INSTALL_PREFIX' defaults to /usr/local on UNIX and c:/Program Files/${PROJECT_NAME} on Windows
+set(WL_INSTALL_BINDIR "${CMAKE_INSTALL_PREFIX}/games" CACHE PATH "Destination of executable files (if installing)")
```
andy5995, line 563: Does anyone think we need a condition  here to specify a different default directory if installing on Windows? If I read the cmake docs right, the binaries would get installed to `c:/Program Files/widelands/games/`

matthiakl (approved):
Looks good now.
@Noordfrees, we need to inform the package maintainers about this change, so they can update their paths configuration.

CMakeLists.txt
matthiakl: Windows builds do not get installed by `make` but instead use the innosetup, mac os also has a separate install method. This should be unaffected by this change. However to be sure I will test the Windows installer and maybe @klaus-halfmann can check on MacOS (go to "Checks" -> "Artifacts")

Edit: Windows installer is still working as expected

Noordfrees:
The daily PPA is failing now:
https://launchpadlibrarian.net/599328127/buildlog_ubuntu-impish-amd64.widelands_2%3A1.1~bzr10257-202205011033~ubuntu21.10.1_BUILDING.txt.gz
How do the paths in `debian/*` need to be adapted?

CMakeLists.txt
```diff
@@ -555,6 +555,13 @@ install(
...
```
matthiakl, line 563: > The daily PPA is failing now

Uh, oh. The value for WL_INSTALL_BINDIR here is set after including the src directory, so it is not picked up properly if not specified on the command line.

PR #5360, opened 2022-04-12
third_party/CMakeLists.txt:only use embedded minizip if not installed

### LINKED ISSUE #399: Remove widelands-bundled minizip sources for Debian
Based on bug 536161

We should check if we can remove widelands-bundled minizip sources.

This can be done if all major distributions are shipping a minizip library.

Currently at least Gentoo is doing so. We could use this bug to track distributions bundling a minizip library.

 * date created: 2012-05-25T20:09:41Z
 * owner: qcumber-some
 * the launchpad url was https://bugs.launchpad.net/bugs/1004684

andy5995:
(fixes #399)

So far, this finds minizip but the lib and include dir info isn't added to the build. I'm not sure where to add it though. Any suggestions on how to finish this?

```
...
-- Found Doxygen: /usr/bin/doxygen (found version "1.9.3") found components: doxygen dot 
-- Found PkgConfig: /usr/bin/pkg-config (found version "1.8.0") 
-- Checking for module 'minizip'
--   Found minizip, version 1.2.11
-- Configuring done
...
```


andy5995:
Seems to work as intended now after my last commit, but there's this warning I'm not sure about

```
/home/andy/src/widelands/src/io/filesystem/CMakeLists.txt:3: io_filesystem misses DEPENDS on MINIZIP_STATIC_LIBRARIES, because it includes third_party/minizip/unzip.h,third_party/minizip/zip.h
/home/andy/src/widelands/src/io/filesystem/CMakeLists.txt:3: io_filesystem DEPENDS on ${MINIZIP_STATIC_LIBRARIES}, but includes none of its headers.
-- Using CMake version 3.22.3
-- Building with XDG support
```

src/io/filesystem/CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ wl_library(io_filesystem
     base_exceptions
     base_macros
     io_stream
-    third_party_minizip
+    ${MINIZIP_STATIC_LIBRARIES}
```
matthiakl:
```suggestion
    MINIZIP_STATIC_LIBRARIES
```
Should resolve the codecheck problem. Like the other library names it does not need varible expansion.
However you should switch the includes in zip_filesystem.h to the system headers (`<zip.h>`, `<unzip.h>`) somehow when the external minizip is available, otherwise you will still get a codecheck error.

andy5995: > Should resolve the codecheck problem. Like the other library names it does not need varible expansion.

The build fails when I remove the variable expansion:

```
[31/189] Linking CXX executable src/base/test/test_base
FAILED: src/base/test/test_base 
...
usr/bin/ld: cannot find -lMINIZIP_STATIC_LIBRARIES: No such file or directory
```

andy5995:
After https://github.com/widelands/widelands/pull/5360/commits/4e93474141071f9ef544f54b48eee8ba3092d8ee this is the output when minizip isn't found

```
###########################################################
 
Call 'compile.sh -h' or 'compile.sh --help' for help.

For instructions on how to adjust options and build with
CMake, please take a look at
https://www.widelands.org/wiki/BuildingWidelands/.
 
###########################################################
 
/home/andy/src/widelands/src/io/filesystem/disk_filesystem.cc:1: (CRITICAL) File not mentioned in any build rule.
/home/andy/src/widelands/src/io/filesystem/disk_filesystem.h:1: (CRITICAL) File not mentioned in any build rule.
/home/andy/src/widelands/src/io/filesystem/filesystem.cc:1: (CRITICAL) File not mentioned in any build rule.
/home/andy/src/widelands/src/io/filesystem/filesystem.h:1: (CRITICAL) File not mentioned in any build rule.
/home/andy/src/widelands/src/io/filesystem/filesystem_exceptions.h:1: (CRITICAL) File not mentioned in any build rule.
/home/andy/src/widelands/src/io/filesystem/layered_filesystem.cc:1: (CRITICAL) File not mentioned in any build rule.
/home/andy/src/widelands/src/io/filesystem/layered_filesystem.h:1: (CRITICAL) File not mentioned in any build rule.
/home/andy/src/widelands/src/io/filesystem/zip_exceptions.h:1: (CRITICAL) File not mentioned in any build rule.
/home/andy/src/widelands/src/io/filesystem/zip_filesystem.cc:1: (CRITICAL) File not mentioned in any build rule.
/home/andy/src/widelands/src/io/filesystem/zip_filesystem.h:1: (CRITICAL) File not mentioned in any build rule.
```

When it's found:

```
...
CMake, please take a look at
https://www.widelands.org/wiki/BuildingWidelands/.
 
###########################################################
 
/home/andy/src/widelands/src/io/filesystem/CMakeLists.txt:3: io_filesystem DEPENDS on ${MINIZIP_STATIC_LIBRARIES}, but includes none of its headers.
-- Using CMake version 3.23.0
-- Building with XDG support
```

I don't know how to fix it yet.

And @matthiakl , I'm still using the variable expansion just so I can proceed a bit. I'll wait to hear back from you on how I should solve that. https://github.com/widelands/widelands/pull/5360#discussion_r856592434

andy5995:
Actually, it seems that my results were skewed incorrectly because I wasn't cleaning each time. I'll leave this for a bit, it's making me pretty sick.

And sometimes I get those messages about "/home/andy/src/widelands/src/io/filesystem/disk_filesystem.cc:1: (CRITICAL) File not mentioned in any build rule." when minizip is found, but then running compile again those messages don't appear, but the build fails.

andy5995:
It's a bit more robust now... 

When minizip is installed, the build works when ./compile.sh is run 2 or 3 times

```
###########################################################
 
/home/andy/src/widelands/src/io/filesystem/disk_filesystem.cc:1: (CRITICAL) File not mentioned in any build rule.
/home/andy/src/widelands/src/io/filesystem/disk_filesystem.h:1: (CRITICAL) File not mentioned in any build rule.
/home/andy/src/widelands/src/io/filesystem/filesystem.cc:1: (CRITICAL) File not mentioned in any build rule.
/home/andy/src/widelands/src/io/filesystem/filesystem.h:1: (CRITICAL) File not mentioned in any build rule.
/home/andy/src/widelands/src/io/filesystem/filesystem_exceptions.h:1: (CRITICAL) File not mentioned in any build rule.
/home/andy/src/widelands/src/io/filesystem/layered_filesystem.cc:1: (CRITICAL) File not mentioned in any build rule.
/home/andy/src/widelands/src/io/filesystem/layered_filesystem.h:1: (CRITICAL) File not mentioned in any build rule.
/home/andy/src/widelands/src/io/filesystem/zip_exceptions.h:1: (CRITICAL) File not mentioned in any build rule.
/home/andy/src/widelands/src/io/filesystem/zip_filesystem.cc:1: (CRITICAL) File not mentioned in any build rule.
/home/andy/src/widelands/src/io/filesystem/zip_filesystem.h:1: (CRITICAL) File not mentioned in any build rule.
[andy@prometheus widelands](iss-399)$ ./compile.sh -t
 ...
 ###########################################################
 
/home/andy/src/widelands/src/io/filesystem/CMakeLists.txt:19: io_filesystem misses DEPENDS on base_exceptions, because it includes base/wexception.h
/home/andy/src/widelands/src/io/filesystem/CMakeLists.txt:19: io_filesystem misses DEPENDS on base, because it includes base/i18n.h,base/log.h,base/string.h
/home/andy/src/widelands/src/io/filesystem/CMakeLists.txt:19: io_filesystem misses DEPENDS on base_macros, because it includes base/macros.h
/home/andy/src/widelands/src/io/filesystem/CMakeLists.txt:19: io_filesystem misses DEPENDS on io_stream, because it includes io/streamread.h,io/streamwrite.h
/home/andy/src/widelands/src/io/filesystem/CMakeLists.txt:19: io_filesystem DEPENDS on ${io_filesystem_depends}, but includes none of its headers.
-- The C compiler identification is GNU 11.2.0
-- The CXX compiler identification is GNU 11.2.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
```

If minizip is not installed, the build works when ./compile.sh is run 2 times.

```
###########################################################
 
/home/andy/src/widelands/src/io/filesystem/CMakeLists.txt:19: io_filesystem misses DEPENDS on base_macros, because it includes base/macros.h
/home/andy/src/widelands/src/io/filesystem/CMakeLists.txt:19: io_filesystem misses DEPENDS on base_exceptions, because it includes base/wexception.h
/home/andy/src/widelands/src/io/filesystem/CMakeLists.txt:19: io_filesystem misses DEPENDS on base, because it includes base/i18n.h,base/log.h,base/string.h
/home/andy/src/widelands/src/io/filesystem/CMakeLists.txt:19: io_filesystem misses DEPENDS on io_stream, because it includes io/streamread.h,io/streamwrite.h
/home/andy/src/widelands/src/io/filesystem/CMakeLists.txt:19: io_filesystem DEPENDS on ${io_filesystem_depends}, but includes none of its headers.
-- The C compiler identification is GNU 11.2.0
-- The CXX compiler identification is GNU 11.2.0
```

matthiakl:
To make the codecheck rule happy, we have to shift some definitions around. See my suggestion in the patch below (tested and working with and without system lib). 
System includes (with <>) should be grouped together at the top (sorted alphabetically).
[suggestion.patch.txt](https://github.com/widelands/widelands/files/8546651/0001-suggestion.patch.txt)

andy5995:
Thank you very much @matthiakl 

I added a FORCE_EMBEDDED_MINIZIP option so it can be tested a bit easier in the future. How would you like it added to the CI? Set up a new individual job with FORCE_EMBEDDED_MINIZIP=ON or do you want it integrated some other way into the existing jobs?

That option (when on) just means that the check for installed minizip will be skipped and the embedded minizip will be used instead. Off by default.

src/CMakeLists.txt
```diff
@@ -69,6 +69,16 @@ else()
   )
 endif()
 
+option(FORCE_EMBEDDED_MINIZIP "Use embedded minizip sources" OFF)
+if(NOT FORCE_EMBEDDED_MINIZIP)
```
matthiakl: Try to mimic existing style for cmake options: 
- define option at beginning of top level CMakeLists.txt
- name it `OPTION_...`
- add it to the table in Readme

Edit: Ah, just saw your comment. So is this only temporary or do you want to leave it in?

matthiakl:
> How would you like it added to the CI?

How about compiling debug build with FORCE_EMBEDDED_MINIZIP=ON and release build with the system lib? Then we test both options without generating new jobs.

src/CMakeLists.txt
andy5995: My comment was probably unclear. I think best to leave it in permanent.

andy5995:
> > How would you like it added to the CI?
> 
> How about compiling debug build with FORCE_EMBEDDED_MINIZIP=ON and release build with the system lib? Then we test both options without generating new jobs.

@matthiakl Should I do this anytime or wait for feedback from @Noordfrees and @hessenfarmer ?

Noordfrees:
We shouldn't add such hidden differences between release and debug builds, they make debugging CI errors harder. If I understand correctly then Ubuntu builds should use the system-wide minizip now and all other OSs the embedded one? If so let's add `FORCE_EMBEDDED_MINIZIP=ON` only to the *Windows* builds to see that it doesn't generate any errors; *MacOS* should use embedded without having to tell it to explicitly; and Ubuntu will use the system libs.

andy5995:
> We shouldn't add such hidden differences between release and debug builds, they make debugging CI errors harder. If I understand correctly then Ubuntu builds should use the system-wide minizip now and all other OSs the embedded one? If so let's add `FORCE_EMBEDDED_MINIZIP=ON` only to the _Windows_ builds to see that it doesn't generate any errors; _MacOS_ should use embedded without having to tell it to explicitly; and Ubuntu will use the system libs.

Well, I'd say the main point of having the FORCE_EMBEDDED option is for testing, or perhaps if someone can't compile with the system minizip for some reason. But using FORCE_EMBEDDED in the CI mostly assures us that a year or two or five years from now, the embedded sources are still building and working.

I could add that option to the Windows builds, but unless minizip is installed for the Win builds, the embedded minizip will be used without needing the FORCE option.

Though... looks like the Windows build is failing as is, even though embedded is still being used... And all the clang builds. I don't know why yet. Suggestions welcome...

matthiakl:
Regarding the errrors:
- clang has more warnings enabled than gcc, so it is complaining about something in the ioapi source. Placing a `-Wno-reserved-id-macro` in the TARGET_COMPILE_FLAGS should fix this. However it should be only ignored for compiling this part and not globally.
**Edit:** Just found, that there is already a macro for it in `base/macros.h`: `CLANG_DIAG_RESERVED_IDENTIFIER_OFF` and `CLANG_DIAG_RESERVED_IDENTIFIER_ON`

- When using the embedded minizip, the library is named `third_party_minizip` (we unintentionally were still linking to the system lib...). So the linking should be
```
  if(ARG_USES_MINIZIP)
      if(MINIZIP_STATIC_LIBRARIES)
          target_link_libraries(${NAME} minizip)
          message(STATUS "Link ${NAME} with minizip")
      else()
          target_link_libraries(${NAME} third_party_minizip)
          message(STATUS "Link ${NAME} with third_party_minizip")
      endif()
  endif()
```
- When setting the OPTION_FORCE_EMBEDDED_MINIZIP, any previously found minizip should be removed with `set(MINIZIP_STATIC_LIBRARIES "")`

andy5995:
>     * When using the embedded minizip, the library is named `third_party_minizip` (we unintentionally were still linking to the system lib...). So the linking should be
> 
>...
> 
>     * When setting the OPTION_FORCE_EMBEDDED_MINIZIP, any previously found minizip should be removed with `set(MINIZIP_STATIC_LIBRARIES "")`


MINIZIP_STATIC_LIBRARIES never gets set if the embed option is used. It would get set on L5 here, which is skipped if OPTION... is ON.

```
if(NOT OPTION_FORCE_EMBEDDED_MINIZIP)
   # If pkg-config is ever needed to find other libraries, move the include outside the if() block
  include(FindPkgConfig)
  if(PKG_CONFIG_FOUND)
    pkg_check_modules(MINIZIP minizip)
  endif()
endif()
```

cmake/WlFunctions.cmake
```diff
@@ -79,6 +79,9 @@ macro(_common_compile_tasks)
     if(MINIZIP_STATIC_LIBRARIES)
       wl_include_directories(${NAME} ${MINIZIP_INCLUDE_DIRS})
       target_compile_options(${NAME} SYSTEM PUBLIC ${MINIZIP_CFLAGS})
+      if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
+        target_compile_options(${NAME} PUBLIC "-Wno-reserved-id-macro")
+      endif()
```
matthiakl: Sorry, if I wasn't clear with my edit:
It should be sufficient to use the macros around the problematic includes (zip.h, unzip.h).
https://github.com/widelands/widelands/blob/efd890fd01556a957a0312b3c8a19e446f11a716/src/base/macros.h#L103-L104
Usage example:
https://github.com/widelands/widelands/blob/efd890fd01556a957a0312b3c8a19e446f11a716/src/base/i18n.cc#L49-L51

matthiakl (commented):
> MINIZIP_STATIC_LIBRARIES never gets set if the embed option is used. It would get set on L5 here, which is skipped if OPTION... is ON.

It does in the CMake cache if you first build with auto-detection and then want to switch to the embedded one. In this case specifying OPTION_FORCE_EMBEDDED_MINIZIP does not have the expected effect.

cmake/WlFunctions.cmake
andy5995: > Sorry, if I wasn't clear with my edit:

No worries. You've been immensely helpful @matthiakl

andy5995:
clang builds failing again after https://github.com/widelands/widelands/pull/5360/commits/922abeb7a236e5fcd82fbcbaf3caa0403f3bc101 although it worked locally on my system with clang 13.0.1

src/CMakeLists.txt
```diff
@@ -69,6 +69,18 @@ else()
   )
 endif()
 
+# unset these two if they are cached
+unset(MINIZIP_FOUND)
+unset(MINIZIP_STATIC_LIBRARIES)
```
matthiakl: This still links to the system minizip after enabling OPTION_FORCE_EMBEDDED_MINIZIP in an existing build dir.
This worked for me:
```suggestion
unset(MINIZIP_FOUND)
set(MINIZIP_STATIC_LIBRARIES "")
```

matthiakl (commented):
> clang builds failing again after https://github.com/widelands/widelands/commit/922abeb7a236e5fcd82fbcbaf3caa0403f3bc101 although it worked locally on my system with clang 13.0.1

That's strange, but I'm out of ideas now as well, so your solution is fine for me.

matthiakl (approved):
Fine for me now, thanks for your effort and patience :)

andy5995:
> Fine for me now, thanks for your effort and patience :)

You're welcome :)

PR #5364, opened 2022-04-16
Fix touchpad map scroll direction

Adriankhl:
At least on my laptop, after enabling map scrolling by touchpad, the horizontal scroll and vertical scroll directions are inconsistent. This PR fixes the default scroll direction.

### DIFF
--- a/src/wlapplication_mousewheel_options.cc
+++ b/src/wlapplication_mousewheel_options.cc
@@ -183,11 +183,12 @@ constexpr int32_t kSignIncreaseRight = -1;
 constexpr int32_t kSignIncreaseUp = 1;
 constexpr int32_t kSignNextRight = -1;
 constexpr int32_t kSignNextDown = -1;
-constexpr int32_t kSignScroll = -1;
+constexpr int32_t kSignScrollRight = 1;
+constexpr int32_t kSignScrollUp = -1;
 
 static const Sign2D kDefaultSignValue(kSignIncreaseRight, kSignIncreaseUp);
 static const Sign2D kDefaultSignMove(kSignNextRight, kSignNextDown);
-static const Sign2D kDefaultSignScroll(kSignScroll, kSignScroll);
+static const Sign2D kDefaultSignScroll(kSignScrollRight, kSignScrollUp);
 
 static std::map<MousewheelHandlerConfigID, MousewheelHandlerOptions> mousewheel_handlers = {
    {MousewheelHandlerConfigID::kChangeValue,


tothxa:
This is very strange, and your solution is most certainly not universal or even common. Please do not change the defaults. If this is really needed, then it must be an option.

What OS do you have? What is the system setting for scrolling direction? How does horizontal and vertical scrolling work in other programs? How do they work for spinboxes in Widelands?

Adriankhl:
OS: openSUSE
Desktop environment: sway

@tothxa I use "natural_scroll", which means my touchpad scrolls like what we expect on a touch screen. And it works well in my other programs.

In Widelands, if I hover over the "Maximum FPS" setting box, the number increases when I scroll right/down, decreases when I scroll left/up.

For map scrolling, it moves up when I scroll up, but it moves left when I scroll right. 

Am I the only one who experience this issue? If this behavior is not universal, I can close this PR and add a setting to invert the scroll optionally.

tothxa:
> OS: openSUSE Desktop environment: sway
> 
> @tothxa I use "natural_scroll", which means my touchpad scrolls like what we expect on a touch screen. And it works well in my other programs.

Does horizontal scroll work correctly (in "natural" mode) in other programs? (eg. GIMP)

> In Widelands, if I hover over the "Maximum FPS" setting box, the number increases when I scroll right/down, decreases when I scroll left/up.

If neither direction is flipped in Widelands scroll options, then these would be up/right increase in normal mode, decrease in "natural" mode. What I don't understand is why your horizontal and vertical directions don't match. Unless your system allows setting scroll mode independently for vertical and horizontal, and you only have it turned to "natural" for vertical. (in which case this should be the case in other programs as well)

> Am I the only one who experience this issue? If this behavior is not universal, I can close this PR and add a setting to invert the scroll optionally.

Well, I only had my Debian oldstable / X11 system for testing when I implemented this in #5068 and @Noordfrees was the only reviewer. I don't know how many users use this and on what systems.

SDL is supposed to report the system scroll direction in scroll events, but on my system it doesn't.

I assumed that map scrolling wouldn't need inverse options, because the system settings would be used.

Adriankhl:
> > OS: openSUSE Desktop environment: sway
> > @tothxa I use "natural_scroll", which means my touchpad scrolls like what we expect on a touch screen. And it works well in my other programs.
> 
> Does horizontal scroll work correctly (in "natural" mode) in other programs? (eg. GIMP)

Yes, it works correctly on GIMP and Firefox. Perhaps some SDL stuffs are not setup properly in Widelands? I am using wayland (xwayland for Widelands) and SDL might have issues with wayland, but my other SDL games (e.g., wesnoth) work fine with touchpad.

frankystone:
Just to be sure: Did you tried to enable the _"Invert click-and-drag map movement direction"_ in the options window, tab "In Game"?

tothxa:
I tried to look SDL vs. XWayland scrolling up with little success, but the few results I got included Wesnoth's struggle with this issue: https://github.com/wesnoth/wesnoth/issues/2218 and https://github.com/wesnoth/wesnoth/issues/3362 – the latter shows that the hardcoded "fix" for the former was not sufficient, and it is still open.

Adding the existing neither/horiz/vert/both invert option for scrolling too would be easy, straightforward and guaranteed to allow a workaround for every system. Though I still don't expect that many users would want to invert vertical, but who knows… But ideally we should be able to detect this and set it automatically, but I don't even know whether it should be a compile time or a runtime test…

@Adriankhl Could you please test with "natural" scrolling disabled, whether it also inverts horizontal or only vertical scrolling? **edit:** Also in other programs, most notably in Wesnoth.

@frankystone AFAIR that setting is not used in the mousewheel codepaths.

Adriankhl:
@frankystone No

@tothxa disabling "natural" scroll invert both the vertical and horizontal scroll direction, so they are still inconsistent. 

I did some more testing in wesnoth. The flatpak version scrolls fine. However, I did build wesnoth from source and found that the same problem happened in wesnoth.

I dug further in SDL and found https://github.com/libsdl-org/SDL/pull/4700, https://github.com/libsdl-org/SDL/issues/5202. Depending on the version of SDL and whether it is x11 or wayland, the horizontal scroll behaves differently. I am on openSUSE tumbleweed so I am using the latest 2.0.20, but I believe Debian will keep the old version for a while. Since it is not possible to run Widelands under pure wayland for now, it further complicates the problem, I am not sure if SDL can handle xwayland correctly. 

I have already added the invert option for map scrolling in my own fork, and I can create a PR for that if you guys think this is the right way to fix the issue. It would be better if it can be detected automatically, but I don't know how it can be done though.

tothxa:
> I dug further in SDL and found [libsdl-org/SDL#4700](https://github.com/libsdl-org/SDL/pull/4700), [libsdl-org/SDL#5202](https://github.com/libsdl-org/SDL/issues/5202). Depending on the version of SDL and whether it is x11 or wayland, the horizontal scroll behaves differently. I am on openSUSE tumbleweed so I am using the latest 2.0.20, but I believe Debian will keep the old version for a while. Since it is not possible to run Widelands under pure wayland for now, it further complicates the problem, I am not sure if SDL can handle xwayland correctly.

Thank you very much for the testing and the research.

I re-checked the SDL headers and the test I wrote for developing the mousewheel PR. I remembered that I had noticed the disagreement between the documentation in the header comment and the actual behaviour, but I got confused yesterday when I only ran the test. Now I can confirm that the behaviour of the new SDL version matches the documentation, and my old SDL version doesn't.

It would be nice to get some testing from w*ndows users too, because this suggests that the current behaviour in Widelands is wrong for them too. Also from apple users, because now I'm kind of lost what to expect there.

So it looks like the SDL fix was merged on 1 Sep, and 2.0.18 seems to be the first release afterwards. SDL is dynamically linked, so I think we should be able to use `SDL_GetVersion()` for checking. And based on the test results on all OSes, we will need compile time checks whether this workaround is needed.

> I have already added the invert option for map scrolling in my own fork, and I can create a PR for that if you guys think this is the right way to fix the issue. It would be better if it can be detected automatically, but I don't know how it can be done though.

After some more thinking I realised that this needs a single toggle that deals with all horizontal scroll handlers at once, because they are all affected (as confirmed by your testing too). With that in place (be it manual user setting or automatic), scrolling would probably not need the invert option after all.

So, as it looks like old SDL (on linux/X11) is the buggy one, I propose that we change both horizontal sign constants (`kSignNextRight` and `kSignIncreaseRight`) to 1, separate scroll signs as per your patch, but please use names `kSignScrollX` and `kSignScrollY` (these are short, always accurate, and neutral to what one considers right/up movement), and introduce the new toggle for the calculation of `current_sign_x_` in `MousewheelHandlerOptions::update_settings()`.

We should probably allow some time for testing, so the toggle should be manual for now. OTOH that needs to touch the GUI code too, unlike automatic...

klaus-halfmann:
cloning and compiling for macOS ...

klaus-halfmann:
OK, I use /opt/homebrew/Cellar/sdl2/2.0.20/lib/libSDL2-2.0.0.dylib 
It works for me but uses "inverse" scrolling which I prefer.
The system settings about "natural scrolling" are ignored.

1.1~git25721[b71af45@fix-touchpad]
macOS 12.3.1 , Apple M1 Pro ARM Processor.

anyhting more I can do?

tothxa:
> OK, I use /opt/homebrew/Cellar/sdl2/2.0.20/lib/libSDL2-2.0.0.dylib It works for me but uses "inverse" scrolling which I prefer.

So both horizontal and vertical behave the same?

> The system settings about "natural scrolling" are ignored.

(That's strange. Is there any chance that you used the equivalent of right button drag instead of the native scroll / mousewheel emulation function of the touchpad?)

> 1.1~git25721[[b71af45](https://github.com/widelands/widelands/commit/b71af4513621c9656ae6ebbc1d206b32cb996ef2)@fix-touchpad] macOS 12.3.1 , Apple M1 Pro ARM Processor.
> 
> anyhting more I can do?

Could you please also report the result with current master and the same SDL version?

klaus-halfmann:
> > OK, I use /opt/homebrew/Cellar/sdl2/2.0.20/lib/libSDL2-2.0.0.dylib It works for me but uses "inverse" scrolling which I prefer.
> 
> So both horizontal and vertical behave the same?
Yes
> 
> > The system settings about "natural scrolling" are ignored.
> 
> (That's strange. Is there any chance that you used the equivalent of right button drag instead of the native scroll / mousewheel emulation function of the touchpad?)

Indeed, as I use a "soft" click this might actually be a right button drag

> > 1.1~git25721[[b71af45](https://github.com/widelands/widelands/commit/b71af4513621c9656ae6ebbc1d206b32cb996ef2)@fix-touchpad] macOS 12.3.1 , Apple M1 Pro ARM Processor.
> > anyhting more I can do?
> 
> Could you please also report the result with current master and the same SDL version?

Will try but need to do lots of other stuff, first.

klaus-halfmann:
Compared master and this brachn but found no difference (on macOS)
Scrolling with two fingers actually zooms in and out. The "pinch" way of zooming does not work.
So either SDL or Widelands works different here...

tothxa:
> Compared master and this brachn but found no difference (on macOS) Scrolling with two fingers actually zooms in and out. The "pinch" way of zooming does not work. So either SDL or Widelands works different here...

I'm sorry, some of the new mousewheel functionality seems to not be well advertised. You need to turn on map scrolling in Options / Interface tab / Edit keyboard and mouse actions / Mouse Scroll tab / Set for touchpad button.

Or you can just test for example with the sound volume sliders: with the default settings in Widelands, with system scrolling set to old style, scrolling up / right should increase the volume, and scrolling down / left should decrease. With system set to "natural", default Widelands should result in reversed behaviour.

(that's why there are invert options – I wanted to make this automatic, but in my environment SDL didn't report the system setting, so I left that part out… if there's interest, maybe there should be another button to reverse directions for all non-scroll functionality at once) (BTW I know as much that w*ndows 10 handles this wrong as well: with inverted scrolling, sliders are also inverted)

tothxa:
> The "pinch" way of zooming does not work.

Right now we only handle mousewheel SDL events (which are in this case emulated by the touchpad driver or so), not proper touch events. Again, on my system I don't think SDL even generates them for the touchpad, so I can't test them.

klaus-halfmann:
Ahh, these options are really kind of hidden.
Configured them for my liking (using CMD+ scroll for zoom),
Gamespeed of 0.2 or 1.7 is kind of funny.

Stiff work the same (AFAIK) in both branches, but maybe I take a deeper look tomorrow.

tothxa:
@Adriankhl I don't think we should wait any longer. As there was no reply for w*ndows, and very little other data, the toggle should be manual. Are you interested in implementing it, or should I take this over?

Here's my previous comment on how I think this should be handled:

> this needs a single toggle that deals with all horizontal scroll handlers at once, because they are all affected.
>
> as it looks like old SDL (on linux/X11) is the buggy one, I propose that we change both horizontal sign constants (kSignNextRight and kSignIncreaseRight) to 1, separate scroll signs as per your patch, but please use names kSignScrollX and kSignScrollY (these are short, always accurate, and neutral to what one considers right/up movement), and introduce the new toggle for the calculation of current_sign_x_ in MousewheelHandlerOptions::update_settings().

Adriankhl:
@tothxa I am interested in working on this, but my main laptop is dead since last week :( So I stopped working on non-essential programming stuffs. 

If you have time, you can take this over. Or I will take a look at this after I receive the new laptop sometime next week.

Noordfrees:
Superseded by #5394

PR #5369, opened 2022-04-22
Update Elders and Chieftains

Noordfrees:
In internal discussion the following changes to the official list of project elders were proposed:
- Myself as Chieftain
- @hessenfarmer for the new post Elder of Scripting
- Elder of Sound: vacant, unless someone volunteers – maybe @klaus-halfmann would be interested?
- Elder of Graphics: Open to volunteers; in particular we'd like to ask @tothxa if you want it.  If nobody volunteers then I'll take the post.

These proposals are now open for general public discussion. @widelands/developers


tothxa:
>     * Elder of Graphics: Open to volunteers; in particular we'd like to ask @tothxa if you want it.

Wow, I'm honoured even to be considered. Are there any specific tasks going with the "title"?

Noordfrees:
There are no additional tasks, only privileges: If there's an argument about something and no consensus or compromise is possible (e.g., use style A or style B with very heated opinions about which one is good and which one is bad), then the Elder of the respective area has the right to make the final decision. And an Elder may veto any proposed change in their area. 

In practice, we're usually quite good at discussing things thoroughly and finding compromises (or at least deciding-to-do-nothing or suspending-things-indefinitely when a proposal is highly debated), and I can't now recall even a single case where an Elder or Chieftain actually had to use their final decision privileges.

jmoerschbach:
fine with me :)

SirVer:
Very much in favor of those changes!

tothxa (approved):
Any particular reason why noone wanted to approve this?

Maybe throw in adding hessenfarmer to Atlantean Campaign Story authors lower down in the file.

Noordfrees:
@tothxa & @hessenfarmer I'm going to write a news post on the homepage to inform the community of these changes. Do you want to include a message about your new Elder positions like [some other new Elders did](https://www.widelands.org/news/category/developers/)? Can be a long text wall or a single sentence, as you prefer ;) Best send it to me by forum PM.

hessenfarmer:
@Noordfrees I was afk a couple of days. I will sent you a short statement via PM. Will take until Monday probably

PR #5372, opened 2022-04-23
Fix Sound Options Panel Styling

Noordfrees:


matthiakl (approved):
Tested and looking good. I'm surprised that this has not been reported before.

PR #5373, opened 2022-04-23
Enable include order codecheck on header files

matthiakl:
Because of the #ifndef include guards in header files they were not processed by the correct_include_order rule.


Noordfrees (approved):
LGTM :)

Issue #5279, opened 2022-02-22
Use after free crash in MapObject::serial()


tothxa:
**Describe the bug**
Game with updated all_buildings_… map from media trunk crashes after a short time.

**To Reproduce**
1. Load attached game or start with attached map
2. Wait about 10 to 20 seconds
3. Game crashes

**Expected behavior**
No crash. :smiley: 

**Version:**
 - OS: antix/Debian buster with backports
 - Widelands Version: Version 1.1~git25646[ef3483a@master]

**Additional context**
[crash_log.txt](https://github.com/widelands/widelands/files/8117037/crash_log.txt)
[crash_game.zip](https://github.com/widelands/widelands/files/8117036/crash_game.zip)
[all_buildings.wmf.zip](https://github.com/widelands/widelands/files/8117035/all_buildings.wmf.zip)

tothxa:
Oops, wrong map. I mean, it crashes just fine :smile:, but it's not the one the savegame was created with, and it has too much logging output.

Here's the one from launchpad (with updates): [all_buildings_emp.wmf.zip](https://github.com/widelands/widelands/files/8123216/all_buildings_emp.wmf.zip)

PR #5269, opened 2022-02-19
remove leftover individual animation frames of buildings

tothxa:

PR #5376, opened 2022-04-25
Add high resolution images for empire buildings

tothxa:
See https://www.widelands.org/forum/topic/5485/

Also done:

- Empire frontier and flag
- destroyed building (model was in Empire directory in launchpad media repo) and ashes (derived from destroyed building)

Adjusted size of some buildings:

- reduced: lumberjack, quarry, fisher, sawmill, ship/ferry yard
- increased: HQ, shipwreck, port, fort, castle, arena, colosseum, piggery, sheepfarm (last 2 to match farm and donkey farm)

Maybe well should also be reduced.

Missing due to problems with the blender models:

- ~~all mines, winery,~~ weaving mill *(build steps and working animation are missing from all model versions on launchpad, so I give up on this for 1.1 unless someone can locate the model that was used for the current images)* ~~and weaponsmith~~


tothxa:
Kept old images of the following for v1.0 savegame compatibility of emp04 scenario:
- brewery, farm, forester, lumberjack, mill (idle added to working), well, warehouse, trainingcamp

**Reminder:** When the new images for these are approved, they will have to be updated in the scenario as well, plus the empire port in fri03 *edit: and empire warehouse in fri02*.

After this PR is merged these will have to be added to #5269

hessenfarmer:
ok i created a test game and compared all buildings with each other (except ship and ferry yard as I chose the wrong map for this) 

- small buildings: ok except fisher and scout could be reduced a bit to match the other small houses. Well are way to big in relation imho.
- medium buildings: tavern and inn could be reduced a bit, vineyard needs to be increased, warehouse could be increased a bit.  - Smelting works: the house is a bit to small however the melting pot is already drawn over a passing street. So the pot must move closer to the house and the house should be increased a bit, but this is probably beyond the current task.     
- big buildings: are ok imho. 
- military buildings: sentry and blockhouse, as well as outpost and barrier seem to be a bit to big in relation to other small / medium buildings and castle / tower.   
- mines: they could be reduced a bit as well, however they do not relate directly top other buildings so this is really of low priority. 

Anyhow this branch is a big improvement.

tothxa:
>     * small buildings: ok except fisher and scout could be reduced a bit to match the other small houses. Well are way to big in relation imho.

Scout should match forester and hunter... I'll re-check.

Finding the right scale for fisher is hard, it has very different proportions from the rest.

>     * medium buildings: tavern and inn could be reduced a bit,

Again different proportions than the rest, but I'll re-check.

> warehouse could be increased a bit.

It's only possible if we let carriers walk over the building, otherwise it already extends over the road behind.
It would be nice if it were possible to set where on the road the carriers appear/disappear to allow bigger medium buildings, but that's most certainly not for 1.1.

>  vineyard needs to be increased, 
>  - Smelting works: the house is a bit to small however the melting pot is already drawn over a passing street.

These are supposed to be the same size (door, roof, end wall) as e.g. bakery and other similar buildings, so here the difference between the simple rectangular and the longer L shaped buildings looks intentional.

> So the pot must move closer to the house and the house should be increased a bit, but this is probably beyond the current task.

Indeed, I'd rather not move geometry around for now, and this is even affected by animation.

>     * big buildings: are ok imho.

I've realised that farms are too big this way for the rest. I think I'll reduce them a little (but still bigger than old piggery/sheepfarm).

>     * military buildings: sentry and blockhouse, as well as outpost and barrier seem to be a bit to big in relation to other small / medium buildings and castle / tower.

I think I didn't change their size and castle was increased. I thought having bigger military buildings slightly bigger than production does not hurt. I may reduce them somewhat.

The tower will need to be edited to make it more slender and have a bigger door, but that's also not for this PR.

>     * mines: they could be reduced a bit as well, however they do not relate directly top other buildings so this is really of low priority.

Now that you mention it, indeed, they look slightly bigger than they should. I'll investigate this also compared to the other tribes as well.

BTW, here's the map I made for testing: [all_buildings_wmf.zip](https://github.com/widelands/widelands/files/8599462/all_buildings_wmf.zip)
(with AIs disabled it doesn't crash any more :smiley: re #5279)

hessenfarmer:
> Scout should match forester and hunter... I'll re-check.

it is only very slightly (about 0.5 to 1 %) larger so might be ok. 
> 
> Finding the right scale for fisher is hard, it has very different proportions from the rest.

minus 3% might be a try  

> 
> > warehouse could be increased a bit.
> 
> It's only possible if we let carriers walk over the building, otherwise it already extends over the road behind. It would be nice if it were possible to set where on the road the carriers appear/disappear to allow bigger medium buildings, but that's most certainly not for 1.1.

maybe we could move it closer to its flag and increase by 3 %     

> > * Smelting works: the house is a bit to small however the melting pot is already drawn over a passing street.
> 
> These are supposed to be the same size (door, roof, end wall) as e.g. bakery and other similar buildings, so here the difference between the simple rectangular and the longer L shaped buildings looks intentional.

directly build in adjacent places it looks a bit smaller then bakery e.g. so a 5% increase for the house would help imho. But for sure  out of scope here. 

> > ```
> > * big buildings: are ok imho.
> > ```
> 
> I've realised that farms are too big this way for the rest. I think I'll reduce them a little (but still bigger than old piggery/sheepfarm).
  
looked ok for me. however a small reduction would be ok as well.
> 
> > ```
> > * military buildings: sentry and blockhouse, as well as outpost and barrier seem to be a bit to big in relation to other small / medium buildings and castle / tower.
> > ```
> 
> I think I didn't change their size and castle was increased. I thought having bigger military buildings slightly bigger than production does not hurt. I may reduce them somewhat.  
  
in the forum you wrote that you increased at least sentry. I followed your advice and built up a trstgame with a lot of comparing possibilities. (this has the advantage of being loadable with current master for comparison as well) and especcially sntry looked to big there. So if we say military could be bigger then prod sites only sentry needs to be reduced a bit further.    
> 
> The tower will need to be edited to make it more slender and have a bigger door, but that's also not for this PR.    
   
that would be eye candy but in its current size it matches its vision range. But in principle I agree.

tothxa:
I created a new branch in my repo where I'll keep track of all hi-res PR branches for testing them together:
https://github.com/tothxa/widelands/tree/all_hi-res

tothxa:
> in the forum you wrote that you increased at least sentry.

Initially yes, but I did the changes I wrote there before creating the PR.

tothxa:
> Kept old images of the following for v1.0 savegame compatibility of emp04 scenario:

Having found out that the new tribes also have single frame animations as `animations = {…}`, but with the new style filenames, and that it works, this may not be necessary after all. Should I remove all `idle_00.png`s?

hessenfarmer:
> > Kept old images of the following for v1.0 savegame compatibility of emp04 scenario:
> 
> Having found out that the new tribes also have single frame animations as `animations = {…}`, but with the new style filenames, and that it works, this may not be necessary after all. Should I remove all `idle_00.png`s?

the problem lies within the lua code of old (1.0 ) savegames which expects these images imho.

hessenfarmer:
latest commits lookm very good. 
still open tavern/inn and vineyard and especially the wells. if these gets changed this would be good to go. Other changes are not essential imho

tothxa:
> > > Kept old images of the following for v1.0 savegame compatibility of emp04 scenario:
> > 
> > 
> > Having found out that the new tribes also have single frame animations as `animations = {…}`, but with the new style filenames, and that it works, this may not be necessary after all. Should I remove all `idle_00.png`s?
> 
> the problem lies within the lua code of old (1.0 ) savegames which expects these images imho.

I'm sorry. Indeed, that's right. The 1.0 campaign scripts refer to `idle_??.png` from the main `data/tribes/…` directories. So they stay for v1.1.

tothxa:
> latest commits lookm very good. still open tavern/inn and vineyard and especially the wells. if these gets changed this would be good to go. Other changes are not essential imho

Well is done and I have increased warehouse too in my local worktree. Before committing though, I'm working on a test increasing most medium buildings to make them better match the small and big ones and bring them closer to the Frisian buildings. If it works out, then tavern/inn may not need to be changed. Final scaling of warehouse and vineyard/smelter/charcoal will also depend on it.

tothxa:
Sorry for doing this last one so late, but I only realised that I could and should do it when I tested the increased warehouse.

Inn is still too big, so I'll reduce it and tavern to match. I'll also reduce farms and mines.

tothxa:
I think it all looks much more even now, so hopefully no more changes are needed.

Should/may I go on to apply the changes to the scenario units (listed above) before merging, or should I do it in a separate PR?

hessenfarmer:
Now it really looks fine and even. Thank you very much for the effort and patience.  
It would be good to do the scenario units in this PR as well imho. After that I'll approve so this can go in

hessenfarmer (changes requested):
Im just saw you used "directory = " in the spritesheets definitions. However our documentation says this is deprecated.   https://www.widelands.org/documentation/animations/
So better to use the animation_directory = again.

tothxa:
> Im just saw you used "directory = " in the spritesheets definitions. However our documentation says this is deprecated. https://www.widelands.org/documentation/animations/ So better to use the animation_directory = again.

I'm sorry, wasn't aware of that. I corrected them for this PR. (but Frisian unit definitions still use `directory =` a lot)

I also did the scenario units. I've kept around the `*_00.png`s for those too for now. I think they're only needed for savegame compatibility of 1.1~… development versions, but that's still useful, isn't it?

(unfortunately old savegames using the old images give away the crippled buildings in emp04 at increased zoom levels, but I guess there's nothing we can do about that)

data/tribes/buildings/productionsites/empire/ferry_yard/init.lua
```diff
@@ -22,12 +22,32 @@ wl.Descriptions():new_productionsite_type {
...
+         rows = 1,
+         hotspot = { 56, 58 },
+      },
+      unoccupied = {
+         frames = 1,
+         columns = 1,
+         rows = 1,
+         hotspot = { 56, 58 },
+      },
+      working = {
```
hessenfarmer, line 45: we do not use a working animation in the production programs so I beleive we won't need this. However we could leave it in for future changes.

hessenfarmer (approved):
LGTM now.
one small nit

hessenfarmer:
> I'm sorry, wasn't aware of that. I corrected them for this PR. (but Frisian unit definitions still use `directory =` a lot)

so maybe we should open an issue for this @Noordfrees 
> 
> I also did the scenario units. I've kept around the `*_00.png`s for those too for now. I think they're only needed for savegame compatibility of 1.1~… development versions, but that's still useful, isn't it?

they are needed for Version 1.0 savegames. we always try to support savegames of the last Version so we need to keep them in for final version 1.1 and can delete them as soon as develeopment has moved to working on 1.2

> 
> (unfortunately old savegames using the old images give away the crippled buildings in emp04 at increased zoom levels, but I guess there's nothing we can do about that). 

no we can't and they have the old sizes as well however the main thing is the savegames are working.

Noordfrees:
> > I'm sorry, wasn't aware of that. I corrected them for this PR. (but Frisian unit definitions still use `directory =` a lot)
> 
> so maybe we should open an issue for this @Noordfrees

Done: #5411

data/tribes/buildings/productionsites/empire/ferry_yard/init.lua
tothxa: I have to admit I didn't check it, just copied all images from shipyard.

tothxa:
@hessenfarmer Thank you for the review!

PR #5377, opened 2022-04-26
Integrate AppImage building into CI

matthiakl:
Supersedes #5354 
Also see https://github.com/andy5995/widelands-appimage/pull/2
Result: https://github.com/matthiakl/widelands/releases/tag/latest

Automatic AppImage building and added it to the automated development releases.
Requires #5359 for correct path configuration (therefore the appimage might not be executable atm, the artifact from https://github.com/andy5995/widelands-appimage/pull/2 uses that branch and works fine)

Should we also provide AppImages for Debug builds?


.github/workflows/build.yaml
```diff
@@ -378,10 +379,63 @@ jobs:
...
+          - clang-10
+    # Must run on the oldest still-supported Ubuntu LTS release
+    runs-on: ubuntu-18.04
+    env:
+      VERSION: "${{ github.sha }}-${{ matrix.config }}-${{ matrix.compiler }}"
+    steps:
+    - name: Installing dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install git cmake ${{ matrix.compiler }} gettext libasio-dev libglew-dev libpng-dev libsdl2-dev libsdl2-image-dev libsdl2-mixer-dev libsdl2-ttf-dev python3 zlib1g-dev -y
```
andy5995: @matthiakl Assuming this will get merged after #5415, the dependencies can be installed via the script as shown here on L172 https://github.com/widelands/widelands/pull/5415/files#diff-d0777657fa3fd81d23aaf7273e58aee453b04e67882517900c56daeef9b3e4c1R172

matthiakl: Thanks for the info. I updated the build file accordingly.

Noordfrees (approved):
LGTM. Don't forget to revert the annotated places in the diff before merging :)

matthiakl:
Thanks for the review :)
> Don't forget to revert the annotated places in the diff before merging :)

Done.

PR #5381, opened 2022-04-29
Calculate max battle animation height

### LINKED ISSUE #5064: Bouncing soldier level indicators during battle
Just noticed while watching a fight between two barbarian soldiers that the level indicators above their heads keep jumping up and down. Probably the battle animations have been cropped to different heights by the spritesheet exporting. Solution would be to add some empty padding to the top or bottom of the animations to get the indicator to stay at the same position all the time. Needs checking whether empire and/or atlanteans are also affected.

matthiakl:
Fixes #5064 
Instead of adding empty pixels, the level indicator is now placed at a fixed offset from the maximum battle animation height. 
The bouncing is present in all tribes and might also transfer to addons if there is no requirement to have all animations in the same proportions.

### DIFF
--- a/src/logic/map_objects/tribes/soldier.cc
+++ b/src/logic/map_objects/tribes/soldier.cc
@@ -99,7 +99,8 @@ SoldierDescr::SoldierDescr(const std::string& init_descname,
      health_(table.get_table("health")),
      attack_(table.get_table("attack")),
      defense_(table.get_table("defense")),
-     evade_(table.get_table("evade")) {
+     evade_(table.get_table("evade")),
+     max_anim_height_(0) {
 
 	// Battle animations
 	// attack_success_*-> soldier is attacking and hit his opponent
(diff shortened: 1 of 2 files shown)


src/logic/map_objects/tribes/soldier.cc
```diff
@@ -308,9 +309,11 @@ Bob& SoldierDescr::create_object() const {
 void SoldierDescr::add_battle_animation(std::unique_ptr<LuaTable> table,
                                         SoldierAnimationsList* result) {
 	for (const std::string& anim_name : table->keys<std::string>()) {
-		if (!is_animation_known(anim_name)) {
-			throw GameDataError("Trying to add unknown battle animation: %s", anim_name.c_str());
-		}
```
tothxa, line 313: (this is not a proper review, just a quick glance, but)

Why are you removing this check? Isn't that dangerous?

matthiakl: I retrieve more info about the animation through ˋget_animationˋ. In this function the same check is already done:

https://github.com/widelands/widelands/blob/master/src/logic/map_objects/map_object.cc#L393-L397

Noordfrees (approved):
LGTM, tested and working. Very good solution; thanks for fixing :)

PR #5388, opened 2022-05-04
Pan map view to keep it centered when resizing game window.

### LINKED ISSUE #4863: Mapview not centered if one clicks on maximize window
If i click on maximize window in the top border of the desktop window the current view of the game is moved to the upper left corner.

**To Reproduce**
1. start a game with a resolution of 800x600
2. click on 'Maximize' in the window upper border (close to the 'x')
3. the view of the map get moved to the upper left corner

**Expected behavior**
The view of the map should be centered inside the current window size.

**Screenshots**
Initial situation:
maximize_window

Result after clicking 'Maximize':
maximize_result

**Version:**
 - OS: archlinux
 - Widelands Version: [cb3bd86@supply-list-remove-crash] (Debug)

attet:
Fixes #4863.

### DIFF
--- a/src/wui/info_panel.cc
+++ b/src/wui/info_panel.cc
@@ -383,7 +383,7 @@ void InfoPanel::set_fps_string(const bool show,
 		text_fps_.set_text("");
 		text_fps_.set_tooltip("");
 	} else {
-		const std::string text = format("%5.1f fps (avg: %5.1f fps)", fps, average);
+		const std::string text = format("%05.1f fps (avg: %05.1f fps)", fps, average);
 		// The FPS string overlaps with the coords string at low resolution.
 		// Therefore abbreviate it if the available width is less than an arbitrary threshold.
 		if (cheating) {
--- a/src/wui/info_panel.cc
+++ b/src/wui/info_panel.cc
@@ -383,7 +383,7 @@ void InfoPanel::set_fps_string(const bool show,
 		text_fps_.set_text("");
 		text_fps_.set_tooltip("");
 	} else {
-		const std::string text = format("%05.1f fps (avg: %05.1f fps)", fps, average);
+		const std::string text = format("%5.1f fps (avg: %5.1f fps)", fps, average);
 		// The FPS string overlaps with the coords string at low resolution.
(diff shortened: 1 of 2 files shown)


src/wui/interactive_base.cc
```diff
@@ -188,6 +188,9 @@ InteractiveBase::InteractiveBase(EditorGameBase& the_egbase, Section& global_s,
 	   [this](const GraphicResolutionChanged& message) {
 		   set_size(message.new_width, message.new_height);
 		   map_view_.set_size(message.new_width, message.new_height);
+		   map_view_.pan_by(Vector2i(-(message.new_width - message.old_width) / 2,
+		                             -(message.new_height - message.old_height) / 2),
```
tothxa:
```suggestion
		   map_view_.pan_by(Vector2i((message.old_width - message.new_width) / 2,
		                             (message.old_height - message.new_height) / 2),
```
sorry, I'm a math geek ;)

src/wui/info_panel.cc
```diff
@@ -383,7 +383,7 @@ void InfoPanel::set_fps_string(const bool show,
 		text_fps_.set_text("");
 		text_fps_.set_tooltip("");
 	} else {
-		const std::string text = format("%5.1f fps (avg: %5.1f fps)", fps, average);
+		const std::string text = format("%05.1f fps (avg: %05.1f fps)", fps, average);
```
tothxa: (I think this would have been better in a separate PR, though both changes are simple, so no big harm)

As this is 1.2 material, I think `base/format/{number,float}_node.h` should be fixed to use `kDigitWidthSpace` for padding in localised mode, and use `format_l()` instead of padding with zeroes.

(… except this isn't localised, but why? I've just tested it and it has no problem with unicode chars.)

attet: Sorry, I assumed the change was trivially acceptable. I can look into it separately.

tothxa (approved):
@Noordfrees This is much simpler than some of the outstanding PRs targeted for 1.1. Shouldn't it get an exception from the feature freeze?

Issue #5387, opened 2022-05-04
Daily PPA Failing to Build


Noordfrees:
Re https://github.com/widelands/widelands/pull/5359#issuecomment-1114222800 & https://github.com/widelands/widelands/pull/5383

As of 0ce5fb604e9bf884b0c4956500b61282cf6ae869 the [daily PPA](https://code.launchpad.net/~widelands-dev/+recipe/widelands-daily) is still failing. The build logs seem to be unavailable currently.

@matthiakl @andy5995 

---

By the way, the PPA also doesn't use Debian's minizip library yet:
https://github.com/widelands/widelands/blob/0ce5fb604e9bf884b0c4956500b61282cf6ae869/debian/control#L12-L14
(oldest we now support is 16.04)

matthiakl:
[amd64 build of widelands 2:1.1~bzr10260-202205040401\~ubuntu18.04.1 in ubuntu bionic RELEASE](https://code.launchpad.net/~widelands-dev/+archive/ubuntu/widelands-daily/+build/23607664) succeeded, so I am not sure if this is really another bug in the code or something related to launchpad. Also the missing logfiles look like the builds did not properly started. 
Do you have the possibility to retry a build?

Noordfrees:
Yes but it usually takes some time until it starts. I triggered just one new build for now: https://code.launchpad.net/~widelands-dev/+archive/ubuntu/widelands-daily/+build/23607678

matthiakl:
Same again: now kinetic builts successfully. Could it be a problem that the repo is too full with old packages? (20GB/8GB) Maybe launchpad only allows to store a single package per build matrix in this case.

> (oldest we now support is 16.04)

The PPA description still lists trusty as published platform. Is this outdated?

Noordfrees:
Okay, seems to have just been a transient LP error then. 
The repo size limit has been exceeded since about forever :wink: we never had any problems with it though.
We'll know for certain after the next build…

Trusty is listed – along with a lot of other EOL releases – because the PPA contains a package that installs there, but they're no longer kept up to date. The official description text says "Packages are available for all currently supported Ubuntu releases."

Noordfrees:
Failing again:
```
CMake Error at cmake/WlFunctions.cmake:81 (target_compile_options):
  target_compile_options called with invalid arguments
Call Stack (most recent call first):
  cmake/WlFunctions.cmake:232 (_common_compile_tasks)
[...]

PR #5392, opened 2022-05-05
Use Debian Minizip Library for the PPA

Noordfrees:
Re #5387

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -9,12 +9,10 @@ Standards-Version: 3.9.8
 Build-Depends: debhelper (>= 9),
   cmake,
   doxygen,
-#  This package isn't available in older Ubuntu releases.
-#  As long as we support 14.04, we build using the embedded one.
-#  libminizip-dev,
   libasio-dev,
   libglew-dev,
   libicu-dev,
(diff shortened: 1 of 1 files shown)


debian/control
```diff
@@ -9,12 +9,10 @@ Standards-Version: 3.9.8
 Build-Depends: debhelper (>= 9),
   cmake,
   doxygen,
-#  This package isn't available in older Ubuntu releases.
-#  As long as we support 14.04, we build using the embedded one.
-#  libminizip-dev,
   libasio-dev,
   libglew-dev,
   libicu-dev,
+  libminizip-dev,
```
matthiakl, line 15: Does this automatically add `libminizip` to the run/install dependencies or is this configured somewhere else?

andy5995: @matthiakl I think that happens in the 'Depends' field where `${shlibs:Depends}` is used.
https://www.debian.org/doc/manuals/maint-guide/dreq.en.html

andy5995 (approved):

matthiakl (approved):

PR #5394, opened 2022-05-06
Workaround for inverted horizontal scrolling bug in SDL

tothxa:
This is my take on replacing #5364 (with apologies to @Adriankhl)

I'd like to add a note when the workaround is turned on to ask for feedback on affected systems for future automation. Where should I direct it: bugreport or forum topic?


Noordfrees:
I think actually showing a message box for this option is too obtrusive IMHO. Better just print a message to standard output; even if just a small sample of users report back this should still give us a fairly good idea where the bug happens. Instead of asking users to spam us with feedback issues better please a link to https://github.com/widelands/widelands/discussions/5367.

Also the message should make it clear that the OS and the SDL version are the important information, ideally even include the latter in the message so the user can copy-paste the information. (Or add the SDL version to the About → Technical Info section and direct the user to use that feature.)

tothxa:
> I think actually showing a message box for this option is too obtrusive IMHO.

OK, I removed it. May I try changing the checkbox label when the box is checked to remind users to send feedback?

> Instead of asking users to spam us with feedback issues better please a link to #5367.

Can non-members comment there? Hmm, looks so, attet isn't a member yet. OK then, this makes the message simpler.

> Also the message should make it clear that the OS and the SDL version are the important information, ideally even include the latter in the message so the user can copy-paste the information. (Or add the SDL version to the About → Technical Info section and direct the user to use that feature.)

I had actually checked whether I could direct users to the Technical Info for this purpose, but I found it had a very different scope. But adding the info we need for this feature to it would be useful in any case, so I'll do it. Though I'm not sure we can easily get everything in a portable way, like system versions and the like. (I suspect this must have worked right on linux/X11 too when it was added to SDL a long time ago, so something must have changed in either linux or X11 that reversed it... too bad the SDL people didn't investigate it before unconditionally flipping it. And I'm not sure where I should ask them about this – I don't think commenting on the closed PR would get much attention.)

tothxa:
Functionally done, but maybe the look of the report window could be improved.

Noordfrees:
Code LGTM, tested and working, just two UI bugs: 

In the About window, the left column is now so wide that a new layouting glitch appears at low resolutions:
image

And in the feedback window, the box is not centered:
image

Noordfrees (approved):
LGTM, thanks for fixing :)

tothxa:
Thanks for the review.

PR #5396, opened 2022-05-06
remove SYSTEM from target_compile_options

### LINKED ISSUE #5387: Daily PPA Failing to Build

andy5995:
(fixes #5387)

SYSTEM is not mentioned as a valid argument according to the docs for
this function


andy5995:
mentioned in that launchpad log many times

'target_compile_options called with invalid arguments'

That function is only being used for the minizip check I think, on L81 and L85 of WlFunction.cmake

Noordfrees (approved):

PR #5399, opened 2022-05-06
Fix string bugs in lore texts reported on Transifex

tothxa:


klaus-halfmann (approved):
makes perfect sense, where in transifex would I find this?

Noordfrees:
> where in transifex would I find this?

https://www.transifex.com/widelands/widelands/translate/#issues

PR #5401, opened 2022-05-07
Restore iterator after insertion in defaultai

### LINKED ISSUE #5400: Debug builds on MSVC crash frequently due to invalid iterator access in DefaultAI::check_economies
Trace: 
>	widelands.exe!std::_Deque_const_iterator<std::_Deque_val<std::_Deque_simple_types<AI::EconomyObserver *>>>::operator*() line 266	C++
 	widelands.exe!std::_Deque_iterator<std::_Deque_val<std::_Deque_simple_types<AI::EconomyObserver *>>>::operator*() Line 459	C++
 	widelands.exe!AI::DefaultAI::check_economies() Line 4027	C++
 	widelands.exe!AI::DefaultAI::think() Line 412	C++
 	widelands.exe!SinglePlayerGameController::think() Line 78	C++
 	widelands.exe!Widelands::Game::think() Line 836	C++
 	widelands.exe!InteractiveBase::game_logic_think() Line 721	C++
 	widelands.exe!UI::Panel::logic_thread() Line 204	C++
 	[External Code]	
 	widelands.exe!thread_start<unsigned int (__cdecl*)(void *),1>(void * const parameter=0x000001bc85cbb0d0) Line 97	C++
 	[External Code]

The assert in MSVC deque.h:266 is just:
        _STL_VERIFY(_Mycont, "cannot dereference value-initialized deque iterator");

The actual problem is this loop in DefaultAI::check_economies. Notice the call to get_economy_observer:
https://github.com/widelands/widelands/blob/cb85b9b09cad209187c1de526be4b11594cc523a/src/ai/defaultai.cc#L4021-L4029

There we end up hitting a case where 'economies' is modified.
https://github.com/widelands/widelands/blob/cb85b9b09cad209187c1de526be4b11594cc523a/src/ai/defaultai.cc#L6170

Thus obs_iter in check_economies is now invalid (iterators are not valid for std::deque after any insert operations), leading to crash on line 4027 on the next access.

**To Reproduce**
1. Start a new game using barbarian map 2 (won't repro on lvl 1, I assume due to AI / other economies not being present; I've seen the crash on plenty of other maps as well but this is fastest and most reliable repro case so far)
2. Dismiss the intro dialog so the game gets going
3. Wait around 30 seconds
4. Experience crash

**Version:**
 - OS: Windows 11
 - Widelands Version: 1.1~git25741[0ce5fb6@master] (Debug) 

**Additional context**
I assume this won't easily repro on non-msvc/non-debug environments as it's rather frequent and probably would've been reported already. Release build container implementations might be just fine with the usage, but that's no reason not to fix going against the spec.

I won't try to fix since I'm rather unfamiliar with the codebase, perhaps use a different container or maybe it's fine to just restart the loop?

matthiakl:
Fixes #5400 
Restores the observer iterator to the last position after it has been invalidated due to insertion.

### DIFF
--- a/src/ai/defaultai.cc
+++ b/src/ai/defaultai.cc
@@ -4019,9 +4019,8 @@ bool DefaultAI::check_economies() {
 	}
 
 	size_t eco_size = economies.size();
-	size_t obs_idx = 0;
 	for (std::deque<EconomyObserver*>::iterator obs_iter = economies.begin();
-	     obs_iter != economies.end(); ++obs_iter, ++obs_idx) {
+	     obs_iter != economies.end(); ++obs_iter) {
 		// check if any flag has changed its economy
 		std::deque<Widelands::Flag const*>& fl = (*obs_iter)->flags;
 
@@ -4033,12 +4032,7 @@ bool DefaultAI::check_economies() {
 				j = fl.erase(j);
 				if (eco_size != economies.size()) {
 					// economies was modified, so obs_iter is invalid now.
-					// Restore it to the last position
-					eco_size++;
-					obs_idx++;
-					assert(eco_size == economies.size());
-					obs_iter = economies.begin();
-					std::advance(obs_iter, obs_idx);
+					return true;
 				}
 			} else {
 				++j;
--- a/src/ai/defaultai.cc
+++ b/src/ai/defaultai.cc
@@ -4018,8 +4018,10 @@ bool DefaultAI::check_economies() {
 		get_economy_observer(flag.economy(Widelands::wwWORKER))->flags.push_back(&flag);
 	}
 
+	size_t eco_size = economies.size();
+	size_t obs_idx = 0;
 	for (std::deque<EconomyObserver*>::iterator obs_iter = economies.begin();
-	     obs_iter != economies.end(); ++obs_iter) {
+	     obs_iter != economies.end(); ++obs_iter, ++obs_idx) {
 		// check if any flag has changed its economy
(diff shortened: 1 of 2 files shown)


attet:
Thanks for the quick fix, not seeing any related crashes with this patched in.

src/ai/defaultai.cc
```diff
@@ -4029,6 +4031,15 @@ bool DefaultAI::check_economies() {
 				get_economy_observer((*j)->economy(Widelands::wwWORKER))->flags.push_back(*j);
 				// and erase from this economy's observer
 				j = fl.erase(j);
+				if (eco_size != economies.size()) {
+					// economies was modified, so obs_iter is invalid now.
```
tibor95, line 4034: What would modify economies?
There is economies.erase() below, but there is return immediately after it

matthiakl: `get_economy_observer()` may add an item to economies, if `(*j)->economy()` is not yet in there.

tibor95: You are right.
What about using lazy approach? Just return True... We are doing the same few lines below... It would be less code and AI's scheduler is fine with returning True, it will skip some other stuff and delay would be only minimal.

matthiakl: Done. I wasn't sure about that return value.

tibor95: It is ok for me now

tibor95:
looks ok

hessenfarmer (approved):
LGTM

PR #5413, opened 2022-05-11
workflows/build.yaml:fix CI (disable uninstall)

### LINKED ISSUE #5412: testsuite frequently failing with Make install / uninstall task
for some latest builds our ubuntu 20 debug builds are failing in the make instal uninstall task

andy5995:
fixes #5412


tothxa:
Isn't there a way to force single thread for a target in CMakeLists? That would make it safer for our users too.

andy5995:
The commonalities for the failures is they happen on Ubuntu 20.04 with all Debug builds.

In the logs a successful completion looks like:

```
-- Removing empty dir: /home/runner/work/widelands/widelands/build/test_install/usr/local/./data
-- Removing empty dir: /home/runner/work/widelands/widelands/build/test_install/usr/local
-- Removing empty dir: /home/runner/work/widelands/widelands/build/test_install/usr
-- Removing empty dir: /home/runner/work/widelands/widelands/build/test_install
Built target uninstall
```

Failure looks like:

```
 -- Removing empty dir: /home/runner/work/widelands/widelands/build/test_install/usr/local/./doc
-- Removing empty dir: /home/runner/work/widelands/widelands/build/test_install/usr/local/share
-- Removing empty dir: /home/runner/work/widelands/widelands/build/test_install/usr/local/./data
-- Removing empty dir: /home/runner/work/widelands/widelands/build/test_install/usr/local/.
Error removing directory "/home/runner/work/widelands/widelands/build/test_install/usr/local/.".
CMake Error at cmake/cmake_uninstall.cmake:22 (message):
  Failed to remove directory:
  /home/runner/work/widelands/widelands/build/test_install/usr/local/.
```

So the debug build is trying to remove '/usr/local/." and failing.

And that's happening after I removed the '-j' flag.

> Isn't there a way to force single thread for a target in CMakeLists? That would make it safer for our users too.

@tothxa Good suggestion. Not that I know of. I think if that could be specified, it would be set in [add_custom_target()](https://cmake.org/cmake/help/latest/command/add_custom_target.html).

One option for that function is JOB_POOL, but it only works with ninja.

hessenfarmer:
rather then leaving out -j for uninstall we might try -j1 as afaik make remebers the last j option set. 

Edit: deleted the part about the CI as @andy5995 did this while I was commenting

hessenfarmer:
The important question is why did it work previously and why is it failing now? Has something changed in the CI environment?

andy5995:
> rather then leaving out -j for uninstall we might try -j1 as afaik make remebers the last j option set.
> 

I don't see that's (parallel jobs) a problem. As I detailed above, there's a definite pattern. Happens only with Debug builds on Ubuntu 20.04.

>The important question is why did ir work previously and why is it failing now? HAs something changed in the CI environment?

I have no clue yet either. If I get any ideas on the cause or what to try next, I'll definitely let you know.

hessenfarmer:
> > rather then leaving out -j for uninstall we might try -j1 as afaik make remebers the last j option set.
> 
> I don't see that's (parallel jobs) a problem. As I detailed above, there's a definite pattern. Happens only with Debug builds on Ubuntu 20.04.

I gave it a try in a parallel PR anyhow.

andy5995:
One idea I had to diagnose the problem is to get an [Ubuntu 20.04 docker image](https://hub.docker.com/_/ubuntu) and do a release build, back up the install_manifest.txt, then do a debug build. Then do a diff on the two files.

I may get around to trying that in the next few days.

andy5995:
>As I detailed above, there's a definite pattern. Happens only with Debug builds on Ubuntu 20.04.

Wrong. I see it's also happened on release builds on 18.04
https://github.com/widelands/widelands/actions/runs/2310076474

If anyone that wants to try with docker, after you pull the ubuntu image, you can use

* cd widelands (your existing widelands source directory)
* docker run -it -v $PWD:/wl ubuntu:focal

That'll start a container, your local widelands src dir will be mounted at /wl and any changes you make in that directory will be persistent.

hessenfarmer:
closing as @matthiakl found a working solution

PR #5415, opened 2022-05-12
Update testsuite compilers and OSs for Jammy

### LINKED ISSUE #5418: Does not compile using gcc version 12.1.0-1
After updating my system gcc was also updated:

~~~~
[2022-05-13T09:33:53+0200] [ALPM] upgraded gcc (11.2.0-4 -> 12.1.0-1)
[2022-05-13T09:33:57+0200] [ALPM] upgraded lib32-gcc-libs (11.2.0-4 -> 12.1.0-1)
~~~~

Compiling a release build does not work anymore after the update:

~~~~
[ 83%] Building CXX object src/ui_fsmenu/CMakeFiles/ui_fsmenu.dir/multiplayersetupgroup.cc.o
In file included from /home/kaputtnik/Quellcode/widelands-repo/widelands/src/ui_fsmenu/main.cc:30:
/home/kaputtnik/Quellcode/widelands-repo/widelands/src/editor/editorinteractive.h: In constructor 'EditorInteractive::Tools::Tools(const Widelands::Map&)':
/home/kaputtnik/Quellcode/widelands-repo/widelands/src/editor/editorinteractive.h:63:37: error: member 'EditorInteractive::Tools::unset_port_space' is used uninitialized [-Werror=uninitialized]
   63 |                      set_port_space(unset_port_space),
      |                                     ^~~~~~~~~~~~~~~~
[ 83%] Building CXX object src/ui_fsmenu/CMakeFiles/ui_fsmenu.dir/netsetup_lan.cc.o
[ 83%] Building CXX object src/ui_fsmenu/CMakeFiles/ui_fsmenu.dir/options.cc.o
cc1plus: all warnings being treated as errors
make[2]: *** [src/ui_fsmenu/CMakeFiles/ui_fsmenu.dir/build.make:398: src/ui_fsmenu/CMakeFiles/ui_fsmenu.dir/main.cc.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [CMakeFiles/Makefile2:30436: src/ui_fsmenu/CMakeFiles/ui_fsmenu.dir/all] Error 2
make: *** [Makefile:146: all] Error 2
~~~~

Compiling a debug build throws another warning as error:

~~~~
./compile.sh -a -s
…
[ 83%] Building CXX object src/base/CMakeFiles/base.dir/string.cc.o
/home/kaputtnik/Quellcode/widelands-repo/widelands/src/base/log.cc: In function 'void do_log(LogType, const Time&, const char*, ...)':
/home/kaputtnik/Quellcode/widelands-repo/widelands/src/base/log.cc:189:94: error: 'snprintf' output may be truncated before the last format character [-Werror=format-truncation=]
  189 |                 snprintf(buffer_prefix, sizeof(buffer_prefix), "[%02u:%02u:%02u.%03u %s] %s: ", hours,
      |                                                                                              ^
/home/kaputtnik/Quellcode/widelands-repo/widelands/src/base/log.cc:189:25: note: 'snprintf' output 23 or more bytes (assuming 33) into a destination of size 32
  189 |                 snprintf(buffer_prefix, sizeof(buffer_prefix), "[%02u:%02u:%02u.%03u %s] %s: ", hours,
      |                 ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  190 |                          minutes, seconds, t, gametime.is_invalid() ? "real" : "game", to_string(type));
      |                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
make[2]: *** [src/base/CMakeFiles/base.dir/build.make:118: src/base/CMakeFiles/base.dir/log.cc.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [CMakeFiles/Makefile2:6620: src/base/CMakeFiles/base.dir/all] Error 2
make: *** [Makefile:136: all] Error 2
~~~~

Compiling using clang works.

Noordfrees:


Noordfrees:
g++12 Debug builds work now. g++12 Release builds fail with various warnings from included __system headers__ (`std_function.h`, `atomic_base.h`), does anyone know why this happens?

matthiakl:
I do not get these errors with gcc 12.1.0 (CI uses 12.0.1). Instead I get two errors which I only could resolve by using local GCC_DIAG_OFF, but then it compiled (both debug and release):
```
[ 89%] Building CXX object src/logic/map_objects/CMakeFiles/logic_map_objects.dir/tribes/market.cc.o
In file included from /usr/include/c++/12.1.0/bits/shared_ptr_atomic.h:33,
                 from /usr/include/c++/12.1.0/memory:78,
                 from /mnt/storage/Programme/widelands-dev/src/logic/map_objects/tribes/market.h:22,
                 from /mnt/storage/Programme/widelands-dev/src/logic/map_objects/tribes/market.cc:19:
In member function ‘_PTp* std::__atomic_base<_PTp*>::load(std::memory_order) const [with _PTp = Widelands::Player]’,
    inlined from ‘std::__atomic_base<_PTp*>::operator __pointer_type() const [with _PTp = Widelands::Player]’ at /usr/include/c++/12.1.0/bits/atomic_base.h:694:20,
    inlined from ‘std::atomic<_Tp*>::operator __pointer_type() const [with _Tp = Widelands::Player]’ at /usr/include/c++/12.1.0/atomic:433:16,
    inlined from ‘Widelands::Player* Widelands::MapObject::get_owner() const’ at /mnt/storage/Programme/widelands-dev/src/logic/map_objects/map_object.h:252:10,
    inlined from ‘void Widelands::Market::traded_ware_arrived(int, Widelands::DescriptionIndex, Widelands::Game*)’ at /mnt/storage/Programme/widelands-dev/src/logic/map_objects/tribes/market.cc:254:41:
/usr/include/c++/12.1.0/bits/atomic_base.h:820:31: error: ‘long unsigned int __atomic_load_8(const volatile void*, int)’ writing 8 bytes into a region of size 0 overflows the destination [-Werror=stringop-overflow=]
  820 |         return __atomic_load_n(&_M_p, int(__m));
      |                ~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
```
and
```
[ 98%] Building CXX object src/economy/CMakeFiles/economy.dir/idleworkersupply.cc.o
In file included from /mnt/storage/Programme/widelands-dev/src/economy/supply.h:22,
                 from /mnt/storage/Programme/widelands-dev/src/economy/economy.h:26,
                 from /mnt/storage/Programme/widelands-dev/src/economy/economy.cc:19:
In member function ‘void Trackable::Tracker::deref()’,
    inlined from ‘BaseTrackPtr::~BaseTrackPtr()’ at /mnt/storage/Programme/widelands-dev/src/economy/trackptr.h:110:19,
    inlined from ‘TrackPtr<Widelands::Supply>::~TrackPtr()’ at /mnt/storage/Programme/widelands-dev/src/economy/trackptr.h:164:27,
    inlined from ‘Widelands::RequestSupplyPair::~RequestSupplyPair()’ at /mnt/storage/Programme/widelands-dev/src/economy/economy.cc:766:8,
    inlined from ‘void std::push_heap(_RAIter, _RAIter, _Compare) [with _RAIter = __gnu_cxx::__normal_iterator<Widelands::RequestSupplyPair*, vector<Widelands::RequestSupplyPair> >; _Compare = Widelands::RequestSupplyPair::Compare]’ at /usr/include/c++/12.1.0/bits/stl_heap.h:217:5,
    inlined from ‘void std::priority_queue<_Tp, _Sequence, _Compare>::push(const value_type&) [with _Tp = Widelands::RequestSupplyPair; _Sequence = std::vector<Widelands::RequestSupplyPair>; _Compare = Widelands::RequestSupplyPair::Compare]’ at /usr/include/c++/12.1.0/bits/stl_queue.h:741:16,
    inlined from ‘void Widelands::Economy::process_requests(Widelands::Game&, Widelands::RSPairStruct*)’ at /mnt/storage/Programme/widelands-dev/src/economy/economy.cc:845:27:
/mnt/storage/Programme/widelands-dev/src/economy/trackptr.h:58:27: error: pointer may be used after ‘void operator delete(void*)’ [-Werror=use-after-free]
   59 |                         --refcount_;
      |                           ^~~~~~~~~
In member function ‘void Trackable::Tracker::deref()’,
    inlined from ‘void Trackable::Tracker::deref()’ at /mnt/storage/Programme/widelands-dev/src/economy/trackptr.h:56:8,
    inlined from ‘BaseTrackPtr::~BaseTrackPtr()’ at /mnt/storage/Programme/widelands-dev/src/economy/trackptr.h:110:19,
    inlined from ‘TrackPtr<Widelands::Supply>::~TrackPtr()’ at /mnt/storage/Programme/widelands-dev/src/economy/trackptr.h:164:27,
    inlined from ‘Widelands::RequestSupplyPair::~RequestSupplyPair()’ at /mnt/storage/Programme/widelands-dev/src/economy/economy.cc:766:8,
    inlined from ‘void std::push_heap(_RAIter, _RAIter, _Compare) [with _RAIter = __gnu_cxx::__normal_iterator<Widelands::RequestSupplyPair*, vector<Widelands::RequestSupplyPair> >; _Compare = Widelands::RequestSupplyPair::Compare]’ at /usr/include/c++/12.1.0/bits/stl_heap.h:215:23,
    inlined from ‘void std::priority_queue<_Tp, _Sequence, _Compare>::push(const value_type&) [with _Tp = Widelands::RequestSupplyPair; _Sequence = std::vector<Widelands::RequestSupplyPair>; _Compare = Widelands::RequestSupplyPair::Compare]’ at /usr/include/c++/12.1.0/bits/stl_queue.h:741:16,
    inlined from ‘void Widelands::Economy::process_requests(Widelands::Game&, Widelands::RSPairStruct*)’ at /mnt/storage/Programme/widelands-dev/src/economy/economy.cc:845:27:
/mnt/storage/Programme/widelands-dev/src/economy/trackptr.h:60:40: note: call to ‘void operator delete(void*)’ here
   61 |                                 delete this;
      |                                        ^~~~
```

andy5995:
Using Ubuntu 22.04 and g++12.0.1 in a docker container, here's the err output. I think it matches what the CI is doing (build stops at 76%).

I'm configuring with `CC=gcc-12 CXX=g++-12 cmake .. -DCMAKE_BUILD_TYPE:STRING=Release -DOPTION_BUILD_TRANSLATIONS="ON" -DOPTION_BUILD_WEBSITE_TOOLS="ON" -DOPTION_ASAN="ON" -DOPTION_BUILD_CODECHECK="OFF"`

[gcc12_err_out.txt](https://github.com/widelands/widelands/files/8695782/gcc12_err_out.txt)

>g++12 Debug builds work now. g++12 Release builds fail with various warnings from included **system headers** (`std_function.h`, `atomic_base.h`), does anyone know why this happens?

I don't see any references to atomic_base.h but I do for std_function.h. Seems to be most related to the warnings about using variables that may be uninitialized.

This may be a situation where using [fanalyzer](https://github.com/widelands/widelands/discussions/5423) might be useful to help track down why the variables may be uninitialized

andy5995:
Here's the output when fanalyzer is used:

[gcc12_fanalyzer_err_02.log](https://github.com/widelands/widelands/files/8695809/gcc12_fanalyzer_err_02.log)

andy5995:
> I do not get these errors with gcc 12.1.0 (CI uses 12.0.1). Instead I get two errors which I only could resolve by using local GCC_DIAG_OFF, but then it compiled (both debug and release):
> 

When I tried on Arch in a docker container, using 12.1.0, the release build fails at 76%

[arch_gcc12.1.0_01_err.txt](https://github.com/widelands/widelands/files/8695890/arch_gcc12.1.0_01_err.txt)
[stdout_arch_gcc12.1.0_01.log](https://github.com/widelands/widelands/files/8695891/stdout_arch_gcc12.1.0_01.log)

andy5995:
I know my knowledge of cpp is limited, but I've not a clue as to why when I initialize `s` to `""` (on L33) in this example, the error still happens:

```
/wl/src/base/md5.cc: In member function ‘std::string Md5Checksum::str() const’:
/wl/src/base/md5.cc:41:16: error: use of uninitialized value ‘<unknown>’ [CWE-457] [-Werro
r=analyzer-use-of-uninitialized-value]
   41 |         return s;
      |                ^
  ‘std::string Md5Checksum::str() const’: events 1-3
    |
    |   32 | std::string Md5Checksum::str() const {
    |      |             ^~~~~~~~~~~        ~~~~~
    |      |             |                  |
    |      |             |                  (2) region created on stack here
    |      |             (1) entry to ‘Md5Checksum::str’
    |   33 |         std::string s = "";
    |      |                         ~~
    |      |                         |
    |      |                         (3) calling ‘std::__cxx11::basic_string<char>::basic_
string’ from ‘Md5Checksum::str’
```

andy5995:
If it makes you feel any better, I just added a test for gcc-12 on Jammy for one of my small projects. I've had the fanalyzer flag enabled for a while and it's been clean. With 12, not so much.

https://github.com/theimpossibleastronaut/rmw/runs/6443383022?check_suite_focus=true#step:3:239

andy5995:
After reviewing the new warnings for my own project, [I've concluded they're false positives](https://github.com/theimpossibleastronaut/rmw/issues/343).

So I disabled them in https://github.com/theimpossibleastronaut/rmw/commit/998a405d4325385113eb2d2966581d658c74e1b4

@Noordfrees As long as widelands is checking with clang and other versions of g++, it would probably be safe to do the same for the warnings about possible use of uninitialized values, using a condition to check if the compiler is gcc && version 12 or greater. And then try enabling the warnings again when gcc-12.2.0 is released.

andy5995:
> I do not get these errors with gcc 12.1.0 (CI uses 12.0.1). Instead I get two errors which I only could resolve by using local GCC_DIAG_OFF, but then it compiled (both debug and release):
> 

Looks like that can be done by adding a pragma to a header but in this example would affect all versions of gcc https://stackoverflow.com/questions/29320555/how-to-disable-wmaybe-uninitialized-warning

.github/workflows/build.yaml
```diff
@@ -131,15 +127,13 @@ jobs:
           - Release
           - Debug
         os:
-          - ubuntu-20.04
+          - ubuntu-22.04
         compiler:
```
andy5995: You should be able to do something like `compiler: [clang-14, clang-13, g++-12, g++-11]` and remove the 4 lines below.

.github/scripts/install_deps.sh
```diff
@@ -0,0 +1,42 @@
...
+  libasio-dev \
+  libglew-dev\
+  libpng-dev  \
+  libsdl2-dev \
+  libsdl2-image-dev \
+  libsdl2-mixer-dev \
+  libsdl2-ttf-dev\
+  python3\
+  zlib1g-dev \
+  ${ADD_PKG_LIST}
```
andy5995, line 42:
```suggestion
  ${ADD_PKG_LIST}
  
  exit 0
```
Good practice to include an exit statement?

src/base/log.cc
```diff
@@ -177,7 +177,7 @@ void do_log(const LogType type, const Time& gametime, const char* const fmt, ...
 	assert(logger != nullptr);
 
 	// message type and timestamp
-	char buffer_prefix[32];
+	char buffer_prefix[256];
```
andy5995, line 180:
```suggestion
	char buffer_prefix[sizeof "[24:00:00.000 xxxx] XXXXXXX:"];
```

andy5995 (commented):
Just some minor stuff I suggested really. Seems like @Noordfrees has this all under control. LGTM!

.github/workflows/build.yaml
```diff
@@ -131,15 +127,13 @@ jobs:
...
-          - clang-7
-          - clang-6.0
+          - clang-14
+          - clang-13
+          - g++-12
           - g++-11
-          - g++-10
         include:
           # Some compilers need an older ubuntu version to run correctly
           - compiler: g++-5
```
andy5995, line 132: Also, I think you can replace all these lines up to 162 with

```
          - compiler: [g++-5, g++-6]
            os: ubuntu-18.04
            config: [Debug, Release]
          - compiler: [clang-6.0, clang-7]
            os: ubuntu-20.04
            config: [Debug, Release]
```

src/base/log.cc
Noordfrees: If we want to change the format in the future, even for something small, then we'd need to update two places instead of one. It's just one stack-allocated array with short lifespan so this doesn't have any impact resources-wise; also the gametime can have 3 or more digits in the hour in which case this could overflow. So IMHO better to allow some extra space.

.github/scripts/install_deps.sh
Noordfrees: We don't do this in any other script, and it doesn't add any value IMHO since it's implicit

matthiakl (approved):
Looks ok for now. Let's have it, so everyone can compile again.

> This compiler is known to cause false-positive warnings.

Do you know of already reported similar issues? Is this something the compiler devs are already aware of?

Noordfrees:
Thanks for the review :) 

The GCC C++ bug tracker lists several recent regressions around warnings as well as several reports about spurious warnings in v12 so this seems to be a known (and common) problem.

svenstaro:
Can we get a new release with this? Widelands doesn't currently compile in Arch because of this and the patch doesn't apply cleanly.

matthiakl:
I'm not sure what you need. The current development version does compile fine (there is also an PKGBUILD for it in the [AUR](https://aur.archlinux.org/packages/widelands-git)). The last release (1.0) is provided as prebuilt package for many platforms and should not be needed to be compiled. If you still want to compile it, simply install and use gcc11 package.

svenstaro:
Distributions don't usually use prebuilt binaries if it can at all be helped. We like to compile things from source so that we can be sure the all binaries are using the same binary hardening, for instance. 

Also in Arch Linux, we don't really have a way to go back to old compilers usually. As such, I'd still appreciate a new release. Otherwise if that is not possible, I could look into building a recent commit and shipping that but I'd rather not if it can be helped.

Noordfrees:
We're already in feature freeze for the v1.1 release. There are still some open PRs and issues targeted to v1.1 that need testing and review. I expect the release to happen in late summer or early autumn this year.

What exactly is the problem when building v1.0 with latest compilers? Would patching the CMakeLists.txt to remove the `-Werror` flag fix things for you?

svenstaro:
Yeah that'd work and that's what I would have gone with anyway in case you couldn't get a release out short term. We try not to mess with upstream compile options if it can be helped. :)

PR #5421, opened 2022-05-14
Fix save dialog and debug window text input bug

### LINKED ISSUE #5390: Filename input box in save game menu doesn't have initial focus
**Describe the bug**
When opening the menu for saving a game, the filename input box doesn't truly have focus.
The caret is blinking as if the box is focused, yet nothing happens when characters are typed. I need to click on the box before it actually starts accepting input.

Note that this behaviour does not occur if the save game menu is instead opened using keyboard shortcut (Ctrl+ S); in that case the field is immediately accepting input as expected.

**To Reproduce**
1. Start/load any game
2. Click 'main menu'
3. Click 'save game'
4. The filename text box is blinking as if ready for input
5. Type a character. Nothing happens.

**Version:**
 - OS: Windows 11
 - Widelands Version: recent master

akupar:
Fixes #5390 without causing key presses to passing to game by making the save dialog window modal. Also disables it's pinning that becomes useless.


Noordfrees:
Fix confirmed. But please revert the pin button change, it does not do any harm to have the window pinnable and not being able to pin it just introduces an inconsistency.

jmoerschbach:
Hmm, this reintroduces https://github.com/widelands/widelands/pull/5261#issuecomment-1043323031
In editor when saving the map, this problem does not exist, as the edit box is not focused but some item in the table (which does not reflect in the editbox but that is yet another issue)

akupar:
> Fix confirmed. But please revert the pin button change, it does not do any harm to have the window pinnable and not being able to pin it just introduces an inconsistency.

Sure. I disabled pinning because, if the player wants to keep the window open while they play, it is not possible anymore. Although I doubt anyone actually does that.

akupar:
> Hmm, this reintroduces #5261 (comment).

Confirmed.

IIRC, when I was debugging this, this (the focus issue) happens because the dropbox disables text input after the editbox gets focus (which would enable text input). So I think one way to fix this would be to have a counter that keeps track of how many controls want text input, and then disable text input only when the counter goes to zero. So that
- drop box opens (++counter = 1)
- dialog opens (++counter = 2)
- drop box closes (--counter = 1)

jmoerschbach:
I think it is easier to close the dropdown (and hence, disabling sdl text input) BEFORE the selected signal is triggered, see changed code as proposal ;)
Needs further testing (just tried the use case in editor with t,p,t,p and the issue you wanted to fix in the first place ;) )

akupar:
> I think it is easier to close the dropdown (and hence, disabling sdl text input) BEFORE the selected signal is triggered, see changed code as proposal ;)

I actually thought, that disabling text input globally shouldn't be in a control, but at a higher level, but I think that would have been overdoing it. This is much simpler and better. I tested this and it works well!

jmoerschbach (approved):

PR #5422, opened 2022-05-14
combine dependency installation into script

andy5995:


andy5995:
Note that this is a request to merge into https://github.com/widelands/widelands/pull/5415 not master

I figured since @Noordfrees is updating so much in the related files, would be better to suggest it this way.

Noordfrees (approved):

Issue #5407, opened 2022-05-09
Keyboard game menu navigation is broken


attet:
First, the expected behavior from Widelands 1.0; when I navigate to a menu item and press space, it changes from this:

> ![1 0_pre](https://user-images.githubusercontent.com/1911416/167321076-4003459a-d025-4a4d-b013-17602700092e.png)

To this:

> ![1 0_after](https://user-images.githubusercontent.com/1911416/167321073-39efa5e4-0730-47fb-a70a-6ff03e58a6a1.png)

The text changed from 'Show statistics' to 'Hide statistics', and this setting is reflected in game. All is well and good.

Now compare this with the current master. Similarly we start from:
>  ![1 1_pre](https://user-images.githubusercontent.com/1911416/167321079-02ce4386-5b86-4479-8bef-362ca7489cef.png)

By pressing space we get to this:
>  ![1 1_after1](https://user-images.githubusercontent.com/1911416/167321077-45191bbd-b369-4d9d-a614-c933ec0d8d97.png)

Notice that:

- the menu oddly gets quite a bit wider (Some underlying text width gets recalculated?)
- the selection jumps to the topmost item
- most importantly the desired census status change is not triggered (no change in menu text nor in game).

Catastrophically, pressing space once more brings us to this weird state where the menu is there but all content is gone:
>  ![1 1_after2](https://user-images.githubusercontent.com/1911416/167321078-f3a1b84c-5da3-4ad8-a8e1-8116b3f58955.png)

**Version:**
 - OS: Windows 11
 - Widelands Version:  1.1~git25752[993b193@master] (Release)

attet:
Just noticed that the overall main menu suffers from similar problems. E.g. in 'Single Player...' menu the space press makes Campaigns and Tutorials options disappear (and the whole menu is gone on second press):
![mainmenu1](https://user-images.githubusercontent.com/1911416/167323730-4cd256a0-9fe6-4d8e-a0ea-483c176e2f6e.png)
![mainmenu2](https://user-images.githubusercontent.com/1911416/167323732-c5d6276d-aa75-4da8-8d80-e32abf6fde27.png)

Noordfrees:
The reason for this is that pressing character keys now filters the dropdown for the text you enter. Pressing Space once filters for entries containing a whitespace, pressing it again filters for strings containing two whitespaces in a row.

[...]

PR #5441, opened 2022-06-02
Make space available as dropdown hotkey & only trigger hotkey when not filtered

jmoerschbach:
Re  #5407


jmoerschbach:
hmm, false positive? Without this include it does not compile...

jmoerschbach:
Hmm, I think the find_unused_includes script may have an error.
When a header file is scanned (here: dropdown.h), we look into each header file that is included from the header file to be checked and search for class definitions but not for function definitions.
https://github.com/widelands/widelands/blob/45e168b317f6f025f604f16d8476d758930f72f3/utils/find_unused_includes.py#L228

Changing the boolean param from `False` to `True` fixes the problem but I am not sure if this is smart...?

Noordfrees:
Would perhaps adding the file to `DIFFICULT_FILES` (line 42) solve this?

jmoerschbach:
> Would perhaps adding the file to `DIFFICULT_FILES` (line 42) solve this?

yes, of course :) way better :+1:

src/ui_basic/dropdown.h
```diff
@@ -405,17 +406,32 @@ template <typename Entry> class Dropdown : public BaseDropdown {
...
 	bool check_hotkey_match(const std::string& input_text) {
 		for (auto& x : unfiltered_entries) {
-			if (input_text == to_lower(x.hotkey) && is_in_filtered_list(x.value)) {
-				if (hotkey_fn_) {
-					hotkey_fn_(x.value);
-				} else {
-					verb_log_dbg("hotkey match: %s but no hotkey function available!", x.hotkey.c_str());
+			if (" " == input_text) {
+				SDL_Keycode c = SDLK_SPACE;
+				const std::string& localized_space =
```
```suggestion
				const std::string localized_space =
```
`to_lower` does not return a reference

Noordfrees (commented):
One comment inline, but I'm not sure what this is supposed to do actually. <kbd>Space</kbd> is a reserved button, so it cannot be assigned as a hotkey at all. Selecting the highlighted entry with Space is still not working.

jmoerschbach:
> One comment inline, but I'm not sure what this is supposed to do actually. Space is a reserved button, so it cannot be assigned as a hotkey at all. Selecting the highlighted entry with Space is still not working.

there is a space hotkey set to toggle buildhelp in show/hide menu. On master, hitting space modifies the filter, on this branch the hotkey is matched, toggling the buildhelp and closing the dropdown, see screenshot (master):
spaceNoHotkey

If we want to select with space then this hotkey cannot be assigned to space (and nowhere else) imho (as said in https://github.com/widelands/widelands/issues/5407#issuecomment-1140499258)

Noordfrees:
I forgot about that one. Fix confirmed. I noticed two bugs though: 
- Filtering the dropdown with `E` works fine, but filtering it with `O` closes the dropdown instead of filtering it, even though no entry has O as a hotkey. Similarly, filtering with only `D` or only `A` works fine but filtering for the text `DA` closes it.
  For reference here the translated dropdown:
  image

- Hotkeys are interpreted even while the dropdown is already being filtered, which seems strange to me. E.g. in the above dropdown typing `G` `E` `W` starts filtering first for `G`, then for `GE`, and then it selects the workarea overlap entry instead of appending to the filter.

jmoerschbach:
> 

fixed the second issue.
The first issue is on master as well and is not related to this pr. Should be investigated separately

jmoerschbach:
does anybody know what is wrong here?

frankystone:
~~~~
The following packages have unmet dependencies:
 udev : Breaks: systemd (< 249.11-0ubuntu3.3) but 249.11-0ubuntu3.1 is to be installed
E: Error, pkgProblemResolver::Resolve generated breaks, this may be caused by held packages.
Error: Process completed with exit code 100.
~~~~

Looks like the repo is not in sync.  I try to rerun the failed jobs, maybe this helps.

matthiakl:
> 
> Looks like the repo is not in sync. I try to rerun the failed jobs, maybe this helps.

See #5467

Noordfrees (approved):

PR #5442, opened 2022-06-03
Unmark Amazons as Experimental

Noordfrees:


hessenfarmer (approved):
ok. Although they are not balanced against the other tribes I think we can unmark them as zhey are well playabale for themselves.

PR #5443, opened 2022-06-03
Add highlighting of filter text in dropdown list

### LINKED ISSUE #5444: Clearing dropdown filter via backspace not working for special characters
**Describe the bug**
Clearing dropdown filter via backspace not working for special characters

**To Reproduce**
Language is German
1. Click Start "new random game"
2. Open "Clima" Menu and filter via "ä" for "zufällig" - all fine
3. Hit backspace to remove "ä" as filter -> only entries with Umlaute are displayed - a second backspace is needed to clear the filter completely and show all entries as expected

**Expected behavior**
only one hit on backspace is necessary for removing "ä" from filter

**Version:**
 - OS:  Ubuntu 20/04
 - Widelands Version: 1.1~git133 (8b1ac19@master)

**Additional context**
This is for sure caused by UTF encoding... https://github.com/widelands/widelands/pull/4578

akupar:
Adds highlighting of filter text in filtered dropdown list as described [here](https://github.com/widelands/widelands/pull/5040#issuecomment-905821532).

Kuvakaappaus - 2022-06-03 19-58-13


jmoerschbach:
oh nice, thanks for taking this up! :)
Found one bug which clearly is related to https://github.com/widelands/widelands/issues/5444 but the Highlighting gets confused also
widelands

akupar:
This is because at line [545](https://github.com/widelands/widelands/blob/e94e96b1d2dcc8ca0e38c7d78f386e036c4e993b/src/ui_basic/dropdown.cc#L545) of dropdown.cc only the last byte is removed. I fixed it to remove the last character, but there's still an issue with lowercasing. to_lower doesn't change non-ascii characters to lowercase. So in the example picture Ö wouldn't be filtered with ö.

jmoerschbach (approved):
Nice work! :) Let's have it
Fixes #5444
Re #5407

akupar:
I think it would be better to ignore this fix and use [icu::UnicodeString](https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/classicu_1_1UnicodeString.html) for the filter. It is already used in [graphic/text/bidi.cc](https://github.com/widelands/widelands/blob/master/src/graphic/text/bidi.cc) and [graphic/text/rt_render.cc](https://github.com/widelands/widelands/blob/master/src/graphic/text/rt_render.cc). That would fix both the backspace issue and the lowercase issue.

Issue #1083, opened 2019-09-09
AI dismantles militarysites at the border


bunnybot:
The AI frequently dismantles militarysites right at the border, thus losing territory and inviting attacks. Sites are dismantled even if the AI "knows" it´ll lose territory by doing so. 
Observed e.g. with red player in second frisian scenario.

See also bug 1792478 #26.

------------------------------------
Imported from Launchpad using lp2gh.

 * date created: 2018-09-28T11:28:38Z
 * owner: nordfriese
 * the launchpad url was https://bugs.launchpad.net/bugs/1794960

bunnybot:
(by tiborb95)
I ques this is probably because of shortage of soldiers and too much vacant soldier positions....

bunnybot:
(by nordfriese)
I don´t think that´s the cause here. The AI starts with a lot of soldiers, and I could see that it still had several of them left in the HQ and no vacant positions at all…

bunnybot:
(by tiborb95)
I see. This behavior is controlled fully by genetic algorithm, so there is not much that can be done in the code, maybe do some C++ debug to see why AI decided to do it. BTW can you estimate the distance from dismantled militarysite to nearest enemy military site?

bunnybot:
(by nordfriese)
Here´s a screenshot: The AI (red) just dismantled the Sentry, losing some territory that is now held by my Outpost. It also dismantles the Tower a few minutes later, which will not lose any territory but is also an obvious mistake.

bunnybot:
(by tiborb95)
Yes, it seems not reasonable.

Among other things AI counts own vs enemy soldiers around the military building that it probably came to conclusion that it had odds 4:1 so it is fine to dismantle the sentry.
I might had preserve the tower though...

I plan to run another rounds of testing, though I am now very restricted with access to home PC, so it is not simple.... it is not very feasible just to do simple C++ code change...

But generally final result is what matter, not partial action...

Also I presume it was before they started fight and/or AI is not pressed from other side

bunnybot:
(by stephan-lutz)
I have seen this many times now. Best to be seen while watching a 4 players AI only game on several maps. (normally I use full moon for that).
The behaviour is not related to whether the AI has contact with the enemy or not.
Even after having been in fight it dismantles strong military buildings with no obvious reason. 

After b20 is out I will start investigating this any further.

bunnybot:
(by tiborb95)
OK

bunnybot:
(by stonerl)
[...]

PR #5446, opened 2022-06-07
Improving building and dismantling mil. sites

tibor95:
Improving building and dismantling military sites
- creating future_military_loneliness
- modification of can_be_dismantled variable

For discussion, pls check if the behavior is improved. I can train again a bit and than we might merge

re #1083


hessenfarmer:
Code looks reasonable. will test tomorrow.

tibor95:
I will do some more training than and push results shortly after lunchtime

tibor95:
OK. I pushed results of some training

hessenfarmer (approved):
this is definitly a big improvement.   
to see the effect I use a 4 AI, peaceful, endless game on full moon and watch the "clash zone".    
In master the AI is dismantling much to much buildings. While here it only happens in rare occasions (building finished but not manned), which leads to less territory losses.    
@Noordfrees may we have this although we already had feature freeze.

hessenfarmer:
@tothxa as you reported the same bug recently I would like to hear your opinion as well wether this is an improvement as big as I believe it is

tibor95:
I am still doing some training, for case it will be merged...

hessenfarmer:
> I am still doing some training, for case it will be merged...

As @Noordfrees agreed to merge in v1.1 it we will merge when training is finished and @tothxa gave us his opinion. 
I would prefer this to be at the earliest opportunity as we should play all campaigns after AI code is finished.

tothxa:
edit: Sorry, my mistake. I used `--datadir` forgetting about the `.wai` files.

edit2: Do I need to start a new game to get a clean AI state? Using my previous save ([atl02_map0608_see_all.wgf.zip](https://github.com/widelands/widelands/files/8874496/atl02_map0608_see_all.wgf.zip)) I still see strange dismantling and rebuilding behaviour.

~Running this with atl02 (taking advantage of #5447) I still see this:~

shot0003
shot0004

~Dangerous behaviour from the Empire too later, though not losing land this time:~

shot0006

~Then some spurious building and dismantling:~

shot0008

~In the first test the Empire also lost land when it dismantled its fortress, but I hadn't made screenshots of that one.~

tibor95:
content of wai is stored in savegame, so no point changing them. And they themselves are not cardinal now.
New build of widelands should be enough to see the difference.

I can test your savegame, but just confirm you have new build...

tothxa:
Yes, I tested with a release build of 1.1~git25780 (1bc8a10@ms_fix). For the second test I checked out `data/campaigns/atl02.wmf` from atlantean02 over it to be able to load the savegame.

tothxa:
I re-ran the test, but this time there was less spurious dismantle and rebuild. The barbarians still dismantle the citadel and fortresses, losing much territory and several production sites, but this time the empire only built some medium sites to replace the fortress, and left it alone for a while. (Yesterday it kept building and dismantling fortresses.)  Later it expanded with a new fortress, then dismantled it when the military strength of the barbarians surpassed them. So the worse dismantles only happened when the respective player was weaker overall, so they can be interpreted as strategical retreat.

hessenfarmer:
Well from what I saw in my usual test scenarios, the behaviour is not perfect but much better. First of all the AI does not dismantle due to constructionsites anymore, which caused much of the territory losses.     
 I will check atlantean 2 but this isn't a perfect example as there are many military sites in one region which still might be cosidered to be superfluous and dismanteled. we should probably extent the size bonus of big military buildings to the buildings at the front line as well, as it is currently only regarded in inner land decisions, to prevent more territory losses. 
In the current state dismantleing will stop after there are only a few buildings left at the border resulting in high enough military loneliness. 
From my perspective we should have this improvement now and keep the bug open to work on the issue a bit further in 1.2 
Still this is much better then current master I beleive

tibor95:
I would also vote for merging this - AI improvement is endless task anyway... But I would still run some training. How much time can I have for it?

tibor95:
So I would say that should_be_dismantled should be more sensitive to  enemy_owned_land_nearby, not only yes/no, but how many fields...
and unowned_land_nearby as well....

tothxa:
I started a game of atl02 (with the `destruction_blocked` lines removed) with this branch and this time the Barbarians too kept one fortress at the border with the Empire. I didn't watch them closely this time, because I also tried to expand towards the Barbarians to see what they'd do at the border with me, but I didn't get to watch them much because their western border was soon cut off from the main territory by the Empire. But at least in this time they didn't dismantle the military sites near my border. So there is some improvement, but the test I tried to do was inconclusive.

tibor95:
I just pushed some training.
I would make some further changes, but probably it is too late for the release. So it is up to you. To merge or not.

hessenfarmer:
> I just pushed some training. I would make some further changes, but probably it is too late for the release. So it is up to you. To merge or not.

if you already have an idea what to change feel free to do so.

tibor95:
I have an idea, and can prepare something in a week, but I would prefer to have this in master. If you think this change is improvement.

hessenfarmer:
as this is definitly an improvement I'll merge this now.

PR #5449, opened 2022-06-12
Rework of MS dismantlement (should_be_changed calculation)

tibor95:
I prepared some change, this is very basic idea, to provide some list of input variables and let genetic algorithm cope with them.
We can add more.
I did some training also.
You can try, it seems to me that at macro level it behaves well - I mean overall growth of territory and fighting...

re #1083


src/ai/defaultai_warfare.cc
```diff
@@ -985,30 +985,95 @@ bool DefaultAI::check_militarysites(const Time& gametime) {
...
+
+	// available military numbers: 77, 99, 91, 84, 24, 89, 88
+	// available neuron pools: 13,14, 28,29, 30, 45 - 48, 57 - 59
+
+	// We calculate the score - if positive, site will be dismounted
+	// We starting with positive number, most modificators are decreasing th score
+	int32_t dismantle_score = 3 * management_data.get_military_number_at(91) / 2 + 100;
+	dismantle_score -= 2 * management_data.neuron_pool[13].get_result_safe(
+		                        bf.enemy_owned_land_nearby / 2, kAbsValue);
+	dismantle_score -= (bf.enemy_owned_land_nearby) ?
```
hessenfarmer: the bool expression needs to be expressed explicitly as bool.
i.e. bf.enemy_owned_land_nearby > 0 
otherwise clang-tidy should complain

src/ai/defaultai_warfare.cc
```diff
@@ -985,30 +985,95 @@ bool DefaultAI::check_militarysites(const Time& gametime) {
...
+                         std::abs(management_data.get_military_number_at(77)) :
+                         -std::abs(management_data.get_military_number_at(77));
+	dismantle_score -=
+	   management_data.neuron_pool[14].get_result_safe(bf.unowned_land_nearby / 2, kAbsValue);
+	dismantle_score -= (bf.unowned_land_nearby) ?
```
hessenfarmer: see above

src/ai/defaultai_warfare.cc
```diff
@@ -985,30 +985,95 @@ bool DefaultAI::check_militarysites(const Time& gametime) {
...
+	// 	const uint16_t size_bonus =
+	// 	   total_capacity * std::abs(management_data.get_military_number_at(89)) / 5;
+	// 	if (bf.military_score_ + size_bonus < management_data.get_military_number_at(88) * 5 &&
+	// 	    bf.future_area_military_capacity > static_cast<int16_t>(total_capacity)) {
+	// 		should_be_dismantled = true;
+	// 	}
+	// }
+
+	if (can_be_dismantled) {
+		printf("dismantle score: %4d, can and should be dismantled: %s\n",
```
hessenfarmer: please use verb_log_debg function for such cases.

hessenfarmer:
the concept seems to be better then the current one. Will test it.   
some small general remarks inline.

hessenfarmer:
furthermore during review / analysis I found out that "military_last_dismantle_" is only defined but not used. It might be a good idea to use it to ensure the Buioding fields have been updated before the next dismantle decision. (i saw the AI dismantling 2 adjacent Milsites together which should not have been the case if the second site would have considered the first already being dismantled).

tibor95:
military_last_dismantle_ - yes, good comment, I will make a change

src/ai/defaultai_warfare.cc
tibor95: I plan to remove this altogether, but maybe can stay as debug... I will see

src/ai/defaultai_warfare.cc
tibor95: got it

src/ai/defaultai_warfare.cc
tibor95: done

tibor95:
Please retest and if OK, I will clean up the code, remove commented out parts and so on

hessenfarmer:
ok I will test. However one small thing. In the past we had taken the size of the MS into account for dismantling. This was based on the maximum number of soldiers a site could accomodate ("total_capacity") Would it be possible to include this in the decision in a way that bigger MS would be a bit discouraged from dismantling.

tibor95:
We might add to score:
dismantlement_score -= total_capacity * 2 - 8
So bigger capacity => less willing to dismantle
I added 8 - to kind of normalize it, It was trained without this input so to normalize its effect... Site with capacity 4 soldiers will have no impact on the score

tibor95:
Maybe we should shrink 45 seconds to like 30

src/ai/ai_help_structs.cc
```diff
@@ -355,7 +355,7 @@ BuildableField::BuildableField(const Widelands::FCoords& fc)
      ground_water(1),
      space_consumers_nearby(0),
      rangers_nearby(0),
-     area_military_capacity(0),
+     future_area_military_capacity(0),
```
hessenfarmer, line 359: shouldn't we set area_military_capacity to zero here as well.

src/ai/defaultai_warfare.cc
```diff
@@ -981,32 +981,71 @@ bool DefaultAI::check_militarysites(const Time& gametime) {
...
+	// bf.own_military_presence - increases dismantle probability
+
+	// bf.enemy_military_presence - decreases dismantle probability
+	// bf.enemy_military_sites - - decreases dismantle probability
+
+	// We calculate the score - if positive, site will be dismounted
+	// We starting with positive number, most modificators are decreasing th score
+	int32_t dismantle_score = 3 * management_data.get_military_number_at(91) / 2 + 100;
+	dismantle_score -= 2 * management_data.neuron_pool[13].get_result_safe(
+	                          bf.enemy_owned_land_nearby / 2, kAbsValue);
```
hessenfarmer: enemy_owned_land or unowned_land can get quite big as the check radius is 16 + genetic (currently 19). So normalizing with 2 might be to small. 
with the current formula all MS in range 1-13 from the border will get the same (full weight) result.

I would propose to normalize by 10 (or even better by magicnumber / 5 and multiply the neuron result with 3  instead of 2.

src/ai/defaultai_warfare.cc
```diff
@@ -981,32 +981,71 @@ bool DefaultAI::check_militarysites(const Time& gametime) {
...
+	dismantle_score -= (bf.enemy_owned_land_nearby > 0) ?
+                         std::abs(management_data.get_military_number_at(77)) :
+                         -std::abs(management_data.get_military_number_at(77));
+	dismantle_score -=
+	   management_data.neuron_pool[14].get_result_safe(bf.unowned_land_nearby / 2, kAbsValue);
```
hessenfarmer: see above the same applies here. we could use the same magic number as for enemy land as Normalizing factor. And may add a multiplicator of 3/2.

src/ai/defaultai_warfare.cc
```diff
@@ -981,32 +981,71 @@ bool DefaultAI::check_militarysites(const Time& gametime) {
...
+	dismantle_score +=
+	   management_data.neuron_pool[28].get_result_safe(bf.area_military_capacity, kAbsValue);
+	dismantle_score +=
+	   management_data.neuron_pool[57].get_result_safe(bf.future_area_military_capacity, kAbsValue);
+	dismantle_score -=
+	   management_data.neuron_pool[29].get_result_safe(bf.military_loneliness / 50, kAbsValue);
+	// dismantle_score -= management_data.neuron_pool[30].get_result_safe( // NOCOM uncomment this
+	//	                        bf.future_military_loneliness, kAbsValue);
+	if (soldier_status_ == SoldiersStatus::kBadShortage) {
+		dismantle_score += 2 * std::abs(management_data.get_military_number_at(77));
```
hessenfarmer: here we should use different magic number as the number has been used above candidates are 24, 84, 88 or 89 as they were used previously.

src/ai/defaultai_warfare.cc
```diff
@@ -981,32 +981,71 @@ bool DefaultAI::check_militarysites(const Time& gametime) {
...
+	} else if (soldier_status_ == SoldiersStatus::kShortage) {
+		dismantle_score += std::abs(management_data.get_military_number_at(77));
```
hessenfarmer: here we should use a different number as well (meaning different to 77 and the number used for badShortage

src/ai/defaultai_warfare.cc
```diff
@@ -981,32 +981,71 @@ bool DefaultAI::check_militarysites(const Time& gametime) {
...
+	}
+	dismantle_score -=
+	   management_data.neuron_pool[45].get_result_safe(bf.military_score_ / 50, kAbsValue);
+	dismantle_score +=
+	   management_data.neuron_pool[46].get_result_safe(bf.own_military_presence, kAbsValue);
+	dismantle_score -=
+	   management_data.neuron_pool[47].get_result_safe(bf.enemy_military_presence, kAbsValue);
+	dismantle_score -=
+	   management_data.neuron_pool[48].get_result_safe(bf.enemy_military_sites * 3, kAbsValue);
+	dismantle_score -= total_capacity * 2 - 8;
```
hessenfarmer: we should have this dependent on genetics.

How about 
```suggestion
	dismantle_score -= total_capacity * std::abs(management_data.get_military_number_at(89)/20 - std::abs(management_data.get_military_number_at(88)/5;
```

hessenfarmer (changes requested):
Some comments inline.

hessenfarmer:
I did some tests with the atlantean 2 scenario, but I found out it won't work in this scenario, as the calculation of enemy owned land does not take place as the empire and the barbarians are not counted as enemies as they are not allowed to attack each other by scripting in the beginning. 
I will do more tests without this scripting to see what happens, but this will take time as I need to prepare some instrumented version of this branch

src/ai/defaultai_warfare.cc
tibor95: After so many changes retraining will be needed so we do not need a replacement for '8', so this would be enough:

dismantle_score -= total_capacity * std::abs(management_data.get_military_number_at(89)/20

tibor95:
In the middle of the next week - I will not work on this on weekend and some training will be needed.

src/ai/defaultai_warfare.cc
hessenfarmer: well I liked the formula as it is discouraging dismantling buildings bigger then x and courages dismantling these who have less then  x. Therefore I would love to keep the part after the minus

src/ai/ai_help_structs.cc
tibor95: of course

src/ai/defaultai_warfare.cc
tibor95: changed to 10, should be fine

src/ai/defaultai_warfare.cc
tibor95: good catch, using 24

src/ai/defaultai_warfare.cc
tibor95: This is intentional - bad shortage increases the score twice as much as normal shortage. I would leave it as is.
Or I would suggest something like that:

```
if (soldier_status_ == SoldiersStatus::kBadShortage) {
		dismantle_score += std::abs(management_data.get_military_number_at(77));
                dismantle_score += std::abs(management_data.get_military_number_at(XX));
	} else if (soldier_status_ == SoldiersStatus::kShortage) {
		dismantle_score += std::abs(management_data.get_military_number_at(77));
```
Just say if you prefer this

src/ai/defaultai_warfare.cc
tibor95: No, considering that '*' is calculated first and '-' then, that second magic number should not need to be here. Dismantle score is initialized with some value, I would just ignore second magic number - this one would just increase score by some number 0-20. Logically it has no connection to total_capacity.
So I would suggest this:

`dismantle_score -= total_capacity * (1 + std::abs(management_data.get_military_number_at(89)/20);`

tibor95:
I left some points open. Once we agree on code, I will start training...

src/ai/defaultai_warfare.cc
hessenfarmer: I meant to use 24 as magic number for bad shortage and another number (e.g.) for shortage to let the genetics decide on the real relation between both values rather then fix it  to 2. just to prevent fixed numbers and make it as succeptible as possible to genetics

src/ai/defaultai_warfare.cc
hessenfarmer: this would just decrease the value for all buildings. What I liked with the original formula is that small building below the threshhold defined by the part after the minus will be encouraged to dismantle and bigger ones above threshhold will be discouraged from dismantling.  
Edit but I have no strong feelings about this

src/ai/defaultai_warfare.cc
tibor95: This means that bad shortage can get smaller malus then normal shortage - while genetics can cope with this and no tragedy it would be, I just think we should not even allow such possibility

src/ai/defaultai_warfare.cc
tibor95: But to me this is a duplicity to what is in initialization of the variable:

int32_t dismantle_score = 3 * management_data.get_military_number_at(91) / 2 + 100;

If AI 'thinks' some constant need to be added/deleted to/from the score let it change it here.

Further modification of the score should be really impacts of other variables in the game

hessenfarmer: ok, now I get it. You are right sorry for my misunderstanding.

src/ai/defaultai_warfare.cc
hessenfarmer: we still could keep a factor of 2 in front of the kBadShortage number. However it should work with using 1 number as well.
I'll leave it to your decision

tibor95: OK

tibor95:
I think everything has been addressed
I did some training, I will do more, if the code is fine now

hessenfarmer:
code lGTM. Please merge master before doing the training. Finally I think we should have 4 identical files for the release. 
@Noordfrees may we have this for the release? if otherwise we should do some training on master first to be prepared for the release.
You can mark the branch as ready for review now I think. I will do a final test after training is finished.

tibor95:
Ok, I will train and push before tomorrow afternoon

hessenfarmer:
> Ok, I will train and push before tomorrow afternoon

fine. please fix the codechaeck warnings as well

tibor95:
Done, I will check results of checks later

tibor95:
Anything else to do from me? I merged master, but probably there is new commit there today

hessenfarmer:
> Anything else to do from me? I merged master, but probably there is new commit there today

no, fine from your side thank you

hessenfarmer (approved):
code LGTM.
testing looks promising as well.

hessenfarmer:
@Noordfrees many thanks for accepting this, as this might clear 90% of the related bug. 
In fact I only kept it open, because it is hard to say whether such behaviour can be fully avoided, as human players might have different views on when to dismantle.    
Furthermore there is still an issue in some scenarios, which I will discuss in the bug

PR #5457, opened 2022-06-20
add ai basic decision algorithm for accepting diplomacy

hessenfarmer:
- don't accept anything in the first 10 + random minutes
- only accept offers from stronger players

### DIFF
--- a/src/ai/ai_help_structs.h
+++ b/src/ai/ai_help_structs.h
@@ -107,6 +107,7 @@ enum class SchedulerTaskId : uint8_t {
 	kManagementUpdate,
 	kUpdateStats,
 	kWarehouseFlagDist,
+	kDiplomacy,
 	kUnset
 };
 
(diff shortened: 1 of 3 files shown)


Noordfrees (commented):
LGTM, compiling this now.

```
.../src/ai/defaultai.cc:3349: Trailing whitespace at end of line
.../src/ai/defaultai.cc:3358: Trailing whitespace at end of line
```

---

```c++
accept = accept && foo
```
can be better written as
```c++
accept &= foo
```

tibor95:
looks ok. 
So effect of this is that players will become members of the same team?

hessenfarmer:
> looks ok. So effect of this is that players will become members of the same team?

yes, that is the effect of the new diplomacy feature

tibor95 (approved):
Looks 'safe' from AI's standpoint

hessenfarmer:
@Noordfrees if you are happy as well feel free to merge

Noordfrees (approved):

PR #5465, opened 2022-07-06
compile.sh fix and tweaks

### LINKED ISSUE #5463: update.sh script no longer working
Since a few months, the update script hasn't been working for me any more.
Historically, by invoking ./update.sh I could update the game binary (it used to fetch the current version from git, and then recompiled anything that had changed).

Now I see something like.:

~~~~
philipp@notebook6:~/Spiele/widelands$ ./update.sh
…
 po/widelands/zh_TW.po                              |  20 +-
 169 files changed, 10241 insertions(+), 9997 deletions(-)
Call -j/--cores with a number, e.g. '-j 16'
~~~~

I've seen this bug on multiple systems (all of them Debian GNU/Linux on amd64).

Though I wonder if the bug is really in update.sh vs. compile.sh. The line that calls compile.sh from my update.sh is:

~~~~
./compile.sh --cores 15 --default-compiler --with-cross-opt --with-website --with-translations --do-tests --release --no-asan --no-tsan --with-xdg
~~~~

This is the update.sh I got after getting current git a few days ago, and then invoking ./compile-sh -r once.

tothxa:
fixes #5463

Plus adds --dry-run option for testing and breaks up the cmake command line for readibility. I'll revert them if they are too much for this PR.

### DIFF
--- a/compile.sh
+++ b/compile.sh
@@ -580,7 +580,8 @@ buildtool="" #Use ninja by default, fall back to make if that is not available.
       return 0
     fi
       $RUN rm -f update.sh || true
-      $RUN cat > update.sh << END_SCRIPT
+      if [ -z "$RUN" ]; then
+        cat > update.sh << END_SCRIPT
 #!/bin/sh
 echo "################################################"
 echo "#            Widelands update script.          #"
@@ -606,8 +607,17 @@ echo "#      Widelands was updated successfully.     #"
 echo "# You should be able to run it via ./widelands #"
 echo "################################################"
 END_SCRIPT
+
+      else  # dry run
+        echo 'cat > update.sh <<END_SCRIPT'
+        echo "   # ..."
+        echo "   $COMMANDLINE"
+        echo "   # ..."
+        echo "END_SCRIPT"
+      fi
+
       $RUN chmod +x ./update.sh
-      if [ $QUIET -eq 0 ]; then
+      if [ $QUIET -eq 0 -a -z "$RUN" ]; then
         echo "The update script has successfully been created."
       fi
   }
--- a/compile.sh
+++ b/compile.sh
@@ -352,6 +352,18 @@ if [ $QUIET -eq 0 ]; then
   fi
 fi
 
+if [ -n "$RUN" ]; then
+  if [ $QUIET -eq 0 ]; then
+    echo " "
+    echo "#################################################################"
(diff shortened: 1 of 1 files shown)


tothxa:
@spth Please test if it works for you.

spth (commented):
I'm not a shell script expert, but I don't see any obvious problems.

I don't know if the $RUN is really useful, and since I don't use cmake, I have no idea about the changes in the call to cmake.

It works for me: in particular it fixes bug #5463.

tothxa:
> It works for me: in particular it fixes bug #5463.

Thank you.

Sorry, I didn't mean to ask you for a full review, only to test the bugfix. :smiley:

compile.sh
```diff
@@ -653,15 +683,17 @@ if [ $BUILD_WEBSITE = "ON" ]; then
...
+  echo "# You should now be able to run Widelands via             #"
+  echo "# typing ./widelands + ENTER in your terminal             #"
+  echo "#                                                         #"
+  echo "# You can update Widelands via running ./update.sh        #"
+  echo "# in the same directory that you ran this script in.      #"
+fi
 echo "###########################################################"
 
-else
+elif [ -z "$RUN" ]; then
```
hessenfarmer, line 708: the original else was related to line 648. It was the only output if quiet was selected. we should keep this but make the same disitnction as in lines 652 to 656

tothxa: Are you sure it's necessary? `--dry-run` already printed the commands of what would have been done, so printing anything here would be a violation of `--quiet` IMO. Only without `--dry-run` makes it sense to report completion for `--quiet`.

hessenfarmer: Well without the quiet option the output was "Congratulations. Widelands has been built successfully." (line 653) followed by all the verbose messages. without it was only "Widelands has been built succesfully" (line 697) without any further messages. Although using dry run with quiet makes not much sense, we should keep the original logic to show one single success line if quiet is selected.

tothxa: OK, I added a success message.

Actually dry run with quiet makes a lot of sense to me, and I have used it several times (I added `--dry-run` first for #5121 and I still use that version of compile.sh locally) to get the `cmake` invocation that `compile.sh` would use. :smiley:

hessenfarmer (approved):
lgtm now

Issue #5469, opened 2022-07-09
Error loading tribes/initialization/atlanteans/images/flag_00_pc.png: Unknown pixel format on sdl2_image 2.6.0


svenstaro:
Widelands 1.0 crashes when compiled and ran against sdl2_image 2.6.0. Startup says this:

```
❯ widelands
This is Widelands Version 1.0 (Release)
[00:00:00.000 real] INFO: Set home directory: /home/svenstaro/.local/share/widelands
[00:00:00.000 real] INFO: Set configuration file: /home/svenstaro/.config/widelands/config
[00:00:00.000 real] INFO: Adding directory: /usr/share/widelands
[00:00:00.000 real] INFO: selected language: (system language)
[00:00:00.000 real] INFO: using locale en_US.UTF-8
[00:00:00.001 real] INFO: Byte order: little-endian
[00:00:00.055 real] INFO: Graphics: Try to set Videomode 1144x263
[00:00:00.169 real] INFO: Graphics: OpenGL: Version "4.6.0 NVIDIA 515.57"
[00:00:00.169 real] INFO: Graphics: SDL_GL_RED_SIZE is 8
[00:00:00.169 real] INFO: Graphics: SDL_GL_GREEN_SIZE is 8
[00:00:00.169 real] INFO: Graphics: SDL_GL_BLUE_SIZE is 8
[00:00:00.169 real] INFO: Graphics: SDL_GL_ALPHA_SIZE is 0
[00:00:00.169 real] INFO: Graphics: SDL_GL_BUFFER_SIZE is 24
[00:00:00.169 real] INFO: Graphics: SDL_GL_DOUBLEBUFFER is 1
[00:00:00.169 real] INFO: Graphics: SDL_GL_DEPTH_SIZE is 24
[00:00:00.169 real] INFO: Graphics: SDL_GL_STENCIL_SIZE is 0
[00:00:00.169 real] INFO: Graphics: SDL_GL_ACCUM_RED_SIZE is 16
[00:00:00.169 real] INFO: Graphics: SDL_GL_ACCUM_GREEN_SIZE is 16
[00:00:00.169 real] INFO: Graphics: SDL_GL_ACCUM_BLUE_SIZE is 16
[00:00:00.169 real] INFO: Graphics: SDL_GL_ACCUM_ALPHA_SIZE is 16
[00:00:00.169 real] INFO: Graphics: SDL_GL_STEREO is 0
[...]

PR #5472, opened 2022-07-17
Added new SDL2-image dependencies to static linking

matthiakl:
Fixes current mingw build failure.
Re #5469


hessenfarmer (approved):
mingw green again.
thanks for fixing

PR #5476, opened 2022-07-25
Military capacity string refactor

hessenfarmer:
first attempt  to fix the regression of #5475


Noordfrees:
This won't work. For example I quote from the wiki the gaelic plural form rule: 
>    "0 cat", "1 chat", "2 chat", "3 cait", ...

Instead, we'll need to state the raw (unformatted) ngettext strings so we have variables like 
```lua
soldier_text_ctxt = "empire_soldier",
soldier_text_1_untranslated_sg = "%1% soldier (+%2%)",
soldier_text_1_untranslated_pl = "%1% soldiers (+%2%)",
soldier_text_1_translated = npgettext("empire_soldier", "%1% soldier (+%2%)", "%1% soldiers (+%2%)", 0),
soldier_text_2_untranslated_sg = "%1% soldier",
soldier_text_2_untranslated_pl = "%1% soldiers",
soldier_text_2_translated = npgettext("empire_soldier", "%1% soldier", "%1% soldiers", 0),
```
The first two values ensure we have the raw, unlocalized, unformatted strings available in C++. The third, unused value ensures that the .po files contain the full loalized ngettext invocation and not just sg/pl strings. In C++ we can then to
```c++
npgettext(soldier_text_ctxt(), soldier_text_1_untranslated_sg(), soldier_text_1_untranslated_pl(), nr_soldiers)
```
which will use the unlocalized unformatted keys to look up the correct ngettext mapping.

src/logic/map_objects/tribes/tribe_descr.cc
```diff
@@ -252,6 +243,21 @@ TribeDescr::TribeDescr(const Widelands::TribeBasicInfo& info,
 		      you need plural forms here, please let us know. */
 		   _("Workers are coming"));
 
+		// TODO(hessenfarmer): Require these strings after v1.2
```
Noordfrees: We shouldn't require them for add-ons forward compatibility

Noordfrees (commented):
:+1:
Not tested yet.

`soldier_singular` and `soldier_plural` can be removed now. The military_capacity scripts can also go.

hessenfarmer:
Not sure whether this works correctly. I tested with the context in npgettext of the military-capacity script, to trigger the usage of the already contained npgettexts in the translations but they did not show up.  
I had not the time to do any further work yesterday on this and probably won't have any time until next week. 
I kept the singular and polural forms as I wanted to build the untranslated strings by them by concatenation in c++. Just for the case that any tribe designer might want to name his soldier different. (e.g. fighter, warrior, etc.).
What is missing as well is the documentation in the amazons units.lua  for these new strings.

feel free to work on this if you have time. otherwise I will look after it next week

Noordfrees:
The reason why it doesn't work is that the existing strings have both different context and different textdomains. I'll look into this now

Noordfrees:
I tested this with custom strings and manually modified PO files, please review

data/tribes/initialization/atlanteans/units.lua
```diff
@@ -197,6 +197,31 @@
 --
 --    **productionsite_workers_coming**: The string to display over this tribe's productionsites when more than 1 worker is coming.
 --
+--    **soldier_context**, **soldier_0_sg**, **soldier_0_pl**, **soldier_1_sg**, **soldier_1_pl**, **soldier_2_sg**, **soldier_2_pl**,
+--    **soldier_3_sg**, **soldier_3_pl**: The pgettext context for militarysites' soldier strings, and the corresponding **unlocalized**
```
hessenfarmer, line 199: Please add some information that the strings must follow exactly the format below and only the name of the soldier can be defined for each tribe. Also add, that the NPGETTEXT entries need to start with the UNUSED_ part

data/tribes/initialization/atlanteans/units.lua
```diff
@@ -197,6 +197,31 @@
...
+--    singular and plural strings. Additionally, it is necessary to specify the ``npgettext`` calls for all four types of soldier strings
+--    so as to ensure that all keys appear in the PO files (the ``npgettext`` results are unused). Example:
```
hessenfarmer: please reword the brackets (the ``npgettext`` results (i.e. the translated pluralized string is not used but by a call in C++)
Please add: the format must be UNUSED_soldier_X = npgettext(soldier_context, soldier_X_sg, soldier_X_pl, 0),
where X is going from 0 to 3 to matchg the c++ call.

hessenfarmer:
code LGTM. 2 small nits for the documentation. 
we can have this for 1.1 but in the longterm we should search for a solution that is not that hard to understand for tribe designers. (i.e. more intuitive)

hessenfarmer:
before merging please remove the military_capacity.lua files
Edit: I can't approve this as I made the PR

hessenfarmer:
PR is fine for me and should be ready. If it is ok I'll merge after the checks are green. I can't formally approve as I made the PR so either @Noordfrees you approve or we merge without formal approval as I would formally approve the work you have done.

Noordfrees (approved):

PR #5481, opened 2022-07-31
Update Asio to version 1.22.2 in download_asio.sh

tothxa:
Should we update this too for the release?

Not tested actual networking with the last version, but it builds on flathub. Version 1.22.1 has been used by flathub beta builds since May.

Probably install-dependencies.sh should be checked too, at least for the github workflow cases. MSYS in particular. :)


klaus-halfmann:
I have this  installed, but can not test this as of that sdl_imgage bug.
> asio 1.22.2 is already installed and up-to-date.

So compiling works ...

tothxa:
`install-dependencies.sh` matches the github workflow for Debian, MSYS and Mac.

The script installs many more deps than the workflow on Mac though. Apart from some usual development tools: icu4c, jpeg, libogg, libpng, libvorbis, zlib. Are these all part of the standard VM setup?

BTW could the github workflow be modified to use `install-dependencies.sh`? That way it would get constant testing at least for these archs and dependencies wouldn't need to be tracked in 2 places. I guess trying to install dev tools that are already present shouldn't be a problem.

Noordfrees:
> could the github workflow be modified to use `install-dependencies.sh`?

+1

tothxa:
asio 1.24.0 is available now… I'll test it and update the PR accordingly.

PR #5484, opened 2022-08-02
Added static SDL2 targets for transitioning

matthiakl:
Fixes msvc builds.
Newer SDL2 versions use a different naming, which resulted in broken builds. For now we can find the libraries with a small addition to the local find modules, but for the future we should consider using the now provided config.cmake modules `find_package(SDL2_mixer CONFIG)`.
Also see the linked topic in the commit.


hessenfarmer (approved):
LGTM

PR #5488, opened 2022-08-04
4950 fix map preview heights

### LINKED ISSUE #4950: Add heights to map previews
The preview image of a map currently does not show the terrain height. It used to do this in the past. Compare for example:

https://www.widelands.org/maps/europa/

https://www.widelands.org/maps/europa-12/

The upper one (quite old) gives a clear idea what the map looks like, whereas the lower one (newer) looks as if it were completely flat, which I think makes maps look less interesting to many players. 

The map previews (both website and in-game) should show the map heights again.

hessenfarmer:
fixes #4950


Noordfrees (approved):

hessenfarmer:
thanks for the quick review

PR #5492, opened 2022-08-06
Optimized height calculation for map preview

matthiakl:
Optimized solution for #4950. I noticed that the preview now takes much longer, when selecting a map.
This PR reduces the loading time of the map previews: For Europa in Debug builds from ~4900ms to ~450ms in Release builds from ~230ms to ~45ms.


hessenfarmer (approved):
code LGTM

Issue #5453, opened 2022-06-19
Stuck at "Game ending - please wait" when loading a game from inside a game


tothxa:
**Describe the bug**
Since #5421, loading a game directly from a game is broken. When the game to load is selected and OK is clicked the Widelands window is grayed out and the "Game ending - please wait" message is shown, but it never goes away. The controls of the game loading dialog underneath the graying out still respond to mouse clicks and key presses. If cancel is clicked, the loading dialog goes away, and the stopped game is visible under the graying out, but it does not respond to the mouse and the keyboard. If the window is closed by the window manager, then the previous game finally ends, the new one is loaded, then it exits immediately (with a segfault, if the dialog was closed).

**To Reproduce**
1. Start a game
2. Select "Load Game" (ctrl-L) from the main menu
3. Select a game to load
4. Click OK
5. See error

**Expected behavior**
The old game should end and the new game should be loaded.

**Version:**
 - OS: antix / Debian buster + backports
 - Widelands Version: 1.1~git25790 (3b6997d@master) ­– 92a5e0c works / 45e168b is broken

Noordfrees:
Confirmed. Does #5419 change the behaviour?

tothxa:
> Does #5419 change the behaviour?

It doesn't. Sorry, I forgot to mention it in the OP.

PR #5493, opened 2022-08-06
5485 alternate fix

### LINKED ISSUE #5485: Master fails to save File in multiplayer mode with unclear focus
**To Reproduce**

* Open attached File from the Muliplayer Menu and try to save it, after playing a few seconds.
* Notice: that it still still possible to select something during the attempted save.
* Notice that Network connection is lost after a while.
* After pressing Escape Game seems to continue in the Background
* Debug Output does not indicate any real attempt to save.
* Notice that Autosave works as expected.

**Expected behavior**

* Game is stopped during the save.
* No interaction with background possible during save.
* Game is saved in the End.

**Attachements**

[Save File and debug log](https://github.com/widelands/widelands/files/9248779/SaveGameBug.zip)
A Screencast can be found here: https://magentacloud.de/s/e7BHdHH2tgNLXLJ 

**Version:**
 - OS: macOS Monterey 12.5 (21G72)
 - CPU:  M1 ARM
 - Widelands Version: 1.1~git25829 (c58a8c6@master)

**Additional context**

* Is this a focus issue from the GUI? or is this just some side effect?
* I this a racecondition related to Multithreading?
* I case I should debug something please give me a hint where to start

hessenfarmer:
fixes #5485 
fix inspired by #5454

### DIFF
--- a/src/wui/game_main_menu_save_game.cc
+++ b/src/wui/game_main_menu_save_game.cc
@@ -196,6 +196,7 @@ void GameMainMenuSaveGame::ok() {
 		switch (type_) {
 		case Type::kSave: {
 			std::string filename = filename_editbox_.text();
+			modal_.reset();
 			if (save_game(filename, !get_config_bool("nozip", false))) {
 				die();
 			} else {


Noordfrees:
The problem with this would be that the save dialog is kept open in case saving fails, and then it would no longer be modal. A modal panel is not meant to stay around after it stops being modal so this could cause lots of "fun" downstream.

hessenfarmer:
I understand, but there are 3 things I don't understand: 
1. Why is the bug only limited to Multiplayer?
2. Why was it made modal in #5421? Wasn't this just a workaround? Can't this be solved alternativly. 
3. Why is the save in MP working if it is not modal and not working if it is modal?

Noordfrees:
1. In MP, the logic thread sends a blocking `NoteThreadSafeFunction` to the main thread during *every* think() cycle. In SP, `NoteThreadSafeFunction`s (especially blocking ones) are used much more sparingly. Since saving is only possible between think cycles to ensure the saved game will be in a well-defined state, the current think cycle (if any) must complete before we can perform the save.
2. It was **not** made modal. It was made to *pretend* to be modal without *actually* being modal. That's what caused both #5453 and #5485.  
  The bug could probably have been solved without making this panel modal in any way; however a modal save dialog has other advantageous side effects; many games make their save dialogs modal, and it makes sense for us to do so too.
3. The reason why Widelands hangs is that we need to wait for the logic thread's `NoteThreadSafeFunction` to be processed before the save can take place. When the save game dialog is not modal, then the InteractiveBase (being the topmost modal panel) is responsible for that and does it just fine within a few milliseconds. When the save game dialog is fake-modal like in master, it and the IBase both think the other is responsible so neither ever actually does it.

hessenfarmer:
Many thanks for the explanantions. that explains the fix in your branch as well.  I am still and ever will be learning.
In any case I agree about the advantages of having the save dialog modal.

PR #5495, opened 2022-08-07
Consistent UI keyboard selection

### LINKED ISSUE #5440: Inconsistent enter/space use in keyboard navigation
**Describe the bug**
When navigating without mouse, you sometimes have to select a control with space and sometimes with enter key.

**To Reproduce**
1. In any game open save game.
2. Enter a file name
3. Use tab to focus to 'Cancel' 
4. Press enter

This saves the game. To select a button you need to use space.

1. In any game open a menu.
2. Select any item and press space. 

Selection moves to first item and nothing is selected. To select a menu item, you need to use enter. 

**Expected behavior**
Either enter or space or both should work in all situations

**Version:**
 - OS: Ubuntu 20.04
 - Widelands Version: 1.1~git25782 (f98a800)

matthiakl:
Fixes #5440
Unifies the use of enter for UI selection in:
- Buttons
- Checkboxes
- Dropdown menus

Space is part of the dropdown filtering, so removing it from the UI control is a plus in consistency IMHO.
Also fixes the bug, that you couldn't select a dropdown item in the options menu by keyboard.


Noordfrees:
Unfortunately this causes a regression, Enter to select the default button no longer works reliably.
1. In the SP setup screen, click on a tribe dropdown.
2. Click to select a tribe.
3. Press Enter. In master this starts the game, as I want. With this branch it reopens the tribe dropdown instead. 

Having to first press Escape (to clear the focus) before being able to use the default button action is unintuitive, especially since pressing Escape *without* something being focused would cancel the setup.

This is the main reason why it was implemented that dropdowns can be opened only with Space but not with Enter. I don't really know a better solution unfortunately. But how about having these changes plus making Ctrl+Enter always call the default action regardless of what is focused and even if a dropdown is open?

matthiakl:
I implemented the behavior you suggested. Ctrl+Enter is dropped by UI base elements, so  the higher panels can handle it. I thought about doing something similar for Ctrl+Escape as a faster way back, but this requires more changes and will be something for 1.2

Noordfrees:
LGTM :) 

One more thing, the Editor Map Options menu does not respect (Ctrl+)Enter yet. Noticed this while trying to test the multiline-editbox codechange. IMHO it would be good to also add support for this here since it's about UI consistency.

Noordfrees (approved):
LGTM :)

matthiakl:
Sorry to bring this up again, but I just noticed, that my fix for the options menu did not work. Mouse selection wasn't possible anymore because of the `get_open_dropdown()` in `Panel::ui_trackmouse`. Now it is back to the list and key handling forwarded to the actual dropdown.

matthiakl:
Thanks for the review :)

PR #5497, opened 2022-08-08
Warp mouse to target when launching attack

Noordfrees:
Fixes https://www.widelands.org/forum/post/39048/


hessenfarmer (approved):
fix confirmed

PR #5504, opened 2022-08-16
Set SDL hint for relative mouse mode

### LINKED ISSUE #5500: scrolling the map results in unexpected behavior
Just updated to last commit and found this behavior:

When scrolling the map sometimes it feels like the right mouse button gets not released: Although the right mouse button is not pressed the map moves with each mouse move. Nothing can be clicked. In this state pressing the right mouse button, to maybe "release" the button again, does not help immediately. After some tries the button get "released" and all is working again, until the next "locked" right mouse button.
In the state of "locked" button, it seems like the mouse pointer get behind the fullscreen of widelands, and something is going on. E.g. a desktop widget get suddenly opened although i never used that widget. Just saw this after successfully closing widelands and surprisingly see the changes on the desktop.

**To Reproduce**
There is no specific step to reproduce, just move the map with pressed right mouse button and release the button. This happens sometimes, not on every move of the map.

**Expected behavior**
If i release the button, widelands should behave normally.

**Version:**
 - OS: archlinux
 - Widelands Version: 1.1~git25843 (f83cf83@master) Debug

**Additional context**
I guess commit 8b76ebdb098847cec009b8edee27db305f985f44 is the culprit. If i checkout the commit before and compile, all works fine.

matthiakl:
Fixes #5500 
Re #5494
The bug seems to only occur in SDL 2.0.22+

### DIFF
--- a/src/wlapplication.cc
+++ b/src/wlapplication.cc
@@ -416,6 +416,8 @@ WLApplication::WLApplication(int const argc, char const* const* const argv)
 	assert(sdl_video != nullptr);
 	SDL_version sdl_ver = {SDL_MAJOR_VERSION, SDL_MINOR_VERSION, SDL_PATCHLEVEL};
 	SDL_GetVersion(&sdl_ver);
+	// Keep cursor in window while dragging
+	SDL_SetHint(SDL_HINT_MOUSE_RELATIVE_MODE_WARP, "1");
 	bool sdl_scroll_x_bug = false;
 
 	// SDL version < 2.0 is not supported, >= 2.1 will have the changes


frankystone:
Fixes the bug, but scrolling down a widelands map in the editor let my desktop panel popup and the mouse pointer warps onto the panel.

matthiakl:
Hm, previously I had the same behavior you described (panel on the left). But now the cursor always stays inside the window. Does the scrolling continue, when you release the button while the cursor is on the panel? Which desktop environment is it (I use Cinnamon)?

frankystone:
Just found out that this happens if the widelands window is below the bottom Edge of the screen… So to trigger this i have to move the widelands window down until the bottom border of the widelands window  is not visible anymore. Probably on your machine you have to move the widelands window to the left until the left border is not visible anymore and scroll the map to the left.

So not sure if this is really an issue because i think most people do not play widelands if a part of the window is outside the visible area of the screen. Anyway not a correct behavior.

Should i try to make a video?

> Hm, previously I had the same behavior you described (panel on the left). But now the cursor always stays inside the window. 

Yes the cursor is inside the widelands window until i release the right button. Then the cursor warps to the bottom of the screen, opening the panel.

> Does the scrolling continue, when you release the button while the cursor is on the panel?

No. Releasing the button stops scrolling the map, but the cursor get suddenly to the bottom of the screen. In result the panel get opened.

Scrolling just a bit down does not show the behavior. One has to scroll down much more, depending on the height of the visual cursor inside the widelands window.

> Which desktop environment is it (I use Cinnamon)?

I am using Plasma, formerly known as KDE.

Additional observation, probably not related to this issue: If i enable `Grab Input` i can move the mouse out of the widelands window:

1. Open Options
2. Enable `Grab Input`
3. Click Apply
4. Moving the cursor out of the widelands window is possible
5. Move the cursor to the widelands window -> The cursor is grabbed
6. Click Apply again
7. Moving the cursor out of the widelands window is possible again

I can move the cursor out of the widelands window on each click on a button, e.g. when clicking OK in Options window, when clicking on Editor -> open Map, during the map loading screen, after clicking on the main menu button in the editor.
But with `Grab Input` enabled the cursor warp and resulting opened panel does not happen.

matthiakl:
> Scrolling just a bit down does not show the behavior. One has to scroll down much more, depending on the height of the visual cursor inside the widelands window.

With this information I am able to reproduce it, however, I don't think we can do much about it. The mouse moves to fast for SDL to process the events and warp the mouse back. When I move the cursor out of the Widelands window over the screen I can see the cursor shortly appearing outside.

> Additional observation, probably not related to this issue: If i enable Grab Input i can move the mouse out of the widelands window:

Definitely another bug, unrelated to mouse moving, but I will look into it.

Edit: This is an SDL 2.0.22 issue, the same behavior occurs in Widelands 1.0 with this update. I opened a ticket there.

frankystone (approved):
Thanks for fixing the main issue. I can keep on training for the tournament ;)

Approve testing only, because i don't know if it may have side effects on #5494

matthiakl:
> Additional observation, probably not related to this issue: If i enable Grab Input i can move the mouse out of the widelands window:

Now fixed on SDL side. You may try it by building the shared library from main branch at https://github.com/libsdl-org/SDL and start Widelands with `LD_PRELOAD=/path/to/SDL/build/libSDL2-2.0.so widelands`

PR #5515, opened 2022-08-19
Default select the first worker to kick from a production site

### LINKED ISSUE #4887: In a productionsite window the first worker should be autoselected for kicking
To kickj a worker currently you need to selct him with an extra click. Even if there is only one worker available. So the first worker of a buoilding should be autoselcted in the workers tab for kicking

### LINKED ISSUE #6196: 'Terminate employment' should not require manual selection of the only worker.
When I want to 'terminate employment' of a worker in a building, I have to select the worker before pressing the button - even if there is only one worker. Sometimes this is a bit confusing, because clicking the 'worker' button at the top can make it look like the worker is selected, but they still won't leave. Whereas if I do select the worker, and then deselect them by clicking somewhere else, they will still be evicted when I click the button.

Clicking the 'terminate employment' button should evict a worker even if none has been selected yet. Currently, if I select and then deselect a worker, it still evicts them. My suggestion is that when the building window is opened, the game behaves as though the 'previously selected' worker is just whoever is on top of the list - so that opening the window and clicking 'terminate employment' will evict the top worker on the list.

It doesn't have to be the top worker of course. It could be the last if you think that makes more sense. Or maybe only 'automatically' select the worker to evict in cases where there is only one worker present.

**Implementation steps**
 * Build a woodcutter and wait for the worker to arrive.
 * Click the building to open the building's window.
 * Click 'terminate employment of selected worker'.

(Currently this will do nothing, because we did not select the worker. My suggestion is that it terminates the employment of the only worker present even if we didn't manually select them.)

saulkim:
Hi, I'm open to any suggestions, criticisms, recommendations. Thanks

Fixes #4887 

Behavior in single worker production sites: 
-The one worker is auto selected for kicking, same as if the user manually selected the worker slot 

Behavior in multiple worker production sites: 
-The top(upper) worker slot is selected by default and clicking on the "fire worker" button without a selection by the user will result in the top worker being fired. Repeated "fire worker" clicks will still stay on the top worker slot until the user manually selects a different worker slot. If the the user clicks on another worker slot, that slot will stay selected for continued kicking until a different slot is chosen. Once the production site window is closed, reopening a production site window will default to the top worker.

Tested with several buildings from each tribe and they all seem to work.

A problem with this implementation is the worker slot is not highlighted in the way that selecting the slot manually would. With the single worker sites this doesn't matter much. But with the multiple worker sites, a user wouldn't know that the default choice is the topmost slot. Though in that case I assume the user would manually click on a worker to fire in the first place since they wouldn't want to risk firing the wrong person.


hessenfarmer:
Hi, 
thanks for contributing this. Code seems to be straight forward. However we are currently in feature freeze so this has to wait for 1.2 to be merged. For this reason you might find a solution to proper mark the worker to be evicted. Maybe simply program the selction rather the just choosing the topmost in the else path.

saulkim:
I see, thanks for your suggestion. I'll use the extra time for a better solution.

src/wui/productionsitewindow.cc
```diff
@@ -261,16 +261,18 @@ void ProductionSiteWindow::evict_worker() {
...
-		                               ->at(worker_table_->get_selected())
-		                               .worker.get(ibase()->egbase());
-		if (worker != nullptr) {
-			if (game_ != nullptr) {
-				game_->send_player_evict_worker(*worker);
-			} else {
-				NEVER_HERE();  // TODO(Nordfriese / Scenario Editor): implement
-			}
-		}
+	  worker_table_->select(worker_table_->get_selected());
```
Noordfrees: This line is redundant, you can collapse the if-else into a shorter `if (!worker_table_->has_selection())`

Noordfrees (approved):
Code LGTM, tested and working. Thank you for the patch :)

saulkim:
Thank you everyone for the help, this was a great learning experience :)

PR #5524, opened 2022-08-22
Use install-depencies.sh in workflow

matthiakl:
Implements https://github.com/widelands/widelands/pull/5481#issuecomment-1204958405
There were actually some errors in the script, so it is a good idea to use it in the workflow.

### DIFF
--- a/install-dependencies.sh
+++ b/install-dependencies.sh
@@ -178,8 +178,9 @@ elif [ "$DISTRO" = "msys64" ]; then
 
 elif [ "$DISTRO" = "homebrew" ]; then
    echo "Installing dependencies for Mac Homebrew..."
+   # TODO(matthiakl): minizip package of brew fails to link dynamically
    brew install $@ asio git cmake doxygen gettext glew graphviz icu4c jpeg \
-    libogg libpng libvorbis ninja python sdl2 sdl2_image sdl2_mixer sdl2_ttf zlib minizip
+    libogg libpng libvorbis ninja python sdl2 sdl2_image sdl2_mixer sdl2_ttf zlib
 
 elif [ "$DISTRO" = "solus" ]; then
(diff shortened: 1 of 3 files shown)


Noordfrees (approved):
:+1:

If this fixes actual errors in the installer script, any reason not to have it for v1.1?

install-dependencies.sh
```diff
@@ -115,73 +119,74 @@ sudo_or_su() {
...
     mingw-w64-i686-glbinding mingw-w64-i686-glew mingw-w64-i686-icu
 
 elif [ "$DISTRO" = "msys64" ]; then
    echo "Installing dependencies for 64-bit Windows..."
-   pacman -S mingw-w64-x86_64-toolchain git mingw-w64-x86_64-cmake \
+   pacman -S $@ mingw-w64-x86_64-toolchain git mingw-w64-x86_64-cmake \
     mingw-w64-x86_64-ninja mingw-w64-x86_64-asio mingw-w64-x86_64-SDL2_ttf \
     mingw-w64-x86_64-SDL2_mixer mingw-w64-x86_64-SDL2_image \
-    mingw-w64-x86_64-glbinding mingw-w64-x86_64-glew mingw-w64-i686-icu
+    mingw-w64-x86_64-glbinding mingw-w64-x86_64-glew mingw-w64-x86_64-icu
```
tothxa: Could these be unified like it was in the workflow file? Maybe define a shell function with $ARCH as argument?

matthiakl:
> If this fixes actual errors in the installer script, any reason not to have it for v1.1?

This is nothing critical, so I thought it is not a must have for v1.1. Of course you can decide to merge it before :wink:

install-dependencies.sh
tothxa: Thank you. Also for picking this up. :)

PR #5532, opened 2022-08-23
Update Asio to version 1.24.0 in download_asio.sh

tothxa:


Noordfrees:
Has it been tested that multiplayer works fine with this version?
When yes we should have it for the release

tothxa:
So far I only did a minimal test locally between a native build with asio-1.22.2 and flatpak build with asio-1.24.0. The latest package on flathub-beta is built with asio-1.24.0, so hopefully it will get some testing.

hessenfarmer:
if required I would be available for a quick test tonight.

tothxa:
I've just had a 2.5 hour long game with @hessenfarmer using a build with asio 1.24.0 without any problem. So I guess we can merge this?

Noordfrees (approved):

PR #5536, opened 2022-08-26
Don't warp mouse to dropdown if it's already there

### LINKED ISSUE #5535: Mouse pointer jumps to main item if entry in list was clicked
The mouse pointer jumps to an unexpected place when an item of the main menu list entry get clicked.

**To Reproduce**

1. open the editor and create a new map
2. click on "Show/Hide" → the dropdown window(?) is shown → click on e.g. "Hide Resources"
3. the dropdown window(?) stays open, but the mouse pointer jumps to the "Show/Hide" button of the main menu 

This is annoying because the mouse pointer is not at the place where one would expect it to be:
* If one wants after the first click to click on "Hide Animals", which is one entry above the previous clicked entry, he has to move mouse down instead of up.
* If one wants to zoom out of the map with the mouse: Clicking "Zoom –" several times is not possible, because once clicked the second click will close the "Map View", because the mouse pointer had jumped to the main menu button and the second click closes the dropdown.

Same happens with the main menu of a game.

**Expected behavior**
The mouse pointer should stay where the last click had happend and not jumping around.

**Version:**
 - OS: archlinux
 - Widelands Version: 1.1~git25858 (fd13252@master) Debug

Noordfrees:
Fixes #5535 

Although this is a small change I'd feels safer having this for v1.2 in case there are side-effects. Or is this bug annoying enough to warrant risking it?

### DIFF
--- a/src/ui_basic/dropdown.cc
+++ b/src/ui_basic/dropdown.cc
@@ -446,7 +446,7 @@ void BaseDropdown::set_value() {
 }
 
 void BaseDropdown::toggle() {
-	set_list_visibility(!list_->is_visible());
+	set_list_visibility(!list_->is_visible(), is_mouse_away());
 }
 
 void BaseDropdown::set_list_visibility(bool open, bool move_mouse) {


tothxa:
> Although this is a small change I'd feels safer having this for v1.2 in case there are side-effects. Or is this bug annoying enough to warrant risking it?

Well, not only is it annoying, it's also a regression from v1.0. As such it should be eligible for v1.1, shouldn't it?

frankystone:
This fixes the issue, tested main menu at bottom and at top of screen. Tested also dropdowns where i had found them, e.g. singleplayer game setup, multiplayer game setup, editor and random map generator.

No other issues found.

tothxa (approved):
Simple change, looks logical.

There were many changes to the dropdown code, so I couldn't easily track down which one caused this bug, but now I'm quite sure this belongs to v1.1.

PR #5539, opened 2022-09-02
CI system updates

matthiakl:
- ~~Github will drop the ubuntu-18.04 image with end of support April 2023. To prevent confusions with failed jobs during the deprecation phase, I propose to already drop g++-5 and g++-6 builds, which require this version.~~ → #5541
- gcc-12.2.0 fixed the false positives, so -Werror can work again
- appveyor now also uses the install script and the latest image ([testbuild](https://ci.appveyor.com/project/widelands-dev/widelands/builds/44657122))

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -129,17 +129,17 @@ jobs:
         compiler: [clang-14, clang-13, g++-12, g++-11]
         include:
           # Some compilers need an older ubuntu version to run correctly
-          - compiler: g++-5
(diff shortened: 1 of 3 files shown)


Noordfrees (commented):
Would it be possible to split this into two PRs? We should have the updates for g++12.2 and VS2022 now, but I'd like to keep support for Bionic and the older g++ around until after the release to ensure v1.1 will still work with them.

hessenfarmer (approved):
LGTM

PR #5541, opened 2022-09-02
Drop support for g++-5 and g++-6

matthiakl:
Split from #5539
> Github will drop the ubuntu-18.04 image with end of support April 2023. To prevent confusions with failed jobs during the deprecation phase, I propose to already drop g++-5 and g++-6 builds, which require this version.


CMakeLists.txt
```diff
@@ -222,23 +222,23 @@ if(CMAKE_BUILD_TYPE STREQUAL "Debug")
     endif()
   endif()
 elseif(CMAKE_BUILD_TYPE STREQUAL "Release")
-  if(("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU") AND (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 6.2))
-    message(STATUS "GCC >6.2.0 breaks -O3, setting -O2")
+  if(("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU") AND (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 7.5))
+    message(STATUS "GCC >7.5.0 breaks -O3, setting -O2")
```
Noordfrees, line 226: Why is this being changed? Do versions from 6.2 to 7.4 not have problems with -O3 after all?

CMakeLists.txt
```diff
@@ -222,23 +222,23 @@ if(CMAKE_BUILD_TYPE STREQUAL "Debug")
...
     set(WL_OPTIMIZE_FLAGS "-O2")
   elseif(MSVC)
     set(WL_OPTIMIZE_FLAGS "/O2")
   else()
     set(WL_OPTIMIZE_FLAGS "-O3")
   endif ()
-  if(("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU") AND (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 7.0))
-    message(WARNING "GCC <7.0.0 is known to cause segfaults in release builds. Please consider upgrading to GCC version 7 or newer or using Clang. Adding selected overrides to suppress reporting of known bugs.")
+  if(("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU") AND (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 7.6))
+    message(WARNING "GCC <7.6.0 is known to cause segfaults in release builds. Please consider upgrading to GCC version 7 or newer or using Clang. Adding selected overrides to suppress reporting of known bugs.")
```
Noordfrees:
```suggestion
    message(WARNING "GCC <7.6.0 is known to cause segfaults in release builds. Please consider upgrading to GCC version 7.6 or newer or using Clang. Adding selected overrides to suppress reporting of known bugs.")
```

CMakeLists.txt
matthiakl: Both gcc 6 and 7 seem to got a minor update which fixes it.
I tested a few versions:
- gcc 7.5.0 works (current version on GH runner)
- gcc 7.3.0 fails
- gcc 6.5.0 works
I think we should not differentiate between minor versions. If someone uses gcc 7 it should be (at least) 7.5.0 if it is in a proper package system.

Noordfrees (approved):

PR #5547, opened 2022-09-06
Use old style name for SDL pixelformat

### LINKED ISSUE #5546: Daily PPA is Failing on Xenial
#5545 breaks on old SDL versions

[Sample Buildlog](https://launchpadlibrarian.net/622054583/buildlog_ubuntu-xenial-amd64.widelands_2%3A1.1~bzr10377-202209061446~ubuntu16.04.1_BUILDING.txt.gz)

```
/<<PKGBUILDDIR>>/src/logic/map_objects/world/terrain_description.cc: In member function ‘void Widelands::TerrainDescription::replace_textures(const LuaTable&)’:
/<<PKGBUILDDIR>>/src/logic/map_objects/world/terrain_description.cc:194:67: error: ‘SDL_PIXELFORMAT_RGBA32’ was not declared in this scope
    SDL_Surface* converted = SDL_ConvertSurfaceFormat(sdl_surface, SDL_PIXELFORMAT_RGBA32, 0);
                                                                   ^
```

tothxa:
fixes #5546

### DIFF
--- a/src/logic/map_objects/world/terrain_description.cc
+++ b/src/logic/map_objects/world/terrain_description.cc
@@ -191,7 +191,7 @@ void TerrainDescription::replace_textures(const LuaTable& table) {
 		// uint32, so let's not risk endianness issues, when 24 bits would be better off with
 		// conversion anyway
 		if (sdl_surface->format->BitsPerPixel != 32) {
-			SDL_Surface* converted = SDL_ConvertSurfaceFormat(sdl_surface, SDL_PIXELFORMAT_RGBA32, 0);
+			SDL_Surface* converted = SDL_ConvertSurfaceFormat(sdl_surface, SDL_PIXELFORMAT_RGBA8888, 0);
 			SDL_FreeSurface(sdl_surface);
 			if (converted == nullptr) {
 				throw WLWarning("", "Could not convert texture to 32bit RGB");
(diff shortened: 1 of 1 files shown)


Noordfrees:
The SDL doc says
> SDL_PIXELFORMAT_RGBA32 is an alias for SDL_PIXELFORMAT_RGBA8888 on big endian machines and for SDL_PIXELFORMAT_ABGR8888 on little endian machines

So on machines with different endianness we probably should use the other constant instead?

tothxa:
> The SDL doc says
> 
> > SDL_PIXELFORMAT_RGBA32 is an alias for SDL_PIXELFORMAT_RGBA8888 on big endian machines and for SDL_PIXELFORMAT_ABGR8888 on little endian machines
> 
> So on machines with different endianness we probably should use the other constant instead?

Doesn't really matter, as we are converting to it just to be able to parse a 32 bit number with SDL_GetRGB(). If the pixelformat is already 32 bit, we don't need to do anything, as we're passing the pixelformat of the surface already.

Noordfrees:
How sure are you that the R/A and G/B channels won't be switched in the resulting colour? 
(I have no idea and no way to test)

tothxa:
> How sure are you that the R/A and G/B channels won't be switched in the resulting colour? (I have no idea and no way to test)

Well, I only go by the SDL documentation, but it's up to SDL_GetRGB() to sort out endianness and everything else, because it receives a raw pixel along with the accompanying pixelformat, both from the same SDL surface.

Noordfrees (approved):
Okay, that sounds safe :)

PR #5560, opened 2022-09-10
Log output and formatting usage cleanups

### LINKED ISSUE #5584: Invalid printf placeholders in internetgame and crash
After configuring a multiplayer game this morning the game crashed. The console output said:

~~~~
[00:00:01.078 real] INFO: **** END SOUND REPORT ****
[00:00:07.400 real] ERROR: [BufferedConnection] Trying to connect to 2a03:4000:32:524:48cb:feff:feae:b0c6:7395 failed!
[00:00:07.400 real] ERROR: [NetClient]: Trying to connect to 2a03:4000:32:524:48cb:feff:feae:b0c6:7395 failed!
[01:43:32.266 real] ERROR: format error: A string contains invalid printf placeholders
[01:43:32.266 real] ERROR: Error: [/home/kaputtnik/Quellcode/widelands-repo/widelands/src/base/format/tree.h:134] Too many arguments provided (expected only 0)
[01:43:32.266 real] ERROR: String: 1.1~git1 (beedd78@master)
[01:43:32.267 real] ERROR: format error: A string contains invalid printf placeholders
[01:43:32.267 real] ERROR: Error: [/home/kaputtnik/Quellcode/widelands-repo/widelands/src/base/format/tree.h:134] Too many arguments provided (expected only 0)
[01:43:32.267 real] ERROR: String: the-x1
[01:44:44.318 real] ERROR: format error: A string contains invalid printf placeholders
[01:44:44.318 real] ERROR: Error: [/home/kaputtnik/Quellcode/widelands-repo/widelands/src/base/format/tree.h:134] Too many arguments provided (expected only 0)
[01:44:44.318 real] ERROR: String: 1.1~bzr10379-202209071446
[01:44:44.318 real] ERROR: format error: A string contains invalid printf placeholders
[01:44:44.318 real] ERROR: Error: [/home/kaputtnik/Quellcode/widelands-repo/widelands/src/base/format/tree.h:134] Too many arguments provided (expected only 0)
[01:44:44.318 real] ERROR: String: WorldSavior
[01:44:44.517 real] ERROR: format error: A string contains invalid printf placeholders
[01:44:44.517 real] ERROR: Error: [/home/kaputtnik/Quellcode/widelands-repo/widelands/src/base/format/tree.h:134] Too many arguments provided (expected only 0)
[01:44:44.517 real] ERROR: String: maps/Downloaded/Inselkrieger Tournament version.wmf
[01:44:44.722 real] ERROR: format error: A string contains invalid printf placeholders
[01:44:44.722 real] ERROR: Error: [/home/kaputtnik/Quellcode/widelands-repo/widelands/src/base/format/tree.h:134] Too many arguments provided (expected only 0)
[01:44:44.722 real] ERROR: String: maps/Downloaded/Inselkrieger Tournament version.wmf
/usr/src/debug/gcc-build/x86_64-pc-linux-gnu/libstdc++-v3/include/bits/basic_string.h:1304: std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::reference std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::back() [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>; reference = char&]: Assertion '!empty()' failed.
~~~~

**To Reproduce**
Don't know. After starting widelands again except of those printf errors all worked fine.

**Version:**
 - OS: archlinux
 - Widelands Version: 1.1~git25864 (b301eaa@HEAD) Debug

**Additional context**

tothxa:
The network_gaming_messages.cc changes may fix some bugs or even crashes, but only much testing will tell us whether that's so.


tothxa:
re d72a85c (bufferedconnection.cc): see 12ecc24

src/network/bufferedconnection.cc
```diff
@@ -299,9 +299,12 @@ void BufferedConnection::start_receiving() {
...
 		   } else {
 			   if (socket_.is_open()) {
-				   log_err("[BufferedConnection] Error when receiving data from host (error %i: %s)\n",
-				           ec.value(), ec.message().c_str());
-				   log_err("[BufferedConnection] Closing socket\n");
+				   if (ec != asio::error::eof) {
+					   log_err(
+					      "[BufferedConnection] Error when receiving data from host (error %i: %s)\n",
+					      ec.value(), ec.message().c_str());
+					   log_info("[BufferedConnection] Closing socket\n");
```
Noordfrees, line 309: This line should be printed for EOF as well

src/network/network_gaming_messages.cc
```diff
@@ -18,17 +18,23 @@
...
 #include "base/i18n.h"
+#include "base/log.h"
 #include "base/string.h"
 
-static std::map<std::string, std::string> ngmessages;
+struct MessageString {
+	std::string str;
+	size_t n_arg;
+};
+static std::map<std::string, MessageString> ngmessages;
```
Noordfrees: Please add some documentation comments

src/network/network_gaming_messages.cc
```diff
@@ -39,72 +45,45 @@ const std::string NetworkGamingMessages::get_message(const std::string& code,
...
 
-	uint8_t last = strings.size() + 1;
-	// Try merging the strings
-	while (strings.size() > 1 && last > strings.size()) {
-		last = strings.size();
+	// for easier handling
+	std::vector<std::string> args;
+	args.push_back(arg1);
+	args.push_back(arg2);
+	args.push_back(arg3);
```
Noordfrees: Pull this all into one initializer list, and if possible to not copy the strings by using references (or pointers if that doesn't work)
```suggestion
	const std::string& args[] = {arg1, arg2, arg3};
```
```suggestion
	const std::string* args[] = {&arg1, &arg2, &arg3};
```

Noordfrees (commented):
A few comments inline, not tested yet

src/network/network_gaming_messages.cc
tothxa: References don't work, it must be pointers.

tothxa:
I wonder whether `ngmessages` should be `static const`, initialised with `gettext_noop()`-s, so translation would be done on the fly in `get_message()`, and `fill_map()` would be eliminated.

I also wonder whether the number of arguments should be got from the format tree instead of storing it in `ngmessages`. But I don't think it's exposed currently. May I add a public `get_nodes_count()` function to the `Tree` class? (Then I could even check whether the translated string has the correct number of placeholders… though that probably shouldn't be overdone)

tothxa:
OK, thanks. Would this be eligible for v1.1 either with or without the above further changes?

Noordfrees:
I'd prefer to have this for v1.2 either way. The current implementation doesn't have any known bugs except printing non-critical warnings, and such a refactoring always has a risk of regressions.

tothxa:
> I'd prefer to have this for v1.2 either way.

OK.

> The current implementation doesn't have any known bugs except printing non-critical warnings,

Well, they're actually scary looking `log_err()`-s, :) that on inspection turn out to be not errors at all. But I get it, and these were all there in v1.0 too.

src/network/network_gaming_messages.cc
```diff
@@ -53,91 +88,36 @@ const std::string NetworkGamingMessages::get_message(const std::string& code,
 		   format("%s, %s, %s, %s", code, get_message(arg1), get_message(arg2), get_message(arg3)));
 	}
 
-	assert(ngmessages[code].n_arg <= 3);
+	const std::string msg_translated = _(ngmessages.at(code));
+	const size_t n_fmt_arg = format_impl::Tree::get(msg_translated).get_nodes_count();
```
Noordfrees: This should be delegated a bit more. Anything that's in an `impl` namespace is an implementation detail and should not be accessed directly. There should be one non-namespace function that provides the interface between the impl namespace and any code that wishes to use the functionality, like `::format()` which merely forwards its arguments to `format_impl::format()`. 

So please similarly add a function `unsigned ::count_placeholders(const std::string&)` or such to `base/string.cc`, which just calls another new function `unsigned format_impl::count_placeholders(const std::string&)` in `tree.h`, which in turn then queries the `Tree` object.

src/network/network_gaming_messages.cc
```diff
@@ -39,126 +91,55 @@ const std::string NetworkGamingMessages::get_message(const std::string& code,
...
-				strings.push_back(get_message(arg3));
-			}
-		}
+	// clang-tidy says this shouldn't be const because it may be returned directly in case 0 below
+	std::string msg_translated = _(ngmessages.at(code));
+
+	const size_t n_fmt_arg = format_arguments_count(msg_translated);
+	assert(n_fmt_arg <= 3);
+
+	////// TODO(tothxa): remove checks after testing? (before even merging the PR?)
```
Noordfrees: This check really shouldn't be needed, we run extensive translation checks before releases to ensure any printf placeholder mismatches in translations are removed. I can remember only one case where one slipped though. And if we do want to check for problems, we should really take a much more general approach since there are lots of translatable strings with placeholders – these here aren't special.

src/base/string.h
```diff
@@ -138,4 +138,9 @@ inline std::string format_l(const std::string& format_string, Args... args) {
 	return format_impl::format(true, format_string, args...);
 }
 
+/** Query the format tree for the number of required arguments to a format string */
```
Noordfrees: Format tree? Not something a developer reading this comment should have to concern themselves with ;)

```suggestion
/**
 * Count the required arguments to a format string.
 * @param format_string String to analyze.
 * @return Number of formatting placeholders.
 */
```

src/network/network_gaming_messages.cc
tothxa: I just added the new check because I wanted to somehow include the offending format string for the other 2 checks (missing or extra arguments), and I didn't like any option I was able to come up with. I disliked this one the least, though it doesn't do that, but it makes sure that the problem is not in the translation, so the format can be used from the table in this file.

However the comment also refers to the other 2 checks. They shouldn't be needed either, I just added them in the hope that they might catch some bugs in the call sites that were masked before by the try-catch trickery.

(I was about to write a comment here to this effect actually, but you beat me to it :)

Noordfrees (approved):

src/network/network_gaming_messages.cc
Noordfrees: Can be merged when the debugging checks are removed

Issue #5501, opened 2022-08-16
FATAL EXCEPTION: buildingdata: building 2: warehouse: Unknown worker '[RANDOM CHARACTERS HERE]'


jasonkhanlar:
```
[00:00:13.406 real] ERROR: ##############################
[00:00:13.406 real] ERROR:   FATAL EXCEPTION: buildingdata: building 1: warehouse: Unknown worker 'x2'
[00:00:13.406 real] ERROR: ##############################
[00:00:13.406 real] ERROR:   Please report this problem to help us improve Widelands.
[00:00:13.406 real] ERROR:   You will find related messages in the standard output (stdout.txt on Windows).
[00:00:13.406 real] ERROR:   You are using build 1.0 (Release).
[00:00:13.406 real] ERROR:   Please add this information to your report.
[00:00:13.406 real] ERROR:   If desired, Widelands attempts to create an emergency savegame.
[00:00:13.406 real] ERROR:   It is often – though not always – possible to load it and continue playing.
[00:00:13.406 real] ERROR: ##############################

```

This error appears when trying to load a single player saved game that was originally a new random game.

Note: The line

```
 FATAL EXCEPTION: buildingdata: building 2: warehouse: Unknown worker '`@'
```

the characters in between the apostrophes are random/different for each attempt to reproduce this error (unique to each save file)

Steps to reproduce:

- Edit file /usr/share/widelands/tribes/initialization/barbarians/starting_conditions/headquarters.lua
  Change:
  ```
        soldiers = {
           [{0,0,0,0}] = 45,
        }
  ```
  to
  ```
        soldiers = {
           [{0,0,0,0}] = 65536,
        }
  ```
  Note: values 65535 and lower do not reproduce this error. I haven't tested for other workers and wares values, but they might be affected as well.
- Begin a new Single Player: New Random Game, make sure to select Barbarians as a tribe for at least 1 of the players, and start the game.
- Save the game.
- Exit the game.
- Load the saved game.
- You should see the error message above, both in console and dialog in game.

jasonkhanlar:
I tested all the other values, and it appears that this error is only affecting the number of soldiers. I tried changing file:

/usr/share/widelands/tribes/initialization/barbarians/starting_conditions/headquarters.lua

to:

```
...
      wares = {
         ax = 65536,
         ax_battle = 65536,
         ax_broad = 65536,
         ax_bronze = 65536,
         ax_sharp = 65536,
         ax_warriors = 65536,
[...]

PR #5568, opened 2022-09-11
Fix for #5501 - add set_soldiers cap

Kotbenek:
As mentioned by @Noordfrees in #5501 - I added a check in `set_soldiers` so the number won't exceed proposed value of 30000.


Noordfrees:

I tested it and it turns out 30000 is still too much; 15000 also crashes; 10000 seems to work.

Code suggestion: you could make the constant `constexpr` to improve compile-time optimizing.

hessenfarmer:
I would prefer to analyze why it is crashing and fix the root cause rather then using this workaround 
At least we should documetn in a comment that it is a workaround

Noordfrees:
It's not a workaround, it's an inherent limitation. You can't have more than a few thousand seeing objects on a node because their amount will overflow a variable. Making the variable bigger merely increases the maximum (and RAM usage, we can have several millions of instances of it in a big game). It's the same as when you try to add ten billion units of a ware – it overflows and is simply not permissible.

Kotbenek:
Thanks for the quick review and tips :)
I did requested changes in my patch.

Noordfrees (approved):

PR #5571, opened 2022-09-15
compile.sh:increase MAXCORES

andy5995:
This increases MAXCORES so when building with [distcc](https://www.distcc.org/) the benefit is significantly improved.

It would probably be better just to remove the limitation completely. What do you think?

### DIFF
--- a/compile.sh
+++ b/compile.sh
@@ -75,6 +75,10 @@ print_help () {
     echo "                      compiling and linking. Default is to leave 1 core"
     echo "                      free."
     echo " "
+    echo "--allow-more-cores"
+    echo "                      Allow 'j' to exceed the maximum number of cores"
+    echo "                      detected on the local machine."
+    echo ""
     echo "-r or --release       Create a release build. If this is not set,"
     echo "                      a debug build will be created."
     echo "-d or --debug         Create a debug build. This is the default,"
@@ -208,18 +212,12 @@ do
           exit 1
         ;;
       esac
-      if [ "$2" -lt 1 ]; then
-        echo "Invalid number of cores was specified: $2"
-        exit 1
-      fi
-      if [ "$MAXCORES" -ge "$2" -o "$MAXCORES" -eq 0 ]; then
-        CORES="$2"
-      else
-        echo "Cannot set number of cores to $2, because the maximum number"
-        echo "of supported cores is $MAXCORES."
-        exit 1
-      fi
-    shift 2 # past argument and value
+      REQUESTED_CORES=$2
+    shift 2
+    ;;
+    --allow-more-cores)
+      ALLOW_MORE_CORES=ON
+    shift # past argument and value
     ;;
     -r|--release)
       BUILD_TYPE="Release"
@@ -380,6 +378,18 @@ if [ -n "$EXTRA_OPTS" ]; then
   echo "Extra CMake options used: $EXTRA_OPTS"
   echo " "
   CMD_ADD "$EXTRA_OPTS"
+fi
+
+if [ "$REQUESTED_CORES" -lt 1 ]; then
+  echo "Invalid number of cores was specified: $2"
+  exit 1
+fi
+if [ "$MAXCORES" -ge "$REQUESTED_CORES" -o "$MAXCORES" -eq 0 -o -n "$ALLOW_MORE_CORES" ]; then
+  CORES="$REQUESTED_CORES"
+else
+  echo "Cannot set number of cores to $REQUESTED_CORES, because the maximum number"
+  echo "of supported cores is $MAXCORES."
+  exit 1
 fi
 
 echo "Using ${CORES} core(s)."
@@ -556,6 +566,14 @@ buildtool="" #Use ninja by default, fall back to make if that is not available.
                -DUSE_XDG=$USE_XDG                              \
                -DUSE_FLTO_IF_AVAILABLE=${USE_FLTO}
 
+    if [ $CORES -gt $MAXCORES ] && [ "$BUILD_TRANSLATIONS" = "ON" ] ; then
+      if [ $MAXCORES -gt 0 ] ; then
+        local TMP_CORES=$MAXCORES;
+      else
+        local TMP_CORES=1;
+      fi
+      $RUN $buildtool lang -j $TMP_CORES
+    fi
     $RUN $buildtool -j $CORES
 
     return 0


compile.sh
```diff
@@ -145,7 +145,7 @@ fi
 # nproc is a GNU tool, may not be available everywhere, but it is preferable,
 # when it is available
 if hash nproc 2>/dev/null ; then
-  MAXCORES="$(nproc)"
+  MAXCORES="$(nproc)" * 6
```
tothxa: If this is desirable at all

1. Why only here?
2. Why 6?
3. Can't distcc be detected to make this conditional?

tothxa (commented):
It was said before that `compile.sh` is aimed at non-developers.

If you really need this, I think a better solution would be to use `$MAXCORES` from the environment if set.

compile.sh
andy5995: > If this is desirable at all
> 
>     1. Why only here?

Because that's all I needed for testing locally. I wasn't sure what the maintainers here would ultimately want so that's all I did for now.

> 
>     2. Why 6?

 I just used 6, because I wanted to specify 20 jobs for testing locally.

> 
>     3. Can't distcc be detected to make this conditional?

No.

tothxa:
OK, so this was just a first draft.

I thought some more about it, and instead of setting `$MAXCORES` it would be more straightforward to add something like a `-J`/`--force-cores` option that wouldn't be checked against `$MAXCORES`.

-----

On the idea of distcc: I guess it may be very useful, but probably could be helped by making it possible to explicitly tell the script about it. But then again, that's probably too much for a specialised use case for more advanced usage.

There's also the problem that distcc only distributes the C++ compilation tasks. Of course that's by far the longest (except for final linking and stripping of the executable, which can't be parallelised :((( ), but it's probably not desirable to swamp the local machine with a gazillion of translation compile tasks. Like I already skip building the tests because running them in parallel needs more RAM than I have, and I don't want to turn on swap on a SSD.

So you're better off if you just run cmake separately, so then you can use `make`/`ninja` `-j 100 widelands` with distcc, and `-j $MAXCORES lang` for translations, etc.

andy5995:
@tothxa I think I understood all your points. Thanks for explaining. I pushed a change that may address all the issues. What do you think? Or would you still prefer a "--force" option?

compile.sh
```diff
@@ -212,13 +212,14 @@ do
         echo "Invalid number of cores was specified: $2"
         exit 1
       fi
-      if [ "$MAXCORES" -ge "$2" -o "$MAXCORES" -eq 0 ]; then
+      if [ "$MAXCORES" -ge "$2" -o "$MAXCORES" -eq 0 ] || [ -n "$DISTCC_POTENTIAL_HOSTS" ]; then
```
tothxa:
```suggestion
      if [ "$MAXCORES" -ge "$2" -o "$MAXCORES" -eq 0 -o -n "$DISTCC_POTENTIAL_HOSTS" ]; then
```
I think…

Yes, this is much better.

compile.sh
```diff
@@ -212,13 +212,14 @@ do
...
         CORES="$2"
       else
         echo "Cannot set number of cores to $2, because the maximum number"
         echo "of supported cores is $MAXCORES."
         exit 1
       fi
+      CORES="$2"
```
tothxa: I think this is already set in the `then` branch, and the `else` branch exits.

compile.sh
```diff
@@ -556,6 +557,9 @@ buildtool="" #Use ninja by default, fall back to make if that is not available.
                -DUSE_XDG=$USE_XDG                              \
                -DUSE_FLTO_IF_AVAILABLE=${USE_FLTO}
 
+    if [ $CORES -gt $MAXCORES ] && [ "$BUILD_TRANSLATIONS" = "ON" ] ; then
+      $RUN $buildtool lang -j $MAXCORES
```
tothxa: I'm not sure it's only the translations. I don't know about codecheck for example in debug builds, and there may be others.

tothxa (commented):
Thanks for addressing my concerns.

Let's wait for other comments though.

compile.sh
```diff
@@ -212,7 +212,7 @@ do
         echo "Invalid number of cores was specified: $2"
         exit 1
       fi
-      if [ "$MAXCORES" -ge "$2" -o "$MAXCORES" -eq 0 ]; then
+      if [ "$MAXCORES" -ge "$2" -o "$MAXCORES" -eq 0 -o -n "$DISTCC_POTENTIAL_HOSTS" -o -n "$DISTCC_HOSTS" ]; then
```
andy5995: Unfortunately I don't think this way is robust enough. Neither are required to be set. For example, on my system, Manjaro, when distcc is installed, symlinks are setup in /usr/lib/distcc and I add that to my PATH, and then the hosts are read from ~/.distcc/hosts

Maybe someone from @distcc would have a suggestion about this @themightyoarfish @sourcefrog @shawnl @asheplyakov

compile.sh
tothxa: Oh, and `$MAXCORES` can be 0 if the detection fails, so you need to check for that case too.

compile.sh
tothxa: If it gets any more complicated then you should move distcc detection next to maxcores detection.

hessenfarmer:
@tothxa @Noordfrees @andy5995 
sorry I can't be of any help here as I don't understand much from the compile script. I am still a windows guy

compile.sh
sourcefrog: 👋🏻 I have never seen Widelands before (though it looks neat) and have never built it, so I'll just answer in general terms:

It looks like the point of this change is to block the user from scheduling more jobs than they have local cores. I'd just drop that! As @andy5995 said in the first comment. If the user wants 20 jobs on a 16 core machine, let them! I probably wouldn't add this complication into my own project. (In fact that's not hypothetical: I recently added a `-j` option to cargo-mutants and it doesn't check NCPUS: https://github.com/sourcefrog/cargo-mutants/pull/93) 

As @tothxa says, the number of cores is anyhow a poor proxy for "how many tasks can you usefully or safely run in parallel" because RAM, RAM bandwidth, and even IO bandwidth can be more important constraints. Also, at least some tools like Mold will spawn multiple threads internally.

Potentially naive users will overload their machine by running `-j 20000` but that's really on them; it's hard to reliably prevent it.

Also, distcc is now pretty old and there are newer and possibly better programs that carry on similar ideas. So I would avoid hardcoding anything specific to distcc.

tothxa: Hi @sourcefrog! Yes, this is exactly the info we needed. Thank you very much! 

So, this means the current approach to have an user override for the jobs count check is OK.

The check should be kept though by default, because this script is mostly intended for users who'd like to help test pre-releases, and I think it's a reasonable rule of thumb to protect them.

sourcefrog: If you're saying you will use NCPUS-1 by default but let the user set whatever number they want then that sounds good to me: a good default but avoiding too much complication, and letting people set whatever number if they think they know best.

andy5995: Thanks for the feedback @sourcefrog . That code is outdated. I've removed the DISTCC-specific stuff, and the intent of this PR is to simply allow an override option.

@tothxa @Noordfrees I believe this is ready for review.

compile.sh
```diff
@@ -556,6 +566,14 @@ buildtool="" #Use ninja by default, fall back to make if that is not available.
                -DUSE_XDG=$USE_XDG                              \
                -DUSE_FLTO_IF_AVAILABLE=${USE_FLTO}
 
+    if [ $CORES -gt $MAXCORES ] && [ "$BUILD_TRANSLATIONS" = "ON" ] ; then
```
tothxa, line 569:
```suggestion
    if [ $CORES -gt $MAXCORES -a "$BUILD_TRANSLATIONS" = "ON" ] ; then
```

Sorry, I missed this previously.

tothxa (commented):
We still need to check all possible targets, and separate C++ compilation from the rest, but I certainly wouldn't be able to review all cmake files to do this.

What I know to exist:

 - codecheck (python): enabled (by cmake) with debug, but also has its own cmake option. It's independent of compilation, but it is better run before it, like in the github actions. It can't be run if not enabled, but normally runs with compilation if enabled. And `compile.sh` completely ignores it currently, doesn't even have an idea whether it's enabled.
 - tests: I think it involves some compilation but also running the testsuites afterwards, and I don't know whether these can be easily separated (compilation can be parallel with main compilation, while running must come afterwards)
 - website tools: compilation, can be parallel

andy5995:
It all seems like a lot of extra code just to alter the behavior of an option as standard as '-j'. Working on it doesn't interest me at all at this time.

PR #5576, opened 2022-09-16
README.md:remove list of cmake options [skip ci]

andy5995:


tothxa:
Why do you want to remove it?

I don't know whether newer cmake is different, but for me `cmake -L` gives a not particularly readable list of currently set options with no explanation at all. But even the explanations given in the comments in `CMakeLists.txt` are much shorter than those in `README.md`. It also includes lots of ICU options as well as some more from other dependencies.

Also, the cmake documentation says that `-L` prints options from the cache, not from the config, so it needs a cache in the first place. Even if you have run cmake before, reading the documentation shouldn't change any files, but your command does update the build directory and `data/datadirversion`. (though if cmake was run before, `cmake -N -L` can be used instead in the build directory at least)

andy5995:
> Why do you want to remove it?
> 
> I don't know whether newer cmake is different, but for me `cmake -L` gives a not particularly readable list of currently set options with no explanation at all. 

Ah, you're right. -LH prints the descriptions.

>But even the explanations given in the comments in `CMakeLists.txt` are much shorter than those in `README.md`. It also includes lots of ICU options as well as some more from other dependencies.
> 

I could move the explanations from the README into the CMakeLists.txt, but it sounds like you're finding too many problems (problems that can't really be fixed because cmake has no option to list variables from the config, only the cache?) with this PR for this patch to be merged. Is that fairly accurate?

andy5995:
> Why do you want to remove it?

I usually prefer to not have duplicated information to make project maintenance easier. My thinking is that if a developer could get the same information from cmake output, then removing it from the docs would mean that less time is spent maintaining documentation, and the docs would be less error-prone to becoming outdated or incorrect.

tothxa:
> > Why do you want to remove it?
> 
> I usually prefer to not have duplicated information to make project maintenance easier. My thinking is that if a developer could get the same information from cmake output, then removing it from the docs would mean that less time is spent maintaining documentation, and the docs would be less error-prone to becoming outdated or incorrect.

Well, in general that's laudable. Though in this case I'm inclined to think that `CMakeLists.txt` is not the best place for detailed documentation while `README.md` is much more so; and the whole thing is not so important as to be a big problem if it gets out of sync sometimes. Maybe turn it around, and add a comment to the options section in `CMakeLists.txt` to remind developers to update `README.md` if they change anything?

> it sounds like you're finding too many problems (problems that can't really be fixed because cmake has no option to list variables from the config, only the cache?) with this PR for this patch to be merged. Is that fairly accurate?

In a way yes, but it's more like I just wanted to share my opinion for other reviewers to consider. (also applies to the previous comment in this post)

hessenfarmer:
When I started working on widelands and building myself. I used cmake without really knowing what I was doing or that i could have printed a list of options. I simply followed the guide in the homepage and the information in the readme. 
So I am in favor of keeping the readme as the main source of information. 
@andy5995 would you like to include a comment with a hint in the cmake options like @tothxa suggested or better close this PR as not approved?

andy5995:
Understood, thank you for the feedback.

Issue #5582, opened 2022-09-20
reword Waterway length limit


frankystone:
Currently the wording "Waterway length limit" in the map options menu does not have any conjunction to ferries. For a player it is difficult to associate "Waterway" with "Ferry". See https://www.widelands.org/forum/post/39377/

So i want to suggest to rename "Waterway length limit" with something that associates with ferries, e.g.:

* Ferry range
* Waterway length for ferries
* Settings for ferries
* Enable ferries by defining a max length

Additionally it would be nice to have a tooltip, also for the other map options.

**Implementation steps**
- [x] Rewording of "Waterway…"
- [x] Add tooltip
- [x] Check other places in map options menu for useful tooltips

tothxa:
Regarding the 3rd part: I think only the "Tags" tab needs more tooltips, because "Main Options" is fine without them, while "Teams" has the ones it needs. But is there any guide to help interpret the map tags?

 * Is "Free for all" the same as 1v1, but for more than 2 players?
 * What is needed of a map to be considered balanced? We had examples in the 2022 Tournament discussions where the balancing of some proposed maps was disputed.
 * Does theme mean the same as climate in the random game setup? If so, is it still relevant? If so, shouldn't it be set automatically based on the used terrain types? Ie. only set it if one climate is clearly dominant – say 75% of the map – otherwise set it to "none"? Maybe also consider relevant immovables?

frankystone:
> * What is needed of a map to be considered balanced? We had examples in the 2022 Tournament discussions where the balancing of some proposed maps was disputed.

Because of having it the map options it depends on the maps author to declare a map "balanced". So for non official  maps (not shipped with the game) this tag might be meaningless. Regarding a tool tip what about:
"Sets the tag "balanced" for this map. If your map becomes an official map this tag might be reconfigured."

> * Does theme mean the same as climate in the random game setup? If so, is it still relevant? If so, shouldn't it be set automatically based on the used terrain types? Ie. only set it if one climate is clearly dominant – say 75% of the map – otherwise set it to "none"? Maybe also consider relevant immovables?

As far i understand this setting applies ["only" for the used loadscreens](https://github.com/widelands/widelands/commit/0df31aadb80acd22c233159e1353d556c5ae2bda) when starting a game. Maybe this can be reworded also:

Theme → Theme for Loadscreens
Tool tip: "Affects the used pictures during loading a game"

Something like that?

PR #5592, opened 2022-10-03
Updated "Waterway length limit" phrase

ghost:
For issue #5582, I have gone through each of the instances where "Waterway length limit" is displayed and changed the phrasing to "Ferry range" as requested.

### DIFF
--- a/po/widelands/ar.po
+++ b/po/widelands/ar.po
@@ -1058,7 +1058,7 @@ msgid "(none)"
 msgstr ""
 
 #: ../../src/editor/ui_menus/main_menu_map_options.cc:398
-msgid "Waterway length limit:"
+msgid "Ferry range:"
 msgstr ""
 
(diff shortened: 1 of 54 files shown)


tothxa (changes requested):
Please don't update the `po` and `pot` files directly. They are updated automatically with the translation updates. We don't even edit them directly for typo fixes that only affect English (that would need manual pushing the updated `po` files to transifex), but this one is a major change where the translations will need to change too. Changing the English phrase behind the translations would hide the change from translators.

klaus-halfmann:
@kskretteberg any progress on this one? 
You might want to create a new branch to add only the changes not affected by translations.

hessenfarmer:
As there hasn't been any response for over a month we might close this PR to clean the board.    
@kskretteberg @Noordfrees

tothxa:
+1

I think I'll take this over, and do the other tasks in #5582 as well.

PR #5598, opened 2022-10-08
Update screenshots in appdata.xml

tothxa:


tothxa:
see https://www.widelands.org/forum/post/39526/

As Economy Options was voted down (I agree that Info Panel looks better as a general image), I think this set lacks some more UI elements and more of the actual in-game look and feel. To fix this, new Atlantean and Empire images could be created: one with Stocks and General Stats, and maybe one with some building window or so, but not too prominent like Economy Options was.

Noordfrees (approved):

tothxa:
I take it as I shouldn't bother with the new images with some windows then… :)

Thanks for merging.

PR #5600, opened 2022-10-09
Update ChangeLog to v1.1

Noordfrees:
To be merged one or two days before the RC


tothxa:
This is a draft, so you may be aware, but for the record, the following PR's are missing:
#5510
#5512
#5513
#5524
#5534
#5536
#5538
#5543
#5598

Noordfrees:
Thanks, I had started logging from the wrong commit so these were overlooked

PR #5601, opened 2022-10-09
Fixes in dump- and load economy scripts

### LINKED ISSUE #5595: `utils/dump_economy.lua` fails on portdocks
**Describe the bug**
When I try to use `utils/dump_economy.lua` with an economy with ports, I get an assertion failure.

**To Reproduce**
1. Start a debug build of Widelands
2. Start local multiplayer game with e.g. this save: [5t.zip](https://github.com/widelands/widelands/files/9737330/5t.zip)
3. Enter the debug console
4. Load the dump economy script with `dofile()` (as described in the script file)
5. Try to dump either tribe (as described in the script file)
6. See error

**Expected behavior**
The economy should be dumped without an error message.

**Version:**
 - OS: Debian buster

**Additional context**
Error message on stdout:
```
UNKNOWN immovable type: 	nil
[00:12:15.495 real] DEBUG: *** [../src/scripting/lua_errors.cc:21] src/widelands-master/utils/dump_economy.lua:130: assertion failed!
```

The error printing line has a bug too, it uses `immovable.type` instead of `immovable.descr.type_name`. The following patch makes the script run, but just ignores portdocks (haven't tested whether the dump can be loaded this way): 
[dump_economy.patch.txt](https://github.com/widelands/widelands/files/9737369/dump_economy.patch.txt)

BTW the usage help assumes that Widelands is started from the top level of a Widelands source directory. It should mention that the argument to `dofile()` is relative to the directory from where Widelands is started.

tothxa:
Dump:
* fix dumping multiple player economies (clear returned table for each call)
* don't fail on unhandled immovables (e.g. port docks and waterways)
* don't dump treetop sentries which can't be placed by lua (cause unrecoverable error)

Load:
* fix call of place_road() (add missing road type argument)
* add sanity check for tribe

fixes #5595


tothxa:
These scripts are new in 1.1 (added with atl02), and the `place_road()` bug prevented `load_economy.lua` to work at all. They are also completely external. So I wonder whether this should be added to 1.1 even though we're in the very last minute.

edit: Alternatively remove the scripts from 1.1 and re-add the fixed ones in 1.2.

utils/load_economy.lua
```diff
@@ -27,13 +27,16 @@ function place_flags(eco, plr)
 end
 
 function place_roads(eco, plr)
-   for i = 1, #eco.roads do
-      local command = string.format("road = plr:place_road(map:get_field(%i,%i).immovable, %s ,true)", eco.roads[i].x, eco.roads[i].y, eco.roads[i].dirs)
+   -- Second carrier can't be placed currently, so we just fake a carrier
```
hessenfarmer: did you try whether this si still valid?     
according to the actual documentation we can readouit the road_type, which wasn't possible when I converted the script from ancient beginning of Atlantean02.     
So we might simply readout the type and try to set the type and the workers that have been read.     
Otherwise we might readout the type and fake a second carrier for each busy road in the HQ.

hessenfarmer: just tested it we can set them and so we might alter the script.   
By doing so i found an issue in Atlantean 02 as well

hessenfarmer: easiest way would be to read the road type and use road:set_workers(road.valid_workers) with the risk of overdoing some second carriers which hadn't been available yet while the road was already busy

tothxa: I tried to test it, but it looks like I did something wrong, because I got errors when I tried to add second carriers. Now I was able to do it.

Some debugging showed though that sometimes more than the allowed carriers are dumped. I guess these are on their way to somewhere else? Anyway, I decided to keep it simple and use `valid_workers` instead as you suggested.

utils/load_economy.lua
```diff
@@ -27,13 +27,22 @@ function place_flags(eco, plr)
...
-      local command = string.format("road = plr:place_road(map:get_field(%i,%i).immovable, %s ,true)", eco.roads[i].x, eco.roads[i].y, eco.roads[i].dirs)
+   carrier2 = plr.tribe.carriers[2]
+   for i, rd in ipairs(eco.roads) do
+      -- Road type is not dumped
+      -- Waterways are not dumped currently
+      if rd.workers[carrier2] then
+         road_type = "busy"
+      else
+         road_type = "normal"
+      end
```
tothxa: I wonder whether I should change `dump_economy` to dump road type too, so this wouldn't be needed. (and maybe waterways could be included then?) Relying on that would however break the load script for old dumps. Is that OK, this being a new feature for 1.1?

Noordfrees: Since it's new, breaking old dumps is fine. +1 for including all road types.

utils/load_economy.lua
```diff
@@ -86,6 +87,9 @@ function load_eco(plrno, tribe, folder)
    -- hq = (tribe.."_headquarters")
    -- plr:place_building(hq, map.player_slots[plrno].starting_field, false, true)
 
+   assert(plr.tribe.name == eco.details[1].tribe,
```
tothxa: BTW is it even possible to load economy for a non-existent player? If not, then we don't even need the `tribe` argument for `load_eco()`.

tothxa: Here's how I'd do it:

[loadecon_patch_against_PR.txt](https://github.com/widelands/widelands/files/9757832/loadecon_patch_against_PR.txt)

[loadecon_patch_against_master.txt](https://github.com/widelands/widelands/files/9757834/loadecon_patch_against_master.txt)

hessenfarmer (approved):
changes including the patch look good to me. Not tested.

PR #5606, opened 2022-10-09
Add some new main menu background images

tothxa:


tothxa:
See https://www.widelands.org/forum/post/39526/

Newly created images for Amazons and Frisians are from Tournament 2022 replays (plus a practice game), hence the repetitve maps. I hope it's OK.

I also hope this is very low risk for the release.

Noordfrees:
IMHO the two Crater Lake screenshots don't look that good, they're rather empty and bland. The others are nice.

tothxa:
OK, I replaced them with some AI games. At least with these new images we have more variation of terrain too.

Indeed, it was hard to pick the Crater Lake images because of the way those matches went, and I didn't have much choice: Inselkrieger was even worse. :) But I thought some old ones aren't that much different, and that it would be a nice touch to include images from the tournament. :)

Noordfrees (approved):
Great :)

Issue #5610, opened 2022-10-17
utils/macos/build_app.sh fails in OSX 12.6 with Apple Silicon


klaus-halfmann:
**Describe the bug**

Running `utils/macos/build_app.sh` fails with
```
WidelandsRelease/Widelands.app/Contents/MacOS/widelands: replacing existing signature
WidelandsRelease/Widelands.app/Contents/MacOS/widelands: bundle format unrecognized, invalid, or unsuitable
In subcomponent: .../v1.1-rc1/WidelandsRelease/Widelands.app/Contents/MacOS/data/maps/Finnish_Lakes.wmf
  * Error : An error occurred while applying ad-hoc signature to WidelandsRelease/Widelands.app/Contents/MacOS/widelands. Attempting workaround
    cp -p "WidelandsRelease/Widelands.app/Contents/MacOS/widelands" "/var/folders/sq/gy2m486j4nggs6j6w_z658440000gn/T/dylibbundler.NXz3pOjx/widelands"
    mv -f "/var/folders/sq/gy2m486j4nggs6j6w_z658440000gn/T/dylibbundler.NXz3pOjx/widelands" "WidelandsRelease/Widelands.app/Contents/MacOS/widelands"
    rm -rf "/var/folders/sq/gy2m486j4nggs6j6w_z658440000gn/T/dylibbundler.NXz3pOjx"
    codesign --force --deep --preserve-metadata=entitlements,requirements,flags,runtime --sign - "WidelandsRelease/Widelands.app/Contents/MacOS/widelands"
WidelandsRelease/Widelands.app/Contents/MacOS/widelands: replacing existing signature
WidelandsRelease/Widelands.app/Contents/MacOS/widelands: bundle format unrecognized, invalid, or unsuitable
In subcomponent: /Users/klaus/develop/widelands/v1.1-rc1/WidelandsRelease/Widelands.app/Contents/MacOS/data/maps/Finnish_Lakes.wmf
  * Error : An error occurred while applying ad-hoc signature to WidelandsRelease/Widelands.app/Contents/MacOS/widelands
```

**Version:**
 - OS: OSX 12.6
 - Widelands Version: v1.1-rc1

**Ideas what could be done**

* Need to dig into that script what command actually fails
* The resulting .app is actually running (So a workaround might be to ignore that error)
* The .dmg (Volume image) is not created and files like README are missing
* Need to compare with "normal" OSX build which seem to not have Issues?

klaus-halfmann:
[...]

PR #5613, opened 2022-10-17
Ignore errors from dylibbundler

klaus-halfmann:
Workaround for #5610 but not a real fix.
OTOH should not affect normal OSX builds etc.
Lets see how this works out.


Noordfrees:
Can be merged from my side. Do you want me to apply this patch to master as well?

klaus-halfmann:
@Noordfrees please merge only to release 1.1 I hope the author of that dylibbundler will pick up and fix the underlying, original problem.

PR #5625, opened 2022-10-25
Require libpng >= 1.6

Noordfrees:
https://www.widelands.org/forum/post/39746/


Noordfrees:
How can we get the MacOS CI to detect the correct libpng? Apparently both 1.4 and 1.6 are installed but CMake only finds the outdated version. 
@klaus-halfmann

klaus-halfmann:
Checking: `brew info libpng`
`==> libpng: stable 1.6.38 (bottled), HEAD`
Thats my local version, Where Did you see that 1.4 version ?
Checking out this Branch now.

Local ./compile.sh gives me:
` Found PNG: /opt/homebrew/lib/libpng.dylib (found suitable version "1.6.38", minimum required is "1.6") `

In the Build I see:
```
 Could NOT find PNG: Found unsuitable version "1.4.12", but required is at
  least "1.6" (found /usr/local/lib/libpng.dylib)
```

Is there some `brew` call as part of this build?

OK found it in `install-dependencies.sh` but at this point in time libpng is already installed, mmh.

klaus-halfmann:
Tried something lets wait for the build to pick up,
Sorry for the accidential commit to master.

Niektory:
I'm also trying something (messed up the commits a little in the process).

Niektory:
OK, so `set(CMAKE_FIND_FRAMEWORK LAST)` seems to have helped. The build succeeded using libpng 1.6 according to the logs.

klaus-halfmann:
OK Lets revert my change about uninstalling libpng, and thanks @Niektory

klaus-halfmann:
OK lets wait for another Build and then we should be fine

klaus-halfmann:
@Niektory  I think we should drop `-DCMAKE_FIND_DEBUG_MODE=ON`, too?

Niektory:
> @Niektory I think we should drop `-DCMAKE_FIND_DEBUG_MODE=ON`, too?

Yes, drop it, that was just to see what's going on in case the build was still failing.

klaus-halfmann:
@Noordfrees I think I can merge this? With "squash and merge"

mohd-akram:
Is `set(CMAKE_FIND_FRAMEWORK LAST)` necessary? This causes linking with mesa's libGL if it's available rather than OpenGL.framework. I don't think it should be required for libpng to work, it's not provided by a framework AFAIK, and omitting this line allows the builder to control the behavior by passing it via the CLI (eg. `-DCMAKE_FIND_FRAMEWORK=LAST`)

PR #5626, opened 2022-10-25
consume every keydown event in editbox

### LINKED ISSUE #5585: Ctrl+Arrow in Editbox Moves Map View
**Describe the bug**
Pressing Ctrl+(Left or right arrow key) in an editbox (e.g. chat) in-game moves the map view in addition to the editbox's handling of the event.

**To Reproduce**
1. Start an MP game and open the chat box. Type several words.
2. Hold down *first* Ctrl, and *then* press the Left or Right arrow key. The cursor moves one word at a time. No other side effects. As expected.
3. Now *first* press and hold the Left or Right arrow key, and *then* press Ctrl. Now the cursor moves one word at a time (as expected), and additionally the map view jumps by one screen page.

**Expected behavior**
The edit box handles the arrow key event, so it should not be forwarded to the map view.

**Version:**
 - OS: Ubuntu MATE 22.04
 - Widelands Version: https://github.com/widelands/widelands/releases/tag/Tournament2022

**Additional context**
Discovered in a tournament game.

jmoerschbach:
Fixes #5585 
Consume every keydown event like `multilineeditbox` does.
Alternative approach could be to only add handling of ctrl keystrokes to the switch statement:

```
case SDLK_LCTRL:
case SDLK_RCTRL:
	return true;
```
But that would enable other funny side effects like:

- hit `CTRL` + `S` while textbox is in focus -> shortkey for saving the game is triggered (kind of okay maybe)
- hit `CTRL` + `+` while textbox is in focus -> map is zoomed in while `+` is added to the textbook (iirgh)
- hit `CTRL` + `-` while textbox is in focus -> map is zoomed out while `-` is added to the textbook (iirgh)
- probably other stuff

so I am in favor to consume all keystrokes :)

### DIFF
--- a/src/ui_basic/editbox.cc
+++ b/src/ui_basic/editbox.cc
@@ -623,7 +623,7 @@ void EditBox::draw_caret(RenderTarget& dst, const Vector2i& point, const uint16_
 
 	const Image* caret_image =
 	   g_image_cache->get(panel_style_ == PanelStyle::kWui ? "images/ui_basic/caret_wui.png" :
-	                                                         "images/ui_basic/caret_fs.png");
+                                                            "images/ui_basic/caret_fs.png");
 	Vector2i caretpt = Vector2i::zero();
 	caretpt.x = point.x + m_->scrolloffset + caret_x - caret_image->width() + kLineMargin;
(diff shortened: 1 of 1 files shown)


Noordfrees (approved):
LGTM, tested and working :)

PR #5630, opened 2022-10-29
explain clearer what an economy is

SimonHeimberg:
Improve the hint about what is what one economy is, as it is not connected by roads only.


SimonHeimberg:
That ferries and ships connect economies is what I have found out by experience. I might be wrong.
And I am unsure about the terms to use. "Ferries", "water ways", "ferry water ways" or anything better and "ship lines", "ship transportation", "ship transport lines" or something better.

data/txts/tips/general_game.lua
```diff
@@ -62,7 +62,7 @@ tips = {
       seconds = 4
    },
    {
-      text = _("An economy consists of all buildings and flags connected by roads. The supplies and demands of buildings are handled in its economy."),
+      text = _("An economy consists of all buildings and flags connected by roads, ferries or ship lines. The supplies and demands of buildings are handled in its economy."),
```
tothxa: I think "ferries" is OK. How about this:

edited to accomodate @hessenfarmer's comment. The last sentence probably needs improvement.

```suggestion
      text = _("An economy consists of all buildings and flags connected by roads, ferries or ships. The supplies and demands of buildings are handled in its economy. A player may have more than one economies, if those are not connected by any of these means. This is why the configure economy menu can only be accessed from flags and warehouses, not the player’s main menu."),
```

hessenfarmer (commented):
not only the demand and supply is economy dependent, but the economy settings only belong to the economy of the chosen flag. We should add this as well as a hint, that a player can have multiple economies.

hessenfarmer:
@SimonHeimberg 
any plans for incorporating this

tothxa:
I apologise for hijacking, but there had been no response for almost 2 weeks.

hessenfarmer (approved):
LGTM.
Thanks @tothxa

PR #5633, opened 2022-10-30
delete selected text when new text input is done

### LINKED ISSUE #5632: TextInput: Selected text is not replaced
**Describe the bug**
Handling of selected text in editbox and multilineeditbox is counterintuitive. Selected text should be replaced when text is input. Instead, the new text is inserted at cursor position

**To Reproduce**
1. Type in text in a editbox
2. Select all or parts of the text
3. Hit a character on your keyboard
4. The new character is inserted at current cursor position

**Expected behavior**
The selected text should vanish and the new character should replace it

jmoerschbach:
Fixes #5632


Noordfrees (approved):
Many thanks for fixing this :) 
LGTM, tested and working

PR #5634, opened 2022-10-30
Upscaled editor splash image

### LINKED ISSUE #5570: The editor should get a bigger load screen
The editor still has a 800x600 loadscreen only, and it isn't even scaled for bigger screens.

It should get a higher resolution loadscreen, that looks good at various scales, and it should be scaled to the window.

**Additional context**
Current editor loadscreen in a high resolution window:
editor_load

matthiakl:
Fixes #5570


src/logic/editor_game_base.h
```diff
@@ -127,6 +127,7 @@ class EditorGameBase {
 	                                     bool show_game_tips,
 	                                     const std::string& theme,
 	                                     const std::string& background,
+	                                     bool crop = true,
```
Noordfrees:
```suggestion
	                                     bool crop,
```

src/ui_basic/progresswindow.h
```diff
@@ -45,7 +45,10 @@ struct IProgressVisualization {
 
 /// Manages a progress window on the screen.
 struct ProgressWindow : public UI::Panel {
-	explicit ProgressWindow(UI::Panel*, const std::string& theme, const std::string& background);
+	explicit ProgressWindow(UI::Panel*,
+	                        const std::string& theme,
+	                        const std::string& background,
+	                        bool crop = true);
```
Noordfrees, line 51:
```suggestion
	                        bool crop);
```

Noordfrees (commented):
LGTM. I think it would be safer to remove the default value though.

Noordfrees (approved):

PR #5642, opened 2022-11-02
Move empty sign on Atlantean mines to make it visible when status labels are shown

### LINKED ISSUE #1735: Atlantean "mine empty" marker obscured by stats overlay
The "mine exhausted" marker sits underneath the stats percentage for Atlantean mines, making the marker very difficult to notice.

 * date created: 2019-08-10T16:47:27Z
 * owner: ad-simplypeachy
 * the launchpad url was https://bugs.launchpad.net/bugs/1839722

tothxa:
fixes #1735


hessenfarmer (approved):
Immages look good to me.
However I was thinking about removing the wares from the empty mines as well. this would be even better to distinguish, but one might not see without the stats what mine this had been. 
So I am unsure about this.

tothxa:
I can decrease the size of the lumps in empty mines.

hessenfarmer:
> I can decrease the size of the lumps in empty mines.

would be a good compromise :-)

PR #5646, opened 2022-11-03
fix portspace discovery

### LINKED ISSUE #5594: Expedition can build port on port space blocked by rocks
**Describe the bug**
An expedition reports port space even if it is blocked by rocks. It is then possible to build a port, in which case the blocking rocks are instantly removed.

**To Reproduce**
1. See attached savegame: [port_space.zip](https://github.com/widelands/widelands/files/9719614/port_space.zip)

**Expected behavior**
Blocked portspaces shouldn't be found by expeditions.

**Screenshots**
blocked port space found

**Version:**
 - OS: Debian buster

hessenfarmer:
fixes #5594


frankystone:
Using the save game from the bug report and placing a tree at the portspace by script the port space isn't found with this branch.

Save game: https://github.com/widelands/widelands/files/9719614/port_space.zip

Placing tree:
~~~~
wl.Game().map:place_immovable("alder_summer_mature", wl.Game().map:get_field(55,39))
~~~~

hessenfarmer:
well thi specific portspace should not be found as it's flag is blocked by rocks

frankystone:
Hm, maybe i did not understand your fix?

If no tree is placed the port space is found and the rocks are removed if placing a port space. This is what i had expected by your fix. But after placing the tree the port space can't be found anymore. So i think this brings back the tree-seeding problem?

Maybe i am doing something wrong…

hessenfarmer:
the fix intends to not find a portspace if rocks or other immovables with size small or bigger are blocking the space but find it if only trees are present.  this check is done for all 4 port fields and the flag field as well.

hessenfarmer:
I screwed the branch up so closing this and doinmg it all again

PR #5647, opened 2022-11-04
5594 fix portspace detection

### LINKED ISSUE #5155: Expedition can't find port space although it is shown in buildhelp
After destroying/dismantling a port, an expedition ship doesn't find the port space anymore. After some Roads/Buildings around the port space are destroyed the expedition ship finds the port space.

**To Reproduce**
1. Load attached save game
2. Let the expedition ship try to find the open port space -> It can't find it
3. Destroy roads around the port space
4. Let the expedition ship try to find the open port space -> It can't find it
5. Destroy the Armor smithy
6. Let the expedition ship try to find the open port space -> Now it finds the port space and the port can be built

I had a similar situation where destroying roads was enough to let the expedition ship find the port space.

**Expected behavior**
If the buildhelp shows a spot for a port space, an expedition should find it.

**Version:**
 - OS: archlinux
 - Widelands Version: 1.1~git25551[3d01188@4929-portpaths-ordering] (Debug)
 
[no_port_space_found.zip](https://github.com/widelands/widelands/files/7698577/no_port_space_found.zip)

### LINKED ISSUE #5594: Expedition can build port on port space blocked by rocks

hessenfarmer:
fixes #5594 
fixes #5155 

- rocks or ruins will block a portspace from detection
- trees will not block detection
- now all trees including immature ones will be removed at construction
- ownership and availability is rechecked periodically and before constructing a port (might be conquered from detection to construction
- a flag at the flag field and some roads around the 4 needed spaces are allowed (build the expedition port on own territory)


frankystone:
Thanks for a more detailed explanation.

Works as described :-)

tothxa:
I had already built this for testing, just hadn't got around to test it. :)

The change looks OK, but I've found a problem in testing. The port space next to the rocks in the following image isn't found, even though it is shown as buildable in the editor:

portspace_editor

(yes I wanted to push it to the limit by also testing whether clearing nearby area still works ;)

If I remove the rock to the East of the portspace's flag position, then it is found. (and the other rocks are cleared by the expedition when starting to build the port)

I'm not sure though if we could limit the check to the 4 fields of the portspace plus the flag in `can_build_port_here()`. I'm inclined to think we shouldn't.

BTW I've found a bug with the clearing code: Only fully grown trees are removed:

growing_tree_not_removed

Should I open a new bug for this or can it be fixed in this PR?

Here's the map I modified for testing: 
[portspace_test_map.zip](https://github.com/widelands/widelands/files/9947128/portspace_test_map.zip)

hessenfarmer:
Will try to fix both tonight:  
for the clearance issue we currently use the attrib "tree" which only full grown trees have. Will use has_terrain affinity there as well.     
for the other issue we scan each field in the radius of 1 for the reason to be far enough from any border. so we might need to split the check for the owner of a field and the check for being occupied by something solid.

src/logic/map_objects/tribes/ship.cc
```diff
@@ -54,47 +54,42 @@ namespace {
...
+	map.get_brn(coord, &c[3]);  // this is the flag
 	for (const Widelands::FCoords& fc : c) {
-		MapRegion<Area<FCoords>> area(map, Area<FCoords>(fc, 1));
+		if (!can_support_port(fc)) {  // check for blocking immovables
+			return false;
+		}
+		MapRegion<Area<FCoords>> area(map, Area<FCoords>(fc, 1));  // radius 1 for owner check
 		do {
-			if (!can_support_port(player_number, area.location())) {
+			const PlayerNumber owner = coord.field->get_owned_by();
```
tothxa: I think this should be something like

```suggestion
			const PlayerNumber owner = area.location().field->get_owned_by();
```

(not tested)

tothxa (commented):
Tested the changes, and now everything worked as intended. I think I've found a copy-paste mistake in the code, see inline. Testing its effect is hard, because it's a special corner case if I'm not mistaken. (only the base coordinate is checked for ownership, not the surrounding area of all fields occupied by the port)

hessenfarmer:
thanks for spotting my mistakes
Hopefully this will fix the testsuite as well

tothxa (approved):
Thank you. I think it's fine now.

I've found yet another old bug though while testing portspaces near borders that could be fixed here if you don't mind yet another round: `exp_construct_port()` should call `can_build_port_here()` too as a check before placing port. If another player conquers the area while the ship is waiting at a portspace, it is now possible to build a port and take the land away from the other player. Of course if you move the ship away and bring it back it doesn't find the portspace any more. [savegame for illustration](https://github.com/widelands/widelands/files/9961702/portspace_found_conquered.zip)

Ideally ships waiting at a portspace should also regularly re-check the availability of that portspace.

hessenfarmer:
Now that we touched this we will surely fix every bug that arises in testing. no matter how many rounds it takes.   
thanks for finding this very old cornercase bug

tothxa:
I've found another old bug report concerning portspace detection: #5155

Unfortunately the savegame can't be loaded even by v1.1, so I'll try to recreate the situation and check with this branch.

edit: done, bug exists with this branch too, new savegame for testing was uploaded to original bug report

src/logic/map_objects/tribes/ship.cc
```diff
@@ -883,6 +889,19 @@ WalkingDir Ship::get_scouting_direction() const {
...
+	do {
+		const PlayerNumber owner = area.location().field->get_owned_by();
+		if (owner != neutral() && owner != get_owner()->player_number()) {
+			// enemy player was faster conquering the spot
+			set_ship_state_and_notify(
+			   ShipStates::kExpeditionWaiting, NoteShip::Action::kDestinationChanged);
+			return;
+		}
+	} while (area.advance(*map));
+
```
tothxa: I really think `can_build_port_here()` should be used (I'm not sure if I got the syntax right though):

```suggestion
	// recheck ownership and immovables before setting the csite
	Map* map = game.mutable_map();
	if (get_owner()->player_number(), *map, !can_build_port_here(c)) {
		// enemy player was faster conquering the spot or player built something over it
		set_ship_state_and_notify(
		   ShipStates::kExpeditionWaiting, NoteShip::Action::kDestinationChanged);
		return;
	}

```

This way there's no need to maintain the code in 2 places, and checking buildings and flags is necessary, see 
[savegame](https://github.com/widelands/widelands/files/9974405/portspace_found_eliminated.zip)

tothxa (commented):
Thank you for the fixes. See comment inline.

src/logic/map_objects/tribes/ship.cc
hessenfarmer: I am unsure whether a player would want to last its road here or just build the port.   
Furthermore this situation seems to be very constructed, i.e. why keeping the ship around and build something over the port?
Edit: the current situation only works for roads or flags. A building will be save already

tothxa: Yes, it's true that it's unlikely that anyone would do it intentionally. So the most likely use case would be that they click the build port button accidentally instead of a direction button or mistaking it for the window of another ship. But in these cases the buildings, roads and flags shouldn't be removed either.

But in my opinion, removing code duplication alone makes it worth the change.

Are you sure about buildings? I tested it, and the current implementation happily removes them.

hessenfarmer: I am sure.  a building or constructionsite direct on the spot will not be removed.

hessenfarmer: but I see the point about the code duplication and prevent unwanted actions. So i will making a change

tothxa: > I am sure. a building or constructionsite direct on the spot will not be removed.

I'd guess `force_csite()` does it. Please test it yourself. :)

hessenfarmer: line 906 will prevent reaching the csite, but only for a building on the main spot of a port space. Teested it.
however I am working on a periodic check now as you sugested.   
Only it is hard to get the coords of the desired port space. will need some time and debugging

tothxa: Yes, sorry, that's true for the portspace location itself. What I tried was placing the building on one of the neighbouring nodes. That building is removed if the port is placed.

Thanks for trying the more involved feature too.

tothxa:
Sorry, there is a regression in the fix. The neighbouring fields of the 4 main fields (but not the other neighbours of the flag) have to be checked for stones to only find portspaces that are also buildable from land.

hessenfarmer:
> Sorry, there is a regression in the fix. The neighbouring fields of the 4 main fields (but not the other neighbours of the flag) have to be checked for stones to only find portspaces that are also buildable from land.

this would make a fix for #5155 at least very hard. Furthermore why should a port not be buildable from land if one of the neighbours is occupied by a rock

tothxa:
> Furthermore why should a port not be buildable from land if one of the neighbours is occupied by a rock

Well, this is what #5594 was about. :) Because the port space is shown as blocked in the editor and in the game as well if it is conquered from land. Spreading of trees complicates the issue, but at least in the absence of trees, the expedition check should give the same result as the normal buildspace check. (actually I even think that if possible, the tree issue should be solved by not allowing automatic seeding of trees near portspaces instead of special casing them here, to not mislead map designers)

hessenfarmer:
ok I see the point. I forgot that rocks are big.  Will try to fix

tothxa:
I think we need these checks (size 0 and trees count as nothing, unless seeding is solved, in which case we just have to check buildcap and ownership):

 - nothing is allowed on coords, ln, tln, trn
 - only own flag is allowed at flagpos = brn
 - only own road is allowed at bln and rn
 - size 1 except flag is allowed at brn, bln and rn of flagpos
 - only own road or own flag is allowed at trn+rn, trn+trn, trn+tln, tln+tln, ln+tln, ln+ln, ln+bln

hessenfarmer:
yep, will try to call / copy the logic of the editor, or the game. Just need to find it. 
I am still struggling with the periodic check. Need to follow a complete new idea. maybe tonight.

src/logic/map_objects/tribes/ship.cc
```diff
@@ -52,49 +52,62 @@ namespace Widelands {
 
 namespace {
 
-/// Returns true if 'coord' is not occupied or owned by 'player_number' and
-/// nothing stands there.
-bool can_support_port(const PlayerNumber player_number, const FCoords& coord) {
-	const PlayerNumber owner = coord.field->get_owned_by();
-	if (owner != neutral() && owner != player_number) {
+/// Returns true if 'coord' is not blocked by immovables
```
tothxa, line 55:
```suggestion
/// Returns true if 'coord' is not blocked by immovables
/// Trees are allowed, because we don't want spreading forests to block portspaces from expeditions
```

src/logic/map_objects/tribes/ship.cc
```diff
@@ -439,6 +452,19 @@ void Ship::ship_update_expedition(Game& game, Bob::State& /* state */) {
 			             _("An expedition ship found a new port build space."),
 			             "images/wui/editor/fsel_editor_set_port_space.png");
 		}
+	} else if (ship_state_ == ShipStates::kExpeditionPortspaceFound) {
+		if (!expedition_->seen_port_buildspaces.empty()) {
```
tothxa: I don't think it should ever be empty. Maybe log it if it happens anyway?

src/logic/map_objects/tribes/ship.cc
```diff
@@ -883,23 +909,32 @@ WalkingDir Ship::get_scouting_direction() const {
...
+	if (!expedition_->seen_port_buildspaces.empty()) {
+		const FCoords fc = map->get_fcoords(expedition_->seen_port_buildspaces.front());
+		if (!can_build_port_here(get_owner()->player_number(), *map, fc)) {
+			set_ship_state_and_notify(
+			   ShipStates::kExpeditionWaiting, NoteShip::Action::kDestinationChanged);
+			send_message(
+			   game, _("Port Space"), _("Port Space Lost"),
+			   _("A discovered port build space is not available for building a port anymore."),
+			   "images/wui/editor/fsel_editor_set_port_space.png");
+			expedition_->seen_port_buildspaces.clear();
```
tothxa: Please make this a function to avoid duplication.

src/logic/map_objects/tribes/ship.cc
```diff
@@ -883,23 +909,32 @@ WalkingDir Ship::get_scouting_direction() const {
...
+			return;
+		}
 	}
 	get_owner()->force_csite(c, get_owner()->tribe().port()).set_destruction_blocked(true);
 
 	// Make sure that we have space to squeeze in a lumberjack
```
tothxa:
```suggestion
	// Make sure that we have space to squeeze in a lumberjack or a quarry
```

tothxa (commented):
Thank you. I think everything works now. Just some nitpicking inline. :)

For the record, there's a corner case of somewhat inconcistent behaviour with owned land because own flags aren't ignored now:

 - if only trees are present, then the portspace is found
 - if trees and otherwise compatible roads and flags are present then the portspace is not found
 - (of course if the portspace is clear and only compatible roads and flags are present, then the first check ensures that it is found)

But as this is already conquered land, it shouldn't matter much as #5155 is fixed by the first check.

src/logic/map_objects/tribes/ship.cc
hessenfarmer: this would need to be a function of the ship, otherwise a lot of values won't be available in the function and we would need to call the funtion with approx 6 parameters.  
although I favor avoiding duplication in general I simply deem this not enough for an own ship function and not worth creating a function with 6 params. So I think we leave it like it is

src/logic/map_objects/tribes/ship.cc
hessenfarmer: Well I simply was not sure whether there might be a cornercase for it to be empty. e.g. a race between construct and periodic check. so i tried to safeguard the thing without throwing an error.

src/logic/map_objects/tribes/ship.cc
tothxa: I think this is long enough for a private function for `Ship`. Now you even have the log message in addition to the 2 new translatable strings. Plus 2 levels of `if`s and a local variable. The functionality is also quite important.

src/logic/map_objects/tribes/ship.cc
```diff
@@ -883,23 +912,35 @@ WalkingDir Ship::get_scouting_direction() const {
...
+			send_message(
+			   game, _("Port Space"), _("Port Space Lost"),
+			   _("A discovered port build space is not available for building a port anymore."),
+			   "images/wui/editor/fsel_editor_set_port_space.png");
+			expedition_->seen_port_buildspaces.clear();
+			return;
+		}
+	} else {
+		log_warn_time(
+		   game.get_gametime(), "Expedition list of seem üport spaces is unexpectedly empty!\n");
```
tothxa:
```suggestion
		   game.get_gametime(), "Expedition list of seen port spaces is unexpectedly empty!\n");
```

(also the other one – see? its worth the function ;)

tothxa:
Re testsuite regression: Shouldn't we add tests for this?

hessenfarmer:
> Re testsuite regression: Shouldn't we add tests for this?       
   
Well we have the tests and they failed so they are good enough imho. the problem was that we need to check for ownership first, because else the normal port build check fails. What else would you have checked. 

Edit: ok now I understand. Do you want us to check for trees allowed - rocks not?

test/maps/port_space.wmf/scripting/test_portfield_blocked_by_rock.lua
```diff
@@ -0,0 +1,9 @@
+run(function()
+   --  This places to tower so that the port just fits the space.
+   map:place_immovable("greenland_rocks4", map:get_field(7,10))
+   local ship = create_infrastructure(59, 10)
+   game.desired_speed = 1000
+   wait_for_message("Land Ahoy!")
```
tothxa, line 6: Please extend with adding a rock to one of the neighbouring road fields and waiting for the portspace lost message.

edit: Sorry, "Land Ahoy!" means no portspace found. So scrap this one.

test/maps/port_space.wmf/scripting/test_portfield_covered_by_tree.lua
```diff
@@ -0,0 +1,9 @@
+run(function()
+   --  This places to tower so that the port just fits the space.
+   map:place_immovable("alder_summer_old", map:get_field(7,10))
+   local ship = create_infrastructure(59, 10)
+   game.desired_speed = 1000
+   wait_for_message("Port Space")
```
tothxa, line 11: Please also test with a tree and a young tree placed to two of the main fields of the portspace. Maybe with a rock again at (7,10).

tothxa (commented):
Thank you very much. :) (I thought `recheck_portspace()` would suffice for function name. :)

Some requests inline for extending test coverage. Furthermore, please also add a test for #5155 by placing nearest possible road with flags around the portspace. Maybe again test eliminating portspace afterwards by placing a road or some building inside.

src/logic/map_objects/tribes/ship.cc
```diff
@@ -717,6 +732,30 @@ void Ship::set_ship_state_and_notify(ShipStates state, NoteShip::Action action)
...
+			expedition_->seen_port_buildspaces.clear();
+			return false;
+		}
+	} else {
+		log_warn_time(
+		   game.get_gametime(), "Expedition list of seen port spaces is unexpectedly empty!\n");
+		return false;
+	}
+	return true;
+}
```
tothxa: I think this makes it actually more readable:

```suggestion
	if (expedition_->seen_port_buildspaces.empty()) {
		log_warn_time(
		   game.get_gametime(), "Expedition list of seen port spaces is unexpectedly empty!\n");
		return false;
	}
	const FCoords fc = map->get_fcoords(expedition_->seen_port_buildspaces.front());
	if (!can_build_port_here(get_owner()->player_number(), *map, fc)) {
		set_ship_state_and_notify(
		   ShipStates::kExpeditionWaiting, NoteShip::Action::kDestinationChanged);
		send_message(
		   game, _("Port Space"), _("Port Space Lost"),
		   _("A discovered port build space is not available for building a port anymore."),
		   "images/wui/editor/fsel_editor_set_port_space.png");
		expedition_->seen_port_buildspaces.clear();
		return false;
	}
	return true;
}
```

hessenfarmer: indeed. many thanks

hessenfarmer:
Hopefully I am done now ;-)

src/logic/map_objects/tribes/ship.cc
```diff
@@ -717,6 +732,28 @@ void Ship::set_ship_state_and_notify(ShipStates state, NoteShip::Action action)
...
+	if (expedition_->seen_port_buildspaces.empty()) {
+		log_warn_time(
+		   game.get_gametime(), "Expedition list of seen port spaces is unexpectedly empty!\n");
+		return false;
+	}
+	const FCoords fc = map->get_fcoords(expedition_->seen_port_buildspaces.front());
+	if (!can_build_port_here(get_owner()->player_number(), *map, fc)) {
+		set_ship_state_and_notify(
+		   ShipStates::kExpeditionWaiting, NoteShip::Action::kDestinationChanged);
+		send_message(game, _("Port Space Lost!"), _("No Port Can Be Build"),
```
tothxa: Typo

```suggestion
		send_message(game, _("Port Space Lost!"), _("No Port Can Be Built"),
```

tothxa:
> Hopefully I am done now ;-)

I'll need some time to go over the tests in detail. :( I hoped I'd be able to watch them progress, but I'm afraid that's not possible. Checking them blindly will be harder.

src/logic/map_objects/tribes/ship.cc
```diff
@@ -52,49 +52,62 @@ namespace Widelands {
...
+	map.get_rn(coord, &c[4]);
+	map.get_brn(coord, &c[5]);
+	map.get_bln(coord, &c[6]);
+	map.get_bln(c[0], &c[7]);
+	map.get_ln(c[0], &c[8]);
+	map.get_tln(c[0], &c[9]);
+	map.get_tln(c[1], &c[10]);
+	map.get_trn(c[1], &c[11]);
+	map.get_trn(c[2], &c[12]);
+	map.get_rn(c[2], &c[13]);
```
tothxa: I found a mistake in https://github.com/widelands/widelands/pull/5647/commits/e856f8e7a72eb4f383813e3ec11f8c0677df8ca0 doing my usual tests. Base coordinate indexes were not increased when main `coord` was reinserted as `c[0]`.

```suggestion
	map.get_bln(c[1], &c[7]);
	map.get_ln(c[1], &c[8]);
	map.get_tln(c[1], &c[9]);
	map.get_tln(c[2], &c[10]);
	map.get_trn(c[2], &c[11]);
	map.get_trn(c[3], &c[12]);
	map.get_rn(c[3], &c[13]);
```

tothxa: So full test coverage would need trying with a rock or other size 1 immovable at each of these 14 positions one by one, but do we need or want that?

hessenfarmer: I don't think so, cause to test all permutations we might need approx more then 50 tests. as we might need to test with size 1 immos or size 3 (like rocks as they are big actually and block more space).

hessenfarmer:
@tothxa  any further comments

tothxa:
I've modified the test map and one of the test cases to test the port space surrounded by rocks at the nearest possible positions, and added some size small immovables where the buildcap checks allow them: 
[port_space_wmf.zip](https://github.com/widelands/widelands/files/10044710/port_space_wmf.zip)

This however fails, but I think it shouldn't. Here's the new portspace in the editor without trees:

portspace

So adding only trees shouldn't make it fail. I'm not sure it's only the artifacts that make it fail though, because it still fails if I remove everything.

tothxa:
I've finally looked through all tests. (even figured out that I can actually watch them work :) They're OK, I think only my above test case is needed in addition.

I've modified `wait_for_message()` to abort on wrong message to not depend on time-outs. Not sure if it's safe though, but all tests (except for the above failure of course) passed for me with this change. (I also wonder why it has several long delays, but I didn't touch them.) I also added progress messages to the tests which have multiple steps:

[port_space_test_wmf.zip](https://github.com/widelands/widelands/files/10045700/port_space_test_wmf.zip)

tothxa:
Here's a patch that makes detection work the same as in the editor, except for trees (I checked that medium isn't allowed on the Northern/Western 2nd neighbours): 
[size_small_fix_patch.txt](https://github.com/widelands/widelands/files/10046540/size_small_fix_patch.txt) (edit: fixed East->West in comment)

My test case still doesn't find it, but I can't see why. On my big test map the same immovable arrangement works, so it must be some interference from the other port or the owned land, but I couldn't find in the code how that could affect the result.

Sorry for all the trouble I make.

hessenfarmer:
> Here's a patch that makes detection work the same as in the editor, except for trees (I checked that medium isn't allowed on the Northern/Western 2nd neighbours): [size_small_fix_patch.txt](https://github.com/widelands/widelands/files/10046540/size_small_fix_patch.txt) (edit: fixed East->West in comment)
> 
> My test case still doesn't find it, but I can't see why. On my big test map the same immovable arrangement works, so it must be some interference from the other port or the owned land, but I couldn't find in the code how that could affect the result.
> 
> Sorry for all the trouble I make.

no need to apologize: testing my buggy code thoroughly isn't making trouble it is adding quality. So thanks for doing so.   
I added a patch quite similar to yours and ran into the same issue that it is not finding it. So I started to add some debug code but did not succeed yet

tothxa:
It's the portdock check. It isn't allowed on the border. Here's the fixed test case: 
[port_space_test_wmf.zip](https://github.com/widelands/widelands/files/10046571/port_space_test_wmf.zip)

edit: OTOH border with neutral shouldn't fail. At first sight it may seem as a narrow corner case, but actually it may make a neighbouring island unconquerable. I'm afraid though that checking whether other size is neutral or other player wouldn't be trivial.

hessenfarmer:
> It's the portdock check. It isn't allowed on the border. Here's the fixed test case: [port_space_test_wmf.zip](https://github.com/widelands/widelands/files/10046571/port_space_test_wmf.zip)
> 
> edit: OTOH border with neutral shouldn't fail. At first sight it may seem as a narrow corner case, but actually it may make a neighbouring island unconquerable. I'm afraid though that checking whether other size is neutral or other player wouldn't be trivial.

I think we can safely use the force true in ship.cc line 441. will test that. and report back

hessenfarmer:
ok using the force option of find_portdock shpul dreally deliver all theoretical places as intended. 
Should be save as we check them later on . at least it worked locally lets see what the test suite says.

tothxa:
> ok using the force option of find_portdock shpul dreally deliver all theoretical places as intended. Should be save as we check them later on . at least it worked locally lets see what the test suite says.

Indeed. I think it's fine now if all tests pass.

tothxa (approved):
Thank you. :)

hessenfarmer:
> Thank you. :)

Thank you for such good testing and inputs

PR #5653, opened 2022-11-11
Explain waterway setting in Editor Map Options and add toolips to all map tags

### LINKED ISSUE #5582: reword Waterway length limit

### LINKED ISSUE #5666: waterway length limit Spin box buttons miss step definitions
The spin box buttons of the waterway length limit spinbox are not correctly translated see screenshot.    
furthermore with CTRL pressed it jumps to max

shot0000

tothxa:
Fixes #5582 (first 2 tasks for now), replaces #5592

I don't understand why the tooltip is not shown for the textarea…


src/editor/ui_menus/main_menu_map_options.cc
```diff
@@ -394,10 +394,16 @@ MainMenuMapOptions::MainMenuMapOptions(EditorInteractive& parent, Registry& regi
 
 	tags_box_.add_space(labelh_);
 
-	tags_box_.add(new UI::Textarea(&tags_box_, UI::PanelStyle::kWui, UI::FontStyle::kWuiLabel, 0, 0,
-	                               max_w_, labelh_, _("Waterway length limit:")));
+	UI::Textarea* ww_text =
+	   new UI::Textarea(&tags_box_, UI::PanelStyle::kWui, UI::FontStyle::kWuiLabel, 0, 0, max_w_,
+	   labelh_, _("Ferry range:"));
+	std::string ww_tooltip = _("Set maximum length of waterways for ferries. If ferry range is set "
+	   "to 0, then the map will not allow building ferry yards and waterways (ferry routes).");
```
frankystone: The UI already shows "Deactivated" for a ferry range <2. So the last sentence can be skipped imho.

What about: "Allow ferries and ferry yards on this map by defining a maximum range for Ferries. The value reflects the amount of allowed fields."

Noordfrees:
> I don't understand why the tooltip is not shown for the textarea…

You need to call `set_handle_mouse(true)` on the textarea to enable tooltips

tothxa:
@frankystone Please check new wording.

@Noordfrees Thank you. I was just one step from finding it yesterday.

src/editor/ui_menus/main_menu_map_options.cc
```diff
@@ -394,10 +394,17 @@ MainMenuMapOptions::MainMenuMapOptions(EditorInteractive& parent, Registry& regi
 
 	tags_box_.add_space(labelh_);
 
-	tags_box_.add(new UI::Textarea(&tags_box_, UI::PanelStyle::kWui, UI::FontStyle::kWuiLabel, 0, 0,
-	                               max_w_, labelh_, _("Waterway length limit:")));
+	UI::Textarea* ww_text =
+	   new UI::Textarea(&tags_box_, UI::PanelStyle::kWui, UI::FontStyle::kWuiLabel, 0, 0, max_w_,
+	                    labelh_, _("Ferry range:"));
+	std::string ww_tooltip = _("Enable ferries, waterways, and ferry yards on this map by setting"
+	                           " maximum length of waterways for ferries.");
```
frankystone: Fine :-)

But shouldn't it be "… by setting **a** maximum length of waterways for ferries."?

tothxa:
This should be almost ready now, but please suggest improvements to the style of tooltips. Or should I allow optional different tooltip texts for setting and filtering?

frankystone:
Looks good to me. One nit: The tooltip for "Theme" is only shown when hovering the arrow of the dropdown box, not when hovering the text box. This differs from the other drop downs.

Not related to this branch: The "Teams" tags (in second tab) are a bit misleading. For example i can set them for a scenario with just one player and they make not really sense if no teams are specified in the third tab ("Teams"), imho. Probably worth another bug report?

src/wui/map_tags.cc
```diff
@@ -24,33 +24,73 @@
...
+    {/** TRANSLATORS: This is a map tag */
+     gettext_noop("Balanced"),
+     /** TRANSLATORS: This is a tool tip for a map tag */
+     gettext_noop(
+        "Player starting positions were designed to provide equal conditions for all players")}},
+   {"unbalanced",
+    {/** TRANSLATORS: This is a map tag */
+     gettext_noop("Unbalanced"),
+     /** TRANSLATORS: This is a tool tip for a map tag */
+     gettext_noop("No efforts were made to provide equal conditions for the players")}},
```
frankystone: This sounds a bit like the map maker was too lazy for providing equal starting positions. I think this should be a bit more positive. What about:
"Set this to 'Balanced' if you think each player has equal starting conditions, e.g. equal distances to resources."

Something like that, giving also a hint how to achieve equal starting conditions :-)

tothxa:
Now I'm quite satisfied with it. I even sidestepped the need for different tooltips in the editor and in mapselect.

@frankystone:
> Not related to this branch: The "Teams" tags (in second tab) are a bit misleading. For example i can set them for a scenario with just one player and they make not really sense if no teams are specified in the third tab ("Teams"), imho. Probably worth another bug report?

Feature request opened: #5657 :)

frankystone:
Great :+1: 

Just found another few things, but since you are on it :-) :  
Suggestion for the tooltips of the table headers of maps in the select maps screen:

* Number of players -> Sort by Number of Players
* The Name of the map or scenario -> Sort by Name of the Map 
* Size of the Map (Width x Height) -> Sort by Map Dimensions

tothxa:
> Just found another few things, but since you are on it :-) : Suggestion for the tooltips of the table headers of maps in the select maps screen:
> 
>     * Number of players -> Sort by Number of Players
> 
>     * The Name of the map or scenario -> Sort by Name of the Map
> 
>     * Size of the Map (Width x Height) -> Sort by Map Dimensions

These tooltips are actually set elsewhere (`wui/{map,savegame}table.cc`), and they're quite elaborate, so I wouldn't feel comfortable changing them. Also, the behaviour on mouseclick is either sort by given column, or if it is already done, then reverse sorting order. Adapting the tooltip to that would lead very far IMO. (dynamic tooltip handling in table class with format strings to allow user part? well, I think it could be done, but I don't think it's worth the trouble – neither the effort, nor the code complexity)

hessenfarmer:
@frankystone can this be merged from your testing?

frankystone:
> Adapting the tooltip to that would lead very far IMO. (dynamic tooltip handling in table class with format strings to allow user part? well, I think it could be done, but I don't think it's worth the trouble – neither the effort, nor the code complexity)

I didn't meant to have dynamic tooltips. Just adding the word "Sort…" would be enough, because the headers don't appear to be buttons.

@hessenfarmer I am fine with it right now, but i couldn't do a code review. Thanks @tothxa for the changes :+1:

src/wui/map_tags.cc
```diff
@@ -24,33 +24,73 @@
...
-   /** TRANSLATORS: This is a map tag */
-   {"2teams", _("Teams of 2")},
-   /** TRANSLATORS: This is a map tag */
-   {"3teams", _("Teams of 3")},
-   /** TRANSLATORS: This is a map tag */
-   {"4teams", _("Teams of 4")},
+const std::map<std::string, TagTexts> kMapTags = {
+   {"official",
+    {/** TRANSLATORS: This is a map tag */
+     gettext_noop("Official"),
```
hessenfarmer, line 30: Will gettext_noop work with our translations in the same way as Gettext with options?

hessenfarmer (approved):
Code LGTM 
one question inline

src/wui/map_tags.cc
tothxa: Yes, we already have a few instances of it. I also tried `utils/buildcat.py` on this because I was unsure at one point about whether some translator comments would turn out right, and it worked nicely.

hessenfarmer: ok

hessenfarmer:
so this can go in if @tothxa won't do the changes on the other tooltips as @frankystone suggested.   
During testing I found another issue but this does not need to be solved here (could be of course)
see #5666

tothxa:
> so this can go in if @tothxa won't do the changes on the other tooltips as @frankystone suggested. During testing I found another issue but this does not need to be solved here (could be of course) see #5666

I think both should be in separate PRs, as they both affect things in `src/ui_basic/`, independent of map options and map tags. (well, at least I still think that the sorting tooltips should be handled in `UI::Table` for a uniform general solution)

tothxa:
fixes #5666

hessenfarmer (approved):
Code Looks good.
Will test later and merge then

hessenfarmer:
tested and looks fine thanks for fixing

tothxa:
Thanks to all of you for testing and reviewing.

PR #5661, opened 2022-11-18
fix buildings help button

frankystone:
The buildings help button in the buildings window didn't worked. This fixes it.

### DIFF
--- a/data/tribes/scripting/help/format_help.lua
+++ b/data/tribes/scripting/help/format_help.lua
@@ -5,6 +5,7 @@
 -- Functions used in the ingame help windows for formatting the text and pictures.
 
 include "scripting/richtext.lua"
+include "scripting/help.lua"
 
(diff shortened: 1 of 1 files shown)


tothxa:
The fix works, but it would be nice if we could make the links in work (open the selected item in the encyclopedia). I guess the easier approach might be to open the encyclopedia right for the building help if that is acceptable instead of the current standalone window.

frankystone:
Yes, that would be nice. Funnily the links inside the separate window do only work if the tribal encyclopedia help window is open.

I will try that, but i fear this is beyond my capabilities.

hessenfarmer:
enabling the links in the building help would be a bigger task I think. So I will merge this one and any fix for the bigger issue might be a new PR.  Maybe we should open an issue to not forget this

hessenfarmer (approved):

PR #5663, opened 2022-11-19
Fix handling of small immovables and add test case with closest allowed immovables

tothxa:
And here it is all as a PR. But if you have already changed `ship.cc` then it doesn't help much. :)


tothxa:
oops, I forgot that it goes to the project PR-s… So this is meant for #5647

hessenfarmer (approved):

PR #5672, opened 2022-11-25
Switch to C++17

Noordfrees:
We no longer support Xenial, so we should finally be able unlock those goodies like init statements in `if`, enhanced variadic templates, the `optional` and `any` types, and much more.


hessenfarmer (approved):
as the ci is green now this should be good to go.

PR #5673, opened 2022-11-27
5589 traingsite encyclopedia

### LINKED ISSUE #5589: Add wares consumed info for all levels to encyclopedia entries of training sites
Currently the encyclopedia entry of training sites just shows all wares that are consumed by any of the training levels, making it look like all wares are required for all levels.

Each level should be listed separately with the correct amounts, like all produced wares are listed for production sites.

hessenfarmer:
fixes #5589


src/scripting/lua_map.cc
```diff
@@ -3343,26 +3324,11 @@ TrainingSiteDescription
...
-   PROP_RO(LuaTrainingSiteDescription, min_attack),
-   PROP_RO(LuaTrainingSiteDescription, min_defense),
-   PROP_RO(LuaTrainingSiteDescription, min_evade),
-   PROP_RO(LuaTrainingSiteDescription, min_health),
-   PROP_RO(LuaTrainingSiteDescription, weapons_attack),
-   PROP_RO(LuaTrainingSiteDescription, weapons_defense),
-   PROP_RO(LuaTrainingSiteDescription, weapons_evade),
-   PROP_RO(LuaTrainingSiteDescription, weapons_health),
    {nullptr, nullptr, nullptr},
 };
```
Noordfrees: Please don't remove Lua properties or methods for API compatibility. You can mark them as "deprecated since 1.2", but if they can at all be kept around they should be.

src/logic/map_objects/tribes/production_program.h
```diff
@@ -561,6 +561,15 @@ struct ProductionProgram : public MapObjectProgram {
...
 	const Buildcost& recruited_workers() const;
+	const std::string& trained_attribute() const {
+		return trained_attribute_;
+	}
+	const uint8_t& train_from_level() const {
+		return train_from_level_;
+	}
+	const uint8_t& train_to_level() const {
+		return train_to_level_;
+	}
```
Noordfrees:
```suggestion
	uint8_t train_from_level() const {
		return train_from_level_;
	}
	uint8_t train_to_level() const {
		return train_to_level_;
	}
```
For primitive types, returning a copy is preferred over returning a const reference

src/scripting/lua_map.cc
hessenfarmer: I have checked carefully whether they are still in use in our lua code. (i did a fullsearch over the data directory.  It would be possible to keep min and max values as they are still availablein c++ code of trainingsite. keeping the food and weapons values would mean either keeeping the unnecessary definitions in the lua files from where they were parsed in the trainingsite code or try to rebuild them by summarising elements of the program code, Both is not desirable imho.

src/logic/map_objects/tribes/production_program.h
hessenfarmer: did not know about such finesses. However a String is not deemed a primitive in this case correct?

data/tribes/scripting/help/building_help.lua
```diff
@@ -889,41 +821,85 @@ end
...
-         end
+   local checked_programs = {}
+   if (building_description.type_name == "trainingsite") then
+      programs, ware_counters, ware_strings = programs_wares_count(tribe, building_description, nil)
+      if (not checked_programs[program]) then
          for j, program in ipairs(programs) do
-            if (ware_counters[program] > 0) and not checked_programs[program] then
+            trained = building_description:trained_soldiers(program)
+            -- TRANSLATORS: Building Encyclopedia Skill trained by a trainingsite
+            result = result .. h3(_("Trained skill:"))
```
tothxa: How about:

```suggestion
            result = result .. h3(_("Training step:"))
```

tothxa (commented):
Thank you very much! :) Tested and works nicely. The h3→h4 change was much needed. :)

One minor graphical issue: The level indicator images are not aligned well with the arrows:
shot0000
Though I accept that this may not be easily solved.

edit: I only noticed now, looking at the image: I think "`n` to `n+1`" would be better than "up to".

data/tribes/scripting/help/building_help.lua
hessenfarmer: for me the first focus is on the property or skill. only afterwards on the steps. therefore I'd rather stay with skill here.

hessenfarmer:
>

hessenfarmer:
> edit: I only noticed now, looking at the image: I think "`n` to `n+1`" would be better than "up to".

ok will fix the wording. However I need to mention that with my changes it is now possible to define training programs which do an upgrade of n steps at once.

hessenfarmer:
the misalignment of arrow and images is due to different sizes of the images. But I couldn't find a solution how to upscale the soldier level images and downsizing the arrow is not looking good.

tothxa:
> the misalignment of arrow and images is due to different sizes of the images. But I couldn't find a solution how to upscale the soldier level images and downsizing the arrow is not looking good.

Oh, I thought the arrow was an arrow character, but now I see it's an image too. Anyhow, as these are inline images in the richtext, I guess it's not possible to really change the alignment.

But I also see there are already several special `dependencies_…` functions. So if you think it's worth creating another one for this, here's the arrow cropped to the same height as the level indicators (zipped to make sure it won't get mangled):
[arrow-right-trnlvl.zip](https://github.com/widelands/widelands/files/10108458/arrow-right-trnlvl.zip)

Or leave it as it is for now, because eventually the level indicator code should be changed to allow mipmapping too, and then the indicators should get higher scale images.

data/tribes/scripting/help/building_help.lua
tothxa: I see. But "Trained skill" doesn't feel natural English to me (even though the direct translation is perfectly fine in hungarian), but of course I'm not a native speaker either.

Maybe @andy5995 or @RHolton could help us here.

andy5995: I'm not sure yet what's intended. It looks like you might want the equivalent of "Promotion" or "Promoted" level. That term is used in @MegaGlest and @0ad when a unit stays the same (doesn't morph), but receives an armor or hp increase.

hessenfarmer: @andy5995 indeed the headline should denote that a soldier gets a promotion / a rise in level, but I wanted to emphasize on the exact capapbility of our 4 military capabilities that gets promoted. I do not know a better word then skill for it. 
As @tothxa is unsure whether "Trained skill" is the proper way to say this in english maybe it is just a matter of sequence. Would "Skill trained:" sound better?

src/scripting/lua_map.cc
hessenfarmer: I reinstated the min_max values is this ok for @Noordfrees

hessenfarmer:
found a logical error introduced by pushing the internal max value to the real max value 
please review carefully my changes in https://github.com/widelands/widelands/pull/5673/commits/0a8b592fbdc36d1c8b42c56c5c2b9b9dfbe01f9b

hessenfarmer:
> Or leave it as it is for now, because eventually the level indicator code should be changed to allow mipmapping too, and then the indicators should get higher scale images.   
  
tested it with the cropped arrow, but found it even harder to see the small icons as they are dominated by the arrow in this case. having bigger icons would be much preferable, so I will not change anything yet but keep it in mind to be changed in the future

data/tribes/scripting/help/building_help.lua
andy5995: @hessenfarmer "Skill trained" doesn't sound better to me. how about "Skill-step" or "skill-boost" or "skill-booster"? I'll give it some more thought...

hessenfarmer: well this is the help of the building, therefore I wanted to have a short form for "this building trains soldier in [skill]". With my suggestion I just took the other short headlinbes into account which are "Wares produced:" and "Wares consumed:". I will have a look whether and how it would be feasible to include the Skill already in the Headline, resulting in "Health Improvement", " Attack Training" "Defense Training" and "Evade Training:".    
could this be a solution?

data/tribes/scripting/help/building_help.lua
```diff
@@ -889,41 +730,88 @@ end
...
-         end
+   local checked_programs = {}
+   if (building_description.type_name == "trainingsite") then
+      programs, ware_counters, ware_strings = programs_wares_count(tribe, building_description, nil)
+      if (not checked_programs[program]) then
          for j, program in ipairs(programs) do
-            if (ware_counters[program] > 0) and not checked_programs[program] then
+            trained = building_description:trained_soldiers(program)
+            -- TRANSLATORS: Building Encyclopedia Skill trained by a trainingsite
+            if (trained[1] == "Health") then
```
Noordfrees: This is not a translatable string

hessenfarmer: correct. It is just a string value written from c++ to lua which is not intended to be used in a translatable string. I could have used an arbitrary enum but I thought a string would be more human readable in the code.  
Edit: now I got the comment which is about the translators comment will remove that

Noordfrees (approved):
LGTM :)

PR #5674, opened 2022-11-27
Fix cursor flashing

### LINKED ISSUE #5669: Editbox Cursor Blinks Too Much
- When an editbox loses focus and does not accept keystrokes, its caret should stop blinking. 
- While moving the caret with the arrow keys, the caret should not blink either since the temporary invisibility makes it harder to position it accurately.

jmoerschbach:
fixes #5669

The threshold of 200ms is arbitrary but feels good to me.

I did not look too much into the focus details, just using the same mechanism as the rectangle/border drawing uses
(The cursor blinking while not in focus is not a problem of all editboxes, I think it is related to where it is put into individual layouts)


src/ui_basic/editbox.h
```diff
@@ -111,6 +111,10 @@ struct EditBox : public Panel {
 
 	ScopedTimer caret_timer_;
 	uint32_t caret_ms;
+	ScopedTimer cursor_movement_timer;
+	uint32_t cursor_ms;
+	bool cursor_movement_active = false;
```
Noordfrees: Ideally, all private variables should end with a trailing underscore. But since this is already mixed I'm leaving it up to you if you want to refactor this here or not.

Noordfrees (approved):
Tested and working, code LGTM :)

src/ui_basic/editbox.h
jmoerschbach: done. Could that be a codecheck? :)

Issue #5641, opened 2022-11-02
More Compiler Warnings


Noordfrees:
Some new compiler warnings in current master 1549bed6a021b2a96ba40b7ae4f5decb3b7ffad1 with g++ 12.2.0, release build:
```c++
[281/737] Building CXX object src/economy/CMakeFiles/economy.dir/economy.cc.o
In file included from .../src/economy/supply.h:22,
                 from .../src/economy/economy.h:26,
                 from .../src/economy/economy.cc:19:
In member function ‘void Trackable::Tracker::deref()’,
    inlined from ‘BaseTrackPtr::~BaseTrackPtr()’ at .../src/economy/trackptr.h:108:19,
    inlined from ‘TrackPtr<Widelands::Supply>::~TrackPtr()’ at .../src/economy/trackptr.h:162:27,
    inlined from ‘Widelands::RequestSupplyPair::~RequestSupplyPair()’ at .../src/economy/economy.cc:766:8,
    inlined from ‘void std::push_heap(_RAIter, _RAIter, _Compare) [with _RAIter = __gnu_cxx::__normal_iterator<Widelands::RequestSupplyPair*, vector<Widelands::RequestSupplyPair> >; _Compare = Widelands::RequestSupplyPair::Compare]’ at /usr/include/c++/12/bits/stl_heap.h:218:5,
    inlined from ‘void std::priority_queue<_Tp, _Sequence, _Compare>::push(const value_type&) [with _Tp = Widelands::RequestSupplyPair; _Sequence = std::vector<Widelands::RequestSupplyPair>; _Compare = Widelands::RequestSupplyPair::Compare]’ at /usr/include/c++/12/bits/stl_queue.h:741:16,
    inlined from ‘void Widelands::Economy::process_requests(Widelands::Game&, Widelands::RSPairStruct*)’ at .../src/economy/economy.cc:845:27:
.../src/economy/trackptr.h:58:27: warning: pointer may be used after ‘void operator delete(void*)’ [-Wuse-after-free]
   58 |                         --refcount_;
      |                           ^~~~~~~~~
In member function ‘void Trackable::Tracker::deref()’,
    inlined from ‘void Trackable::Tracker::deref()’ at .../src/economy/trackptr.h:56:8,
    inlined from ‘BaseTrackPtr::~BaseTrackPtr()’ at .../src/economy/trackptr.h:108:19,
[...]

PR #5679, opened 2022-12-04
Fix clang-tidy C++17 checks

Noordfrees:
- `modernize-use-nodiscard`: Functions that only perform a calculation and return the result without any side-effects should be annotated `[[nodiscard]]`. This annotation causes compilers to emit a warning when you invoke the function without using its result.
- `readability-redundant-declaration`: Sort of what the includes script already checks, but more sophisticated so it catches more occurrences. (The includes script also does something else which clang-tidy can't.)
- `modernize-concat-nested-namespaces`: More concise and readable declarations of nested namespaces.
- Re #5641

### DIFF
--- a/src/ai/defaultai.cc
+++ b/src/ai/defaultai.cc
@@ -84,15 +84,6 @@ namespace AI {
 
 // Fix undefined references
 
-
-
-
-
-
-
-
-
-
 DefaultAI::NormalImpl DefaultAI::normal_impl;
 DefaultAI::WeakImpl DefaultAI::weak_impl;
 DefaultAI::VeryWeakImpl DefaultAI::very_weak_impl;
--- a/src/ai/defaultai.cc
+++ b/src/ai/defaultai.cc
@@ -83,15 +83,15 @@ constexpr int kMineTypes = 4;
 namespace AI {
 
 // Fix undefined references
-constexpr Duration DefaultAI::kManagementUpdateInterval;
-constexpr Duration DefaultAI::kStatUpdateInterval;
-constexpr Duration DefaultAI::kFlagWarehouseUpdInterval;
-constexpr Duration DefaultAI::kExpeditionMinDuration;
-constexpr Duration DefaultAI::kExpeditionMaxDuration;
-constexpr Duration DefaultAI::kShipCheckInterval;
-constexpr Duration DefaultAI::kCampaignDuration;
-constexpr Duration DefaultAI::kTrainingSitesCheckInterval;
-constexpr Duration DefaultAI::kDiplomacyInterval;
+
+
+
+
+
+
+
+
+
 
 DefaultAI::NormalImpl DefaultAI::normal_impl;
 DefaultAI::WeakImpl DefaultAI::weak_impl;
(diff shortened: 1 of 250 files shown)


tothxa:
> * `modernize-use-nodiscard`: Functions that only perform a calculation and return the result without any side-effects should be annotated `[[nodiscard]]`. This annotation causes compilers to emit a warning when you invoke the function without using its result.

This is quite ugly. And if clang-tidy can find them, compilers should also be able to do it without annotations.

hessenfarmer:
> > * `modernize-use-nodiscard`: Functions that only perform a calculation and return the result without any side-effects should be annotated `[[nodiscard]]`. This annotation causes compilers to emit a warning when you invoke the function without using its result.
> 
> This is quite ugly. And if clang-tidy can find them, compilers should also be able to do it without annotations.

I agree it is ugly on the first view. However I always needed some time to accomodate with the new syntaxes enforced by clang_tidy to see their benefit. In this case this is something like a special comment for me that reminds new coders or not people with limited knowledge like myself to always use the result of such a function call, which is not that bad imho.

src/ai/defaultai.cc
```diff
@@ -83,15 +83,6 @@ constexpr int kMineTypes = 4;
 namespace AI {
 
 // Fix undefined references
```
hessenfarmer, line 85: I am not sure whether the solution for these values would be so easy. I think solving this issue would need more thoughts as the comment implies that without these declarations the compiler would complaion about missing references. In the header file these values are declared as private members of the defaultAI struct, but maybe they are not always used only in a private context.

tothxa:
> In this case this is something like a special comment for me that reminds new coders or not people with limited knowledge like myself to always use the result of such a function call, which is not that bad imho.

The problem is that we don't check header files with clang-tidy, so these annotations only go to the `.cc` files, while they would be much better in the headers for this use case.

Noordfrees:
Turns out we can fix headers too ;) Something that should be done for all other checks as well in a follow-up branch.

src/ai/defaultai.cc
Noordfrees: The compiler warning here says
```
warning: out-of-line definition of constexpr static data member is redundant in C++17 and is deprecated [-Wdeprecated]
```
C++14 and older needs this, C++17 is smarter and makes this superfluous. And apparently it compiles and links fine without them.

tothxa:
> Turns out we can fix headers too ;) Something that should be done for all other checks as well in a follow-up branch.

[I'd told you before](https://github.com/widelands/widelands/discussions/5262#discussioncomment-2212007). :)

Noordfrees:
> > Turns out we can fix headers too ;) Something that should be done for all other checks as well in a follow-up branch.
> 
> [I'd told you before](https://github.com/widelands/widelands/discussions/5262#discussioncomment-2212007). :)

I must have missed that :/ Will implement these suggestions in the next branch

src/ai/defaultai.cc
hessenfarmer: ok. did not know that. I just ran into issues with these definitions a lot of times, so I was alerted. So I am glad that c++17 is better in this.

hessenfarmer (approved):
well, I checked the changes and they seem to be good. and it builds fine.    
So ok from my side.

PR #5682, opened 2022-12-06
Update clang-tidy CI version

Noordfrees:


.github/scripts/install_deps.sh
```diff
@@ -6,9 +6,8 @@ set -ev
 ADD_PKG_LIST=""
 
 if [ "${GITHUB_JOB}" = "clang_tidy" ]; then
-  ADD_PKG_LIST="
-    clang-tidy  \
-    python-yaml"
+  pip install pyyaml
```
hessenfarmer, line 9: if you look at line 12 of the log of the dependencies install task we might not need this.

hessenfarmer (approved):
as clang tidy is green again this should be good to go. one inline comment though

.github/scripts/install_deps.sh
Noordfrees: No harm leaving it in, currently it's a no-op and if pyyaml is no longer preinstalled in a future version of the build image we won't have to update the script again.

PR #5705, opened 2022-12-11
Add shortcut keys for all menu items

tothxa:
* add shortcut to statistics button tooltip in interactive_spectator.cc
 * unify Load and Save hotkeys
 * change some editor tool hotkeys and make all tools use Shift modifier
 * editor new map window: create new map when terrain is doubleclicked
 * editor: remove Alt modifier for third tool function (old TODO item)
 * editor: ask for exit confirmation even if map is not modified (still skip on Ctrl+click)
 * ask for exit confirmation in main menu (can be skipped by Ctrl+clicking)


src/editor/editorinteractive.cc
```diff
@@ -584,17 +597,17 @@ void EditorInteractive::think() {
...
-			UI::WLMessageBox mmb(this, UI::WindowStyle::kWui, _("Unsaved Map"),
-			                     _("The map has not been saved, do you really want to quit?"),
-			                     UI::WLMessageBox::MBoxType::kOkCancel);
-			if (mmb.run<UI::Panel::Returncodes>() == UI::Panel::Returncodes::kBack) {
-				return;
-			}
+void EditorInteractive::exit(const bool force) {
+	if (force) {
+		end_modal<UI::Panel::Returncodes>(UI::Panel::Returncodes::kBack);
+	} else {
```
Noordfrees:
```suggestion
	if (!force) {
```

src/wlapplication_options.cc
```diff
@@ -359,6 +359,21 @@ static std::map<KeyboardShortcut, KeyboardShortcutInfo> shortcuts_ = {
                          keysym(SDLK_F11),
                          "screenshot",
                          []() { return _("Take Screenshot"); })},
+   {KeyboardShortcut::kCommonSave,
+    KeyboardShortcutInfo({KeyboardShortcutInfo::Scope::kGame, KeyboardShortcutInfo::Scope::kEditor},
```
Noordfrees, line 363: These all also need Game scope

src/wlapplication_options.cc
```diff
@@ -419,10 +431,51 @@ static std::map<KeyboardShortcut, KeyboardShortcutInfo> shortcuts_ = {
...
                                                          "editor_tools",
                                                          []() { return _("Tools"); })},
+   {KeyboardShortcut::kEditorChangeHeight,
+    KeyboardShortcutInfo({KeyboardShortcutInfo::Scope::kEditor},
+                         keysym(SDLK_h),
+                         "editor_change_height",
+                         []() { return _("Change Height Tool"); })},
+   {KeyboardShortcut::kEditorRandomHeight,
+    KeyboardShortcutInfo({KeyboardShortcutInfo::Scope::kEditor},
+                         keysym(SDLK_b),
```
Noordfrees: I think having some tools with a modifier and some without is inconsistent. Maybe make all tools use Shift.

src/editor/editorinteractive.cc
tothxa: I just copied the old KMOD_CTRL branch, though I wondered why this is called twice. If you say it's OK to remove it, I'll do it.

src/wlapplication_options.cc
tothxa: I didn't want to change existing ones, but I agree consistency is much preferred, and I practically already selected shift for this purpose. But:

 - Can History become plain 'H', just like the Tools menu is plain "T"?
 - How to resolve Players vs. Port spaces? Can Players remain plain P or make Port spaces shift+E like E for seafaring stats?
 - Can Info remain plain 'I'?

Noordfrees: +1 for <kbd>H</kbd> = History

I think all tools should now use <kbd>Shift</kbd>, so <kbd>Shift</kbd>+<kbd>I</kbd> for Info and <kbd>Shift</kbd>+<kbd>P</kbd> for players. 
<kbd>X</kbd> is soldier stats, seafaring stats are <kbd>E</kbd>. So if <kbd>E</kbd> is available use that for port spaces, otherwise well it doesn't matter that much since it's not 1:1 identical to any other hotkey.

tothxa: OK, thanks. Will change it accordingly.

(Sorry, I forgot to verify Seafaring hotkey before sending the comment, but I also corrected it in the meantime)

tothxa:
I wonder whether I should add exit confirmation to exiting Widelands completely too. Possibly only for `Escape` which is too easy to press one too many times, but probably doesn't hurt for the Exit Widelands button either, with the usual Ctrl override.

Noordfrees (approved):
LGTM :) 

> I wonder whether I should add exit confirmation to exiting Widelands completely too. Possibly only for Escape which is too easy to press one too many times, but probably doesn't hurt for the Exit Widelands button either, with the usual Ctrl override.

Sounds good to me, but only for <kbd>Esc</kbd>, not for <kbd>Alt</kbd><kbd>F4</kbd> and similar

Noordfrees (approved):

Noordfrees (changes requested):
Sorry one more nit:
```
[00:00:00.124 real] WARNING: The default shortcuts for game_sound_options and game_sfstats_open_goto (Strg+O, Strg+O) collide
[00:00:00.190 real] WARNING: The default shortcuts for game_sfstats_open_goto and game_sound_options (Strg+O, Strg+O) collide
```

tothxa:
> ```
> [00:00:00.124 real] WARNING: The default shortcuts for game_sound_options and game_sfstats_open_goto (Strg+O, Strg+O) collide
> [00:00:00.190 real] WARNING: The default shortcuts for game_sfstats_open_goto and game_sound_options (Strg+O, Strg+O) collide
> ```

Oops.

That's an unadvertised feature of the ship stats window (combination of the 2 buttons with hotkeys `j` and `o`). May I change it to Shift to keep the main menu consistent?

edit: and add it to the tooltips of the buttons :) also I'd enable the "go to ship" hotkey in the ship window too

Noordfrees:
The tooltip now states two actions for a single button, leaving it unclear which of the two actions will be triggered by clicking and how to trigger the other action. This needs some better unambiguous phrasing.

tothxa:
I didn't change the ship window after all, because it has other buttons too, and other similar windows (e.g. building windows) don't have keyboard shortcuts either.

tothxa:
> The tooltip now states two actions for a single button, leaving it unclear which of the two actions will be triggered by clicking and how to trigger the other action. This needs some better unambiguous phrasing.

Indeed. Though as you can see, this was the original intention too.

OTOH, do we really need this hotkey after all?

Noordfrees:
```
[00:00:41.713 real] ERROR: Error: [.../src/base/format/tree.h:300] Too few arguments provided: 4 required but only 3 passed
[00:00:41.714 real] ERROR: String: <rt><p>%s %s<br>%s %s</p></rt>
```

Noordfrees (approved):

tothxa:
@Noordfrees Can richtext font tags be nested? Maybe the second hotkey would look better with the hotkey font style, but with the translatable format string it would be hard to embed the font tags otherwise.

PR #5717, opened 2022-12-20
Update keyboard controls help in the game and the editor and generate it semi-automatically

### LINKED ISSUE #5716: Unable to query shortcut for editor_menu
For some reason this error got not exposed to the terminal, so just a screen shot:

missing_shortcut_for_Editor_menu

Looks like an Editor specific Problem, in a game all works fine.

**To Reproduce**
1. Open editor
2. Click on the help button
3. Click on Controls

**Expected behavior**
At least no error :-D

**Version:**
 - OS: archlinux
 - Widelands Version: git25986 (43599b4@master) Debug

**Additional context**
Not specific to this Problem but i remember such errors were formerly exposed to the terminal. Using a debug build with command line option `--verbose` just shows nothing related in the terminal.

tothxa:
**Type of change**

Bugfix / Refactoring

**Issues closed**

fixes #5716

Last fastplace slot can actually be used, not only defined.

**New behaviour**

Keyboard shortcuts help texts in the Encyclopedia and the in-editor help are generated from the same definition that is used for the keyboard options window. This should help make sure that new hotkeys will not be left out in the future.

Fastplace shortcuts for the current player/tribe are now also included in the Encyclopedia.

The hotkeys for debug features (debug console and cheat mode) were made configurable, and a hotkey was added for upload map in the editor.

**Possible regressions**

Anything related to keyboard shortcuts, because the enum class was reordered heavily.


tothxa:
This is a draft because although it already fixes the bug itself by updating the existing hotkeys in the help, it does not add the new ones yet. And I believe there's room for more cleanup too.

data/scripting/editor/editor_controls.lua
```diff
@@ -14,9 +14,7 @@ local r = {
       p(
          -- TRANSLATORS: This is an access key combination. Localize, but do not change the key.
```
tothxa: May I remove "Localize, but do not change the key." from all of these? The key hasn't been part of the translatable text for a long time now…

data/scripting/editor/editor_controls.lua
```diff
@@ -14,9 +14,7 @@ local r = {
...
          dl(help_format_hotkey(wl.ui.get_shortcut("encyclopedia")), _("Help")) ..
-         -- TRANSLATORS: This is an access key combination.
-         dl(help_format_hotkey(wl.ui.get_shortcut("editor_menu")), _("Toggle main menu")) ..
-         -- TRANSLATORS: This is an access key combination. The hotkey is 't'
+         -- TRANSLATORS: This is an access key combination. The default hotkey is 't'
```
tothxa: This too. I don't think the keyboard combination is particularly important.

Actually the "access key combination" part should be changed too, maybe to "This is the command for a keyboard shortcut."

frankystone:
The Key combination for opening the debug console is missing in the editor: SHIFT+CTRL+SPACE

tothxa:
I've added the help for the debug console.

While editing the common helptexts file, I've noticed there was a text for the old 2 part replay format, so I updated that too to reflect the new single file format.

tothxa:
This is done functionally, but some descriptions may still need changing to make them more suitable for the help too.

tothxa:
@Noordfrees May I convert `KeyboardShortcutInfo` to store the English descriptions as `std::string`, instead of the current `std::function`? These would then be initialised using `gettext_noop()`, and translated in `to_string()`.

Either here or in a separate PR, because this one already looks much bigger than it actually is.

tothxa:
@frankystone Could you please test this again?

frankystone:
Some brief testing and observations, but i am a mouse jockey and those shortcuts is a closed book for me :smile:  :

Some shortcut entries in the Editors encyclopedia are only valid for a game. E.g. `,`, `.` or `STRG+SHIFT+BACKSPACE` (cheating). In a game help i found `Entf` (Delete). I think those shortcuts shouldn't be mixed. The editor help should only contain valid shortcuts for the editor and the game help should show only valid shortcuts for a game

Otherwise this works as expected for me.

tothxa:
Thank you for testing.

> Some shortcut entries in the Editors encyclopedia are only valid for a game. E.g. `,`, `.` or `STRG+SHIFT+BACKSPACE` (cheating). In a game help i found `Entf` (Delete). I think those shortcuts shouldn't be mixed. The editor help should only contain valid shortcuts for the editor and the game help should show only valid shortcuts for a game

The new  help generator code uses the same groups as the shortcut settings in the Options menu, so yes, only valid ones should be listed. And actually most of your examples do work in both:

 - `,` and `.` work in the editor: if you jump around by clicking on the minimap, you can then backtrack with `,` and then forward again with `.`.
 - I was surprised you didn't expect the `Delete` key in the in-game menu, because it's used by the messages window. So I searched for it, and it turns out that the actual configurable hotkey is only referenced by that and the save/load game windows, but nothing in the editor code. The `Delete` key itself is hardcoded though in some basic UI elements for text editing.

As for cheating mode, that's a newly configurable hotkey in this PR. I included it in common, because it's used in common code (`interactive_base.cc`) right next to the debug console hotkey, so even if it has no real effect in the editor, it could mask other editor hotkeys if allowed to be set to the same combination.

src/wui/interactive_player.h
```diff
@@ -88,6 +88,10 @@ class InteractivePlayer : public InteractiveGameBase {
 
 	void edit_pinned_note(const Widelands::FCoords& c);
 
+	/** Get the richtext formatted help of the current fastplace shortcuts for the player
+	    including a header. */
+	std::string get_fastplace_help();
```
Noordfrees: Can this be `const`?

src/scripting/lua_ui.cc
```diff
@@ -1056,9 +1056,55 @@ static int L_get_shortcut(lua_State* L) {
 	return 1;
 }
 
+/* RST
+.. method:: get_ingame_shortcut_help()
+
```
Noordfrees, line 1061: These all need a `.. versionadded: 1.2` directive

src/wlapplication_options.cc
```diff
@@ -121,6 +121,20 @@ void set_config_string(const std::string& section,
...
+KeyboardShortcut operator+(const KeyboardShortcut& id, const int i) {
+	return static_cast<KeyboardShortcut>(static_cast<uint16_t>(id) + i);
+}
+
+KeyboardShortcut& operator++(KeyboardShortcut& id) {
+	id = id + 1;
+	return id;
+}
+
+uint16_t operator-(const KeyboardShortcut& id, const KeyboardShortcut& base) {
```
Noordfrees: Operands to + and - operators should be called `lhs` and `rhs` or `a` and `b`

src/wlapplication_options.cc
```diff
@@ -134,17 +148,30 @@ struct KeyboardShortcutInfo {
...
 	   : scopes(s), default_shortcut(sym), current_shortcut(sym), internal_name(n), descname(f) {
 	}
 };
 
+// Aliases are used for help generation to simplify inserting common shortcuts
+struct KeyboardShortcutAlias {
+	const KeyboardShortcut real_shortcut;
+	const std::string descname_override;  // Use gettext_noop() for the init value!
+
+	KeyboardShortcutAlias(const KeyboardShortcut real, const std::string& desc_override = "")
```
Noordfrees:
```suggestion
	KeyboardShortcutAlias(const KeyboardShortcut real, const std::string& desc_override = std::string())
```
just for style consistency :)

src/wlapplication_options.cc
```diff
@@ -1117,24 +1429,24 @@ std::string shortcut_string_for(const SDL_Keysym sym, const bool rt_escape) {
...
 			shortcuts_.emplace(k, kFastplaceDefaults[off]);
 		} else {
 			++counter;
 			shortcuts_.emplace(
 			   k, KeyboardShortcutInfo({KeyboardShortcutInfo::Scope::kGame}, keysym(SDLK_UNKNOWN),
 			                           format("%scustom_%i", kFastplaceGroupPrefix, counter),
-			                           [counter]() { return format(_("Fastplace #%i"), counter); }));
+			                           // descname is not actually used, fastplace_descr() uses a
+			                           // translatable format string instead
+			                           format("Fastplace #%i", counter)));
```
Noordfrees: I think this might actually be used in some cases, or maybe not anymore… Please just keep this translatable anyway, just in case.

src/wlapplication_options.cc
```diff
@@ -159,365 +186,373 @@ static inline SDL_Keysym keysym(const SDL_Keycode c, uint16_t mod = 0) {
...
                          keysym(SDLK_F11),
                          "screenshot",
-                         []() { return _("Take Screenshot"); })},
+                         gettext_noop("Take Screenshot"))},
+   {KeyboardShortcut::kCommonDebugConsole,
+    KeyboardShortcutInfo({KeyboardShortcutInfo::Scope::kGlobal},
+                         keysym(SDLK_SPACE, kDefaultCtrlModifier | KMOD_SHIFT),
+                         "debugconsole",
+                         gettext_noop("Open the Debug Console (only in debug builds)"))},
+   {KeyboardShortcut::kCommonCheatMode,
```
Noordfrees, line 395: Please do not expose the Cheat Mode and Debug Console hotkeys to the user. These only work in release builds by design, and we should not have a shortcut that is defunct for 99.9% of users.

Noordfrees (commented):
Looks cool, not tested yet, some comments inline

src/wlapplication_options.cc
frankystone: > These only work in release builds by design,

I thought they are working only in debug builds, not sure about the cheat mode though. If possible they should be shown in debug builds, but not in release builds, imho.

src/wlapplication_options.cc
tothxa: Do you have any idea how I can still simplify it (with macro or anything else), or do I need to spell them out one by one?

src/wlapplication_options.cc
tothxa: Well, having them here makes sure that no conflicting shortcut can be defined. (of course it doesn't hurt either to have them configurable) This is important, because these are evaluated very early in the `handle_key` chain.

If I make them `#ifndef NDEBUG`, then they will be dropped from the config if a release build is run with the same config dir, and again, conflicting shortcuts could be defined.

I could put them in a new region and only show them in the help in debug builds, but still have them in options.

Or, how about allowing the script console in release builds too with a command line option (e.g. `--script_debugging`)? (I could make it part of solving #5793) Why force content developers to use dead slow debug builds when they only want to fiddle with their scripts? :)

src/wlapplication_options.cc
tothxa: OK, I'll do it, but I still think my names better express what this is actually used for.

src/wlapplication_options.cc
tothxa: BTW I'm pretty sure it's safe. All things, even the struct `KeyboardShortcutInfo` and the map `shortcuts_` are local to this file, not even declared in the header file. The only way to access them from anywhere else is `to_string()`.

src/wui/interactive_player.h
tothxa: It compiled with it for me… :)

src/wlapplication_options.cc
frankystone: For some reason i missed your post...

I didn't thought about the configurable shortcuts and the possibility of conflicting shortcuts. This can also happen if the Cheat Mode and Debug Console isn't shown as a shortcut in the help if one use CTRL+Space for something different.

I see no problems if they are shown in the help, maybe i am just too trustful.

tothxa: @Noordfrees We had some more proof of demand for the script console in release builds recently: https://github.com/widelands/widelands/issues/5852#issue-1667142245:

> The only thing I want from the "Debug" build is the in-game script console. I do not want debugging symbols compiled into the code.

src/wlapplication_options.cc
tothxa: Would it be OK to use just "Fastplace", "Set Toolsize", "Set Landmark" and "Go To Landmark"?

src/wlapplication_options.cc
Noordfrees: Well, then how about making the scripting console available in `RelWithDebInfo` builds as well via a `-Ddefine`. All the benefits of release builds plus additional debugging utilities, of which the scripting console, debug window, and cheating mode are three. Pure release builds really shouldn't have those debugging-only features though IMO.

src/wlapplication_options.cc
tothxa: Or the format string to not pollute the PO file? But that would be somewhat misleading.

src/wlapplication_options.cc
tothxa: @Noordfrees: Would you accept for this PR:

Put the shortcuts for script console and cheat mode in a new region and only show them in the help in debug builds, but still have them configurable in options.

src/wlapplication_options.cc
tothxa: @Noordfrees

Noordfrees:
Hm, tested it again and works fine, but I'm still a bit unhappy about the Console and Cheat hotkeys. Please do hide them also in the options in release builds. Other than that LGTM.

tothxa:
> Hm, tested it again and works fine, but I'm still a bit unhappy about the Console and Cheat hotkeys. Please do hide them also in the options in release builds.

But then how can the user resolve a conflict with it? (ie. free ctrl+shift+space to be able to set it to something else)

tothxa:
@Noordfrees I hope this way is acceptable to you. But I'm doing it against my will only to get the fixes merged at last. :-/

(I forgot that `space` and `backspace` cannot actually be assigned, so testing this needs changing them first in a debug build)

Noordfrees (approved):
Thanks, looking good :)

PR #5720, opened 2022-12-21
Fix heap-use-after-free in `BuildingWindow::init`

Noordfrees:
The bug happens when starting or cancelling an expedition in a port, ASan builds only


hessenfarmer (approved):
LGTM

PR #5728, opened 2022-12-25
Forbid soldiers leaving buildings through the wall

Noordfrees:
May not fix all cases of the bug, but it fixes the one where soldiers who leave a building to attack may exit through the wall instead of through the main entrance.

### DIFF
--- a/src/logic/map_objects/tribes/soldier.cc
+++ b/src/logic/map_objects/tribes/soldier.cc
@@ -962,6 +962,11 @@ void Soldier::attack_update(Game& game, State& state) {
 		}
 	}
 
+	if (location != nullptr && get_position() == location->get_position()) {
+		// Still at home, need to go outside first
+		return start_task_leavebuilding(game, false);
+	}
+
 	if (battle_ != nullptr) {
 		return start_task_battle(game);
 	}


klaus-halfmann (commented):
See no harm in this code, no idea if I can reproduce the fix.

klaus-halfmann (changes requested):
@Noordfrees Please add some comments to Soldier::attack_update
I have no idea what is happening there. e.g.  what `state.ivar3` is used for.
(Or maybe you need to point me to some docs again)

Noordfrees:
This is documented in start_task_attack:

https://github.com/widelands/widelands/blob/efea91e857c0e3885608ff688a7443f0b076fec6/src/logic/map_objects/tribes/soldier.cc#L805-L821

---

Reproducible case: Load this savegame, then wait until the blue barrier is complete, then attack the red sentry east of you with 4 soldiers.
[test.wgf.zip](https://github.com/widelands/widelands/files/10365954/test.wgf.zip)
Buggy result in master: grafik

klaus-halfmann:
The savefile results in a stack-buffer-overflow

```
=================================================================
==68922==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x00016f0020ef at pc 0x0001057153f4 bp 0x00016f001dd0 sp 0x00016f001dc8
READ of size 1 at 0x00016f0020ef thread T0
    #0 0x1057153f0 in ZipFilesystem::file_exists(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) const zip_filesystem.cc:236
    #1 0x105718f4c in ZipFilesystem::load(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, unsigned long&) zip_filesystem.cc:398
    #2 0x10572c088 in FileRead::open(FileSystem&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) fileread.cc:31
```

Looks like this ugly minizip bug

klaus-halfmann (approved):
I un- and then re-compressed the file and can confirm the bug is fixed.
@Noordfrees Will it help to have a feature request to refactor this state machine?

Noordfrees:
Thanks for the reviews

> Will it help to have a feature request to refactor this state machine?

Please please please let's _not_ try that. All Bob tasks are handled in the same way, individual variable names or a linear sequence of specialized subtasks would be nearly impossible to implement, and if they were, I'm pretty sure they would be as poorly readable as the existing code ;) not to mention that this will certainly give us plenty of new bugs; bob acts are extremely fiddly and once they work it is best not to glance at them askew or they'll break.

PR #5740, opened 2023-01-08
Fix map pixel to triangle computation regression

### LINKED ISSUE #5738: Editor: Can't reach some fields when placing terrains
Some fields can't be reached with the mouse cursor anymore. In this screenshot the field below the mouse can't be reached:

cant_reach_field

**To reproduce**
1. Open Editor and create a new map, choose water as terrain
2. Try to place some other terrains

Some fields can't be set with the chosen terrain.

**Expected behavior**
All field s should be reachable

**Version:**
 - OS: archlinux
 - Widelands Version: 1.2~git26011 (3571958@master) Debug

**Additional context**
Some of the latest commits may lead tot this. But i can't say with which version exactly causes it.

Noordfrees:

**Type of change**
Bugfix

**Issues(s) closed**
Fixes #5738

**Additional context**
Previously, we found the node and triangle closest to the cursor by starting at the upper edge of the screen and searching row-by-row downwards. Now we start at the bottom of the screen moving upwards. So there was a mismatch in the rows used for the triangle computation after the node computation successfully stopped.

### DIFF
--- a/src/wui/mapviewpixelfunctions.cc
+++ b/src/wui/mapviewpixelfunctions.cc
@@ -160,6 +160,10 @@ MapviewPixelFunctions::calc_node_and_triangle(const Widelands::Map& map, uint32_
 	Widelands::TCoords<> result_triangle(Widelands::Coords::null(), Widelands::TriangleIndex::D);
 
 	//  Find out which of the 4 possible triangles (x, y) is in.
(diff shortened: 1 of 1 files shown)


frankystone (commented):
approve testing

hessenfarmer (approved):
LGTM

PR #5746, opened 2023-01-09
Scenario "Island hopping" fix

### LINKED ISSUE #5743: Scenario scripting in MP Scenario Islands Hopping is not working correctly
**Describe the bug**
The scripting seems to have problems with soldiers, as no soldiers are taken to the third island.      
There may be other problems as well, as the scenario is very old

hessenfarmer:
**Type of change**
Bugfix

**Issues(s) closed**
Fixes #5743 

**Testing instruction**
1. start a local lan game of "island hopping" with 2 instances
2. on the instance of player 1 build a warehouse.
3. by warehouse settings send all soldiers to the warehouse (remove from HQ setting) 
4. after all soldiers 35 moved set the warehouse setting sto normal again
5. by scripting console add add least 2 different types of soldiers to the HQ
6. by scripting console send "hop_to_next_island(game.players[1], 1) (don't mind tzhe desync just continue playing)
7a. in trunk the HQ on island 2 contains only the 35 soldiers sent to the warehouse
7b. in this branch the new HQ contains the 35 soldiers plus the added soldiers as intended. 

reason for the bug was that set_soldiers is overwriting previous values so we need to always preserve the settings of the new HQ before adding new soldiers.


data/maps/MP_Scenarios/Island_Hopping.wmf/scripting/hop_island.lua
```diff
@@ -14,9 +14,17 @@ function add_workers(hq, workers)
 end
 
 function add_soldiers(hq, soldiers)
-   local setpoints = {}
+   local setpoints = hq:get_soldiers("all")
    for sdescr, count in pairs(soldiers) do
-      setpoints[ {sdescr:match("(%d):(%d):(%d):(%d)")} ] = count
+      local desc = {sdescr:match("(%d):(%d):(%d):(%d)")}
```
tothxa: I tested this PR, but it doesn't fix the soldier problem for me. After some debugging, I've found that the problem is that `match()` returns strings, not numbers.

tothxa: Not very nice, but maybe the simplest solution?

```suggestion
      local desc = {}
      for i, v in ipairs({sdescr:match("(%d):(%d):(%d):(%d)")}) do
         desc[i] = tonumber(v)
      end
```

data/maps/MP_Scenarios/Island_Hopping.wmf/scripting/hop_island.lua
```diff
@@ -14,9 +14,17 @@ function add_workers(hq, workers)
...
+      local did_add = false
+      for key, value in pairs(setpoints) do
+         if key[1] == desc[1] and key[2] == desc[2] and key[3] == desc[3] and key[4] == desc[4] then
+            setpoints[desc] = value + count
```
tothxa: Also this:

```suggestion
            setpoints[key] = value + count
```

I think the problem here was that because `desc` and `key` are separate tables, they become different indexes.

tothxa:
As for the terrain changes, I think the arrows should be kept and the balance should be ensured in some other way. Actually, as the arrow tips are the HQ places, the terrain there doesn't even matter for the number of possible trees, because the affected nodes are blocked anyway.

(BTW the whole map suffers from the usual balancing problem of maps mirrored over the vertical axis, whereas the two players who expand along the ~NW-SE (ie. building flag) direction are at a disadvantage compared to the other two who expand along ~SW-NE. But that's unfixable without completely redoing the whole map.)

hessenfarmer:
> As for the terrain changes, I think the arrows should be kept and the balance should be ensured in some other way. Actually, as the arrow tips are the HQ places, the terrain there doesn't even matter for the number of possible trees, because the affected nodes are blocked anyway.

oops. I did not realize these are arrows until you mention it now. Will undo these changes (a small imbalance won't hurt that much)    

> 
> (BTW the whole map suffers from the usual balancing problem of maps mirrored over the vertical axis, whereas the two players who expand along the ~NW-SE (ie. building flag) direction are at a disadvantage compared to the other two who expand along ~SW-NE. But that's unfixable without completely redoing the whole map.)    
   
Agreed, but this is out of scope.

data/maps/MP_Scenarios/Island_Hopping.wmf/scripting/hop_island.lua
hessenfarmer: well, it worked on my windows machine. will do some retest however.   
what was your result. did you have any soldiers on the second island at all. What were the numbers in the warehouse and what were it in the HQ on the 1st island?

tothxa: If I have 25 in the old HQ and 9 in the warehouse (1 goes to the new castle) then the new HQ gets 25.
Savegames before and after: [hopping.zip](https://github.com/widelands/widelands/files/10497844/hopping.zip)

tothxa: With the scenario in current master, the new HQ only gets the soldiers from the warehouse (9).

tothxa: Though my testcase differs from yours in that all soldiers are at the same level (0).

data/maps/MP_Scenarios/Island_Hopping.wmf/scripting/hop_island.lua
hessenfarmer: yes that was the root cause

data/maps/MP_Scenarios/Island_Hopping.wmf/scripting/hop_island.lua
hessenfarmer: found a solution even more simple ;-)

hessenfarmer: found and fixed thanks for pointing me to use the correct key

tothxa: Indeed. :)

tothxa (approved):
Thank you.

There's one more thing I forgot yesterday: The message to the players should say that (1) all wares, workers and soldiers will be transferred (2) from the HQ and all warehouses as well.

hessenfarmer:
> Thank you.
> 
> There's one more thing I forgot yesterday: The message to the players should say that (1) all wares, workers and soldiers will be transferred (2) from the HQ and all warehouses as well.

done like suggested. I would be available for a testplay tommorow evening

hessenfarmer:
@tothxa could you please do a proofread of my changes. Feel free to merge if you are happy

tothxa (approved):
Text changes are OK. Thank you.

PR #5748, opened 2023-01-11
Make richtext.lua styleable and allow partial theme addons

tothxa:
**Type of change**
Refactoring

**Issues(s) closed**
Re forum thread https://www.widelands.org/forum/post/40083/

These changes will make `richtext.lua` styleable and allow theme addons to only change some selected styles. (but still only one theme can be active at a time)

Errors in the active addon theme no longer crash the game on startup. If the theme addon cannot be loaded, the default theme will be used.

**New behavior**
Formatting scenario messages and in game help with functions from `richtext.lua` like `h1()` etc. and various forms of `p()` don't use hardcoded style any more, and the same styles are available for C++ code as well. Style info is moved to the default template.

The behaviour of the `vspace` richtext tag is changed (fixed?) too to force a new line before the gap. The old behaviour did not render the vertical space if the tag was at the end of a paragraph.

All Lua scripts in `data` were converted to use styling.

Some formatting was fixed or improved as well when I noticed problems during the conversion. See the commit messages for details.

Theme addons no longer need to include definitions for all styles. Missing styles will fall back to the default theme.

Errors in the active addon theme no longer crash the game on startup. If the theme addon cannot be loaded, the default theme will be used.

**Possible regressions**
Different look of scenario messages and help texts, that may break some layouting.

There may be some places where the `vspace` tags weren't adjusted to the new behaviour, also breaking the layouting.

Reading of the default theme and theme addons had big changes, which may have introduced bugs.

**Additional context**
~I'd also like to change `FontStyleInfo` to make `face_`, `color_` and `size_` `std::optional`. At least `color_` is a direct problem, because the variants of `p()` are used in both the `WUI` and `FSmenu` contexts which have different text colours, so it would be best to not force a text colour, so that a common style could be used.~
*(solved in other way, so not important any more)*


src/graphic/styles/paragraph_style.cc
```diff
@@ -0,0 +1,48 @@
...
+}
+
+std::string ParagraphStyleInfo::open_paragraph(const std::string& attrib) const {
+	return format("<p %s>%s%s", attrib, as_vspace(space_before_), font_->as_font_open());
+}
+
+std::string ParagraphStyleInfo::close_paragraph() const {
+	const int gap2 = space_after_ / 2;
+	const int gap1 = space_after_ - gap2;
+	return format("%s</font>%s</p>", as_vspace(gap1), as_vspace(gap2));
```
tothxa: Is this a bug of the richtext code? The non-mandatory behaviour of the `vspace` tag is not documented, and I couldn't figure out what makes it work or not, and I can't see in the code either what makes it behave like it does. The old lua functions for `h1`…`h3` actually return
```
"<p><font …><vspace …>text<vspace gap=1><vspace gap=6></font><vspace gap=6></p>
```

If I put a single `vspace` after the text, then there is no gap at all, no matter if the `vspace` is before, after or between the closing `font` and `p` tags. If I put a small and a big `vspace` inside and outside of the `font` tags, then I think only the small one is used. So even this code results in slightly different layout than the original. I guess that's not much of a problem, but having to use 2 `vspace`s is.

tothxa:
Now I wonder again (as in the original post) whether we would be better off if we added style handling to the richtext code itself. Though that would also have hard problems of its own.

tothxa:
> A fundamental difficulty is that styles are indexed by enums, so they can't be exposed easily to lua like suggested in https://www.widelands.org/forum/post/40088/

To solve this, I'd change the primary key in `style_manager.cc` to be the string table key, and add another map to match the enums to the string keys, then make the lookup functions accept either the string or the enum keys. Is that OK?

tothxa:
> To solve this, I'd change the primary key in `style_manager.cc` to be the string table key, and add another map to match the enums to the string keys, then make the lookup functions accept either the string or the enum keys.

I went the minimal change route, so the lookup map does string→enum, and is filled on the fly instead of being predefined.

edit: The reason is that there are more styleinfo types that will probably never be needed by lua than those that are touched by this PR.

data/templates/default/init.lua
```diff
@@ -780,5 +797,234 @@ return {
...
+         space_before = unknown_font_size,
+         space_after = unknown_font_size,
+      },
    },
+
+   colors = {
+      campaign_bar_thron = campaign_leader_color,
+      campaign_bar_boldreth = {0xFF, 0x00, 0x00},
+      campaign_bar_khantrukh = {0x80, 0x80, 0xFF},
+
```
frankystone, line 1013: Can you please use decimal values here? (and also below)

So each color definition has the same format and the values are not mixed up.

frankystone:
Does this change mean all theme addons have to be modified?

tothxa:
> Does this change mean all theme addons have to be modified?

I think we will need a solution to allow using old or partial theme addons, but currently it is not part of this PR. First I'd like to know whether the new functions and naming schemes are acceptable, or what changes they need.

data/templates/default/init.lua
tothxa: Do we need decimal, or can hexadecimal be the unified format? These are hexadecimal in the current campaign scripts (which is needed for richtext formatting), so reviewing should be easier this way.

frankystone: In my opinion the lua files are the part of the game where people without much knowledge of programming can work with. So they should be as easy as possible for them imho.

frankystone:
with version 1.2~git26070 (f50c24b@lua-text-style) Debug widelands does not start for me:

~~~~
[00:00:02.607 real] INFO: **** END SOUND REPORT ****
terminate called after throwing an instance of 'LuaTableKeyError'
  what():  [/home/kaputtnik/Quellcode/widelands-repo/widelands/src/scripting/lua_errors.cc:21] unknown is not a field in this table.
~~~~

data/templates/default/init.lua
tothxa: OK, I will convert all to decimal then.

tothxa:
> with version 1.2~git26070 ([f50c24b](https://github.com/widelands/widelands/commit/f50c24b18c99087301de3825f59ede47ab4b927b)@lua-text-style) Debug widelands does not start for me:
> 
> ```
> [00:00:02.607 real] INFO: **** END SOUND REPORT ****
> terminate called after throwing an instance of 'LuaTableKeyError'
>   what():  [/home/kaputtnik/Quellcode/widelands-repo/widelands/src/scripting/lua_errors.cc:21] unknown is not a field in this table.
> ```

I think this is because of your theme add-on missing the new style definitions. OK, then I'll definitely need to solve falling back to the default theme in this PR, because otherwise the users don't even get the chance to turn off add-ons after upgrading.

data/scripting/richtext.lua
```diff
@@ -187,10 +246,43 @@ end
...
+   else
+      _p_style_ = "ingame_text"
+   end
+end
+
+
+-- RST
+-- .. function:: fs_color(text)
+--
+--    *Deprecated*. Use `set_fs_style()` instead.
```
Noordfrees:
```suggestion
--    .. deprecated:: 1.2 Use `set_fs_style()` instead.
```

src/graphic/style_manager.cc
```diff
@@ -118,6 +118,21 @@ RGBAColor read_rgba_color(const LuaTable& table) {
 	return RGBAColor(rgbacolor[0], rgbacolor[1], rgbacolor[2], rgbacolor[3]);
 }
 
+// Read optional bool value from LuaTable
+bool read_bool(const LuaTable& table, const std::string& key) {
```
Noordfrees: These functions should take the default value as a third argument for readability

src/scripting/lua_styles.cc
```diff
@@ -0,0 +1,233 @@
...
+
+namespace LuaStyles {
+
+/* RST
+:mod:`styles`
+=============
+
+.. module:: styles
+   :synopsis: Provides helper functions for text formatting using themeable styles
+
```
Noordfrees, line 33: Add a `..versionadded 1.2` tag

src/scripting/lua_styles.cc
```diff
@@ -0,0 +1,233 @@
...
+	const std::string style(luaL_checkstring(L, 1));
+	const std::string attr(luaL_checkstring(L, 2));
+	const std::string text(luaL_checkstring(L, 3));
+	const std::string result = g_style_manager->paragraph_style(style).as_paragraph(text, attr);
+	lua_pushstring(L, result.c_str());
+	return 1;
+}
+
+/* RST
+.. function:: open_p()
```
Noordfrees: Missing argument

src/scripting/lua_styles.cc
```diff
@@ -0,0 +1,233 @@
...
+*/
+static int L_open_p(lua_State* L) {
+	const std::string style(luaL_checkstring(L, 1));
+	const std::string result = g_style_manager->paragraph_style(style).open_paragraph();
+	lua_pushstring(L, result.c_str());
+	return 1;
+}
+
+/* RST
+.. function:: open_p_with_attr(attributes)
```
Noordfrees: Missing argument

src/scripting/lua_styles.cc
```diff
@@ -0,0 +1,233 @@
...
+static int L_open_p_with_attr(lua_State* L) {
+	const std::string style(luaL_checkstring(L, 1));
+	const std::string attr(luaL_checkstring(L, 2));
+	const std::string result = g_style_manager->paragraph_style(style).open_paragraph(attr);
+	lua_pushstring(L, result.c_str());
+	return 1;
+}
+
+/* RST
+.. function:: close_p()
```
Noordfrees: Missing argument

src/scripting/lua_styles.cc
```diff
@@ -0,0 +1,233 @@
...
+*/
+static int L_color(lua_State* L) {
+	const std::string style_name(luaL_checkstring(L, 1));
+	const std::string result = g_style_manager->color(style_name).hex_value();
+	lua_pushstring(L, result.c_str());
+	return 1;
+}
+
+/* RST
+.. function:: dimension(style_name)
```
Noordfrees: Maybe consider a different name? To me "dimension" sounds like "extent", i.e. 2-dimensional width and height. Maybe it's just that I work too much with Java where Dimension is a much-used class that represents an Extent, but IMHO "gap" or "spacing" or "padding" or "size" would sound less confusing.

Noordfrees (commented):
LGTM, a few comments inline, not tested yet

src/scripting/lua_styles.cc
tothxa: Originally this was `size`, but I was afraid it may clash or could be confused with standard functions. Can't name it `gap`, `spacing` or `padding` because it's already used for window sizes as well, and may be used for other UI element sizes when we convert all C++ code.

OK, then how about `styled_size` which is already the name of the enum?

data/templates/default/init.lua
```diff
@@ -780,5 +797,234 @@ return {
...
+      campaign_fri_reebaud = {85, 187, 85},
+      campaign_fri_hauke = {119, 153, 187},
+      campaign_fri_maukor = {153, 0, 0},
+      campaign_fri_murilius = {187, 119, 85},
+      campaign_fri_claus = {28, 194, 0},
+      campaign_fri_henneke = {247, 255, 0},
+      campaign_fri_iniucundus = {255, 0, 89},
+      campaign_fri_angadthur = {0, 22, 255},
+      campaign_fri_amazon = {108, 0, 151},
+      campaign_fri_ketelsen = {68, 85, 204},
```
frankystone, line 1037: I am not sure if those campaign specific colors should be in this file. From my understanding this file is used for general UI styles, not for campaign related things. E.g. Custom themes are based on this file.

src/graphic/style_manager.cc
tothxa: I started to change them, but now I'm not sure it's an improvement. They're just shorthand convenience functions, so adding another option that will always have the same (quite obvious) value is contrary to the original purpose. These are only used by the next 2 functions, they're not even prototyped in the header.

tothxa: … unless you mean adding them to `lua_table.h`.

data/templates/default/init.lua
tothxa: There are several reasons why I think they belong here:

1. there is no other place where the scenarios of a campaign can store common values
2. they may need changing if someone wants to create a theme with a very different colour scheme, especially if it is aimed at visually impaired users
3. these campaigns are official game content

Actually for similar reasons (except for 3. of course, or even that if an addon aims to become official content), it could be useful to come up with a way to allow addons to add their own style elements which theme addons can modify as well in the long run.

tothxa:
@Noordfrees Please check last commit (https://github.com/widelands/widelands/pull/5748/commits/a55a7f46a19ef9217aab798e67b3f8bc0215ad48) whether this is OK to allow old and partial theme addons to work.

src/graphic/style_manager.cc
```diff
@@ -651,8 +660,19 @@ void StyleManager::add_style(UI::PanelStyle style, const LuaTable& table, PanelS
...
+	} else {
+		if (default_style == nullptr) {
+			throw wexception("The default template does not have font style '%s'", table_key.c_str());
+		}
+		assert(g_template_dir != kDefaultTemplate);
+
+		// We can fall back on the default template
+		fontstyle = new UI::FontStyleInfo(default_style->font_style(font_key));
+	}
+	fontstyles_.emplace(std::make_pair(font_key, std::unique_ptr<UI::FontStyleInfo>(fontstyle)));
```
tothxa, line 950: or turn them into `std::shared_ptr`s?

frankystone:
With version 1.2~git26082 (f77d90d@lua-text-style) Debug it does not work yet for me:

~~~~
[00:00:02.615 real] INFO: **** END SOUND REPORT ****
terminate called after throwing an instance of 'LuaTableKeyError'
  what():  [/home/kaputtnik/Quellcode/widelands-repo/widelands/src/scripting/lua_errors.cc:21] paragraphs is not a field in this table.
~~~~

tothxa:
> With version 1.2~git26082 ([f77d90d](https://github.com/widelands/widelands/commit/f77d90d9249fd779946c2e765d3d30b2638495e2)@lua-text-style) Debug it does not work yet for me:

I'm still working on the fallback code, the previous commit was just the first partial prototype. Actually I'm compiling the first attempt at allowing entire sections (like `paragraphs`) to be missing from theme add-ons right now. :)

tothxa:
@frankystone Old theme addons should now work at last. :)

Though there's still some more styles to convert for partial theme addons.

@Noordfrees Again, I'm not sure my ugly hack helper functions are acceptable.

tothxa:
Now this should be complete as far as Lua scripts are concerned.

data/maps/SP_Scenarios/The_Green_Plateau.wmf/scripting/texts.lua
```diff
@@ -48,21 +48,25 @@ obj_defeat_jomo = {
...
 
+mb_w = messagebox_w_step(-1)
+mb_h = messagebox_h_step(-2)
+mb_y = styles.get_size("map_plateau_message_pos_y")
+
 briefing_1_the_forbidden_island = {
    title = _("First Briefing"),
-   width = 400,
-   height = 300,
-   posy = 1,
```
tothxa, line 55: Actually I think this one's just the map author's preference, so `map_plateau_message_pos_y` could be left out, because `campaign_message_box_top_pos_y`  can be changed by a customised theme addon for anyone who doesn't like the default.

tothxa:
@hessenfarmer This affects campaign scripting in a significant way, so I think it's also your "territory". :)

hessenfarmer:
i did not find time for a proper review yet. However I tested if the scenarios load fine and the first texts are displayed. From our structure richtext or better lua failures in richtext arose at load time normall, so the result is promising as only BAR 01 failed the loading. All other tribes campaigns scenarios opened fine. and the text was displayed properly

tothxa:
> the result is promising as only BAR 01 failed the loading.

Oops, that was a lame mistake. Thank you. Sorry, I should have tested all scenarios myself…

At least I tested the theme loading changes much more thoroughly… Also most of richtext.lua. Let's hope there will be no such problems with those.

frankystone (dismissed):
I have played the first three empire campaigns without any observations. There was a crash but i think it's not related to this branch.
Played them with my own Theme.
LGTM

hessenfarmer (dismissed):
Code LGTM.
Did not test any further but I trust kaputtnik here

data/templates/default/init.lua
```diff
@@ -31,6 +31,23 @@ local wui_font_color = {255, 255, 0}
 local wui_font_face = "sans"
 local wui_font_size = 14
 
+local ingame_font_face = "sans"
```
Noordfrees: Can you please rename `ingame` to `wui` for consistency

doc/sphinx/source/themes.rst
```diff
@@ -263,4 +271,104 @@ Some of the above style tables define custom fonts. Additionally, there is a ``f
...
+Optional elements:
+
+* **align** (string): The horizontal alignment of the paragraph, possible values: ``left`` (default), ``center`` or ``middle``, ``right``
+* **valign** (string): The vertical alignment of the paragraph, possible values: ``top`` (default), ``center`` or ``middle``, ``bottom``
+* **indent** (int): Adds an indentation to the first line of the paragraph
+* **spacing** (int): Vertical line spacing as a pixel value
+* **space_before** (int): Padding to be used above the paragraph
+* **space_after** (int): Padding to be used below the paragraph
+
+**Note:** Paddings of consecutive paragraphs are added, not collapsed.
```
Noordfrees:
```suggestion
.. note:: Paddings of consecutive paragraphs are added, not collapsed.
```

data/templates/default/init.lua
tothxa: OK, though I didn't think of these as part of the UI, more like content.

tothxa: The heading styles are used by the About window as well as campaigns and the Encyclopedia. Can they be `text_heading_<n>`?

`text_layout.cc` uses the `…InfoPanelHeading` font styles in the `as_heading…()` functions. Can I rename these functions to `as_infopanel_heading…()`? They seem to be used only in 4 source files, 32 times all together.

Noordfrees: About and Encyclopedia mustn't be mixed, in this case please split into a Wui and an FsMenu style (even if they might be identical in the default theme)

tothxa: Good point. :)

Noordfrees (approved):
LGTM :) 
Thanks for this feature :) 

not tested again

tothxa:
Thanks for all testing and reviews.

Issue #5757, opened 2023-01-16
MSVC CI Builds are failing 


hessenfarmer:
**Describe the bug**
All MSVC builds in our CI chain are failing due to no proper version of the libtool src can be downloaded. 
this is maybe a vcpkg issue but I don't know.

**To reproduce**
Just look at the logs of any recent PR or master

@matthiakl: any idea?

matthiakl:
I already looked at it, but did not create an issue (maybe I should next time, sorry...).
The libtool version vcpkg wants to use is no longer available in the msys repository. This creates a failing dependency chain.
Just waiting for https://github.com/microsoft/vcpkg/pull/28919 to be merged, then deployed on Github runner.

hessenfarmer:
ok the PR got merged but is not yet deployed.  
However while checkiong the deployment I found a log message in the cache section, that a command used by our setup is deprecated maybe this will affect our build actions

matthiakl:
I also noticed the deprecation warning, but it does not affect anything currently. We have time until September 27th to update our actions.
I am working on a branch to update our workflow to resolve the warnings.

PR #5759, opened 2023-01-18
CI workflow update

matthiakl:
**Type of change**
Maintenance / Refactoring

**Issues(s) closed**
Re #5757 

**New behavior**
Resolves the CI deprecation warnings:

> Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/setup-python@v1. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.

> The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

**How it works**
All external actions are updated to the latest version.
Because the release action we are using is not actively maintained, it is not updated to the new requirement. So I scripted the creation directly using `git` and `gh` (Github CLI) commands. In this way we have also more freedom in formatting the release. The release notes will list the last 10 commits instead of only the commits since the last build to provide a better overview.

**Additional context**
Releases from this branch are temporarily deployed to my fork: https://github.com/matthiakl/widelands/releases/tag/latest
Before merging, the TODOs should be addressed to restore the master configuration.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -6,7 +6,8 @@ on:
   pull_request:
     types: [ opened, reopened, synchronize ]
   push:
-    branches: [ master ]
+    # TODO remove branch again
+    branches: [ master, workflow_update ]
 
 jobs:
   # This job runs multiple smaller checks where having several jobs would be overkill.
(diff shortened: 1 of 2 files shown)


hessenfarmer:
the prerelease seems to miss a token, rest lokks fine

matthiakl:
> the prerelease seems to miss a token, rest lokks fine

Maybe this token is not loaded for pull request builds? The run on my fork succeded: 
https://github.com/matthiakl/widelands/actions/runs/3951126186

hessenfarmer (approved):
ok let's have it and fix if it still does not work

matthiakl:
> Before merging, the TODOs should be addressed to restore the master configuration.

Thanks for the review, but I guess you merged it a bit too early. Please see #5761

PR #5763, opened 2023-01-25
Fix add-ons admin dialog Enter key handling

Noordfrees:
**Type of change**
Bugfix

**Issue(s) closed**
When the OK button is disabled, <kbd>Enter</kbd> should not do anything.


tothxa (approved):
Looks straightforward.

PR #5764, opened 2023-01-25
Support compilation with musl libc

### LINKED ISSUE #5760: Widelands master fails to compile with musl libc
The issue is the use of __WORDSIZE. 
https://github.com/widelands/widelands/blob/0a899171bffec4f62c028b74cf5c90953963f6ff/src/base/macros.h#L138
glibc defines it in `bits/wordsize.h`, but musl in `bits/reg.h`. I don't quickly see the former being included, so maybe it is incuded implicitly somewhere...anyway the musl header is not.

How about avoiding its use and platform detection altogether and doing instead
```
#include <climits>
#if (SIZE_MAX == 0xffffffffffffffff)
 #define PRIuS "lu"
 #else
 #define PRIuS "u"
 #endif
```
?

matthiakl:
**Type of change**
Bugfix

**Issues(s) closed**
Fixes #5760 

**New behavior**
- Successfully compiles and runs with musl libc
- Added dependencies for void Linux

**How it works**
Removed usage of internal define `__WORDSIZE` and checked for GNU libintl (see https://invent.kde.org/frameworks/ki18n/-/commit/543b8ae67c57500bf3cabde75bf1599904d6e482)

**Possible regressions**
32 bit builds not tested


src/base/i18n.cc
```diff
@@ -374,9 +374,10 @@ void set_locale(const std::string& name) {
 		// maybe, do another check with the return value (?)
 		locale = lang;
 	}
-
+#if defined(HAS_NL_MSG_CAT_CNTR)
```
thesamesam: This is a glibc internal so it'd be better to just `#if defined(__GLIBC__) && defined(__USE_GNU_GETTEXT)` , I think. See what I did at https://github.com/anope/anope/pull/258.

matthiakl: This would again use an internal define (double `__`), which caused issues with `__WORDSIZE`. So I'd rather go the safe way and try to use a libc independent approach without assumptions about internals. Maybe there is a way to remove this `++_nl_msg_cat_cntr;` completely?

thesamesam: You would be checking an internal __ purely to see if you're using an internal, so it's probably fine (because if it was undefined, who cares?) but I take your point.

i.e. _nl_msg_cat_cntr is purely only ever available with glibc combined with GNU gettext.

Interestingly, https://www.gnu.org/software/gettext/manual/html_node/gettext-grok.html seems to define-it-then-use it. I think what you did is okay, I just don't ever expect it to pass on non-glibc, and if it did, it might have unexpected results.

I'll ask the GNU gettext maintainer what he thinks.

bhaible: I agree with Sam: When you use `_nl_msg_cat_cntr` (for the purpose as here, for notifying the gettext cache that the locale is changing), you should test for it through a configure-time test. Otherwise you will risk compilation errors on platforms that don't have it. The only platforms that have it are glibc and Solaris 11 platforms.

Testing `defined(__USE_GNU_GETTEXT)` is not the right way to test for this feature, because `__USE_GNU_GETTEXT` is defined
- on glibc systems
- on musl libc (which does not support several GNU gettext features: charset conversion, ISO C 99 <inttypes.h> format string directives, and the 'I' format string flag)
- when GNU libintl is installed.

Therefore with this test you would get a link error on musl libc and missed functionality on Solaris 11.

Testing `defined(__GLIBC__) && defined(__USE_GNU_GETTEXT)` is also not right, because while avoiding the link error on musl libc, it results in missed functionality on Solaris 11 and on systems with GNU libintl.

thesamesam: Thanks for the explanation Bruno, let's go for the configure check then. I'll also see about fixing Anope.

matthiakl: Thank you both for the feedback and insights.
So if I understand correctly, you say that I did it already correctly here? cmake checks the presence of `_nl_msg_cat_cntr` during configuration. It does not use any defines, but simply checks, whether it can compile such a source and defines `HAS_NL_MSG_CAT_CNTR` if it can. 

https://github.com/widelands/widelands/blob/c0cda98f81606115f2f722d75c11bc3e9e2d6cf3/src/CMakeLists.txt#L1-L3

bhaible: The `HAS_NL_MSG_CAT_CNTR` approach is the correct one.

But does the check_cxx_source_compiles test actually work in the third case (when GNU libintl is installed)? I don't know what check_cxx_source_compiles does (a compilation check or a compilation+link check?) and whether it considers the need to use `-Lsome/path -libintl` in that case.

matthiakl: It does a compile+link check. However libintl was not considered, so it already failed in the Windows builds (see [here](https://github.com/widelands/widelands/actions/runs/4007249397/jobs/6883529931#step:7:111)), so this was a good hint. 
I added it as library for the check, if it is found during the library discovery.

thesamesam (approved):
Thank you for this!

matthiakl:
It turns out, that the check is not as easy when the link targets are considered, so I converted the PR to draft until I figure it out.

tothxa:
> It turns out, that the check is not as easy when the link targets are considered, so I converted the PR to draft until I figure it out.

Are you referring to the msvc failure? Because I don't think it's related to this PR. At first glance it looks like it's the same webp.lib problem that hit the latest build of my [PR](https://github.com/widelands/widelands/pull/5748) too. It must be the next round of the usual w*ndows VM breakage… :(

matthiakl:
> Are you referring to the msvc failure? 

No. I'm referring to the `-- Performing Test HAS_NL_MSG_CAT_CNTR - Failed` during configuration on Windows and Mac. It should succeed on all platforms in the CI, since we do not have a musl system in there. The build still continues, but it will likely produce unexpected behavior.
https://github.com/widelands/widelands/actions/runs/4072745480/jobs/7017789996#step:4:58

But I will also take a look into the other problem.

thesamesam:
Why do you expect the test to succeed on Windows or macOS? It's not supposed to. It's a glibcism (almost).

matthiakl:
Before this change `_nl_msg_cat_cntr` was used in all build configurations, so it is available in the Windows and Mac builds. The new check should not affect these versions and it should be still compiled in, therefor `HAS_NL_MSG_CAT_CNTR` should be `true`.

thesamesam:
No, it's covered by `#ifdef __linux__`. That code was therefore never built for Windows or macOS.

matthiakl:
Oh, you are right. That ifdef was too high up, I overlooked it. But with the latest commit it did [succeed](https://github.com/widelands/widelands/actions/runs/4075239553/jobs/7024626046#step:4:59).

Now I can either leave it as is for completeness or revert the last commit to reduce cluttering in the CMakeLists.txt.

thesamesam:
If it passes without the last commit / with it reverted, I'd say drop it, so we don't wonder what it's for in a few years' time

hessenfarmer:
may I ask what is the status here.

matthiakl:
> may I ask what is the status here.

@akdjka reported that it indeed resolves the compilation issues ([comment](https://github.com/widelands/widelands/issues/5760#issuecomment-1403372934))
The handling of `_nl_msg_cat_cntr` is also clear now and approved.

So this PR only misses a review from a member to be able to merge, you are welcome to do that ;-)

hessenfarmer:
ok I could do that. However it is unclear whether the check should be true for windows which it isn't currently.

matthiakl:
> However it is unclear whether the check should be true for windows which it isn't currently.

Yes, there was a bit confusion. Let me try to summarize:
Generally speaking, the check should be true for windows because it uses GNU libintl which has `_nl_msg_cat_cntr`. For the check to succeed, linking information of libintl must be provided, which I did successfully in b31a2dd (see this [run](https://github.com/widelands/widelands/actions/runs/4075239553/jobs/7024626046#step:4:59)).
However, I reverted the commit to keep the CMakeLists.txt simple, because the C code using `_nl_msg_cat_cntr` is fenced by `__linux__` and therefore has no impact on Windows builds and the result of the check is irrelevant.

hessenfarmer:
> > However it is unclear whether the check should be true for windows which it isn't currently.
> 
> Yes, there was a bit confusion. Let me try to summarize: Generally speaking, the check should be true for windows because it uses GNU libintl which has `_nl_msg_cat_cntr`. For the check to succeed, linking information of libintl must be provided, which I did successfully in [b31a2dd](https://github.com/widelands/widelands/commit/b31a2dd79b2b98f1ce79da269ffe62d6994cdbc3) (see this [run](https://github.com/widelands/widelands/actions/runs/4075239553/jobs/7024626046#step:4:59)). However, I reverted the commit to keep the CMakeLists.txt simple, because the C code using `_nl_msg_cat_cntr` is fenced by `__linux__` and therefore has no impact on Windows builds and the result of the check is irrelevant.

ok but in this case we should not enable the check fo windows at all.

tothxa:
> the C code using `_nl_msg_cat_cntr` is fenced by `__linux__`

~Couldn't the `extern` declaration be moved to the same `#ifdef` block too? Otherwise shouldn't it get its own `#ifdef __linux__` block?~

~OTOH if it's available on w*ndows, why can't it be used? Is there some other linux-specific trickery in the guarded code?~

Scrap that. The `__linux__` block is just that. But shouldn't `++_nl_msg_cat_cntr` be outside, only guarded by your new `#ifdef`?

bhaible:
> [Check for _nl_msg_cat_cntr only on linux systems](https://github.com/widelands/widelands/pull/5764/commits/67f3d288acd1d400372c2adf4ace17f56659c3aa)

It is a mistake to do i18n stuff only on Linux. Linux is the *kernel*, i18n is totally unrelated to what happens in the kernel. I explained in https://github.com/widelands/widelands/pull/5764#discussion_r1092431508 that `_nl_msg_cat_cntr` is available in a number of situations. Currently it is present on 3 kernels, not just Linux:
- Linux (with glibc, not musl),
- Hurd (also has glibc),
- Solaris 11.

But you shouldn't hardcode that list, because there are plans to make `_nl_msg_cat_cntr` available also on macOS and Windows systems.

> it is used in i18n.cc surrounded by #ifdef __linux__, so checking is only required on linux systems

The `#ifdef __linux__` in i18n.cc is already wrong. The correct way to do things is to do the configure test on all platforms and in i18n.cc test the result of the configure test. This way you don't hardcode any list of platforms.

matthiakl:
Well, now I'm thinking about removing this strange variable completely and remove the complicated check with it…
After some code digging I found that in `setlocale`, which is called on all platforms, the counter is already incremented (https://codebrowser.dev/glibc/glibc/locale/setlocale.c.html#389). So it shouldn't be necessary in our code in the first place and Windows builds already worked without it. A quick test on my linux system showed no issues without that line.
So maybe we can just remove it and not have to use this platform or library specific tests?

The introduction into the code was in 2008 (9cbfed1), so there is a chance the related issues have been fixed in dependencies or distributions. This should be tested on an Ubuntu system.

bhaible:
The purpose of the variable is to invalidate loaded message catalogs when the locale is changed through `setlocale`. In some cases in code in glibc or libintl already does it; in some cases it doesn't.

To test whether it works: 1. Run the program. Make it show some messages. 2. Switch to a different language. 3. Make it show the same messages. Verify that they are now displayed in the new language.

tothxa:
I briefly tested removing `_nl_msg_cat_cntr` from `base/i18n.cc` completely on my Debian buster system too, and it didn't cause any problem here either.

I'm happy you've found the place in glibc where it is incremented, because it belongs there anyway.

tothxa (approved):
I think this should be fine, but I'm not sure whether it needs more testing.

hessenfarmer (approved):
tested a windows 32 bit build briefly. works nice. no problems with switching locales.

As our testsuite is based on Ubuntu and is green this should be fine.

PR #5765, opened 2023-01-26
Fix build with gcc 13 by including <cstdint>

heirecka:
Like other versions before, gcc 13 moved some includes around and as a result <cstdint> is no longer transitively included. Explicitly include it for uint*_t.


Noordfrees (approved):
LGTM, thanks for the patch :)

PR #5773, opened 2023-02-03
Using helper functions in performance texts.

gpalino:

**Type of change**
Refactoring

Usage of format_seconds() and format_minutes_seconds() instead of direct ngettext() in the lua scripts. This improves readability and maintainability.

**Additional context**
Simple usage converted with this script:
`find data -name '*.lua' -exec sed -i -e 's/^\([^%]*\)%s\([^%"]*"):bformat(\)ngettext("%d second", "%d seconds", \([0-9]\+\)):bformat([0-9]\+))/\1%1%\2format_seconds(\3))/g' \{\} \;`


Noordfrees (commented):
If it wouldn't be too much trouble please change only the `format_seconds` part but leave the placeholders in the strings unchanged. All changed strings will have to be newly translated so this causes extra work for translators for no reason.

Also, Generally, `%s` is preferred over `%1%` since it enforces the correct data type. If you pass a number to `%s` by accident the program will complain. `%1%` is to accommodate any data type. The rule is not strictly enforced, but wildcards (`%1%`) should be used only if you don't know in advance what data type will be used.

gpalino:
I agree, should be ok now.

hessenfarmer:
Although it looks like before, I would appreciate if you could change all times above 60 to minutes seconds format.

gpalino:
done

data/tribes/initialization/amazons/units.lua
```diff
@@ -1411,7 +1411,7 @@ wl.Descriptions():new_tribe {
             -- TRANSLATORS: Purpose helptext for an amazon production site: Charcoal Kiln
             purpose = pgettext("amazons_building", "Burns logs into charcoal."),
             -- TRANSLATORS: Performance helptext for an amazon production site: Charcoal Kiln
-            performance = pgettext("amazons_building", "The charcoal kiln needs %s on average to produce one lump of coal."):bformat(ngettext("%d second", "%d seconds", 121):bformat(121))
+            performance = pgettext("amazons_building", "The charcoal kiln needs %s on average to produce one lump of coal."):bformat(format_seconds(121))
```
hessenfarmer:
```suggestion
            performance = pgettext("amazons_building", "The charcoal kiln needs %s on average to produce one lump of coal."):bformat(format_minutes_seconds(2,1))
```

hessenfarmer (approved):
one small nit.
rest LGTM

Issue #5631, opened 2022-10-29
Launch Game window does not resize contents properly


tothxa:
**Describe the bug**
The "Launch Game" window can increase the size of some parts, but not decrease them when the game window is resized.

**To Reproduce**
1. Start new single player game and select map to get to the "Launch Game" window
2. Resize the game window by dragging by the edges so that only the width or only the height changes. Try increasing then decreasing the size
3. See error

**Expected behavior**
All window elements should be resized properly for all changes

**Screenshots**
![shot0026](https://user-images.githubusercontent.com/80712255/198855931-b41729ce-0765-49d9-a5db-7586629e7d06.png)
![shot0027](https://user-images.githubusercontent.com/80712255/198855933-a8d90932-0b29-4051-8430-ea9ddc046dfa.png)
![shot0028](https://user-images.githubusercontent.com/80712255/198855934-500b2e84-2418-4931-9c5b-404a72e45864.png)
![shot0029](https://user-images.githubusercontent.com/80712255/198855936-0cc15886-604f-4c2c-ab0f-a1952d2207ea.png)
![shot0030](https://user-images.githubusercontent.com/80712255/198855937-e776ff76-a2b8-4c32-8469-c2487769e513.png)
![shot0031](https://user-images.githubusercontent.com/80712255/198855938-24c25f60-28fa-429d-8480-00efddd659cd.png)

**Version:**
 - OS: antix/Debian buster with backports
 - Widelands Version: v1.1

**Additional context**
Multiplayer launch window only has problem with vertical resize. If both width and height change (dragging by window corner), then the contents are resized properly. Also if e.g. width is decreased, when height is changed afterwards, the widths will be resized properly.

PR #5774, opened 2023-02-05
Ai diplomacy

hessenfarmer:
**Type of change**
first attempt to make the AI actively managing Diplomacy

**New behavior**
After a certain time the AI asks for alliance, accepts or rejects and leaves alliances based on a diplomatic score.

**How it works**
together with the usual playerstats we now calculate a diplomatic score based on different parameters but mainly centered on players strength. In the handling of diplomacy actions we base acceptance on this score. For players with very high scores we ask for alliance and if a player gets a really bad score we break alliance.  

**Possible regressions**
everything related to AI using playerstatisitcs. 

**Screenshots**
If applicable, add screenshots to help explain the feature.

Any input on better scoring, or better handling of asking for alliance is welcome. Needs a lot of testing


hessenfarmer:
@Noordfrees 
I am quite happy with this already. However an AI player might keep sending requests although it got denied a couple of thimes whis is cluttering the messages window. 
Is there any possibility for an AI player to check for the answer on a request.

Noordfrees:
> Is there any possibility for an AI player to check for the answer on a request.

You could create a new Notification type `NoteDiplomacy`, which is published by `CmdDiplomacy::execute` and subscribed to by the AI.

hessenfarmer:
> > Is there any possibility for an AI player to check for the answer on a request.
> 
> You could create a new Notification type `NoteDiplomacy`, which is published by `CmdDiplomacy::execute` and subscribed to by the AI.

thanks for the answer. In the meantime I had an idea just to limit requests to the same player, and implemented that.

hessenfarmer:
@widelands/developers 
I would be happy if you could test whether this is what we expect from an AI regarding diplomacy.

tothxa:
I tested this a little, and it works quite well. My first observations:

 - the AIs got a bit too active for my liking :)
 - they're amusingly predictable as they all want to join the strongest player

hessenfarmer:
> I tested this a little, and it works quite well. My first observations:
> 
>     * the AIs got a bit too active for my liking :)     
   
well, currently they are allowed to send a request to the same player only every 10 minutes. should this be changed and to which interval or shall we increase the threshhold for diploscore to make a request only when score is high enough.  

> 
>     * they're amusingly predictable as they all want to join the strongest player     
   
basically I have implemented a scoring to make them team against the strongest player. But it may not be considered in every case. https://github.com/widelands/widelands/blob/d2cb9a21295d32d743dff132c6f20fba72d1ab8c/src/ai/defaultai.cc#L6878

tothxa:
> well, currently they are allowed to send a request to the same player only every 10 minutes. should this be changed and to which interval or shall we increase the threshhold for diploscore to make a request only when score is high enough.

Maybe add a random extra delay between requests too, like for the first request? It felt like they came in waves. OTOH if the situation changes significantly, then it may make sense to reduce the delay.

> basically I have implemented a scoring to make them team against the strongest player. But it may not be considered in every case.

I continued the test game, and I think I saw some of that too, but not enough. OTOH I also saw the 2 weakest AIs fighting each other instead of teaming up.

I don't think team strength is considered in the decisions, or is it? Both the current situation and the possible outcome of each request should be evaluated. Also diplomacy and military decisions should be in line with each other.

hessenfarmer:
> Maybe add a random extra delay between requests too, like for the first request? It felt like they came in waves. OTOH if the situation changes significantly, then it may make sense to reduce the delay.

Good idea will implement the randomness. However adapting this to the game situation would be hard. only thing I might do is feed the random function with the diploscore.   

> 
> > basically I have implemented a scoring to make them team against the strongest player. But it may not be considered in every case.
> 
> I continued the test game, and I think I saw some of that too, but not enough. OTOH I also saw the 2 weakest AIs fighting each other instead of teaming up.     
   
some kind of irrationality is very human isn't it? ;-) but for sure i might improve the scoring.    
  
> 
> I don't think team strength is considered in the decisions, or is it? Both the current situation and the possible outcome of each request should be evaluated. Also diplomacy and military decisions should be in line with each other.    
  
Ok I will implement some scoring considering team strength as well. What do you mean by aligning diplomacy and military decisions?

tothxa:
> However adapting this to the game situation would be hard. only thing I might do is feed the random function with the diploscore.

Does the AI have a memory of its past decisions?

> What do you mean by aligning diplomacy and military decisions?

E.g. not attacking the player it tries to make peace with. :) But also whether the other player is blocking a desired direction of expansion, and of course possible common enemies (not only by strength, but also by map positions), but this may be hard to evaluate.

Also, given 2 or more neighbours, choose which ones to attack and which ones to team up with, and do both attacks and diplomacy based on the same decision.

hessenfarmer:
> Does the AI have a memory of its past decisions?   
   
No, it doesn't have any memories of its decisions and actions.      

> 
> > What do you mean by aligning diplomacy and military decisions?
> 
> E.g. not attacking the player it tries to make peace with. :) But also whether the other player is blocking a desired direction of expansion, and of course possible common enemies (not only by strength, but also by map positions), but this may be hard to evaluate.

The AI does not have the knowledge of its relative map position vs. other players. We only know where we have a frontier and whether we see or have seen an enemy. Generally the AI is discouraged to attack stronger players and it is encouraged to team with stronger players. I don't have any clue (yet) how to allign these scores, though, as the military decisions are field based while the diplomacy is more general. But I agree that would be a potential improvement.   
> 
> Also, given 2 or more neighbours, choose which ones to attack and which ones to team up with, and do both attacks and diplomacy based on the same decision.    
    
Currently not possible with the information available to the AI

tothxa:
> No, it doesn't have any memories of its decisions and actions.

I guessed so. I think the resulting instability is a weakness of the AI, but obviously it's much more general than to be addressed in this PR.

> The AI does not have the knowledge of its relative map position vs. other players. We only know where we have a frontier and whether we see or have seen an enemy.

That's already a lot of data to consider. Human players work with the same info after all.

> I don't have any clue (yet) how to allign these scores, though, as the military decisions are field based while the diplomacy is more general.

Don't all decisions have a strength by which the results could be weighted against each other after the initial calculations?

> > Also, given 2 or more neighbours, choose which ones to attack and which ones to team up with, and do both attacks and diplomacy based on the same decision.
> Currently not possible with the information available to the AI

But this is just about neighbours, and you've just wrote that the AI knows about them.

hessenfarmer:
> Don't all decisions have a strength by which the results could be weighted against each other after the initial calculations?

No, but to explain this I need to explain better how the AI works.  Basically all decisions are field based. Which means we have a couple of all of our and neighbouring fields (we don'r know the coordinates though) then we scan for enemy sites within a scan radius a bit bigger then our militarysites vision and if we find an enemy militarysite we decide whether we are abele to attack in terms of Soldiers can walk there and if it makes sense to attack there. the sense to attack this one site is calculated from general strength parameters like the diplomacy and local parameters specific to this one site and our and the enemies local strength. 
it might be doable to pull out the general weighing factors and use diplomacy score instead but this is out of scope of this PR I am afraid.    
> 
> > > Also, given 2 or more neighbours, choose which ones to attack and which ones to team up with, and do both attacks and diplomacy based on the same decision.
> > > Currently not possible with the information available to the AI
> 
> But this is just about neighbours, and you've just wrote that the AI knows about them. 

yers but not exactly where they are. we don't even really  which player we are attacking, just if we have chances to win the local fight. However we have a function to count them and I used this in the last commit.  Don't know whether this is an improvement and I need to test whether we need to adopt threshholds or scoring based on the logged values.  
I would be happy dor any feedback whether the last commit is an improvement or not. 
if you are really interested we could use Hasi50's widelands chat on a trustworthy Jitsi server and i can explain you the AI better by showing the code if you are interested, but not tonight.

tothxa:
I tested the last commit (this time I only watched the AIs). I'm not sure how much their behaviour changed, though at least now they seem to prefer requests to neighbours. But there's still a lot of random jumping in and out of teams, as well as declining a request then sending an equivalent one.

hessenfarmer:
> as well as declining a request then sending an equivalent one.    

Well, I believe this might be due to that I kept over the mechanism to accept only on a random basis only evaluate 50% of the requests and deny the other 50% categorically. this seems to be inappropriate. I will base this purely on diploscore now. in a way that if the score is high enough to send a request it will accept in any case. Thanks for testing and feedback. 

for the random jumps I need to put in some debug logging but I think this is just a matter of tweaking the parameters. However I feel some random jumps are not to bad as we shouldn't make the AI too predictable.

src/ai/defaultai.cc
```diff
@@ -3330,18 +3330,21 @@ void DefaultAI::check_flag_distances(const Time& gametime) {
...
-			accept &= player_statistics.get_player_power(pda.sender) *
-			             (player_statistics.get_player_land(pda.sender) / 100) >
-			          player_statistics.get_player_power(player_number()) *
-			             (player_statistics.get_player_land(player_number()) / 100);
+		if (pda.other == mypn) {
+			// accept if diploscore high, else accept only 50% but not if player is seen lately
+			bool accept = player_statistics.get_diplo_score(pda.sender) >= 18 ||
+			              ((player_statistics.player_seen_lately(pda.sender, gametime) ||
+			                RNG::static_rand(2) == 0) &&
+			               player_statistics.get_diplo_score(pda.sender) > 5);
```
tothxa: How about making the probability depend on the score?

```suggestion
			               player_statistics.get_diplo_score(pda.sender) > 5 &&
			               RNG::static_rand(player_statistics.get_diplo_score(pda.sender)) > 5);
```

Also it would be more readable if it was done in several `if`s instead of a single assignment.

src/ai/defaultai.cc
```diff
@@ -3352,6 +3355,26 @@ void DefaultAI::diplomacy_actions(const Time& gametime) {
...
+		    !player_statistics.player_diplo_requested_lately(opn, gametime)) {
+			if (player_statistics.get_diplo_score(opn) >= 18) {
+				player_statistics.set_last_time_requested(gametime, opn);
+				if (other_player->team_number() == 0) {
+					game().send_player_diplomacy(mypn, Widelands::DiplomacyAction::kInvite, opn);
+				} else if (other_player->team_number() != me->team_number()) {
+					game().send_player_diplomacy(mypn, Widelands::DiplomacyAction::kJoin, opn);
+				}
+			}
+			if (player_statistics.get_diplo_score(opn) < -10 &&
```
tothxa: This is practically an `else if`.

src/ai/defaultai.cc
```diff
@@ -3352,6 +3355,26 @@ void DefaultAI::diplomacy_actions(const Time& gametime) {
 			                             pda.sender);
 		}
 	}
+	for (Widelands::PlayerNumber opn = 1; opn <= game().map().get_nrplayers(); ++opn) {
+		const Widelands::Player* other_player = game().get_player(opn);
+		// other player needs to exist and different from me
+		// we need to be still alive and we don't have send a request in last 5 + X minutes
+		if (other_player != nullptr && opn != mypn && !me->is_defeated() &&
+		    !player_statistics.player_diplo_requested_lately(opn, gametime)) {
```
tothxa: You could invert the condition and use `continue` to spare one level of indentation for the rest of the loop.

src/ai/defaultai.cc
```diff
@@ -6754,9 +6805,139 @@ void DefaultAI::update_player_stat(const Time& gametime) {
...
+					   (old60_strength * std::abs(management_data.get_military_number_at(184) / 7)) +
+					            old60_land >
+					         (me_old60_strength *
+					          std::abs(management_data.get_military_number_at(184) / 7)) +
+					            me_old60_land ?
+                     2 :
+                     -2;
+					inputs[19] = me_cass > cass ? 2 : -1;
+					inputs[20] = this_player->team_number() == 0 ? 4 : -3;
+					inputs[21] = this_player->team_number() == 0 ?
```
tothxa: So a teamless player is counted differently from a player who is alone on a team? (or whose team-mates were defeated)

src/ai/ai_help_structs.cc
```diff
@@ -1110,6 +1124,23 @@ bool PlayersStrengths::player_seen_lately(Widelands::PlayerNumber pn, const Time
...
+		// Should happen only rarely so we print a warning here
+		verb_log_warn("AI %d: player has no statistics yet\n", this_player_number);
+		return false;
+	}
+	if (all_stats[pn].last_time_requested.is_invalid()) {
+		return false;
+	}
+
+	Duration request_interval = Duration(
+	   (RNG::static_rand(std::abs(all_stats[pn].players_diplomacy_score) + 1) + 5) * 60U * 1000U);
```
tothxa: What is intended here? Something like weak players won't get requests anyway, so might as well re-evaluate them sooner, while we don't want to nag strong players too much, because they have all reasons to ignore us?
And don't bother with hopeless players either, due to the abs()?

tothxa:
This time I ran a test with 4 AIs on Elven Forest with verbose logging enabled. There were much fewer diplomacy requests with the latest changes.

The diploscores start out in the massively negative region (around -36), so they leave the initial teams at the first opportunity. I don't think that should be so. Shouldn't all players get scores around 0 when all of them are almost equal?

Looking at the graphs, I also realised that the number of buildings is at least as important as the land area, but it is not considered at all.

Here's the standing after 1:40 with the last diploscores from the log (none of the players started much military production yet):
diplomacy

[01:39:53.896 game] DEBUG: For player(4), the player(1) has the diploscore: -29
[01:39:53.896 game] DEBUG: For player(4), the player(2) has the diploscore: 29
[01:39:53.896 game] DEBUG: For player(4), the player(3) has the diploscore: -9
[01:39:53.896 game] DEBUG: For player(4), the player(4) has the diploscore: -23
[01:39:55.906 game] DEBUG: For player(3), the player(1) has the diploscore: -17
[01:39:55.906 game] DEBUG: For player(3), the player(2) has the diploscore: 14
[01:39:55.906 game] DEBUG: For player(3), the player(3) has the diploscore: -19
[01:39:55.906 game] DEBUG: For player(3), the player(4) has the diploscore: 13
[01:40:03.715 game] DEBUG: For player(1), the player(1) has the diploscore: -41
[01:40:03.715 game] DEBUG: For player(1), the player(2) has the diploscore: 6
[01:40:03.715 game] DEBUG: For player(1), the player(3) has the diploscore: -15
[01:40:03.715 game] DEBUG: For player(1), the player(4) has the diploscore: -1
[01:40:04.518 game] DEBUG: For player(2), the player(1) has the diploscore: -31
[01:40:04.518 game] DEBUG: For player(2), the player(2) has the diploscore: -10
[01:40:04.518 game] DEBUG: For player(2), the player(3) has the diploscore: -9
[01:40:04.518 game] DEBUG: For player(2), the player(4) has the diploscore: 33

I don't think 1 and 3 should hate each other so much, because their best hope is to work together. OTOH, 2 and 4 should be more wary of each other. I don't mean not to team up, only to not so eagerly.

There had already been some short-lived alliances, but I missed the break-ups, so I haven't analysed them yet.

tothxa:
I ran the previous test some more. It seems that when 2 players with very high scores for each other finally make an alliance (it takes too long, so the delay should be limited for high diploscores), suddenly their respective scores go very negative. This is definitely a bug, and I think this must be the reason why they seldom stay in an alliance for long.

[edit:] Here's a plot of player 2 and 4's scores for each other against whether they are allied to prove it:
2vs4

Otherwise the scores are pretty stable, as they should be, only changing as the power balance is shifting:

(all players' scores for player 2)

However it shows that whether a player is in a team has a too big impact, even when they're alone in the "team". Probably this is the root cause of the flip-flopping, which is then made worse by being applied in the wrong direction for own teammates.

BTW here's the savegame at the end of this test run so that you can see the game stats:
[ai_diplo_state_0259.wgf.zip](https://github.com/widelands/widelands/files/10788068/ai_diplo_state_0259.wgf.zip)

tothxa:
BTW it would help a lot analysing the behaviour of this PR if diplomatic actions were logged with the game time (now only team changes are logged with real time). Here's a crude patch for it:
[dipl_log_patch.txt](https://github.com/widelands/widelands/files/10788046/dipl_log_patch.txt)
(crude because `text` is the translated message here, with the names of the players instead of player numbers as in the diploscore logs)

tothxa:
> This time I ran a test with 4 AIs on Elven Forest with verbose logging enabled. There were much fewer diplomacy requests with the latest changes.

I have to retract this. It only lasted until they met (which may be a good thing), but in the end it got too busy again:
dipl_req_3

BTW the test ended in a team win with all 4 AIs accidentally getting on the same team. Maybe the main diplomacy code should be changed to invalidate pending requests when there is a change in the line-up of the team to be joined. Also probably team wins shouldn't be allowed for games with diplomacy enabled.

src/ai/defaultai.cc
hessenfarmer: currently yes but that may one of the contributors to tzhe toggling

src/ai/defaultai.cc
hessenfarmer: ok about the readability. in your suggestion a score of 6 would have zero prob. and a score of 11 would have a 50% chance. so I will think about this but I might change values.

src/ai/ai_help_structs.cc
hessenfarmer: yes exactly

src/ai/defaultai.cc
hessenfarmer: several ifs resulted in clang Tidy complaining

hessenfarmer:
I tried to fix and tweak the behaviour as far as I could. AI only games do not end always in a team win of all members now. diplomacy actions seem to be acceptable in terms of frequency.  
please test again.

src/ai/defaultai.cc
```diff
@@ -3350,6 +3352,35 @@ void DefaultAI::diplomacy_actions(const Time& gametime) {
...
+	for (Widelands::PlayerNumber opn = 1; opn <= game().map().get_nrplayers(); ++opn) {
+		const Widelands::Player* other_player = game().get_player(opn);
+		// other player needs to exist and different from us
+		// we need to be still alive and we don't have send a request in last 5 + X minutes
+		if (other_player == nullptr || opn == mypn || me->is_defeated() ||
+		    player_statistics.player_diplo_requested_lately(opn, gametime)) {
+			continue;
+		}
+		if (player_statistics.get_diplo_score(opn) >= 30)
+			&&other_player->team_number() != me->team_number() {
```
tothxa: There's a compile error here.

```suggestion
		if (player_statistics.get_diplo_score(opn) >= 30 &&
		    other_player->team_number() != me->team_number()) {
```

src/ai/defaultai.cc
```diff
@@ -3350,6 +3352,34 @@ void DefaultAI::diplomacy_actions(const Time& gametime) {
...
+	for (Widelands::PlayerNumber opn = 1; opn <= game().map().get_nrplayers(); ++opn) {
+		const Widelands::Player* other_player = game().get_player(opn);
+		// other player needs to exist and different from us
+		// we need to be still alive and we don't have send a request in last 5 + X minutes
+		if (other_player == nullptr || opn == mypn || me->is_defeated() ||
+		    player_statistics.player_diplo_requested_lately(opn, gametime)) {
+			continue;
+		}
+		if (player_statistics.get_diplo_score(opn) >= 30 &&
+		    other_player->team_number() != me->team_number()) {
```
tothxa: I think a check that team number is not 0 is required here. At least in my latest tests they never made any requests, despite the high diploscores.

src/ai/defaultai.cc
```diff
@@ -3350,6 +3352,34 @@ void DefaultAI::diplomacy_actions(const Time& gametime) {
...
+			                  player_statistics.get_diplo_score(opn));
+			player_statistics.set_last_time_requested(gametime, opn);
+			if (other_player->team_number() == 0) {
+				game().send_player_diplomacy(mypn, Widelands::DiplomacyAction::kInvite, opn);
+			} else if (other_player->team_number() != me->team_number()) {
+				game().send_player_diplomacy(mypn, Widelands::DiplomacyAction::kJoin, opn);
+			}
+		} else if (player_statistics.get_diplo_score(opn) < -15 &&
+		           other_player->team_number() == me->team_number()) {
+			game().send_player_diplomacy(mypn, Widelands::DiplomacyAction::kLeaveTeam, opn);
```
tothxa: Maybe checking that they're not both in team 0 would be better here too?  And team leaves should be logged too.

src/ai/defaultai.cc
```diff
@@ -3330,18 +3330,20 @@ void DefaultAI::check_flag_distances(const Time& gametime) {
...
-			accept &= gametime > Time((10 + RNG::static_rand(10)) * 60 * 1000);
-			// only accept if asking player is stronger (based on mil power and land)
-			accept &= player_statistics.get_player_power(pda.sender) *
-			             (player_statistics.get_player_land(pda.sender) / 100) >
-			          player_statistics.get_player_power(player_number()) *
-			             (player_statistics.get_player_land(player_number()) / 100);
+		if (pda.other == mypn) {
+			// accept if diploscore high, else accept only 50%
+			bool accept =
+			   player_statistics.get_diplo_score(pda.sender) >= 20 ||
```
tothxa: Shouldn't this be in sync with the send threshold? That is now 30. Maybe a constexpr is needed at least for this one.

src/ai/defaultai.cc
```diff
@@ -6754,9 +6816,180 @@ void DefaultAI::update_player_stat(const Time& gametime) {
...
+                     -5;
+					inputs[33] =
+					   cur_strength * 2 + cur_land < player_statistics.get_max_power() * 2 +
+					                                    player_statistics.get_max_land() &&
+					         player_statistics.get_max_power() * 2 + player_statistics.get_max_land() <
+					            (cur_strength + me_strength) * 2 + cur_land + me_land ?
+                     4 :
+                     0;
+					inputs[34] = this_player->team_number() == me->team_number() ? 5 : -5;
+					inputs[35] = this_player->team_number() == me->team_number() ? 3 : -3;
```
tothxa, line 7132: `this_player->team_number != 0 && …`?

src/ai/ai_help_structs.cc
```diff
@@ -1003,19 +1005,22 @@ void PlayersStrengths::add(Widelands::PlayerNumber pn,
...
+		   std::make_pair(opn, PlayerStat(pltn, pp, op, o60p, cs, land, oland, o60l, ds)));
 	} else {
 		all_stats[opn].players_power = pp;
 		all_stats[opn].old_players_power = op;
 		all_stats[opn].old60_players_power = o60p;
 		all_stats[opn].players_casualities = cs;
 		all_stats[opn].players_land = land;
 		all_stats[opn].old_players_land = oland;
-		all_stats[opn].old60_players_land = oland;
+		all_stats[opn].old60_players_land = o60l;
```
tothxa: Shouldn't buildings get the same treatment?

src/ai/ai_help_structs.cc
```diff
@@ -1188,6 +1234,22 @@ uint32_t PlayersStrengths::get_enemies_max_land() {
...
 
+uint32_t PlayersStrengths::get_max_power() {
+	uint32_t power = 0;
+	for (auto& item : all_stats) {
+		power = std::max<uint32_t>(power, item.second.players_power);
+	}
+	return power;
+}
+
+uint32_t PlayersStrengths::get_max_land() {
```
tothxa, line 1401: Shouldn't buildings get an equivalent too?

tothxa:
I ran another test with the AIs put in teams to work around the bug that prevents requests between teamless AIs. It is indeed much better now. I'll do the diagrams and deeper analysis tomorrow.

For now, one quick comment: Team diploscores are needed very much. Most obvious example: 3 players are teamed up against the strongest one, player(2). P(1) quits the team on a whim. Later on it requests to join P(3), but gets rejected. Almost at the same time, P(4) invites P(1), but it rejects, even though this would get it into the desired team of P(3). It could also help decide whether to invite or request join.

But I'd be OK with leaving team scores to another PR too, if the previous comments are resolved.

src/ai/defaultai.cc
hessenfarmer: actually I did this on purpose. 
Asking only if a player has a very high score to not generate too much requests, but accepting if the other player has not that much trust, to help getting some teams at last. (if in team the diploscore should get some boost over time)

src/ai/defaultai.cc
hessenfarmer: well I think if both are team 0 this encourages some teambuildings as well as it helps keeping some trust.  So I won't change this.

src/ai/defaultai.cc
hessenfarmer: oops you are right, overlooked that, reason for this is I found a lot of instances, where after accepting some lines above a player send a new invitation to the same player he already accepted.

src/ai/ai_help_structs.cc
hessenfarmer: actually these stats are needed in the AI warfare code, they are not invented for the diplomacy.     
I am not sure whether we need to consider development of buildings over time for AI as I think they are less important and actual buildings should be sufficient. 
as they are only needed for the diploscore we could have them but we could easily keep them internal.

src/ai/ai_help_structs.cc
hessenfarmer: maybe, do you think number of buildings is that important to determine whether to team up?

hessenfarmer:
> I ran another test with the AIs put in teams to work around the bug that prevents requests between teamless AIs. It is indeed much better now. I'll do the diagrams and deeper analysis tomorrow.       
    
please wait for me to fix the bug you spotted.     
  
> 
> For now, one quick comment: Team diploscores are needed very much. Most obvious example: 3 players are teamed up against the strongest one, player(2). P(1) quits the team on a whim. Later on it requests to join P(3), but gets rejected. Almost at the same time, P(4) invites P(1), but it rejects, even though this would get it into the desired team of P(3). It could also help decide whether to invite or request join.      
     
I see the point but this needs thinking to avoid having cpu hungry nested for loops.      
    
> 
> But I'd be OK with leaving team scores to another PR too, if the previous comments are resolved.

src/ai/defaultai.cc
tothxa: OK.

src/ai/defaultai.cc
hessenfarmer: basically only the logging was incorrect as an invitstion was logge where in fact nothing happened and i misinterpreted this as a bug and the bugfix introduced the real bug. *head-Desk*

src/ai/ai_help_structs.cc
tothxa: I think the number of buildings is a good indicator of the strength of a player's economy, so it's at least as important as land owned. Land owned in itself is only potential and cushion against attacks. So military, land and buildings are the 3 main indicators of a player's strength for me, all of them with about the same weight.

src/ai/defaultai.cc
hessenfarmer: done

src/ai/defaultai.cc
tothxa: Indeed, you're right. At least this way there's no drop or jump in score when they first team up.

OTOH, then we get different treatment for players who are alone in a team, so maybe check for that too?
Or let that be a source of randomness and a way to affect AI behaviour for human players? Yes, it makes sense, and the checks would be too complicated to consider all possibilities.

src/ai/defaultai.cc
```diff
@@ -3350,6 +3352,44 @@ void DefaultAI::diplomacy_actions(const Time& gametime) {
...
+				player_statistics.set_last_time_requested(gametime, opn);
+				verb_log_dbg_time(gametime,
+				                  "Player(%d), invites player (%d) to join with diploscore: %d\n",
+				                  static_cast<unsigned int>(mypn), static_cast<unsigned int>(opn),
+				                  player_statistics.get_diplo_score(opn));
+			} else if (other_player->team_number() != me->team_number()) {
+				game().send_player_diplomacy(mypn, Widelands::DiplomacyAction::kJoin, opn);
+				player_statistics.set_last_time_requested(gametime, opn);
+				verb_log_dbg_time(gametime,
+				                  "Player(%d), requests player (%d) to join with diploscore: %d\n",
```
tothxa:
```suggestion
				                  "Player(%d), requests to join player (%d) with diploscore: %d\n",
```

src/ai/ai_help_structs.cc
hessenfarmer: ok I will implement something like this then

tothxa:
I got an interesting test game this time. The 2 weak players (red and yellow) got stuck between the 2 strong players (green was strongest initially, but as blue got allied with red and yellow, it could gain strength undisturbed, to become the strongest). Blue and green never met each other, so red and yellow had a very hard diplomatic situation. Whichever of blue or green they helped, the other got stronger on their backs. This ended in a team win too, but it actually made a lot of sense for the weak players this time. They had been begging green for a peace for a while without success too, before it finally accepted.

shot0000

hessenfarmer:
> I got an interesting test game this time. The 2 weak players (red and yellow) got stuck between the 2 strong players (green was strongest initially, but as blue got allied with red and yellow, it could gain strength undisturbed, to become the strongest). Blue and green never met each other, so red and yellow had a very hard diplomatic situation. Whichever of blue or green they helped, the other got stronger on their backs. This ended in a team win too, but it actually made a lot of sense for the weak players this time. They had been begging green for a peace for a while without success too, before it finally accepted.
> 
     
uhm, is this good for you or do I need to do anything? Reagrding the team wins, I am not 100% content with the result yet but at least I had some matches where one player won an autocrat game.  maybe I should increase the malus if the team is big. 

I implemented some safeguards to not end with one player teams. Mainly to make it better in interaction with human players.   
I have an idea for better distinction of inviting and joining, but this will only happen tomorrow

tothxa:
> > This ended in a team win too, but it actually made a lot of sense for the weak players this time.
> 
> uhm, is this good for you or do I need to do anything?

No need to change anything based on this test run. I'll run at least another test though, and I'd like to do one where I play too, but I'm not sure I'll be able to do it.

> Reagrding the team wins, I am not 100% content with the result yet but at least I had some matches where one player won an autocrat game. maybe I should increase the malus if the team is big.
 
Actually, why not check directly whether joining player is last one outside of team when the request is decided?

tothxa:
Meanwhile, here are my plotting scripts: [dipl_score_plot_scripts.zip](https://github.com/widelands/widelands/files/10947334/dipl_score_plot_scripts.zip) (run `dipl_plot.sh <logfile>`, requires `gnuplot` and possibly GNU `sed` and `basename`, ie. I didn't check posix compatibility)

And here are the logs and plots of the previous test: [dipl_test_0310.zip](https://github.com/widelands/widelands/files/10947335/dipl_test_0310.zip) along with the replay: [dipl_test_0310_replay.zip](https://github.com/widelands/widelands/files/10947336/dipl_test_0310_replay.zip)

Sorry, no analysis yet. I've been busy reading gnuplot docs and figuring out some of its intricacies. :)

hessenfarmer:
> > Reagrding the team wins, I am not 100% content with the result yet but at least I had some matches where one player won an autocrat game. maybe I should increase the malus if the team is big.
> 
> Actually, why not check directly whether joining player is last one outside of team when the request is decided?

that would be good, however currently I don't have an idea yet how to do that. I could check whether team_members is nr_players -1 but I believe nr_players is not considering players that have lost. need to investigate

tothxa:
> > Actually, why not check directly whether joining player is last one outside of team when the request is decided?
> 
> that would be good, however currently I don't have an idea yet how to do that. I could check whether team_members is nr_players -1 but I believe nr_players is not considering players that have lost. need to investigate

I thought you could just use `team_members`. Now, as you say that defeated players may be included, actually that's a bug. So all iterations over players should skip defeated players, and then `team_members` could also be trusted.

hessenfarmer:
> > > Actually, why not check directly whether joining player is last one outside of team when the request is decided?
> > 
> > 
> > that would be good, however currently I don't have an idea yet how to do that. I could check whether team_members is nr_players -1 but I believe nr_players is not considering players that have lost. need to investigate
> 
> I thought you could just use `team_members`. Now, as you say that defeated players may be included, actually that's a bug. So all iterations over players should skip defeated players, and then `team_members` could also be trusted.        
     
No team members only includes the members of a team and with my latest commit defeated players quit their team. the problem is that in a 4 player game 3 team members would be allowed. but id one is defeated this needs to be reduced to 2. if nr_players would only count active players (exclude defeated players) this would be easy as just limit team_members to (nr_players - 1) however I doubt nr_players is implemented this way.

tothxa:
> > I thought you could just use `team_members`. Now, as you say that defeated players may be included, actually that's a bug. So all iterations over players should skip defeated players, and then `team_members` could also be trusted.
> 
> No team members only includes the members of a team and with my latest commit defeated players quit their team. the problem is that in a 4 player game 3 team members would be allowed. but id one is defeated this needs to be reduced to 2. if nr_players would only count active players (exclude defeated players) this would be easy as just limit team_members to (nr_players - 1) however I doubt nr_players is implemented this way.

But I think you can check the `size()` of `team_members` to see how many teams are there. Teamless players need to be counted too, and I don't think they are currently, so that needs to be done. Either in `team_members`, but then all uses need to special case team 0, or in a separate variable?

Or you could keep track of the number of active players in a separate variable, also updated together with the stats.

edit: Basically do what the helper functions for the autocrat win condition do in lua.

src/ai/defaultai.cc
```diff
@@ -3330,18 +3330,30 @@ void DefaultAI::check_flag_distances(const Time& gametime) {
...
+	}
+
+	const Widelands::PlayerNumber mypn = player_number();
+	const Widelands::Player* me = game().get_player(mypn);
+
+	// if we are defeated or the last one in a team leave team
+	if (me->team_number() != 0 &&
+	    (player_statistics.members_in_team(me->team_number()) == 1 || me->is_defeated())) {
+		game().send_player_diplomacy(mypn, Widelands::DiplomacyAction::kLeaveTeam, 0 /* ignored */);
+		verb_log_dbg_time(gametime, "Player(%d), leaves team (%d) as last one.\n",
```
tothxa: Please include something to make it easy to filter the log in one pass. E.g.

```suggestion
		verb_log_dbg_time(gametime, "AI diplomacy: Player(%d), leaves team (%d) as last one.\n",
```

hessenfarmer: done

src/ai/defaultai.cc
```diff
@@ -3330,18 +3330,34 @@ void DefaultAI::check_flag_distances(const Time& gametime) {
...
-			             (player_statistics.get_player_land(player_number()) / 100);
+		if (pda.other == mypn) {
+			// accept if diploscore high, else accept only 50%
+			bool accept =
+			   player_statistics.get_diplo_score(pda.sender) >= 30 ||
+			   (player_statistics.get_diplo_score(pda.sender) > 20 && RNG::static_rand(2) == 0);
+			accept = accept &&
+			         player_statistics.members_in_team(pda.sender) <
+			            player_statistics.players_active() - 1 &&
+			         player_statistics.members_in_team(mypn) < player_statistics.players_active() - 1;
```
tothxa: Only the team to be joined needs to be checked. Why prevent leaving the big team? ;)
I hope I didn't mix up the directions.

```suggestion
			accept = accept &&
			         player_statistics.members_in_team(
			            pda.action == Widelands::DiplomacyAction::kInvite ? pda.sender : mypn) <
			         player_statistics.players_active() - 1;
```

src/ai/ai_help_structs.cc
```diff
@@ -1041,15 +1053,22 @@ void PlayersStrengths::remove_stat(const Widelands::PlayerNumber pn) {
...
 				team_powers[item.second.team_number] += item.second.players_power;
 			} else {
 				team_powers[item.second.team_number] = item.second.players_power;
 			}
+			++team_members[item.second.team_number];
+		}
+		if (!item.second.defeated) {
+			++active_players_;
 		}
 	}
```
tothxa:
```suggestion
	for (auto& item : all_stats) {
		if (item.second.defeated) {
			continue;
		}
		++active_players_;
		if (item.second.team_number > 0) {  // is a member of a team
			if (team_powers.count(item.second.team_number) > 0) {
				team_powers[item.second.team_number] += item.second.players_power;
			} else {
				team_powers[item.second.team_number] = item.second.players_power;
			}
			++team_members[item.second.team_number];
		}
	}
```

tothxa: This is actually just moving the check for defeated to the beginning of the loop and skip everything for defeated players. I think github shows it as a big diff because previously unchanged lines are affected too.

src/ai/defaultai.cc
```diff
@@ -3330,18 +3330,34 @@ void DefaultAI::check_flag_distances(const Time& gametime) {
...
+	if (gametime < Time((10 + RNG::static_rand(10)) * 60 * 1000)) {
+		return;
+	}
+
+	const Widelands::PlayerNumber mypn = player_number();
+	const Widelands::Player* me = game().get_player(mypn);
+
+	// if we are defeated or the last one in a team leave team
+	if (me->team_number() != 0 &&
+	    (player_statistics.members_in_team(me->team_number()) == 1 || me->is_defeated())) {
```
tothxa: The `me->is_defeated()` check should be moved to as early as possible, and if defeated, then only team membership should be checked and cancelled if needed, and everything else could be skipped.

src/ai/defaultai.cc
```diff
@@ -6754,9 +6851,208 @@ void DefaultAI::update_player_stat(const Time& gametime) {
 					}
 				}
 
-				player_statistics.add(pn, j, me->team_number(), this_player->team_number(),
-				                      cur_strength, old_strength, old60_strength, cass, cur_land,
-				                      old_land, old60_land);
+				// determine the diplomacy score of each player
+				int32_t diplo_score = 0;
+				// we need to be sure all magic numbers have been initialized so wait 30s
+				if (gametime > Time(30000)) {
```
tothxa: I think the diploscore calculation should be skipped for defeated players (if either the calculating or the calculated player is defeated) and probably for the current player too, because I don't think the score for self is used for anything, nor that it would be needed for team calculations. And it doesn't look reliable either.

src/ai/defaultai.cc
```diff
@@ -3330,18 +3330,34 @@ void DefaultAI::check_flag_distances(const Time& gametime) {
 // Dealing with diplomacy actions
 void DefaultAI::diplomacy_actions(const Time& gametime) {
 
+	// don't do any diplomacy for the first 10 + x minutes to avoid click races for allies
+	if (gametime < Time((10 + RNG::static_rand(10)) * 60 * 1000)) {
+		return;
+	}
```
tothxa: Shouldn't the initial delay be moved to the AI init function where diplomacy actions are first scheduled instead of checking every time?

src/ai/defaultai.cc
hessenfarmer: In this section it is only about accepting, so no leave is prevented. However your solution looks nicer so I'll take it

src/ai/defaultai.cc
hessenfarmer: ok good idea

src/ai/defaultai.cc
hessenfarmer: good idea

src/ai/defaultai.cc
hessenfarmer: well in this case we would need to check for open requests as well and this would mean duplicating most of the code that is not already safeguarded by me->is_defeated() already so I see no benefit here

tothxa: OK.

src/ai/defaultai.cc
```diff
@@ -3330,18 +3331,31 @@ void DefaultAI::check_flag_distances(const Time& gametime) {
 // Dealing with diplomacy actions
 void DefaultAI::diplomacy_actions(const Time& gametime) {
 
+	const Widelands::PlayerNumber mypn = player_number();
+	const Widelands::Player* me = game().get_player(mypn);
+
+	// if we are defeated or the last one in a team leave team
+	if (me->team_number() != 0 &&
+	    (player_statistics.members_in_team(me->team_number()) == 1 || me->is_defeated())) {
```
tothxa: I've just checked that `is_defeated()` is expensive, so all checks should use the cached value from`player_statistics`, or store it at the first use for the rest of the function body.

hessenfarmer: done

src/ai/defaultai.cc
hessenfarmer: done

tothxa:
I've ran another test with [9e62c7a](https://github.com/widelands/widelands/pull/5774/commits/9e62c7aaca41e8db5db80c9b0791d048dcead863). Here are the logs, graphs and the replay:
[ai_dipl_test_0312.zip](https://github.com/widelands/widelands/files/10952185/ai_dipl_test_0312.zip)
[ai_dipl_test_0312_replay.zip](https://github.com/widelands/widelands/files/10952186/ai_dipl_test_0312_replay.zip)

This time the main problem was that when yellow overpowered the others, and defeated red, blue and green failed to team up. When red was about to be beaten, it was allied with blue. Blue invited green twice, it accepted both times, but soon left because of red.
Then when red is defeated, yellow and blue team up for a short time, and green has the bad luck of requesting to join blue during this period, so it is rejected because of the team win rule, even though it has better score for blue than yellow.
So it has to wait a long time before it can request blue again, and by that time its score for blue is barely enough to be considered (24), and it gets denied. From then on, green is weakened by both blue and yellow, and its repeated requests to blue all get rejected. There's extreme bad luck involved here, because all this time its score for blue remains 24-26, and it makes 4 requests.
So green is defeated in the end, and blue remains alone against a much stronger yellow.

The main take-away from this is that **I suggest this tweak** to the decision code: When all players are teamed up against one, and that one makes a request, it should be an invite, not a request to join.
[edit:] *Also, members of the big team shouldn't send invites to the lone player that it can't accept. Either just skip it, or if not too complicated, evaluate whether it would be better teaming up with the lone player than to be in the current team.*

I've noticed one more strange thing with scoring in the graphs: Red's score for yellow rises in the end as it's getting beaten. It even has 2 peaks above 40. This didn't result in invitation, but it could have. The reason for this rise may have been that green was yellow's other neighbour, and it still had a good military at this time. (but no match for yellow) … So maybe it's not so strange after all.

hessenfarmer:
> The main take-away from this is that **I suggest this tweak** to the decision code: When all players are teamed up against one, and that one makes a request, it should be an invite, not a request to join. [edit:] _Also, members of the big team shouldn't send invites to the lone player that it can't accept. Either just skip it, or if not too complicated, evaluate whether it would be better teaming up with the lone player than to be in the current team._

made it an automatism so far. May change it, however we won't able to cope with every possible circumstances. In the end this will be a hopefully very educated fake of some strategie

tothxa:
I forked the previous test with the latest changes at the point where red's decline begins. This time green got a little more beating from yellow before red is defeated, but the power balance remains similar enough to make the comparison useful. 
[ai_dipl_test_0313.zip](https://github.com/widelands/widelands/files/10958039/ai_dipl_test_0313.zip)
[ai_dipl_test_0313_replay.zip](https://github.com/widelands/widelands/files/10958040/ai_dipl_test_0313_replay.zip)

Now blue and green team up nicely at around 2:48, and they manage to stop yellow's growth, until blue leaves the team. From then on both of them suffer. From time to time, green invites blue, and it always accepts, but then leaves soon. It's the flip-flopping issue again:
1vs4
dipl_actions

Otherwise scores mostly tend to be quite stable. The big disruption is when players start to meet each other, which always shuffles the line-ups a lot, but that's actually quite natural. Other than that, the scores involving the strongest-by-far player yellow are very unstable even during periods without line-up changes (also others' scores for it, which are not shown here, but are in the attached zip):

hessenfarmer:
Thanks for the analysis. I believe this might be due to the scoring on teammembers. Will analyse and fix.

src/ai/defaultai.cc
```diff
@@ -1147,9 +1147,10 @@ void DefaultAI::late_initialization() {
 	                                                   SchedulerTaskId::kWarehouseFlagDist, 5,
 	                                                   "flag warehouse Update"));
 
-	taskPool.push_back(std::make_shared<SchedulerTask>(std::max<Time>(gametime, Time(40 * 1000)),
-	                                                   SchedulerTaskId::kDiplomacy, 7,
-	                                                   "diplomacy actions"));
+	// don't do any diplomacy for the first 10 + x minutes to avoid click races for allies
+	taskPool.push_back(std::make_shared<SchedulerTask>(
+	   std::max<Time>(gametime, Time((10 + RNG::static_rand(10)) * 60 * 1000)),
+	   SchedulerTaskId::kDiplomacy, 7, "diplomacy actions"));
```
tothxa:
```suggestion
	if (game().diplomacy_allowed()) {
		// don't do any diplomacy for the first 10 + x minutes to avoid click races for allies,
		// and allow statistics to settle after loading game
		taskPool.push_back(std::make_shared<SchedulerTask>(
		   std::max<Time>(
		      gametime + kStatUpdateInterval * 5, Time((10 + RNG::static_rand(10)) * 60 * 1000)),
		   SchedulerTaskId::kDiplomacy, 7, "diplomacy actions"));
	}
```

src/ai/defaultai.cc
```diff
@@ -3350,6 +3365,61 @@ void DefaultAI::diplomacy_actions(const Time& gametime) {
...
+
+	for (Widelands::PlayerNumber opn = 1; opn <= game().map().get_nrplayers(); ++opn) {
+		const Widelands::Player* other_player = game().get_player(opn);
+		// other player needs to exist and different from us
+		// we need to be still alive and we don't have send a request in last 5 + X minutes
+		if (other_player == nullptr || opn == mypn ||
+		    player_statistics.player_diplo_requested_lately(opn, gametime)) {
+			continue;
+		}
+		if (player_statistics.get_diplo_score(opn) >= 40) {
```
tothxa:
```suggestion
		if (player_statistics.get_diplo_score(opn) >= 40) &&
		    me->team_number() != other_player->team_number()) {
```

src/ai/defaultai.cc
```diff
@@ -6754,9 +6864,209 @@ void DefaultAI::update_player_stat(const Time& gametime) {
 					}
 				}
 
+				// determine the diplomacy score of each player
+				int32_t diplo_score = 0;
+				// we need to be sure all magic numbers have been initialized so wait 30s
+				if (gametime > Time(30000) && pn != j) {
```
tothxa:
```suggestion
				if (game().diplomacy_allowed() && gametime > Time(30000) && pn != j) {
```

src/ai/defaultai.cc
```diff
@@ -6754,9 +6864,209 @@ void DefaultAI::update_player_stat(const Time& gametime) {
...
+
+					for (uint8_t i = 0; i < kFNeuronBitSize; ++i) {
+						if (management_data.f_neuron_pool[28].get_position(i)) {
+							diplo_score += inputs[i];
+						}
+						if (management_data.f_neuron_pool[29].get_position(i)) {
+							diplo_score += inputs[i + kFNeuronBitSize];
+						}
+					}
+				}
```
tothxa, line 7261:
```suggestion
					verb_log_dbg_time(
					   gametime, "AI Diplomacy: For player(%d), the player(%d) has the diploscore: %d\n",
					   static_cast<unsigned int>(pn), static_cast<unsigned int>(j), diplo_score);
				}
```

src/ai/defaultai.cc
```diff
@@ -6754,9 +6864,209 @@ void DefaultAI::update_player_stat(const Time& gametime) {
...
+
 				player_statistics.add(pn, j, me->team_number(), this_player->team_number(),
 				                      cur_strength, old_strength, old60_strength, cass, cur_land,
-				                      old_land, old60_land);
+				                      old_land, old60_land, diplo_score, buildings, player_def);
+				verb_log_dbg_time(
+				   gametime, "AI Diplomacy: For player(%d), the player(%d) has the diploscore: %d\n",
+				   static_cast<unsigned int>(pn), static_cast<unsigned int>(j), diplo_score);
```
tothxa:
```suggestion
```

tothxa (commented):
Some more optimisation when `!diplomacy_allowed()`, fix players leaving teams on game loading, and fix regression of team members inviting each other.

src/ai/defaultai.cc
hessenfarmer: this change would prevent 2 players in team 0 tom team. Did you see a lot of regressions of players inviting each other. 
I intended to prevent this by requesting different team powers for the invitation. If this happens we need some logging because that means team powers are calculated differently.

tothxa: Oops, indeed. We've been through these, and I also remember thinking the same about comparing team powers. :)

But yes, it happened a lot in the [previous test](https://github.com/widelands/widelands/files/10958040/ai_dipl_test_0313_replay.zip). E.g. see the graph above at around 2:25, but there were more, but those are harder to identify by the graph.

hessenfarmer: ok I will ad some logging and run a test,

src/ai/defaultai.cc
```diff
@@ -3350,6 +3369,61 @@ void DefaultAI::diplomacy_actions(const Time& gametime) {
...
+				verb_log_dbg_time(
+				   gametime,
+				   "AI Diplomacy: Player(%d), requests to join player (%d) with diploscore: %d\n",
+				   static_cast<unsigned int>(mypn), static_cast<unsigned int>(opn),
+				   player_statistics.get_diplo_score(opn));
+			}
+		} else if (player_statistics.get_diplo_score(opn) < -15 &&
+		           other_player->team_number() == me->team_number() &&
+		           other_player->team_number() != 0) {
+			game().send_player_diplomacy(mypn, Widelands::DiplomacyAction::kLeaveTeam, opn);
```
tothxa: Could we loop here through all other team members and reset `last_time_requested` to allow inviting the good ones over into a new team quickly?

src/ai/defaultai.cc
tothxa: BTW instead of military power only, we should sum or average the diploscores of unaffected members of both teams, and decide which team is better comparing that.

tothxa: BTW instead of military power only, we should sum or average the diploscores of the unaffected members of both teams, and compare that. Or even if both are negative, then leave current team and send invite in the same step.

src/ai/defaultai.cc
hessenfarmer: that might result in a lot of requests to human players from computerplayers that just left the team.   
Remember that one goal is not to spam human players with requests.

src/ai/defaultai.cc
hessenfarmer: We could do that but I think this might be a lot of effort just to distinguish between invite and join.

src/ai/defaultai.cc
tothxa: I don't think that would happen all that often. It's not very likely that you'd often end up in a team with both strong and weak players. But yes, this would pretty much require the diploscore based invite/join distinction to prevent flip-flopping.

It's actually a shortcoming of the current diplomacy possibilities that it's not possible to kick out a weak player from a strong team, so this workaround is needed to achieve it. (it would be a more complicated UI though – I almost created a feature request for such a thing yesterday, but then thought better of it)

src/ai/defaultai.cc
tothxa: I think (based on the test games too) that it's a very important distinction. :)

src/ai/defaultai.cc
```diff
@@ -6754,9 +6873,209 @@ void DefaultAI::update_player_stat(const Time& gametime) {
...
+					                   !player_statistics.strong_enough(pn) ?
+                               5 :
+                               0;
+					inputs[39] = player_statistics.strong_enough(pn) ? -10 : 2;
+					inputs[40] = player_statistics.strong_enough(pn) ? -5 : 2;
+					inputs[41] = player_statistics.get_is_enemy(j) &&
+					                   player_statistics.player_seen_lately(j, gametime) &&
+					                   player_statistics.enemies_seen_lately_count(gametime) > 1 ?
+                               5 :
+                               0;
```
tothxa: These checks with `get_is_enemy()` are major contributors to the swings on join/leave.
edit: oops, `inputs[42]` was left out

edit2: Maybe remove `get_is_enemy()` from one of the pair? Or make it a multiplier instead of a condition?
Like
```
inputs[37] = (player_statistics.get_is_enemy(j) ? 2 : 1) *
                   (player_statistics.player_seen_lately(j, gametime) &&
                   !player_statistics.strong_enough(pn) ?
                               4 :
                               -1);
```

src/ai/defaultai.cc
hessenfarmer: ok so I'll exchange military team power with team diploscores correct?

src/ai/defaultai.cc
hessenfarmer: if we get the flip flopping under control, we should have teams that last longer then diplo_requested interval. So I still beleive that this isn't worth it. the last player that requested a diplo action has the longest interval and if this player decides to leave I think he should take some time to make up his mind. The players in the team that only accepted are not affected at all.

src/ai/defaultai.cc
hessenfarmer: Ok I will reduce the gap between the 2 values as suggested. However please keep in mind, that not all the inputs are evaluated. Only the inputs that are chosen to be valuable based on AI training values.      
So this whole thing is subject to AI Training as well.

tothxa: Actually my suggestion now is edit2 above. Keep the bigger gap, because it's an important factor, but the part whether we're allied is less important for this factor.

Unfortunately I don't know how the AI training works. How do you set a target function for a thing as complex as diplomacy?

src/ai/defaultai.cc
tothxa: Well, this is indeed less critical. The other half of the solution is improving invite/join, so that players don't have to join teams with players they don't like in order to get allied with those who they like.

src/ai/defaultai.cc
tothxa: Leaving out the player making the request and the player who is requested, because they will be both in the selected team if all works out. So they have to evaluate the other members of both teams to see which ones they're better off with. I think I'll write an invite or join function proposal, because I have several things in mind how to handle various cases.

src/ai/defaultai.cc
hessenfarmer: Well, usually training has been done by @tibor95. I think we can't evaluate diplomacy directly but might rely on the outcome (i.e. winning the game).

src/ai/defaultai.cc
hessenfarmer: Basically I thought about just summing the diplo_scores in recalculate team power as it is least costly.

tothxa:
I think I've found the root cause of the wild fluctuations of yellow's scores in the test game. It turns out that `input[58]` (along with `input[31]`) is the most frequent variable involved in sudden changes. It's a simple check, that shouldn't change at all, as it compares both players supposedly to the biggest military strength, and yellow is firmly the strongest, even though it's strength may have fluctuations.

So I believe there's a time-of-check/time-of-use problem here, which is shared by many other inputs: `get_max_power()` uses the values stored by the previous run, but `me_strength` and `cur_strength` use the new values to be stored.

Even if the ToC/ToU problem is handled, I've already wondered whether some of the inputs that are based on comparisons would be better off if they had a soft(er) transition, instead of binary true/false.

hessenfarmer:
> I think I've found the root cause of the wild fluctuations of yellow's scores in the test game. It turns out that `input[58]` (along with `input[31]`) is the most frequent variable involved in sudden changes. It's a simple check, that shouldn't change at all, as it compares both players supposedly to the biggest military strength, and yellow is firmly the strongest, even though it's strength may have fluctuations.
> 
> So I believe there's a time-of-check/time-of-use problem here, which is shared by many other inputs: `get_max_power()` uses the values stored by the previous run, but `me_strength` and `cur_strength` use the new values to be stored.
> 
> Even if the ToC/ToU problem is handled, I've already wondered whether some of the inputs that are based on comparisons would be better off if they had a soft(er) transition, instead of binary true/false.

You are right if a new soldier is trained or recruited then actual me strength (for p3) or other strength (for the others) is greater then old max strength of the strongest player. 
solution would be to use the old strength as well using the get_player_power() funtion. will do that. Thanks for finding.

src/ai/defaultai.cc
hessenfarmer: made the enemy status a factor

tothxa:
I tested the latest improvements with my decision branch, and the behaviour is indeed much more stable now. (makes it harder to test ;) The strongest player's scores still fluctuate, but much less than before. I'll investigate it tomorrow.

Accepting requests still needs the same group evaluation as invite/join in my branch (weak player was able to get itself on strong team because its score for one player wasn't entirely bad, just to force disbanding and recreating the team again), but I'd rather do it separately.

tothxa:
I've found some more remaining toc/tou: inputs 59 to 63 still compare `buildings` with `get_max_buildings()`.

frankystone:
Will this be an option which can be enabled optionally?

In the long run there might be a need of something like a `single player won` counter. If the counter is low some features (like ai diplomacy) are disabled, if the counter increases some features get automatically enabled (ai diplomacy, Trading Outpost for the ai opponents) . Beginners, who are likely struggling with the game handling, should not be forced to  struggle with a too strong ai imho.

tothxa:
> Will this be an option which can be enabled optionally?

I think it should be possible to disable Diplomacy in general when starting a game. But in a separate PR please. :)

> Beginners, who are likely struggling with the game handling, should not be forced to struggle with a too strong ai imho.

Actually this is a bigger problem for experienced players who want to play a challenge against many AIs, because now the AIs will happily turn against each other, in spite of you putting them in the same team initially. Actually they even tend to leave the initial teams, because the first time diplomacy actions are run they still have low scores for each other.

@hessenfarmer Maybe we could consider lengthening the initial diplomacy delay. It would be best if we could consider map size vs. number of players. There's already something similar in `compute_max_players()` in `editor/ui_menus/main_menu_random_map.cc`.

hessenfarmer:
definitly we will need a game option to turn diplomacy of, for the reasons provided above.  An unexperienced player might be confused by all the calls and might find himself in a game against all AI. So I will evaluate whether it will be in  this PR or in a seperate one.

tothxa:
> definitly we will need a game option to turn diplomacy of, for the reasons provided above. An unexperienced player might be confused by all the calls and might find himself in a game against all AI. So I will evaluate whether it will be in this PR or in a seperate one.

I suggested separate PR because it's all just UI code, so it touches entirely different source files.

src/ai/defaultai.cc
```diff
@@ -3330,26 +3338,239 @@ void DefaultAI::check_flag_distances(const Time& gametime) {
...
+
+			default:
+				NEVER_HERE();
+			}
+		}
+		game().send_player_diplomacy(mypn, planned_action, planned_opn);
+		player_statistics.recalculate_team_power();
+
+		// Team change invalidates statistics
+		return;
```
tothxa: `recalculate_team_power()` is useless here, it's effectively a no-op that just wastes some cycles on iterating over the old statistics again, and the (same as before) result will never be used (at least by the diplomacy code), because by the next time this function is called it will have been updated several times over.

tothxa (commented):
I don't think this needs to be further solved here. We could revisit the scheduling in a follow up PR instead. This is a big change already, that is now ready to merge.

If you want to still decrease the probability of a possible race, then these functions should be scheduled some fractions of a second apart at the end of the update stats scheduler task. The check site functions would run each time, while diplo actions should get a new variable `diplomacy_actions_next_due_`, so that it would only run `if (gametime > diplomacy_actions_next_due_)`. That way both the initial delay could be easily set and the newly introduced jitter could stay.

src/ai/defaultai.cc
hessenfarmer: oops you are right will undo this

tothxa:
To sum up my above outstanding comments:

From my POW, there's only one change to be considered: the initial delay before the AI starts diplomacy should be increased.

Related issues to be solved by other PRs are adding the diplomacy allowed option to the game start menus (this in itself shouldn't be a big deal, but those windows need a general overhaul, and they already don't fit in the lowest allowed resolution / window size #5631), and cleaning up diplo requests on team change (related to #5776). And possibly for scheduling: maybe we could try running the check site functions less frequently too.

hessenfarmer:
> To sum up my above outstanding comments:
> 
> From my POW, there's only one change to be considered: the initial delay before the AI starts diplomacy should be increased. 
    
I think it is already quite long in small maps, and adopting to map size may be part of another PR.   

> 
> Related issues to be solved by other PRs are adding the diplomacy allowed option to the game start menus (this in itself shouldn't be a big deal, but those windows need a general overhaul, and they already don't fit in the lowest allowed resolution / window size #5631), and cleaning up diplo requests on team change (related to #5776). And possibly for scheduling: maybe we could try running the check site functions less frequently too.    
   
Agreed

hessenfarmer:
As we both contributed here, do we need a third person as reviewer?    
@Noordfrees functionally @tothxa and me are happy for the moment, could you do an independent code review?

tothxa:
> > To sum up my above outstanding comments:
> > From my POW, there's only one change to be considered: the initial delay before the AI starts diplomacy should be increased.
> 
> I think it is already quite long in small maps, and adopting to map size may be part of another PR.

OK.

For the record, I have a new idea for this: Make it somewhere between 45 and 60 minutes (based on observing how long it takes for the AIs to build up a reasonable economy), but have a way to start scheduling it earlier if an enemy is seen or there are pending diplo actions (presumably from human players). After all this is what matters, and anything else is just an approximation for this.

Noordfrees (approved):
Changes look safe, but I'm not that deeply familiar with the AI code. Not tested. If you are both satisfied that it works well feel free to merge. And thanks for this big feature :)

hessenfarmer:
@Noordfrees thanks for the review

PR #5781, opened 2023-02-14
Carry over ship names in Atlantean campaign

### LINKED ISSUE #5777: Atl01→02 ship names should be carried over
**Describe the bug**
The starting ship names of atl02 are random, but the ships at the end of atl01 have names, so they should be carried over.

**To reproduce**
1. Play atl01 to end
2. Note the names of the built ships
3. Start atl02
4. See ship names

**Expected behavior**
Ship names should be carried over from atl01 (2 of the 3 are required) as campaign data. (save all 3 and drop random one on atl02 start? ;)

**Version:**
 - OS: endless (flatpak)
 - Widelands Version: v1.1

tothxa:
**Type of change**
Bugfix

**Issues closed**
Fixes #5777

Also remove manually set ship- and warehouse names from the list of remaining names if present

**New behavior**
The names of the ships that are required by the win condition of atl01 are saved. At the start of atl02 one name is dropped randomly (the ship lost in the storm), and the remaining 2 names are assigned to the starting ships.

Such carried over names are now removed from the list of the available names to avoid duplicate names in case the player builds many ships. The same is done for warehouse names, although currently no campaign makes similar use of randomly assigned warehouse names.

**Possible regressions**
Things related to ship and warehouse name assignment?


hessenfarmer (approved):
Code LGTM.
Tested and working.

PR #5784, opened 2023-02-16
Work around glewInit crash on wayland

### LINKED ISSUE #4958: Crash when using SDL's Wayland backend
**Describe the bug**
The game refuses to start when using the Wayland backend of SDL.

**To Reproduce**
```
$ SDL_VIDEODRIVER=wayland widelands
[00:00:00.000 real] INFO: Set home directory: /home/kitchu/.local/share/widelands
[00:00:00.000 real] INFO: Set configuration file: /home/kitchu/.config/widelands/config
[00:00:00.000 real] INFO: Adding directory: /usr/share/widelands
[00:00:00.000 real] INFO: selected language: (system language)
[00:00:00.000 real] INFO: using locale en_IN.UTF-8
[00:00:00.002 real] INFO: Byte order: little-endian
[00:00:00.068 real] INFO: Graphics: Try to set Videomode 800x600
[00:00:00.075 real] ERROR: glewInit returns 4
[00:00:00.075 real] ERROR: Your OpenGL installation must be __very__ broken. Unknown error

Caught exception (of type '10WException') in outermost handler!
The exception said: [../src/graphic/gl/initialize.cc:141] glewInit returns 4: Broken OpenGL installation.

This should not happen. Please file a bug report on version 1.0(Release).
```

**Version:**
 - OS: Archlinux
 - Widelands Version: 1.0 (Release)

**Additional context**
I'm on a Wayland session of KDE's Plasma desktop.

Adriankhl:
**Type of change**
Fix glewInit crash on wayland. This is an issue of glew, see https://github.com/nigels-com/glew/issues/172

**Issue(s) closed**
Fixes https://github.com/widelands/widelands/issues/4958

**To reproduce**
1. Run widelands with `SDL_VIDEODRIVER=wayland widelands`
2. Crash

**New behavior**
Now it works.

**How it works**
Honestly I don't really know :) This is how other people fixed it for https://github.com/nigels-com/glew/issues/172


src/graphic/gl/initialize.cc
```diff
@@ -133,7 +133,9 @@ SDL_GLContext initialize(
 	// LeakSanitizer reports a memory leak which is triggered somewhere above this line, probably
 	// coming from the gaphics drivers
 
-	if (err != GLEW_OK) {
+	// err != GLEW_ERROR_NO_GLX_DISPLAY is a workaround for crash on wayland
+	// https://github.com/nigels-com/glew/issues/172
+	if (err != GLEW_OK && err != GLEW_ERROR_NO_GLX_DISPLAY) {
```
Noordfrees:
```suggestion
	if (err == GLEW_ERROR_NO_GLX_DISPLAY) {
		log_info("Using glewInit workaround for Wayland");
	} else if (err != GLEW_OK) {
```

Noordfrees (commented):
Thanks for the patch :) 

Please print a log message when the workaround is active as proposed inline.

Could someone with Wayland please test this?

src/graphic/gl/initialize.cc
Adriankhl: Ok

src/graphic/gl/initialize.cc
```diff
@@ -133,7 +133,11 @@ SDL_GLContext initialize(
 	// LeakSanitizer reports a memory leak which is triggered somewhere above this line, probably
 	// coming from the gaphics drivers
 
-	if (err != GLEW_OK) {
+	// err == GLEW_ERROR_NO_GLX_DISPLAY is a workaround for crash on wayland
+	// https://github.com/nigels-com/glew/issues/172
+	if (err == GLEW_ERROR_NO_GLX_DISPLAY) {
```
tothxa: SDL is already initialised here, so we can check whether we're actually using wayland: (not tested)

```suggestion
	if (err == GLEW_ERROR_NO_GLX_DISPLAY && std::strcmp(SDL_GetCurrentVideoDriver(), "wayland") == 0) {
```

tothxa (approved):
I don't think we'll get more testing for this, and it shouldn't affect other use cases, so we might as well merge it?

Issue #4689, opened 2021-02-16
Watch window jumps to center of screen


frankystone:
Related: #1232

If you open a watch window to watch an enemies military site and click on the military site inside the watch window, the watch windows jumps to the center of the screen.

**To Reproduce**
1. Open the watch window and move the content to see a foreign military building
2. Move the watch window to e.g. the upper left corner
3. Eventually wait until soldiers arrive at the watched military site
4. Click the military inside the watch window

**Expected behavior**
The watch window should stay where it is.

**Version:**
 - OS: archlinux
 - Widelands Version: [9d5409e@master] (Debug)

Noordfrees:
This bug has been around for some time

PR #5785, opened 2023-02-16
Fix collapsing input queues and other window handling fixes and enhancements

### LINKED ISSUE #5605: Ware priority section get widened on start stop
I guess once a player has a specific situation reached during a game, he would likely have the window of the Training camp constantly open to start/stop the training camp depending on the stored wares. But this window covers some of the maps area, especially if the priority sliders are visible. Fortunately those sliders can be hidden so the window is much smaller and does not cover so many parts of the map. But unfortunately if one starts/stop the training camp the priority slider section of the window get widened again.

On big screens one may do not use the possibility to hide the priority sliders but since i am on my laptop with a screen resolution of 1280x800 i use this feature a lot to see more of the map. But the current behavior of widen the window on start/stopping the training camp becomes really annoying. After every click on stop and every click on start i have to make an additionally click to hide the sliders again. Since i have this window in the top right corner the window will move to the left if the priority sliders get widened (start/stop). So i additionally i have to move the window to the right again.
The reason to have this window on the top right might be that the inventory window is opened at the top left of the screen.

Once the sliders got hidden they should only be widened again if the corresponding button is clicked.

Maybe there could also be a configuration option which applies to all buildings which have priority sliders. Something like: "Hide priority sliders on startup"?

tothxa:
**Type of change**
New feature

**Issue(s) closed**
Fixes #5605

Also fixes remembering the position of UniqueWindows (e.g. statistics and editor tools).

**To reproduce**
*If it's a bugfix:*

Production sites:

1. Open a productionsite window
2. Collapse the inputqueues
3. Click stop

Other windows:

1. Open the general statistics window
2. Move it from its default position
3. Toggle it off from the Statistics menu or with its hotkey (default `A`)
4. Reopen it → it opens in the default position again
5. Move it again
6. Close it by right-clicking or with the close button
7. Reopen it →it opens in the position where it was closed – I believe this is the intended behaviour

**New behavior**
The collapsed state of inputqueue displays is stored in the building window's registry to remember it when the window is closed. (stopping/restarting practically closes and reopens the window) The state is also saved with the game, ~but currently it only works~ for open windows ~– that's why this is a draft.~ (like other window types)

As a side effect, construction sites now have the same collapsed state for both the building materials and the settings for the future production site. I think this actually makes sense.

Initial positioning of building (and ship and field action) windows is changed so that the window always opens near the clicked position, but if possible, with a clearance to allow seeing the surroundings.

**Possible regressions**
Handling, display and saving/loading of building and other windows.


src/wui/buildingwindow.cc
```diff
@@ -715,17 +716,20 @@ void BuildingWindow::update_expedition_button(bool expedition_was_canceled) {
...
 		const uint16_t packet_version = fr.unsigned_16();
-		if (packet_version == kCurrentPacketVersion) {
+		if (packet_version >= 1 && packet_version <= kCurrentPacketVersion) {
 			const int32_t x = fr.signed_32();
 			const int32_t y = fr.signed_32();
 			const bool workarea = fr.unsigned_8() != 0u;
 			BuildingWindow& bw = dynamic_cast<BuildingWindow&>(
 			   *ib.show_building_window(Widelands::Coords(x, y), true, workarea));
 			bw.tabs_->activate(fr.unsigned_8());
+			if (packet_version >= 2) {
```
Noordfrees, line 732: Needs a "TODO savegame compat v1.2" comment

src/wui/buildingwindow.h
```diff
@@ -145,6 +155,7 @@ struct BuildingWindow : public UI::UniqueWindow {
 
 	bool showing_workarea_;
 	bool avoid_fastclick_;
+	bool* priority_collapsed_;
```
Noordfrees:
```suggestion
	bool* priority_collapsed_;  ///< Owned by the BuildingWindow::Registry
```

src/wui/inputqueuedisplay.h
```diff
@@ -115,7 +117,8 @@ class InputQueueDisplay : public UI::Box {
 	                  Widelands::InputQueue*,
 	                  Widelands::ProductionsiteSettings*,
 	                  bool,
-	                  bool);
+	                  bool,
+	                  bool*);
```
Noordfrees: When you're at it could you please add the parameter names here

src/wui/inputqueuedisplay.h
```diff
@@ -150,7 +153,7 @@ class InputQueueDisplay : public UI::Box {
 	UI::Panel spacer_;
 	const Widelands::WarePriority* slider_was_moved_;
 
-	bool collapsed_;
+	bool* collapsed_;
```
Noordfrees:
```suggestion
	bool* collapsed_;  ///< Owned by the window creating the input queue
```

Noordfrees (commented):
Very elegant solution :) 
Needs a few documentary comments.

One thing I'm worried about is that it's all too easy to confuse `bool*` with `bool` and treat the pointer like the underlying boolean. Maybe replacing `bool` with an `enum class CollapsedState { kExpanded, kCollapsed }` would be safer?

Not tested yet.

tothxa:
> Very elegant solution :)

Thank you. :) Actually I didn't have much choice though. :) The problem turned out to be much harder than I expected, and I started out with more complicated code in `BuildingWindow`, but as I went on to `InputQueueDisplay` it just fell into place in this form.

tothxa:
> The state is also saved with the game, but currently it only works for open windows – that's why this is a draft.

I've checked the window state saving code and tested it, and it turns out that only the state of currently open windows is saved for other window types too. I think collapsed state is less important than say minimap state, so I shouldn't go out of my way to implement saving state for closed windows in this PR.

So this PR is ready after all.

frankystone:
Works very well :-)

Loading an older save game the behavior is always collapsed (if i remember correctly): E.g. when building a new tavern the state is collapsed or when enhancing a mine the state is collapsed although it was previously not collapsed.

Otherwise a very good enhancement, imho :+1:

tothxa:
> Loading an older save game the behavior is always collapsed (if i remember correctly): E.g. when building a new tavern the state is collapsed or when enhancing a mine the state is collapsed although it was previously not collapsed.

I couldn't reproduce it, but added code to make sure windows are expanded when loaded from old games.

-----

(sorry for the botched master merge / rebase / push)

frankystone:
Found another thing: If a window with collapsed prio section is close to the right screen edge and you click stop/start the window jumps to left. I guess the window jumps the amount of pixels the ware priority sliders would have need.

tothxa:
> Found another thing: If a window with collapsed prio section is close to the right screen edge and you click stop/start the window jumps to left. I guess the window jumps the amount of pixels the ware priority sliders would have need.

Fixed, and on the side I also fixed saving and restoring positions of all `UniqueWindow`s like the various statistics windows and tool windows in the editor.

It stressed my abilities to hunt it down though. ;) I needed `gdb`'s help to find the culprits.

~Now I need to debug why building windows are still not reopened at the last position.~ *(scrapped, because it turned out to be annoying for most buildings)*

tothxa:
Last commit does not work right in watch windows. I believe it's the same problem as #4689.

frankystone:
From my side commit [58aa436](https://github.com/widelands/widelands/pull/5785/commits/58aa43618bd9abf5abd36f14cb4658c629cc811a) works as expected. Maybe you should not try to solve too many bugs in this pr?

Anyway many thanks for your work!

tothxa:
> From my side commit [58aa436](https://github.com/widelands/widelands/pull/5785/commits/58aa43618bd9abf5abd36f14cb4658c629cc811a) works as expected.

Unfortunatley it was inconsistent, and for the wrong reason.

When new (including just reopened) building windows are first drawn, `move_out_of_the_way()` is called, which despite the name, moved them first to the centre of the main window, and then, if they covered the position of the mouse pointer (that was presumably clicked to bring up the window), were moved to just above/below the pointer, and then `move_inside_parent()` was called to make sure that the whole building window is inside the main window. 
But [58aa436](https://github.com/widelands/widelands/pull/5785/commits/58aa43618bd9abf5abd36f14cb4658c629cc811a) changed `move_inside_parent()` to restore the previous position, so instead of moving to centre then, if needed, out of the way but making sure it stays inside, it became either move to centre, or if centre covers pointer, to old position (regardless of old position covering the pointer).

So I tried the easy way, to always move to the old position if there was one, but as I wrote above, it was annoying when you opened and closed windows of minor buildings.

> Maybe you should not try to solve too many bugs in this pr?

I know, but stomping out one problem leads to the next one, so it's hard to stop. :)

Now it turns out that I must solve the initial sizing of collapsed building windows anyway, because it is needed for `move_out_of_the_way()` in either case to centre the window correctly, and because the mouse pointer is moved to the rightmost button, but then previously collapsed windows shrink under it, so it's left outside the window.

tothxa:
It seems to work now… Probably https://github.com/widelands/widelands/pull/5785/commits/66bb34fe6fd973ec71913ae9cdbdb1337eb0c987 both fixed and broke it. It should have kept `set_collapsed()` in the constructor too… I can't remember if I tested it this way back then…

frankystone:
I think this is fine now.

hessenfarmer (approved):
Code looks reasonable (although I don't understzand it fully).
Tested and working

tothxa:
> Code looks reasonable (although I don't understzand it fully). Tested and working

Thank you for the review. Please feel free to point out where it may need explanatory comments.

hessenfarmer:
> > Code looks reasonable (although I don't understzand it fully). Tested and working
> 
> Thank you for the review. Please feel free to point out where it may need explanatory comments.    
   
the problem is not missing comments but lack of knowledge on my side ;-) I simply don't have all parts of our code understood completely.

PR #5794, opened 2023-02-28
Disable apt upgrade in workflow

### LINKED ISSUE #5792: Workflow checks are failing
**Describe the bug**
All workflows are failing codecheck and clang tidy due to an error on the Ubuntu-22.04 runner during dependency installation.

Upstream bug: https://github.com/actions/runner-images/issues/7192

**To reproduce**
1. See recent build logs

**Additional context**
Should we deploy the workaround to reset grub boot devices from https://github.com/actions/runner-images/issues/7192#issuecomment-1445902492 ?

matthiakl:
**Type of change**
Bugfix / Maintenance

**Issue(s) closed**
Fixes #5792 

**New behavior**
~Disable phased updates as suggested by Github Action Maintainer~
Removing `sudo apt-get upgrade` from install_deps.sh.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -40,10 +40,7 @@ jobs:
       with:
         fetch-depth: 1
     - name: Installing dependencies
-      run: |
-        # https://github.com/actions/runner-images/issues/7192
-        echo 'APT::Get::Always-Include-Phased-Updates "false";' | sudo tee /etc/apt/apt.conf.d/99-phased-updates
-        sh ./.github/scripts/install_deps.sh
+      run: sh ./.github/scripts/install_deps.sh
     - name: Test
       run: |
         ./cmake/codecheck/run_tests.py
@@ -93,10 +90,7 @@ jobs:
       with:
         fetch-depth: 1
     - name: Installing dependencies
-      run: |
-        # https://github.com/actions/runner-images/issues/7192
-        echo 'APT::Get::Always-Include-Phased-Updates "false";' | sudo tee /etc/apt/apt.conf.d/99-phased-updates
-        sh ./.github/scripts/install_deps.sh
+      run: sh ./.github/scripts/install_deps.sh
     - name: CMake
       run: |
         echo "Will use $(nproc) cores."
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yaml
```diff
@@ -40,7 +40,10 @@ jobs:
       with:
         fetch-depth: 1
     - name: Installing dependencies
-      run: sh ./.github/scripts/install_deps.sh
+      run: |
+        # https://github.com/actions/runner-images/issues/7192
+        echo 'APT::Get::Always-Include-Phased-Updates "false";' | sudo tee /etc/apt/apt.conf.d/99-phased-updates
+        sh ./.github/scripts/install_deps.sh
```
tothxa: Shouldn't the workaround be put in `install_deps.sh` in a single place?

But my vote goes to removing `apt-get upgrade` anyway. :)

matthiakl: Sorry, I was a bit overacting here.
And with your other comment, I think you are right. So let's remove it and see how long it works ;)

tothxa:
200MB + firefox snap download saved per Ubuntu 22.04 task. ;) IIRC I had builds actually delayed by the snap download, for an unused component, no less. I think this is well worth the possible minor risk of not forcing a system-wide upgrade.

Now there's just a small nit that had been lost in the noise so far: `install-dependencies.sh` uses `apt` instead of `apt-get`, so it emits a warning about unstable CLI interface. Maybe that could also be changed in this PR.

tothxa (approved):
You might as well ignore my other comments.

One more thing that may worth a try though, to shave off 1 more minute from testsuite builds, is to see whether we could remove `linux-generic` from the dependencies. I don't think we need even `linux-headers`, and that's the only dependency that might be required from it.

matthiakl:
> Now there's just a small nit that had been lost in the noise so far: install-dependencies.sh uses apt instead of apt-get, so it emits a warning about unstable CLI interface.

> One more thing that may worth a try though, to shave off 1 more minute from testsuite builds, is to see whether we could remove linux-generic from the dependencies. 

I will try this separately, first get the CI green again ;)
Thanks for the comments.

PR #5799, opened 2023-03-02
Workflow improvements

matthiakl:
**Type of change**
Maintenance

**Issue(s) closed**
Re https://github.com/widelands/widelands/pull/5794#issuecomment-1449085994

**New behavior**
- (Should) resolve workflow warning in i18n.yml
- Use apt-get instead of apt
- Remove unused `linux-generic` dependency
- Use `install-dependencies.sh` also for vcpkg packages
- Check / Build vcpkg cache before compilation to reduce overall time.

**How it works**
The vcpkg cache is built in a separate job. If it is already present, fine the job finishes in ~2 minutes. Otherwise the packages are built and stored in the cache, if the Widelands compilation then fails, the cache will still be valid for the next run.
When the main compilation jobs start, the packages are already available, reducing the msvc build time.

### DIFF
--- a/.github/scripts/install_deps.sh
+++ b/.github/scripts/install_deps.sh
@@ -14,7 +14,6 @@ if [ "${GITHUB_JOB}" = "testsuite" ]; then
   ADD_PKG_LIST="
     ${CXX} \
     libwayland-egl1-mesa  \
-    linux-generic \
     mesa-utils  \
     xserver-xorg  \
     xserver-xorg-core \
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -6,7 +6,7 @@ on:
   pull_request:
     types: [ opened, reopened, synchronize ]
   push:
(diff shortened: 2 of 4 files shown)


Noordfrees (commented):
LGTM :) 
One request, please make the job `vcpkg-packages` depend on `clang-tidy` et al so we only launch these two workflows by the time we know we will actually perform the entire build matrix.

matthiakl:
Actually this is intended to not depend on the other jobs to decrease the total runtime of the complete workflow. Currently the critical path is `clang-tidy` (1h40m) → `windows-msvc` (1h20m with packages, 40m without) ≈ 3h total. By building the packages parallel to `clang-tidy` it should save about 40m.
In case the `clang-tidy` fails and the compilation is not started, the chances are, that the cache is still valid on the next push and the job finishes quickly.

hessenfarmer (approved):
Lgtm

PR #5804, opened 2023-03-03
Fix undefined `GLEW_ERROR_NO_GLX_DISPLAY`

Noordfrees:
**Type of change**
Bugfix

**Issue(s) closed**
PPA build error with older GLEW: https://launchpadlibrarian.net/654359026/buildlog_ubuntu-bionic-i386.widelands_2%3A1.2~bzr10541-202303031701~ubuntu18.04.1_BUILDING.txt.gz

**Additional context**
Re #5784

### DIFF
--- a/src/graphic/gl/initialize.cc
+++ b/src/graphic/gl/initialize.cc
@@ -131,14 +131,17 @@ SDL_GLContext initialize(
 	glewExperimental = GL_TRUE;
 	GLenum err = glewInit();
 	// LeakSanitizer reports a memory leak which is triggered somewhere above this line, probably
-	// coming from the gaphics drivers
+	// coming from the graphics drivers
 
+#ifdef GLEW_ERROR_NO_GLX_DISPLAY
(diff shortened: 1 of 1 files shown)


tothxa:
But do we want to still support ubuntu 18.04 or not? IIRC it was dropped from the github testsuite because support is ending soon. Shouldn't it be dropped on launchpad too?

Not that the change in this PR is a problem, but if we evaluate PRs without test builds on ubuntu 18.04, and then expect the committed version to build on it, then we may get into similar problems again.

Noordfrees:
Ubuntu 18.04 will be EOL in two months, so in my opinion, support for it becomes a nice-to-have that we don't really need to enforce in the testsuite but which we should patch up when the PPA complains, until keeping it in support becomes too much effort to maintain at which point it can be dropped.

tothxa (approved):
OK then.

~The indentation of line 144 is strange though. Should this be ran through auto format to spare one formatting commit in master?~

PR #5806, opened 2023-03-05
Add timeout support to regression_test.py and improve output

### LINKED ISSUE #1996: Allow the regression test suites to time out if nothing happens for a long while
I recently pushed some code for review, and as a good developer I wanted to run the test suite first to make sure I didn't break anything. Most ran ok, but one test suite kept running for ages before I finally closed the window it was running in. The test in question was «test_check_transportation_works_two_ships», though I guess other test suites could cause this problem too. (And I ran into the same issue with that particulart test suite with an unpatched trunk too)

It would be nice if there was a default timeout threshold where the testsuite would stop, mark the test as timed out and move on to the next test. Most tests seem to run fairly quickly so I guess a reasonable threshold would be 5-10 minutes in case it runs slow for some reason. When tests a marked as timeout a developer may then investigate why the test fails (or of course, mark it as ignored until another developer can look at it).

It would also ensure that a testrun will end at some point, rather than being stuck at a particular test step until the end of time. (Unless we at some point encounter the halting problem :P )

Done some more rethinking regarding the latter part. From what I can see, failures will currently send an in-game message with a description of expected and actual result. So this state seems to also keep a window open without closing. That probably means a naive implementation of a timeout would mark failures as timeouts too. Hm...

I guess I should probably mention that one of the long-term goals with this approach would be to make it possible to run the test suite in a continous integration setup*. I'm hesitant to do so at the moment, because it seems like a failing test or becoming stuck would create a process running forever which would then need to be manually stopped somehow. 

Widelands r6945

* I've actually set one up at some time in the past :) It built latest revision with GCC and Clang on I believe up-to-date Debian Sid and Arch Linux. Main reasons I discontinued it was that it was a) I don't have spare hardware to keep it running 24/7 so it lived in a set of virtual machines I ran and b) nothing broke for the time it was running, so it didn't seem high priority.

 * date created: 2014-04-16T14:57:15Z
 * owner: hjd
 * the launchpad url was https://bugs.launchpad.net/bugs/1308604

### LINKED ISSUE #5805: Workflow testsuite log does not show which test timed out
**Describe the bug**
#5519 builds time out in the testsuite, but the last log entry is a success report, so it's hard to tell that it is actually a test that times out, and which one it is.

**To reproduce**
1. See https://github.com/widelands/widelands/actions/runs/4307730442/jobs/7543142088

**Expected behavior**
When I run `regression_test.py` manually, the output shows e.g.:
```
test/maps/lua_testsuite.wmf/scripting/editor_test_lua_in_editor.lua ... 
  Running Widelands ... done.
ok
```

This is how it should be in the workflow log too, ie. the last log entry should show e.g.:
```
test/maps/lua_testsuite.wmf/scripting/editor_test_lua_in_editor.lua ... 
  Running Widelands ...
```

**Additional context**
Related bug: #1996

tothxa:
**Type of change**
Bugfix / New feature

**Issue(s) closed**
Fixes #1996
Fixes #5805

**New behavior**
 - Test cases are stopped and marked as failed if they don't finish within the set timeout. Default is 10 minutes.
 - Debug builds in workflow runs use 20 minutes timeout, as was found necessary.
 - Elapsed time is reported for each test case.
 - The name of the test case is flushed to stdout before the Widelands binary is started to make it easier to identify which test is running.
 - Tests are run in a deterministic order. (all `glob()`s are now `sorted()`)
 - Finding the Widelands binary in the top level of the source tree actually works.

**Possible regressions**
Test suite problems

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -168,7 +168,6 @@ jobs:
       ASAN_OPTIONS: 'detect_leaks=0'
       LSAN_OPTIONS: 'suppressions=${{ github.workspace }}/asan_3rd_party_leaks'
       CXX: '${{ matrix.compiler }}'
-      CONF: '${{ matrix.config }}'
     steps:
     - name: Checkout
       uses: actions/checkout@v3
@@ -208,7 +207,7 @@ jobs:
     - name: Testsuite
       run: |
         TIMEOUT=""
-        if [ "$CONF" = "Debug" ]
+        if [ "${{ matrix.config }}" = "Debug" ]
         then
           TIMEOUT="-t 20"
         fi
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -168,6 +168,7 @@ jobs:
       ASAN_OPTIONS: 'detect_leaks=0'
       LSAN_OPTIONS: 'suppressions=${{ github.workspace }}/asan_3rd_party_leaks'
       CXX: '${{ matrix.compiler }}'
+      CONF: '${{ matrix.config }}'
     steps:
     - name: Checkout
       uses: actions/checkout@v3
@@ -205,7 +206,13 @@ jobs:
         build/src/website/wl_map_object_info temp_web || [ $? -eq 2 ]
         build/src/website/wl_map_info data/maps/Archipelago_Sea.wmf || [ $? -eq 2 ]
     - name: Testsuite
-      run: ./regression_test.py -b build/src/widelands
+      run: |
+        TIMEOUT=""
+        if [ "$CONF" = "Debug" ]
+        then
+          TIMEOUT="-t 20"
+        fi
+        ./regression_test.py -b build/src/widelands $TIMEOUT
 
   vcpkg-packages:
     # If the vcpkg cache is not present, build packages and populate the cache
--- a/regression_test.py
+++ b/regression_test.py
@@ -34,6 +34,15 @@ else:
     def bytes_to_str(bytes):
         return str(bytes)
 
+has_timeout = "TimeoutExpired" in dir(subprocess)
+
+if "monotonic" in dir(time):
+    get_time = time.monotonic
+elif "perf_counter" in dir(time):
+    get_time = time.perf_counter
+else:
+    get_time = time.time
+
 def datadir():
     return os.path.join(os.path.dirname(__file__), "data")
 
@@ -109,17 +118,20 @@ class WidelandsTestCase(unittest.TestCase):
               stdout_file.write(" ")
             stdout_file.write("\n")
 
-            start_time = time.monotonic()
+            start_time = get_time()
             widelands = subprocess.Popen(
                     args, shell=False, stdout=stdout_file, stderr=subprocess.STDOUT)
-            try:
-                widelands.communicate(timeout = self.timeout)
-            except subprocess.TimeoutExpired:
-                widelands.kill()
+            if has_timeout:
+                try:
+                    widelands.communicate(timeout = self.timeout)
+                except subprocess.TimeoutExpired:
+                    widelands.kill()
+                    widelands.communicate()
+                    self.wl_timed_out = True
+                    stdout_file.write("\nTimed out.\n")
+            else:
                 widelands.communicate()
-                self.wl_timed_out = True
-                stdout_file.write("\nTimed out.\n")
-            end_time = time.monotonic()
+            end_time = get_time()
             stdout_file.flush()
             self.duration = datetime.timedelta(seconds = end_time - start_time)
             stdout_file.write("\nReturned from Widelands in {}, return code is {:d}\n".format(
@@ -205,9 +217,13 @@ def parse_args():
         help = "Assume success on return code 1, to allow running the tests "
         "without ASan reporting false positives."
     )
-    p.add_argument("-t", "--timeout", type=float, default = "10",
-        help = "Set the timeout duration for test cases in minutes. Default is 10 minutes."
-    )
+    if has_timeout:
+        p.add_argument("-t", "--timeout", type=float, default = "10",
+            help = "Set the timeout duration for test cases in minutes. Default is 10 minutes."
+        )
+    else:
+        p.epilog = "Python version does not support timeout. -t, --timeout is disabled. " \
+                   "Python >=3.3 is required for timeout support."
 
     args = p.parse_args()
 
@@ -268,7 +284,11 @@ def main():
     WidelandsTestCase.do_use_random_directory = not args.nonrandom
     WidelandsTestCase.keep_output_around = args.keep_around
     WidelandsTestCase.ignore_error_code = args.ignore_error_code
-    WidelandsTestCase.timeout = args.timeout * 60
+    if has_timeout:
+        WidelandsTestCase.timeout = args.timeout * 60
+    else:
+        out("Python version does not support timeout on subprocesses,\n"
+            "test cases may run indefinitely.\n\n")
 
     all_files = [os.path.basename(filename) for filename in sorted(glob("test/test_*.py")) ]
     if args.regexp:
--- a/regression_test.py
+++ b/regression_test.py
@@ -229,13 +229,19 @@ def parse_args():
 
     if args.binary is None:
         potential_binaries = (
-            glob("widelands") +
-            glob("src/widelands") +
-            glob("../*/src/widelands")
+            glob(os.path.join(os.curdir, "widelands")) +
+            glob(os.path.join(os.path.dirname(__file__), "widelands")) +
+            glob(os.path.join("src", "widelands")) +
+            glob(os.path.join("..", "*", "src", "widelands"))
         )
-        if not potential_binaries:
+        if potential_binaries:
+            args.binary = potential_binaries[0]
+        elif "which" in dir(shutil):
+            args.binary = shutil.which("widelands")
+
+        if args.binary is None:
             p.error("No widelands binary found. Please specify with -b.")
-        args.binary = potential_binaries[0]
+
     return args
 
 
--- a/regression_test.py
+++ b/regression_test.py
@@ -49,7 +49,7 @@ class WidelandsTestCase(unittest.TestCase):
     path_to_widelands_binary = None
     keep_output_around = False
     ignore_error_code = False
-    timeout = []
+    timeout = 600
 
     def __init__(self, test_script, **wlargs):
         unittest.TestCase.__init__(self)
@@ -71,6 +71,7 @@ class WidelandsTestCase(unittest.TestCase):
             else:
                 os.makedirs(self.run_dir)
         self.widelands_returncode = 0
+        self.wl_timed_out = False
 
     def run(self, result=None):
         self.currentResult = result # remember result for use in tearDown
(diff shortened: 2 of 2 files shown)


tothxa:
In the first run, one test timed out. While that was a nice demonstration that the timeout works as intended, it's also a sign that we may need to increase the default timeout for the sake of the workflow, if we don't want frequent failures due to timeouts. Not sure though how much it should be. I think 30 minutes would be too much. 15 minutes may not help much if 10 minutes was too little. Maybe 20 minutes is tolerable? Or should the default be 10 minutes and the workflow should set it to 20 minutes with a comment?

tothxa:
BTW most of `regression_test.py` is just tooling to force the square peg of running a binary with arguments and check its return value into the round hole of python unit tests (and then some more to iterate over items in directories), so I tried to reimplement it in posix sh: 
[regression_test.sh.txt](https://github.com/widelands/widelands/files/10898616/regression_test.sh.txt)

Though I had to implement some basic tooling that is available in python, and almost gave up several times in the process, the end result is actually smaller in terms of characters even if it has more lines. But the main benefit is better control over its output, and (IMO) better readibility due to doing much more things natively in the implementation language, and needing much less bending over backwards to make the whole thing work.

The only missing major functionality is the ability to run additional python scripts from the test directory. Using a fixed directory for output files isn't implemented either, but that would be easy to add.

I have already tested most of the script.

matthiakl:
I do not know the circumstances, when these timeouts happen, but maybe it would be helpful to have some kind of time output in the log as well? Simply adding timestamps to see which tests run longer than expected.

Your script is indeed better readable (for me ;)). You might want to add a logging function to reduce the number of `echo` lines.

> The only missing major functionality is the ability to run additional python scripts from the test directory.

Is this actually used? There are no python scripts in the test directory. Maybe for additional manual tests?

tothxa:
> I do not know the circumstances, when these timeouts happen,

I believe the timeout in the workflow run of this PR was just caused by the many parallel runs taking up VM time from each other. But I don't know how VM provisioning works.

> but maybe it would be helpful to have some kind of time output in the log as well? Simply adding timestamps to see which tests run longer than expected.

Indeed. Unfortunately both `time` and `date` seem to be vastly incompatible between GNU and BSD, and the common posix subset is unusably limited, as always… (`time` and `timeout` would probably clash anyway, and I don't think I can run shell functions through either)

So probably timestamps are the most I could do, but that wouldn't be very helpful. Maybe test for GNU `date +%s` and BSD `date -r`, and fall back on timestamps only if neither works?

> Your script is indeed better readable (for me ;)). You might want to add a logging function to reduce the number of `echo` lines.

How do you think that would help? It would just replace all `echo`s with a call to it. Not using `\n` for empty lines is just my preference for readibility, though it seems to be shared by other shell scripts we have.

tothxa:
Sorry, I forgot that the python script is the real one. :) (for now?) Added timing info to it.

Maybe this should be taken over by somebody who actually knows python. ;) While hunting for documentation for the last commit, I've found out that `subprocess` actually supports timeout, but as far as I could tell, not the way we use it: at least `run()` and `communicate()` do, but we read stdout line by line instead. `readline()` doesn't seem to have a simple way to do timeouts, but even if it did, that wouldn't be sufficient for this use case, because there may be problems where the test script fails to finish, but the game does not hang, so it keeps outputting due to `--verbose`.

[edit:] Oops, I've just realised that I could test the elapsed time in the `while` loop. That, with a timeout on `readline()`, could work.

regression_test.py
```diff
@@ -115,13 +118,16 @@ def run_widelands(self, wlargs, which_time):
                 stdout_file.write(bytes_to_str(line))
                 stdout_file.flush()
```
matthiakl: Essentially this writes stdout to the file, right? So it should be possible to give `stdout_file` directly to `Popen`. Maybe one of `encoding` or `text` has to be set also to handle the encoding.
Then you can remove the infinite while loop and use the timeout in `communicate` instead.

matthiakl (commented):
> How do you think that would help? It would just replace all echos with a call to it. Not using \n for empty lines is just my preference for readibility, though it seems to be shared by other shell scripts we have.

Ok, I understand your point. It just seemed a bit verbose to me with `echo`ing empty lines frequently.

regression_test.py
tothxa: :+1:  Thanks for the tip. :smiley: I'll wait for the current test run to finish (to collect some more data on timeouts and running times), then try to do it.

tothxa: Thank you again. It works for me without any other options to Popen(), let's see in the workflow run.

tothxa:
Timeout support was added in python 3.3. Should I add compatibility checks for python 2.x (just disable timeout and timing support), or may I remove the other compatibility block from the beginning? (on my Debian oldstable system /usr/bin/python is still python2)

matthiakl:
If disabling it for Python 2 does not introduce too much special handling I would prefer this way.

tothxa:
I've analysed the logs of the last 2 runs a bit. Some tests in both runs pushed the 10min limit with >8min elapsed time, but the vast majority doesn't even come close to it. It turns out, that the all tribes tests on debug builds always go above 5min, all the rest are <=2min.

```
$ rgrep -oh 'Elapsed time: 0:0.' log? | sort | uniq -c
   2412 Elapsed time: 0:00
    369 Elapsed time: 0:01
      1 Elapsed time: 0:02
     26 Elapsed time: 0:05
     19 Elapsed time: 0:06
     11 Elapsed time: 0:07
      7 Elapsed time: 0:08
      1 Elapsed time: 0:09
$ rgrep -B2 'Elapsed time: 0:0[2-9]:..' log? | grep -o 'test/.*' | sort | uniq -c
     17 test/maps/all_tribes.wmf/scripting/test_atlanteans.lua
     16 test/maps/all_tribes.wmf/scripting/test_barbarians.lua
     16 test/maps/all_tribes.wmf/scripting/test_empire.lua
     16 test/maps/all_tribes.wmf/scripting/test_frisians.lua
$ rgrep -B2 'Elapsed time: 0:0[7-9]:..' log? | grep -o 'test/.*' | sort | uniq -c
      5 test/maps/all_tribes.wmf/scripting/test_atlanteans.lua
     11 test/maps/all_tribes.wmf/scripting/test_barbarians.lua
      2 test/maps/all_tribes.wmf/scripting/test_empire.lua
      1 test/maps/all_tribes.wmf/scripting/test_frisians.lua
$ rgrep -c 'Elapsed time: 0:0[7-9]:..' log? | grep '[1-9]$' | sort
log1/testsuite (clang-7, ubuntu-20.04, Debug)/7_Testsuite.txt:4
log1/testsuite (clang-8, ubuntu-20.04, Debug)/7_Testsuite.txt:1
log1/testsuite (Debug, ubuntu-22.04, clang-13)/7_Testsuite.txt:3
log1/testsuite (Debug, ubuntu-22.04, g++-11)/7_Testsuite.txt:1
log1/testsuite (Debug, ubuntu-22.04, g++-12)/7_Testsuite.txt:1
log1/testsuite (g++-8, ubuntu-20.04, Debug)/7_Testsuite.txt:1
log2/testsuite (clang-7, ubuntu-20.04, Debug)/7_Testsuite.txt:1
log2/testsuite (clang-8, ubuntu-20.04, Debug)/7_Testsuite.txt:1
log2/testsuite (Debug, ubuntu-22.04, clang-14)/7_Testsuite.txt:1
log2/testsuite (Debug, ubuntu-22.04, g++-11)/7_Testsuite.txt:2
log2/testsuite (g++-7, ubuntu-20.04, Debug)/7_Testsuite.txt:2
log2/testsuite (g++-8, ubuntu-20.04, Debug)/7_Testsuite.txt:1
$ rgrep -c 'Elapsed time: 0:0[5-9]:..' log? | grep '[1-9]$' | sort
log1/testsuite (clang-7, ubuntu-20.04, Debug)/7_Testsuite.txt:4
log1/testsuite (clang-8, ubuntu-20.04, Debug)/7_Testsuite.txt:4
log1/testsuite (Debug, ubuntu-22.04, clang-13)/7_Testsuite.txt:4
log1/testsuite (Debug, ubuntu-22.04, clang-14)/7_Testsuite.txt:4
log1/testsuite (Debug, ubuntu-22.04, g++-11)/7_Testsuite.txt:4
log1/testsuite (Debug, ubuntu-22.04, g++-12)/7_Testsuite.txt:4
log1/testsuite (g++-7, ubuntu-20.04, Debug)/7_Testsuite.txt:4
log1/testsuite (g++-8, ubuntu-20.04, Debug)/7_Testsuite.txt:4
log2/testsuite (clang-7, ubuntu-20.04, Debug)/7_Testsuite.txt:4
log2/testsuite (clang-8, ubuntu-20.04, Debug)/7_Testsuite.txt:4
log2/testsuite (Debug, ubuntu-22.04, clang-13)/7_Testsuite.txt:4
log2/testsuite (Debug, ubuntu-22.04, clang-14)/7_Testsuite.txt:4
log2/testsuite (Debug, ubuntu-22.04, g++-11)/7_Testsuite.txt:4
log2/testsuite (Debug, ubuntu-22.04, g++-12)/7_Testsuite.txt:4
log2/testsuite (g++-7, ubuntu-20.04, Debug)/7_Testsuite.txt:4
log2/testsuite (g++-8, ubuntu-20.04, Debug)/7_Testsuite.txt:4
```

tothxa:
It worked nicely. This time there were 3 parallel builds running, and 1 test went over 10 minutes, but not too much:

```
testsuite (Debug, ubuntu-22.04, clang-14)/7_Testsuite.txt:2023-03-08T02:12:32.4418024Z     Elapsed time: 0:10:31.438279
```

So I think 20 minutes for debug builds will work with a good safety margin, and 10 minutes is more than enough for release builds.

tothxa:
Regarding the line-by-line flushing that we removed to allow native timeouts: I think that makes sense for tests that may crash the whole testing rig, to get logging up to a point as close to the crash as possible. But that's not the case here, so I don't think we introduced any new risk of missing some of the logs.

.github/workflows/build.yaml
```diff
@@ -205,7 +206,13 @@ jobs:
         build/src/website/wl_map_object_info temp_web || [ $? -eq 2 ]
         build/src/website/wl_map_info data/maps/Archipelago_Sea.wmf || [ $? -eq 2 ]
     - name: Testsuite
-      run: ./regression_test.py -b build/src/widelands
+      run: |
+        TIMEOUT=""
+        if [ "$CONF" = "Debug" ]
```
matthiakl: Is there a special reason, why you do not use matrix.config here directly?
```suggestion
        if [ "${{ matrix.config }}" = "Debug" ]
```

matthiakl (commented):
> Regarding the line-by-line flushing that we removed to allow native timeouts: I think that makes sense for tests that may crash the whole testing rig, to get logging up to a point as close to the crash as possible. 

AFAIK, python flushes the streams on newline characters, so we should already get near the last complete line.

One more comment inline, the rest looks fine to me.

.github/workflows/build.yaml
tothxa: Do you mean other than not being familiar with it? :smiley:

Thanks!

.github/workflows/build.yaml
```diff
@@ -168,6 +168,7 @@ jobs:
       ASAN_OPTIONS: 'detect_leaks=0'
       LSAN_OPTIONS: 'suppressions=${{ github.workspace }}/asan_3rd_party_leaks'
       CXX: '${{ matrix.compiler }}'
+      CONF: '${{ matrix.config }}'
```
tothxa:
```suggestion
```

matthiakl (approved):

PR #5811, opened 2023-03-14
add function to decide between invite and request to join

tothxa:
@hessenfarmer re https://github.com/widelands/widelands/pull/5774#discussion_r1135276600

### DIFF
--- a/src/ai/ai_help_structs.cc
+++ b/src/ai/ai_help_structs.cc
@@ -1102,8 +1102,9 @@ uint32_t PlayersStrengths::team_power(Widelands::TeamNumber tn) {
 }
 
 // Make the decision whether own or player's team is better
-void PlayersStrengths::join_or_invite(
-   const Widelands::PlayerNumber pl, Widelands::Game& game, const Time& gametime) {
+void PlayersStrengths::join_or_invite(const Widelands::PlayerNumber pl,
+                                      Widelands::Game& game,
+                                      const Time& gametime) {
 	if (this_player_number == pl || active_players_ < 3) {
 		return;
 	}
@@ -1127,7 +1128,7 @@ void PlayersStrengths::join_or_invite(
 	if (!me_alone || !other_alone) {
 		for (auto& item : all_stats_) {
 			if (item.first == pl || item.first == this_player_number || item.second.team_number == 0 ||
-				 item.second.defeated) {
+			    item.second.defeated) {
 				continue;
 			}
 			if (can_invite && item.second.team_number == this_player_team) {
@@ -1208,8 +1209,8 @@ void PlayersStrengths::join_or_invite(
 	if (invite) {
 		verb_log_dbg_time(
 		   gametime, "AI Diplomacy: Player(%d)%s invites player (%d) to join with diploscore %d%s\n",
-		   static_cast<unsigned int>(this_player_number), myts_s.c_str(), static_cast<unsigned int>(pl),
-		   get_diplo_score(pl), ots_s.c_str());
+		   static_cast<unsigned int>(this_player_number), myts_s.c_str(),
+		   static_cast<unsigned int>(pl), get_diplo_score(pl), ots_s.c_str());
 		game.send_player_diplomacy(this_player_number, Widelands::DiplomacyAction::kInvite, pl);
 		set_last_time_requested(gametime, pl);
 		return;
--- a/src/ai/ai_help_structs.cc
+++ b/src/ai/ai_help_structs.cc
@@ -1104,10 +1104,15 @@ uint32_t PlayersStrengths::team_power(Widelands::TeamNumber tn) {
 // Make the decision whether own or player's team is better
 Widelands::DiplomacyAction PlayersStrengths::join_or_invite(const Widelands::PlayerNumber pl,
                                                             const Time& gametime) {
-	const Widelands::TeamNumber other_tn = all_stats_[pl].team_number;
-	if (this_player_number == pl || active_players_ < 3 || this_player_team == other_tn) {
+	if (this_player_number == pl || active_players_ < 3) {
 		return Widelands::DiplomacyAction::kRefuseJoin;
 	}
+
+	const Widelands::TeamNumber other_tn = all_stats_[pl].team_number;
+	if (this_player_team != 0 && this_player_team == other_tn) {
+		return Widelands::DiplomacyAction::kRefuseJoin;
+	}
+
 	int my_team_sc = 0;
 	int other_team_sc = 0;
 	bool me_alone = false;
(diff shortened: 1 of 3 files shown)


tothxa:
I'd add a similar team score check to consider accepting invites, and a team based check for leaving too (leave if team average is negative).

src/ai/ai_help_structs.cc
```diff
@@ -1100,6 +1101,82 @@ uint32_t PlayersStrengths::team_power(Widelands::TeamNumber tn) {
...
+			                  "player (%d) with diploscore %d\n",
+			                  static_cast<unsigned int>(this_player_number), my_team_sc,
+			                  static_cast<unsigned int>(pl), other_team_sc);
+			return Widelands::DiplomacyAction::kRefuseJoin;
+		}
+	} else {
+		other_team_sc /= members_in_team(other_tn) - 1;
+	}
+
+	if (!can_invite || !can_join) {
```
hessenfarmer: why don't we want to log this if both are true. for example both teams have 2 players in a 4 player game

src/ai/ai_help_structs.cc
```diff
@@ -1100,6 +1101,82 @@ uint32_t PlayersStrengths::team_power(Widelands::TeamNumber tn) {
 	return team_powers_[tn];
 }
 
+// Make the decision whether own or player's team is better
+Widelands::DiplomacyAction PlayersStrengths::join_or_invite(const Widelands::PlayerNumber pl,
+                                                            const Time& gametime) {
+	if (this_player_number == pl || active_players_ < 3) {
```
hessenfarmer, line 1108: the check for both being the same player is not really necessary as it is already checked in defaultai.cc line 3390

src/ai/defaultai.cc
```diff
@@ -3392,26 +3392,16 @@ void DefaultAI::diplomacy_actions(const Time& gametime) {
...
 		if (player_statistics.get_diplo_score(opn) >= 35) {
-			if (((other_player->team_number() == 0 ||
-			      player_statistics.team_power(me->team_number()) >
-			         player_statistics.team_power(other_player->team_number())) &&
-			     player_statistics.members_in_team(me->team_number()) <
-			        player_statistics.players_active() - 1) ||
-			    (player_statistics.members_in_team(other_player->team_number()) ==
-			        player_statistics.players_active() - 1 &&
-			     me->team_number() != other_player->team_number())) {
+			Widelands::DiplomacyAction action = player_statistics.join_or_invite(opn, gametime);
```
hessenfarmer: as we return the diplo action we could directly use it in the send_player diplomacy command making the if loop here superfluous. Logging could be done in the help function

hessenfarmer (approved):
very good idea some things inline. would you like to fix them or shall I merge and fix them myselfs

src/ai/ai_help_structs.cc
tothxa: It is logged to have some trace if then the action is skipped because of the score. If both are allowed, then either choice will result in an action that is already logged.

With the other logging brought here, this will only log skipped actions.

src/ai/ai_help_structs.cc
tothxa: It's a cheap check, so why not be safe in case this gets later called from somewhere else?

src/ai/defaultai.cc
tothxa: OK. I almost did it, but it complicated the return code here, and we need to check the return value anyway. But using it directly was part of the original plan.

tothxa: OTOH it's cleaner to have the logging in the same place as the actual execution… I'll bring that too, and then we don't even have to fiddle with reporting skipped actions to the caller. Is that OK with you?

hessenfarmer: as far as I understand we could use exactly the return codes already provided as a kRefuseJoin without a Join request gets simply ignored.

src/ai/ai_help_structs.cc
hessenfarmer: I don't get it as far as I read the code nothing is skipped here but only logging takes place. the possible states of the 2 variables are either one of them is true or mostly both of them will be true they can't be both negative. but the condition is excluding the both true case.

src/ai/defaultai.cc
hessenfarmer: > OTOH it's cleaner to have the logging in the same place as the actual execution… I'll bring that too, and then we don't even have to fiddle with reporting skipped actions to the caller. Is that OK with you?

yes please. Please note I made some changes to the original branch.

src/ai/ai_help_structs.cc
tothxa: The return value for the not allowed case is set to kRefuseJoin, so if it is selected based on the scores, then the action will be skipped. If both are allowed, then either choice will result in a valid action that will be executed. (and we have already excluded the possibility of both of them being invalid by checking that there are at least 3 players)

tothxa: But this will be cleaned up then with the execution brought here.

hessenfarmer: I understand the attempt but I think the logic might be wrong. case 0 0 can't happen but we would log. cases 1 0 or 0 1 we will log and some action happens. case 1 1 we won't log but action happens.  this seems to be inconsistent.

tothxa:
This turned out to be harder than I expected. Some decision branches may not have worked quite as I intended in the original version, and I had to be careful not to miss any possibility.

src/ai/ai_help_structs.cc
```diff
@@ -1102,79 +1102,119 @@ uint32_t PlayersStrengths::team_power(Widelands::TeamNumber tn) {
...
+			   gametime,
+			   "AI Diplomacy: Player(%d), requests to join player (%d) with diploscore: %d\n",
+			   static_cast<unsigned int>(this_player_number), static_cast<unsigned int>(pl),
+			   get_diplo_score(pl));
+			game.send_player_diplomacy(this_player_number, Widelands::DiplomacyAction::kJoin, pl);
+			set_last_time_requested(gametime, pl);
+			return;
+		}
+	} else {  // Own team is more desirable
+		if (other_alone) {
```
hessenfarmer, line 1190: why not trying to recruit a lone player with a diploscore of higher then 35 just because the average of my other team mates is even higher.

tothxa: Because it will decrease the team average, especially if we consider that team members are penalized somewhat in the scoring. And bigger teams are not necessarily better, because they can be more unstable.

tothxa: But we may remove the random penalty for lone players, or even make it a small bonus to also invite them if they're close enough. Or make it `score - penalty + random bonus` or even `score - random penalty + random bonus` (even distribution vs. favouring the mean).

hessenfarmer: ok so let's have it and see how it performs. We can adjust values afterwards.

PR #5813, opened 2023-03-16
Consider team strengths for accepting requests plus cleanups

tothxa:
@hessenfarmer  re https://github.com/widelands/widelands/pull/5774#issuecomment-1469228092

### DIFF
--- a/src/ai/ai_help_structs.cc
+++ b/src/ai/ai_help_structs.cc
@@ -1096,57 +1096,58 @@ uint8_t PlayersStrengths::members_in_team(Widelands::TeamNumber tn) {
 	return team_members_[tn];
 }
 
+bool PlayersStrengths::get_is_alone(const Widelands::PlayerNumber pn) {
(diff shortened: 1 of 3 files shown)


tothxa:
There's still some instability when 2 players like each other, but there's a third one who is liked by one of them and disliked by the other. I think leaving teams will have to consider strength of team too (allow worse players to have the better ones). And possibly team evaluation should consider the weakest of the team too. But handling this properly will need team based AI thinking, that we agreed not to do in this PR.

Another thing is that I ran a 6 player test. In turbulent times, it often happened that dangling requests were accepted, resulting in unexpected team line-ups. So only one team change should be done in one diplo action run, but the hard part is that the first acceptable one may not be the best, and all decided rejects should be done to clean up the queue as much as possible. Sometimes leave should also be considered before pending requests to help with the first point.

hessenfarmer (approved):

PR #5814, opened 2023-03-17
consider team score for leaving team and some fixes and optimisations

tothxa:
@hessenfarmer Last part of team based decisions.
Also implements checking all pending requests before doing maximum one team change (and rejecting all others) in one run of diplomacy actions.

**Additional context**
I tested your last scoring fix with this, and it's now stable. I think this works very well now.

[outdated] ~Here's today's test game with this branch:
[dipl_test_0317.zip](https://github.com/widelands/widelands/files/10997108/dipl_test_0317.zip)
[dipl_test_0317_replay.zip](https://github.com/widelands/widelands/files/10997109/dipl_test_0317_replay.zip)~

### DIFF
--- a/src/ai/defaultai.cc
+++ b/src/ai/defaultai.cc
@@ -3380,23 +3380,24 @@ void DefaultAI::diplomacy_actions(const Time& gametime) {
 	if (!have_plan && player_statistics.get_worst_ally_score() < -15) {
 		const int32_t team_vs_worst = my_team_score + player_statistics.get_worst_ally_score();
 		if ((team_vs_worst > 0 && player_statistics.members_in_team(mytn) > 3) ||
-			    (team_vs_worst > -10 &&
-			     RNG::static_rand(player_statistics.members_in_team(mytn) * 2 - 2) > 0)) {
-				verb_log_dbg_time(gametime,
-				                  "AI Diplomacy: Player(%d) tolerates player (%d) with diploscore "
-				                  "%d in team (%d)%s\n",
-				                  static_cast<unsigned int>(mypn),
-		                        static_cast<unsigned int>(player_statistics.get_worst_ally()),
-				                  player_statistics.get_worst_ally_score(),
-				                  static_cast<unsigned int>(mytn), myts_s.c_str());
+		    (team_vs_worst > -10 &&
+		     RNG::static_rand(player_statistics.members_in_team(mytn) * 2 - 2) > 0)) {
+			verb_log_dbg_time(gametime,
+			                  "AI Diplomacy: Player(%d) tolerates player (%d) with diploscore "
+			                  "%d in team (%d)%s\n",
+			                  static_cast<unsigned int>(mypn),
+			                  static_cast<unsigned int>(player_statistics.get_worst_ally()),
+			                  player_statistics.get_worst_ally_score(),
+			                  static_cast<unsigned int>(mytn), myts_s.c_str());
 		} else {
 			planned_action = Widelands::DiplomacyAction::kLeaveTeam;
 			have_plan = true;
 			plan_priority = std::max(0, -team_vs_worst);
 			if (g_verbose) {
-				planned_log_append_text = format("%s because of player (%d) with diploscore %d",
-			                  myts_s, static_cast<unsigned int>(player_statistics.get_worst_ally()),
-			                  player_statistics.get_worst_ally());
+				planned_log_append_text =
+				   format("%s because of player (%d) with diploscore %d", myts_s,
+				          static_cast<unsigned int>(player_statistics.get_worst_ally()),
+				          player_statistics.get_worst_ally());
 			}
 		}
 	}
@@ -3452,33 +3453,33 @@ void DefaultAI::diplomacy_actions(const Time& gametime) {
 
 	// Execute only one action that changes team line-ups
 	if (have_plan) {
-		assert (planned_action == Widelands::DiplomacyAction::kLeaveTeam ||
-		        planned_action == Widelands::DiplomacyAction::kAcceptInvite ||
-		        planned_action == Widelands::DiplomacyAction::kAcceptJoin);
+		assert(planned_action == Widelands::DiplomacyAction::kLeaveTeam ||
+		       planned_action == Widelands::DiplomacyAction::kAcceptInvite ||
+		       planned_action == Widelands::DiplomacyAction::kAcceptJoin);
 		if (g_verbose) {
 			std::string action_str = "join request";
 			switch (planned_action) {
-				case Widelands::DiplomacyAction::kLeaveTeam:
-					verb_log_dbg_time(gametime, "AI Diplomacy: Player(%d) leaves team (%d)%s.\n",
-					                  static_cast<unsigned int>(mypn), static_cast<unsigned int>(mytn),
-					                  planned_log_append_text.c_str());
-					break;
+			case Widelands::DiplomacyAction::kLeaveTeam:
+				verb_log_dbg_time(gametime, "AI Diplomacy: Player(%d) leaves team (%d)%s.\n",
+				                  static_cast<unsigned int>(mypn), static_cast<unsigned int>(mytn),
+				                  planned_log_append_text.c_str());
+				break;
 
-				case Widelands::DiplomacyAction::kAcceptInvite:
-					action_str = "invitation";
-					FALLS_THROUGH;
-				case Widelands::DiplomacyAction::kAcceptJoin:
-					verb_log_dbg_time(
-					   gametime,
-					   "AI Diplomacy: Player(%d)%s accepts the %s of player (%d) with diploscore %d%s.\n",
-					   static_cast<unsigned int>(mypn), myts_s.c_str(), action_str.c_str(),
-					   static_cast<unsigned int>(planned_opn), planned_other_score,
-					   planned_log_append_text.c_str());
-					break;
+			case Widelands::DiplomacyAction::kAcceptInvite:
+				action_str = "invitation";
+				FALLS_THROUGH;
+			case Widelands::DiplomacyAction::kAcceptJoin:
+				verb_log_dbg_time(
+				   gametime,
+				   "AI Diplomacy: Player(%d)%s accepts the %s of player (%d) with diploscore %d%s.\n",
+				   static_cast<unsigned int>(mypn), myts_s.c_str(), action_str.c_str(),
+				   static_cast<unsigned int>(planned_opn), planned_other_score,
+				   planned_log_append_text.c_str());
+				break;
 
-				default:
-					NEVER_HERE();
-					break;
+			default:
+				NEVER_HERE();
+				break;
 			}
 		}
 		game().send_player_diplomacy(mypn, planned_action, planned_opn);
--- a/src/ai/defaultai.cc
+++ b/src/ai/defaultai.cc
@@ -3389,10 +3389,11 @@ void DefaultAI::diplomacy_actions(const Time& gametime) {
 
 	// Check for undesirable teammates
 	if (planned_action == kNoAction && player_statistics.get_worst_ally_score() < -15) {
-		const int32_t team_vs_worst = my_team_score + player_statistics.get_worst_ally_score();
-		if ((team_vs_worst > 0 && player_statistics.members_in_team(mytn) > 3) ||
-		    (team_vs_worst > -10 &&
-		     RNG::static_rand(player_statistics.members_in_team(mytn) * 2 - 2) > 0)) {
+		const uint8_t my_team_size = player_statistics.members_in_team(mytn);
+		const int32_t team_vs_worst =
+		   (my_team_size < 3 ? 0 : my_team_score) + player_statistics.get_worst_ally_score();
+		if ((team_vs_worst > 0 && my_team_size > 3) ||
+		    (team_vs_worst > -10 && RNG::static_rand(my_team_size * 2 - 2) > 0)) {
 			verb_log_dbg_time(gametime,
 			                  "AI Diplomacy: Player(%d) tolerates player (%d) with diploscore "
 			                  "%d in team (%d)%s\n",
--- a/src/ai/defaultai.cc
+++ b/src/ai/defaultai.cc
@@ -3426,7 +3426,9 @@ void DefaultAI::diplomacy_actions(const Time& gametime) {
 		           other_player->team_number() == mytn && other_player->team_number() != 0) {
 			if (player_statistics.members_in_team(mytn) > 2) {
 				const int32_t team_vs_this = my_team_score + player_statistics.get_diplo_score(opn);
-				if (team_vs_this > 0 || (team_vs_this > -10 && RNG::static_rand(8) > 0)) {
+				if ((team_vs_this > 0 && player_statistics.members_in_team(mytn) > 3) ||
+				    (team_vs_this > -10 &&
+				     RNG::static_rand(player_statistics.members_in_team(mytn) * 2 - 2) > 0)) {
 					verb_log_dbg_time(
 					   gametime,
 					   "AI Diplomacy: Player(%d) tolerates player (%d) with diploscore %d in team (%d)%s\n",
--- a/src/ai/defaultai.cc
+++ b/src/ai/defaultai.cc
@@ -545,7 +545,9 @@ void DefaultAI::think() {
 			break;
 		case SchedulerTaskId::kDiplomacy:
 			diplomacy_actions(gametime);
-			set_taskpool_task_time(gametime + kDiplomacyInterval, SchedulerTaskId::kDiplomacy);
+			set_taskpool_task_time(
+			   gametime + kDiplomacyInterval + Duration(RNG::static_rand(30) * 1000),
+			   SchedulerTaskId::kDiplomacy);
 			break;
 		case SchedulerTaskId::kUnset:
 			NEVER_HERE();
@@ -1152,7 +1154,8 @@ void DefaultAI::late_initialization() {
 		// and allow statistics to settle after loading game
 		taskPool.push_back(std::make_shared<SchedulerTask>(
 		   std::max<Time>(
-		      gametime + kStatUpdateInterval * 5, Time((10 + RNG::static_rand(10)) * 60 * 1000)),
+		      gametime + kStatUpdateInterval * 4 + Duration(RNG::static_rand(5 * 60) * 1000),
+			   Time((10 + RNG::static_rand(10)) * 60 * 1000)),
 		   SchedulerTaskId::kDiplomacy, 7, "diplomacy actions"));
 	}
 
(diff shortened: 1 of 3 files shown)


src/ai/defaultai.cc
```diff
@@ -3353,23 +3353,30 @@ void DefaultAI::diplomacy_actions(const Time& gametime) {
...
 		if (pda.other == mypn) {
 			const int32_t diploscore = player_statistics.get_diplo_score(pda.sender);
 			const Widelands::TeamNumber other_tn = player_statistics.get_team_number(pda.sender);
 
-			// accept if diploscore high, else accept only 50%
-			bool accept = diploscore >= std::max(my_team_score, 25) ||
-			              (diploscore > std::max(my_team_score / 2, 15) && RNG::static_rand(2) == 0);
+			// consider only if we are not defeated and if not resulting in a team win
+			bool accept = !me_def &&
+			              std::max<uint8_t>(
```
hessenfarmer: what is the purpose of this std::max?

hessenfarmer (commented):
looks good so far. one question inline

src/ai/defaultai.cc
tothxa: I believe team 0 returns 0. ~But indeed, checking it directly isn't that much longer and is at least as efficient, so I'll do that instead.~ I'll try to refactor it to check that directly.

hessenfarmer: from my perspective 0 wouldn't hurt here as it is smaller then active_players -1

tothxa: Oops, indeed, I mixed up what we're checking. :) Meanwhile I made `team_members_[0]` be 1 if there's at least one teamless player. I think that might be useful anyway.

src/ai/defaultai.cc
```diff
@@ -3341,70 +3341,202 @@ void DefaultAI::diplomacy_actions(const Time& gametime) {
...
+			                  static_cast<unsigned int>(mytn), myts_s.c_str());
+		} else {
+			planned_action = Widelands::DiplomacyAction::kLeaveTeam;
+			have_plan = true;
+			plan_priority = std::max(0, -team_vs_worst);
+			if (g_verbose) {
+				planned_log_append_text =
+				   format("%s because of player (%d) with diploscore %d", myts_s,
+				          static_cast<unsigned int>(player_statistics.get_worst_ally()),
+				          player_statistics.get_worst_ally());
```
tothxa, line 3406: Already corrected locally to `get_worst_ally_score_()`, but waiting for result of clang-tidy.

tothxa:
~This is now ready, if it passes the tests.~ I have no more change requests for the main PR either.

Too bad that a lot of testing should be done with debug builds to validate all asserts properly, but can't run those with high speed. :(

src/ai/defaultai.cc
```diff
@@ -3341,70 +3344,210 @@ void DefaultAI::diplomacy_actions(const Time& gametime) {
...
+                                             Widelands::DiplomacyAction::kRefuseInvite :
+                                             Widelands::DiplomacyAction::kRefuseJoin),
+				                             pda.sender);
+			} else {
+				if (have_plan) {
+					verb_log_dbg_time(
+					   gametime,
+					   "AI Diplomacy: Player(%d) replaces plan: old priority: %d, new priority: %d",
+					   static_cast<unsigned int>(mypn), plan_priority, priority);
+				}
```
tothxa: Oops, forgot rejecting previous planned accept.

tothxa:
OK, now it should be done.

src/ai/defaultai.cc
```diff
@@ -545,7 +545,9 @@ void DefaultAI::think() {
 			break;
 		case SchedulerTaskId::kDiplomacy:
 			diplomacy_actions(gametime);
-			set_taskpool_task_time(gametime + kDiplomacyInterval, SchedulerTaskId::kDiplomacy);
+			set_taskpool_task_time(
+			   gametime + kDiplomacyInterval + Duration(RNG::static_rand(30) * 1000),
```
hessenfarmer, line 549: I believe this is to decouple the Diplomacy of each player from the other players correct?

src/ai/defaultai.cc
```diff
@@ -3341,70 +3344,219 @@ void DefaultAI::diplomacy_actions(const Time& gametime) {
...
+			planned_log_append_text = myts_s;
+			/* verb_ */ log_dbg_time(gametime,
+			                         "AI Diplomacy: Player(%d) plans to leave team with priority %d",
+			                         static_cast<unsigned int>(mypn), plan_priority);
+		}
+	}
+
+	// Check for undesirable teammates
+	if (planned_action == kNoAction && player_statistics.get_worst_ally_score() < -15) {
+		const int32_t team_vs_worst = my_team_score + player_statistics.get_worst_ally_score();
```
hessenfarmer: wouldn't this be ally score * 2 for a team with only 2 players. If yes we would leave earlier as before, wouldn't we

src/ai/defaultai.cc
```diff
@@ -3341,70 +3344,219 @@ void DefaultAI::diplomacy_actions(const Time& gametime) {
...
+					verb_log_dbg_time(
+					   gametime,
+					   "AI Diplomacy: Player(%d) replaces plan: old priority: %d, new priority: %d",
+					   static_cast<unsigned int>(mypn), plan_priority, priority);
+					if (planned_action != Widelands::DiplomacyAction::kLeaveTeam) {
+						verb_log_dbg_time(gametime,
+						                  "AI Diplomacy: Player(%d)%s denies the %s of player (%d) with "
+						                  "diploscore %d%s\n",
+						                  static_cast<unsigned int>(mypn), myts_s.c_str(),
+						                  planned_action == Widelands::DiplomacyAction::kAcceptInvite ?
```
hessenfarmer, line 3493: I believe this will always fall to the second part as we only assign kAcceptInvite further down.

src/ai/defaultai.cc
```diff
@@ -3341,70 +3344,219 @@ void DefaultAI::diplomacy_actions(const Time& gametime) {
...
+                                       "invitation" :
+                                       "join request",
+						                  static_cast<unsigned int>(planned_opn), planned_other_score,
+						                  planned_log_append_text.c_str());
+						game().send_player_diplomacy(
+						   mypn,
+						   (planned_action == Widelands::DiplomacyAction::kAcceptInvite ?
```
hessenfarmer, line 3500: same here I could not find an assignment further up where this could have resulted

hessenfarmer (changes requested):
looks good in principle some small remarks though

src/ai/defaultai.cc
tothxa: Exactly.

src/ai/defaultai.cc
tothxa: I missed this one. But if we only have one bad ally, then the previous teamscore check also triggers earlier than before (-10 vs. -15), but I thought that makes sense. I'll add a check for this case to not overpenalise single bad ally.

src/ai/defaultai.cc
tothxa: Starting at line 3503:
```
				planned_action = pda.action == Widelands::DiplomacyAction::kInvite ?
                                Widelands::DiplomacyAction::kAcceptInvite :
                                Widelands::DiplomacyAction::kAcceptJoin;
```

src/ai/defaultai.cc
tothxa: It was annoying that the actions graph was always unreadabe because many accepts/denies were overlapping with the request. :) But also all checks depended on player ordering, which may sometimes favour one player over the other.

BTW, speaking of the graphs, should we include the scripts in `utils`?

src/ai/defaultai.cc
tothxa: Actually this didn't affect when we leave single ally, only the priority for the plan to leave. The whole block depends on the old check of `diploscore < -15`, and neither of the inner conditions are met then, so the `else` branch is always executed for single ally.

src/ai/defaultai.cc
hessenfarmer: yes I saw that but here in line 3491 we already evaluate for a value that might get assigned only later in the Program run. so here the DiplomacyAction never can be KAcceptInvite can't it?

tothxa: This is inside of the loop over pending diplomacy actions, and `planned_action` is defined outside, and will be finally used outside again. During the first execution it may be `kNoAction` or `kLeaveTeam`, then that may get overridden by `kAcceptInvite` or `kAcceptJoin`. During the later iterations, it could be any of these 4 values.

```
void diplomacy_actions(…) {
   …
   Wl::DiplAction planned_action = kNoAction;
   …
   if (<condition_to_leave>) {
      planned_action = kLeaveTeam;
   }
   …
   for (… pda : game().pending_diplomacy_actions()) {
      <evaluate action>
      accept = accept && <no plan or previously planned is better>
      if (!accept) {
         <log and execute deny>
      } else {
         if (planned_action != kNoAction) {
            <log replacing old action and execute its deny counterpart if not kLeaveTeam>
            <line 3491 is here>
         }
         <replace action>
         <now line 3504 is here:> planned_action = <new action>
      }
   }
   <execute planned_action if any>
}
```

hessenfarmer: ok now I get it. Lets have it then

src/ai/defaultai.cc
```diff
@@ -3334,77 +3337,229 @@ void DefaultAI::check_flag_distances(const Time& gametime) {
 
 // Dealing with diplomacy actions
 void DefaultAI::diplomacy_actions(const Time& gametime) {
+	// Make sure we use reasonably up to date stats
+	update_player_stat(gametime);
```
tothxa: I've just seen a stupid move by an AI because of stale stats in a test. :)

tothxa:
```
[01:54:19.707 game] DEBUG: AI Diplomacy: For player(1), the player(2) has the diploscore: 45
[01:54:19.707 game] DEBUG: AI Diplomacy: For player(1), the player(3) has the diploscore: 27
[01:54:19.707 game] DEBUG: AI Diplomacy: For player(1), the player(4) has the diploscore: 53
[01:54:21.077 game] DEBUG: AI Diplomacy: Player(2) plans to accept the invitation of player (1) with priority 17
[01:54:21.077 game] DEBUG: AI Diplomacy: Player(2) accepts the invitation of player (1) with diploscore 23.
[01:54:28.277 game] DEBUG: AI Diplomacy: For player(3), the player(1) has the diploscore: 6
[01:54:28.277 game] DEBUG: AI Diplomacy: For player(3), the player(2) has the diploscore: -34
[01:54:28.277 game] DEBUG: AI Diplomacy: For player(3), the player(4) has the diploscore: 23
[01:54:30.977 game] DEBUG: AI Diplomacy: Player(1) plans to leave team with priority 0
[01:54:30.977 game] DEBUG: AI Diplomacy: Player(1) leaves team (1) as last one.
```

If I'm not mistaken, that's less than the minimum allowed update interval of 15 seconds, so the above still does not make it impossible. Extreme bad luck for AI player 1, extreme good luck for me to catch problem. ;)

I made it to at least use consistent data in [63bdc6a](https://github.com/widelands/widelands/pull/5814/commits/63bdc6ab842d5efcdf6378c71e143ba007136476). I haven't checked thoroughly the rest of the function for similar possibilities though.

hessenfarmer: I am not too happy with calling this outside its own scheduler. How do you define stale stats? it is updated every 60 seconds anyway, So the maximum stale time can be 60 seconds., while diplomacy is checked at an interval of 90 to 119 seconds. If you really feel we should update the diplomacy score just before making the decisions then we should move this part to the diplomacy action part. Furthermore we already tried to only compare data that is from the same point in time.  
I think it is basically impossible to prevent the AI from doing weird things from time to time and a silly diplomacy move is no problem from my side as long as it does not happen too often. Remember the whole thing might be affected by the mutations applied at game startup anyway. 

Edit: what we could do is to add the call to diplomacy actions to the update stats task, and add some small time jitter there (i.e. update interval between 50 and 70 seconds). and by this delete the diplomacy task from  the scheduler

src/ai/defaultai.cc
```diff
@@ -3342,7 +3342,7 @@ void DefaultAI::diplomacy_actions(const Time& gametime) {
 
 	const Widelands::PlayerNumber mypn = player_number();
 	const Widelands::Player* me = game().get_player(mypn);
-	const Widelands::TeamNumber mytn = me->team_number();
+	const Widelands::TeamNumber mytn = player_statistics.get_team_number(mypn);
```
hessenfarmer: wouldn't this mean we are using outdated team information. As our team number might have been modified in the mean time by another player. Would it be good to base the evaluation forther down on such data.     
   
Edit: Furthermore I believe team_number wasn't the problem in the case from the log. me_alone is the culprit here. 

Edit2: I think the best solution for this issue is to recalculate team powers after a team change has been done. And to make this safe for the interaction with humans this might be best done in Diplomacy action code.

src/ai/defaultai.cc
tothxa: > I am not too happy with calling this outside its own scheduler.

I wasn't either, but I searched for it and it turned out that 2 functions in `defaultai_warfare.cc` do it too. (`check_enemy_sites()` and `check_militarysites()`)
git blame says it's been so since 2017.

> How do you define stale stats? it is updated every 60 seconds anyway, So the maximum stale time can be 60 seconds.

If you have a look at some logs, you can see that it's actually 17 to 19 seconds between stat updates of a player, probably because of the above. (it has a guard to prevent running it more often than once per 15 seconds)

So we have a more general problem here, that is out of scope for AI Diplo.

> Edit: what we could do is to add the call to diplomacy actions to the update stats task, and add some small time jitter there (i.e. update interval between 50 and 70 seconds). and by this delete the diplomacy task from the scheduler

I've thought about similar possibilities too. I'm not sure about threading and such, maybe we should just schedule it (at the end of every second run) and the other 2 from there, instead of directly calling? But the others are currently scheduled more often (19 and 15 secs). Can we increase that?

src/ai/defaultai.cc
tothxa: > wouldn't this mean we are using outdated team information. As our team number might have been modified in the mean time by another player. Would it be good to base the evaluation forther down on such data.

My reasoning was that if we can't prevent stale data, at least we should use it consistently. I thought I left a comment to that effect…
 
> Edit: Furthermore I believe team_number wasn't the problem in the case from the log. me_alone is the culprit here.

It's both of them interacting. The stale team number would have been 0, so leave wouldn't have been triggered.

> Edit2: I think the best solution for this issue is to recalculate team powers after a team change has been done. And to make this safe for the interaction with humans this might be best done in Diplomacy action code.

Actually the best solution would be to change the main diplomacy code to delete all pending requests by or to a player when their team changes. So again, out of scope for AI Diplo.

src/ai/defaultai.cc
hessenfarmer: > > I am not too happy with calling this outside its own scheduler.
> 
> I wasn't either, but I searched for it and it turned out that 2 functions in `defaultai_warfare.cc` do it too. (`check_enemy_sites()` and `check_militarysites()`) git blame says it's been so since 2017.      
    
I will have a look, that seems to be a bloody workaround then. 
> 
> > How do you define stale stats? it is updated every 60 seconds anyway, So the maximum stale time can be 60 seconds.
> 
> If you have a look at some logs, you can see that it's actually 17 to 19 seconds between stat updates of a player, probably because of the above. (it has a guard to prevent running it more often than once per 15 seconds)
> 
> So we have a more general problem here, that is out of scope for AI Diplo.
> 
> > Edit: what we could do is to add the call to diplomacy actions to the update stats task, and add some small time jitter there (i.e. update interval between 50 and 70 seconds). and by this delete the diplomacy task from the scheduler
> 
> I've thought about similar possibilities too. I'm not sure about threading and such, maybe we should just schedule it (at the end of every second run) and the other 2 from there, instead directly calling? But the others are currently scheduled more often (19 and 15 secs). Can we increase that?   
   
currently AI isn't threaded at all. i.e. all AI calcualtions run in the same thread as the complete Economy and logic code. Only the display code is threaded so far AFAIK. Therefore AI Code directly affects Performance of our code.   I will have a look in the 2 issues. However if they only call the function update_player_stat at such an interval anyway we could simply lengthen the update state scheduler interval and call update_player_stat() and diplomacy_actions() really in one task,

Edit: ok we effectively are executing the update_player_stat function every 15 seconds due to the call in check_militarysites. every other call will be superfluous imho. so the solution will be to have an update interval of 15 secs for the scheduler and delete all other calls. thus we guarantee a stale time of max 15 seconds.    
I think we will merge this and I'll fix that in the main PR.

src/ai/defaultai.cc
hessenfarmer: > > wouldn't this mean we are using outdated team information. As our team number might have been modified in the mean time by another player. Would it be good to base the evaluation forther down on such data.
> 
> My reasoning was that if we can't prevent stale data, at least we should use it consistently. I thought I left a comment to that effect…    
   
understood, but I still beleive this would not help for the specific issue. 
      
> 
> > Edit2: I think the best solution for this issue is to recalculate team powers after a team change has been done. And to make this safe for the interaction with humans this might be best done in Diplomacy action code.
> 
> Actually the best solution would be to change the main diplomacy code to delete all pending requests by or to a player when their team changes. So again, out of scope for AI Diplo.         
    
well Leaving is not dependent on a open action isn't it.       
   
We should ask @Noordfrees whether we can call a function defined for a specific computer player (namely the player_statistics.recalculate_team_powers() function as defined in default_ai.h  (defining the PLayers_strenghts Struct for the computerplayer) and ai_help_structs.h (defining the actual function of the struct)) from the CmdDiplomacy::execute() code

tothxa:
Meanwhile I ran a 7 AI players test in a debug build without problems. There were 2 cases of an accept replacing another, and several accepts replacing leaves.

[dipl_log_0319.zip](https://github.com/widelands/widelands/files/11011346/dipl_log_0319.zip)
[dipl_test_0319_replay.zip](https://github.com/widelands/widelands/files/11011347/dipl_test_0319_replay.zip)

I didn't follow the game, but when I sometimes glanced at it, the line-ups looked reasonable.

src/ai/defaultai.cc
tothxa: > understood, but I still beleive this would not help for the specific issue.

That's why I called it a mitigation, not a solution. :)
 
> well Leaving is not dependent on a open action isn't it.

True, but at least it's halfway detectable by comparing the current team with the stored one. Later on we have much less chance to detect outdated stats.
 
> We should ask @Noordfrees whether we can call a function defined for a specific computer player (namely the player_statistics.recalculate_team_powers() function as defined in default_ai.h (defining the PLayers_strenghts Struct for the computerplayer) and ai_help_structs.h (defining the actual function of the struct)) from the CmdDiplomacy::execute() code

I don't think such complications are needed. Only loop through pending actions right there in the accept handling code, and remove matching ones. Pending actions is always as fresh as it can be on our side.

hessenfarmer: Well I understood the problem above as Player 1 invited Player 2 and afterwards missed the acception of Player 2. So he left the team shortly after. I can't see any open pending action involved here.  
Only thing that should help is to recalculate all values realted to teams as soon as we change the teams. And as a team change can happen due to a human player as well we need to find a way to trigger the recalculation after every team change. 
for the current problem it would have been suitable to recalculate after the player 2 accepted. so as a first mitigation we might at least recalculate after an Ai triggered team change.

tothxa: Outdated pending actions are a similar possibility. I'm almost sure I saw that too in some of my tests. It's also something that is a more general issue (papered over for AIs by this PR actually), that also affects human players, even more than AIs, because they may postpone dealing with diplomacy requests for much longer. So this is worth solving in general, by the main diplomacy code. (at least I believe it's going to be simple)

As you said, these race conditions in `diplomacy_actions()` are not all that important, so we shouldn't go out of our way to solve them, only apply simple mitigations where possible, or just ignore them. So yes, please merge this as it is now (without mitigation) and then we can discuss more in the main PR if needed.

PR #5820, opened 2023-03-22
Watch window fixes and tweaks

### LINKED ISSUE #4689: Watch window jumps to center of screen

### LINKED ISSUE #5830: Single watch window mode does not work
**Describe the bug**
In single watch window mode the watch window only remembers the last map position.

**To reproduce**
1. Turn on "Use single watchwindow mode" in Options
2. Start a game
3. Start watching a map position
4. Start watching another map position
5. Change watch window back to view 1
6. See error

**Expected behavior**
The view should go back to the first position.

**Version:**
 - Widelands Version: v1.1 as well as 1.2~git26082 (fa28389@master)

**Additional context**
If noone noticed before, then probably noone actually uses single watchwindow mode. Could we remove it completely instead of fixing it? That would also mean one less config option. :)

tothxa:
**Type of change**
Bugfix / New feature

**Issue(s) closed**
Fixes #4689
Fixes #5830

**New behavior**
 * watch window positions are remembered, ie. if you close a watch window then open a new one, it will open in the previous position
 * watch windows can be scrolled by the keyboard arrow keys
 * watch windows can be closed by right-clicking inside too (not only on the border)

**Possible regressions**
Things related to map view key- and mouse handling and scrolling


frankystone:
Very nice that a new watch windows get slightly moved to bottom right, so it does not cover the formerly watch window.

> watch windows can be scrolled by the keyboard arrow keys

How can this be done?

What irritates me everytime: A watch window can't be closed with a right click, except one makes right click on the border. I know that right clicking is used to scroll the map inside the watch window, but can't there be check if a map is really moved? Something like: If right mouse button is released and not map moved then close window?

tothxa:
> > watch windows can be scrolled by the keyboard arrow keys
> 
> How can this be done?

Windows (and all UI elements) only handle keys when they have focus, ie. when they are selected. (Maybe we should have a focus-follows-mouse option? I prefer that anyway. ;) When a window is newly opened, it also has focus, until something else is clicked, so with this change you can more easily fine-tune the position and zoom for new watch windows.

> What irritates me everytime: A watch window can't be closed with a right click, except one makes right click on the border. I know that right clicking is used to scroll the map inside the watch window, but can't there be check if a map is really moved? Something like: If right mouse button is released and not map moved then close window?

I thought of changing all clicks to do this (currently when you press the left mouse button in a window, it both executes the `clicked()` action and starts dragging the window – that was the cause of this bug after all). AFAIK a click event should only be triggered on releasing the mouse, if a drag didn't happen. I chose not to because of accessibility issues, ie. users with disabilities often move the mouse a little between pressing and releasing the button. (and it happens sometimes with everybody) I think the movement threshold is normally handled by the system, but unfortunately SDL only gives us the raw mouse data. So I decided against it, but I'll try it for this specific case.

tothxa:
@Noordfrees  Please review https://github.com/tothxa/widelands/pull/1 whether it would be OK for master.

tothxa:
I converted it to draft because I will try to solve #5830 here too.

frankystone:
This works nicely :+1:

hessenfarmer (approved):
Code looks reasonable

PR #5821, opened 2023-03-23
Make Amazon treetop sentry placeable by scripts

tothxa:
**Type of change**
Bugfix

**Issue(s) closed**
Treetop sentry can now be placed by scripts.

**To reproduce**
*If it's a bugfix:*
1. Try to place a treetop sentry over a tree that allows it from a script
2. See error

**New behavior**
The problem was that `terraform_for_building()` used to remove immovables unconditionally. Now it checks whether the immovable is needed by the building.

**Possible regressions**
Placing buildings by scripts


src/logic/player.cc
```diff
@@ -69,8 +69,20 @@ void terraform_for_building(Widelands::EditorGameBase& egbase,
...
+			immovable->remove(egbase);
+		}
+	} else {
+		if (!c[0].field->get_immovable()->has_attribute(required_immovable)) {
+			// TODO(tothxa): ideally this should pick a matching immovable and place it
+			throw wexception(
+			   "Attempting to build %s at %dx%d - no immovable with required attribute %i found",
+			   descr->name().c_str(), location.x, location.y, required_immovable);
+		}
+		return;
```
tothxa: Shouldn't return here, because conquer happens in next round.

hessenfarmer:
as this is only called if place building is forced, the expectation is that we place a suitable immovable rather then throwing an error. 
So we should iterate over descriptions.immovables(), get the description at each position and check for has_attribute(), with the first suitable immovable found break the iteration and place it .

tothxa:
> as this is only called if place building is forced, the expectation is that we place a suitable immovable rather then throwing an error.

Done.

BTW, I didn't expect that we can force this: :smiley: 
shot0000

hessenfarmer (commented):
Code LGTM.
but for the new test shouldn't we check, whether place_building really worked ?

hessenfarmer (approved):
Thanks a lot for making this possible now.

PR #5823, opened 2023-03-23
5667 player end status fix

### LINKED ISSUE #5667: `PlayersManager::add_player_end_status` fails when player slots are closed
>   what():  vector::_M_range_check: __n (which is 3) >= this->size() (which is 3)
> 

What could cause that? Occurs only in release build. In debug build, this error doesn't accour.

I use Multplayer game, with only AI players (Empire, Barbarians and Europeans) and me as Observer. Game mode is landlord.

I use parameter --ai_training and --auto_speed.

### LINKED ISSUE #5809: Assertion failed when destroying opponent HQ in multiplayer
Tested with 1.2~git26073 (e082880@master) Debug on Ubuntu 20.04 and Windows 10. Save game and reply attached.

When the yellow player's HQ at coordinates 142,182,4 is destroyed in a multiplayer game, the game crashes. Sometimes the server, other times the client.

`../src/logic/playersmanager.cc:97: void Widelands::PlayersManager::add_player_end_status(const Widelands::PlayerEndStatus&, bool): Assertion 'status.player > 0 && status.player <= number_of_players_' failed.`

**To reproduce**
1. [Download](https://www.widelands.org/maps/new-england/) the "New England" map.
1. Start a LAN/direct multiplayer game with the attached save. You do not need additional players to join.
2. The yellow HQ has been (or will soon be) destroyed. Wait for 10 seconds or so.
3. Game crashes.
[assertion_failed_bug.zip](https://github.com/widelands/widelands/files/10947337/assertion_failed_bug.zip)

hessenfarmer:

**Type of change**
Bugfix

**Issue(s) closed**
Fixes #5667 
Fixes #5809

**How it works**
the players end status table is now always as big as kMaxPlayers (16) to make it possible to address it by playernumber as introduced in #5604 whic was the culprit for the bug

**Possible regressions**
everything with regard to winconditions and game summary and game speed setting possibility


src/network/gamehost.cc
```diff
@@ -1935,7 +1935,7 @@ bool GameHost::client_may_change_speed(uint8_t playernum) const {
 		return false;
 	}
 	const Widelands::PlayerEndStatus* pes =
-	   d->game->player_manager()->get_player_end_status(playernum + 1);
+	   d->game->player_manager()->get_player_end_status(playernum);
```
tothxa: I think this shouldn't have been changed. You don't change the meaning of it anywhere else. (I remember from reviewing #5604 that there are 2 kinds of player numbers, one is 0 based, the other is 1 based, and this is a conversion between them)

I ran a test, and it regresses on #5604 in a way that I believe is because of this: if host is player 1 and it loses, then the host can change the speed, but player 2 cannot. If host is player 2, then both player 1 and 2 can still change the speed after host / player 2 loses.

I've checked, and it turns out that `players_end_status_` is actually always indexed by `player - 1`.

tothxa: Yes, tested and restores correct behaviour.

```suggestion
	   d->game->player_manager()->get_player_end_status(playernum + 1);
```

tothxa (changes requested):

tothxa:
~Rest looks OK, fix confirmed with savegame from #5809.~

I tested it with LAN game with 3 instances, with player slots 3, 4 and 5 on a 5 player map, player 5 was host. First player 5 resigned. Then when player 3 resigned, the other 2 instances crashed with segfault. Instance of player 3 reported the end status correctly.

With players 3 and 4, player 3 gives up, both crash with same error:
```
../src/ui_basic/table.h:229: void* UI::Table<void*>::operator[](uint32_t) const: Assertion `i < entry_records_.size()' failed.
```

Backtrace:
```
#3  0x00007ffff59f21a2 in __assert_fail () from /lib/x86_64-linux-gnu/libc.so.6
#4  0x00005555556c1641 in UI::Table<void*>::operator[] (this=0x555559891a40, i=3) at ../src/ui_basic/table.h:229
#5  0x00005555556c168c in UI::Table<unsigned long>::operator[] (this=0x555559891a40, i=3)
    at ../src/ui_basic/table.h:521
#6  0x0000555555ba2987 in GameSummaryScreen::player_selected (this=0x5555581f9b80, entry_index=3)
    at ../src/wui/game_summary.cc:269
#7  0x0000555555ba05a7 in GameSummaryScreen::<lambda(uint32_t)>::operator()(uint32_t) const (
    __closure=0x5555598b4a38, i=3) at ../src/wui/game_summary.cc:119
#8  0x0000555555ba3761 in std::_Function_handler<void(unsigned int), GameSummaryScreen::GameSummaryScreen(InteractiveGameBase*, UI::UniqueWindow::Registry*)::<lambda(uint32_t)> >::_M_invoke(const std::_Any_data &, unsigned int &&) (__functor=..., __args#0=@0x7fffffff80b4: 3) at /usr/include/c++/8/bits/std_function.h:297
#9  0x00005555556e8a88 in std::function<void (unsigned int)>::operator()(unsigned int) const (
    this=0x5555598b4a38, __args#0=3) at /usr/include/c++/8/bits/std_function.h:687
#10 0x00005555557fcd6b in Notifications::Signal<unsigned int>::operator() (this=0x555559891c20, args#0=3)
    at ../src/notifications/signal.h:62
#11 0x000055555594de5e in UI::Table<void*>::select (this=0x555559891a40, i=3) at ../src/ui_basic/table.cc:591
#12 0x0000555555ba26cd in GameSummaryScreen::fill_data (this=0x5555581f9b80) at ../src/wui/game_summary.cc:254
#13 0x0000555555ba1759 in GameSummaryScreen::GameSummaryScreen (this=0x5555581f9b80, parent=0x55555880a3b0, 
    r=0x55555880c860) at ../src/wui/game_summary.cc:124
#14 0x0000555555aa086c in InteractiveGameBase::<lambda()>::operator()(void) const (__closure=0x7fffd813e240)
    at ../src/wui/interactive_gamebase.cc:707
#15 0x0000555555aa1e7f in std::_Function_handler<void(), InteractiveGameBase::show_game_summary()::<lambda()> >::_M_invoke(const std::_Any_data &) (__functor=...) at /usr/include/c++/8/bits/std_function.h:297
#16 0x00005555556fb57e in std::function<void ()>::operator()() const (this=0x7fffd813e240)
    at /usr/include/c++/8/bits/std_function.h:687
#17 0x00005555561392db in NoteThreadSafeFunction::<lambda()>::operator()(void) const (__closure=0x7fffd813e240)
    at ../src/base/multithreading.cc:92
#18 0x000055555613a047 in std::_Function_handler<void(), NoteThreadSafeFunction::instantiate(const std::function<void()>&, bool, bool)::<lambda()> >::_M_invoke(const std::_Any_data &) (__functor=...)
    at /usr/include/c++/8/bits/std_function.h:297
#19 0x00005555556fb57e in std::function<void ()>::operator()() const (this=0x7fffd813eb90)
    at /usr/include/c++/8/bits/std_function.h:687
#20 0x000055555592f90d in UI::Panel::handle_notes (this=0x55555880a3b0) at ../src/ui_basic/panel.cc:220
#21 0x000055555593043e in UI::Panel::do_run (this=0x55555880a3b0) at ../src/ui_basic/panel.cc:378
#22 0x00005555555f8156 in UI::Panel::run<UI::Panel::Returncodes> (this=0x55555880a3b0)
    at ../src/ui_basic/panel.h:147
#23 0x00005555557d7a26 in Widelands::Game::run (this=0x5555581c6cd0, 
    start_game_type=Widelands::Game::StartGameType::kMap, script_to_run="", prefix_for_replays="nethost")
    at ../src/logic/game.cc:777
#24 0x000055555588d4d4 in GameHost::run_callback (this=0x55555830ac90) at ../src/network/gamehost.cc:569
#25 0x000055555588c5e5 in GameHost::<lambda()>::operator()(void) const (__closure=0x5555586e69a8) at ../src/network/gamehost.cc:468
#26 0x0000555555898957 in std::_Function_handler<void(), GameHost::run()::<lambda()> >::_M_invoke(const std::_Any_data &) (__functor=...) at /usr/include/c++/8/bits/std_function.h:297
#27 0x00005555556fb57e in std::function<void ()>::operator()() const (this=0x5555586e69a8) at /usr/include/c++/8/bits/std_function.h:687
#28 0x00005555559627d8 in FsMenu::LaunchMPG::clicked_ok (this=0x5555586e2640) at ../src/ui_fsmenu/launch_mpg.cc:243
#29 0x000055555598ff07 in FsMenu::TwoColumnsFullNavigationMenu::<lambda()>::operator()(void) const (__closure=0x555559710f78) at ../src/ui_fsmenu/menu.cc:146
#30 0x0000555555991bd1 in std::_Function_handler<void(), FsMenu::TwoColumnsFullNavigationMenu::TwoColumnsFullNavigationMenu(FsMenu::MenuCapsule&, const string&, double)::<lambda()> >::_M_invoke(const std::_Any_data &) (__functor=...) at /usr/include/c++/8/bits/std_function.h:297
#31 0x00005555556fb57e in std::function<void ()>::operator()() const (this=0x555559710f78) at /usr/include/c++/8/bits/std_function.h:687
#32 0x00005555556fa2a3 in Notifications::Signal<>::operator()() const (this=0x5555586e3d30) at ../src/notifications/signal.h:62
#33 0x00005555559122a9 in UI::Button::handle_mouserelease (this=0x5555586e3b70, btn=1 '\001') at ../src/ui_basic/button.cc:381
#34 0x00005555559333cd in UI::Panel::do_mouserelease (this=0x5555586e3b70, btn=1 '\001', x=99, y=12) at ../src/ui_basic/panel.cc:1319
#35 0x0000555555933cc1 in UI::Panel::ui_mouserelease (button=1 '\001', x=99, y=12) at ../src/ui_basic/panel.cc:1532
#36 0x00005555555e7f42 in WLApplication::handle_mousebutton (this=0x55555691c300, ev=..., cb=0x55555672c520 <UI::Panel::do_run()::input_callback>) at ../src/wlapplication.cc:1047
#37 0x00005555555e79dd in WLApplication::handle_input (this=0x55555691c300, cb=0x55555672c520 <UI::Panel::do_run()::input_callback>) at ../src/wlapplication.cc:954
#38 0x0000555555930457 in UI::Panel::do_run (this=0x7fffffffa730) at ../src/ui_basic/panel.cc:382
#39 0x000055555598ad44 in UI::Panel::run<int> (this=0x7fffffffa730) at ../src/ui_basic/panel.h:147
#40 0x000055555597e256 in FsMenu::MainMenu::main_loop (this=0x7fffffffa730) at ../src/ui_fsmenu/main.cc:226
#41 0x00005555555e6aa3 in WLApplication::run (this=0x55555691c300) at ../src/wlapplication.cc:822
#42 0x00005555555e0ee9 in main (argc=2, argv=0x7fffffffe398) at ../src/main.cc:45
```

hessenfarmer:
i have teste some changes but it seems they were wild guesses. I am not sure yet what causes the crashes.

tothxa:
I think this should be solved in `playersmanager` only. I'll try turning `players_end_status_` into a `std::map` to handle non-continuous player numbers…

hessenfarmer:
> I think this should be solved in `playersmanager` only. I'll try turning `players_end_status_` into a `std::map` to handle non-continuous player numbers…    
   
From what I understand it is not the call to players_end_status that is crashing. it is the indexed call to the players table in game summary.cc, which has its own index, that is causing it.

tothxa:
> From what I understand it is not the call to players_end_status that is crashing. it is the indexed call to the players table in game summary.cc, which has its own index, that is causing it.

Yes, you're right. After some experimentation I believe this is a separate bug in `GameSummaryScreen::fill_data()`, also introduced by #5604. I thought I had an idea how to solve it, but it failed, so now I'm quite confused what might cause it. I'll do some deeper debugging on it.

hessenfarmer:
> > From what I understand it is not the call to players_end_status that is crashing. it is the indexed call to the players table in game summary.cc, which has its own index, that is causing it.
> 
> Yes, you're right. After some experimentation I believe this is a separate bug in `GameSummaryScreen::fill_data()`, also introduced by #5604. I thought I had an idea how to solve it, but it failed, so now I'm quite confused what might cause it. I'll do some deeper debugging on it.

I will try to do so as well tonight

tothxa:
The problem is that `players_table_` treats the index as if it was the player number in `player_selected()` and I believe also in `compare_status()`. The problem I found in `fill_data()` is also real, we need to keep track of assigned indexes separately from the player number used in the iteration. So that mapping needs to be stored too. I'm not sure whether we can do it in the table itself, or we need to create a separate vector or map for it. Actually it may make sense to do it right in players manager (maybe it already exists?), and use it everywhere for end statuses and possibly other uses.

tothxa:
See https://github.com/widelands/widelands/pull/5826#issuecomment-1485557381 and https://github.com/widelands/widelands/pull/5826#issuecomment-1485629593:

There are more places where player numbers are iterated only up to number of players in the game, so maybe it would be best to revert https://github.com/widelands/widelands/commit/2a27e73cf68a2394d18649112f35f7bddc909300 and do https://github.com/widelands/widelands/pull/5604#pullrequestreview-1169102400 (with map instead of vector :) but at this point I'm not sure which is easier: untangle https://github.com/widelands/widelands/pull/5604 or hunt down all downstreams that need updating.

PR #5824, opened 2023-03-23
Fix segfault in Panel::move_to_top()

### LINKED ISSUE #5822: Segfault in Panel::move_to_top() in the testsuite
**Describe the bug**
There are 2 similar segfaults in unrelated builds in panel.cc:621:43 in function move_to_top() after saving game according to stdout in both cases.

**To reproduce**
See https://github.com/widelands/widelands/actions/runs/4496332047/jobs/7915749110
and https://github.com/widelands/widelands/actions/runs/4496287708/jobs/7912371533?pr=5821

**Expected behavior**
No segfault. :)

**Version:**
 - OS: Ubuntu 20.04, clang 7 and 8
 - Widelands Version: 2 separate PRs with probably unrelated changes, so I believe it comes from master (https://github.com/widelands/widelands/commit/37c6fee60bbfe5c4cb1e3b4f453a1045c6d48856)

**Additional context**
The failures happened in different tests too, so it's probably a race condition.

tothxa:
**Type of change**
Bugfix

**Issue(s) closed**
Fixes #5822 hopefully...

**Possible regressions**
Z ordering issues of panels? Other crashes?


Noordfrees (commented):
If `p` is nullptr then the `p = p->next` will still segfault, so I guess it should be `break`?

tothxa:
> If `p` is nullptr then the `p = p->next` will still segfault

Oops, sorry. I thought I checked the iteration, but then I was wrong.

> so I guess it should be `break`?

Would that be enough? I'm not sure what can cause that and what is the best way to handle it.

tothxa:
I'm not sure if this is entirely safe, but in most cases it will result in a return, and retry in next run of `do_think()`. I think. But I'm sick of the failures in other builds.

Noordfrees (approved):

PR #5826, opened 2023-03-27
Fix player end status handling

### LINKED ISSUE #5667: `PlayersManager::add_player_end_status` fails when player slots are closed

### LINKED ISSUE #5809: Assertion failed when destroying opponent HQ in multiplayer

tothxa:
**Type of change**
Bugfix

**Issue(s) closed**
Fixes #5667
Fixes #5809

**To reproduce**
1. Start a game with closed player slots (test both with single and multiplayer)
2. Resign all players
3. master crashes, with these changes it works

**Possible regressions**
Win conditions, game summary, game speed setting

**Additional context**
Alternative and extension of #5823

### DIFF
--- a/src/game_io/game_player_info_packet.cc
+++ b/src/game_io/game_player_info_packet.cc
@@ -185,7 +185,7 @@ void GamePlayerInfoPacket::write(FileSystem& fs, Game& game, MapObjectSaver* /* 
 	const std::map<Widelands::PlayerNumber, Widelands::PlayerEndStatus>& end_status_map =
 	   game.player_manager()->get_all_players_end_status();
 	fw.unsigned_8(end_status_map.size());
-	for (auto it : end_status_map) {
+	for (const auto it : end_status_map) {
 		const Widelands::PlayerEndStatus& status = it.second;
 		fw.unsigned_8(status.player);
 		fw.unsigned_8(static_cast<uint8_t>(status.result));
--- a/src/game_io/game_player_info_packet.cc
+++ b/src/game_io/game_player_info_packet.cc
@@ -185,7 +185,7 @@ void GamePlayerInfoPacket::write(FileSystem& fs, Game& game, MapObjectSaver* /* 
 	const std::map<Widelands::PlayerNumber, Widelands::PlayerEndStatus>& end_status_map =
 	   game.player_manager()->get_all_players_end_status();
 	fw.unsigned_8(end_status_map.size());
-	for (const auto it : end_status_map) {
+	for (const auto& it : end_status_map) {
 		const Widelands::PlayerEndStatus& status = it.second;
 		fw.unsigned_8(status.player);
 		fw.unsigned_8(static_cast<uint8_t>(status.result));
--- a/src/game_io/game_player_info_packet.cc
+++ b/src/game_io/game_player_info_packet.cc
@@ -185,7 +185,7 @@ void GamePlayerInfoPacket::write(FileSystem& fs, Game& game, MapObjectSaver* /* 
 	const std::map<Widelands::PlayerNumber, Widelands::PlayerEndStatus>& end_status_map =
 	   game.player_manager()->get_players_end_status();
 	fw.unsigned_8(end_status_map.size());
-	for (auto it: end_status_map) {
+	for (auto it : end_status_map) {
 		const Widelands::PlayerEndStatus& status = it.second;
 		fw.unsigned_8(status.player);
 		fw.unsigned_8(static_cast<uint8_t>(status.result));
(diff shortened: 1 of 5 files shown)


src/logic/playersmanager.cc
```diff
@@ -87,34 +87,33 @@ Player* PlayersManager::add_player(PlayerNumber const player_number,
...
-
-	{
-		PlayerEndStatus& s = players_end_status_.at(status.player - 1);
-		if (!change_existing && s.player != 0) {
+	assert(status.player > 0);
+	const PlayerNumber& pn = status.player;
+	auto it = players_end_status_.find(pn);
+
+	if (it == players_end_status_.end()) {
+		players_end_status_.insert({pn, status});
```
tothxa: @Noordfrees Should/could this be emplace?

tothxa:
There are more places where player numbers are iterated only up to number of players in the game. One I've found, and where I could trigger the problem is saving the game. (last player's end status is not saved if there is a closed slot: start multiplayer on a map that allows at least 4 players, pick last player, close a lower numbered slot, set rest to AI, start game, give up, save, reload: your end status is not set)

tothxa:
OK, maybe it would be best to revert https://github.com/widelands/widelands/pull/5604/commits/2a27e73cf68a2394d18649112f35f7bddc909300 and do https://github.com/widelands/widelands/pull/5604#pullrequestreview-1169102400 (with map instead of vector :) but at this point I'm not sure which is easier: untangle #5604 or hunt down all downstreams that need updating.

tothxa:
OK, I think this is the best of both worlds. Based on the diff of https://github.com/widelands/widelands/commit/2a27e73cf68a2394d18649112f35f7bddc909300 there shouldn't be more affected users…

src/logic/playersmanager.h
```diff
@@ -92,12 +92,15 @@ class PlayersManager {
 	void add_player_end_status(const PlayerEndStatus& status, bool change_existing = false);
 
 	[[nodiscard]] const PlayerEndStatus* get_player_end_status(PlayerNumber player) const;
+	[[nodiscard]] const std::map<PlayerNumber, PlayerEndStatus>& get_players_end_status() {
```
hessenfarmer: may we name this function get_all_players_end_status to avoid logical error due to a silly typo. (or any other name more distinguishable)

hessenfarmer (approved):
nice and much more elegant solution that I could have imagined. 
one small nit inline. 
Code LGTM. not tested yet.
will close my PR then

src/logic/playersmanager.h
tothxa: Indeed. Thanks.

tothxa:
> nice and much more elegant solution that I could have imagined.

Thank you. :)

src/game_io/game_player_info_packet.cc
```diff
@@ -182,19 +182,15 @@ void GamePlayerInfoPacket::write(FileSystem& fs, Game& game, MapObjectSaver* /*
...
-	fw.unsigned_8(end_status_list.size());
-	for (const PlayerEndStatus* status : end_status_list) {
-		fw.unsigned_8(status->player);
-		fw.unsigned_8(static_cast<uint8_t>(status->result));
-		status->time.save(fw);
-		fw.c_string(status->info.c_str());
+	const std::map<Widelands::PlayerNumber, Widelands::PlayerEndStatus>& end_status_map =
+	   game.player_manager()->get_all_players_end_status();
+	fw.unsigned_8(end_status_map.size());
+	for (const auto it : end_status_map) {
```
hessenfarmer:
```suggestion
	for (auto it = end_status_map.begin(), it != end_status_map.end(), ++it) {
```
maybe this fixes clang tidy

hessenfarmer (approved):
tested it and works perfectly.

small suggstion for clang tidy

PR #5831, opened 2023-03-29
Remove libvorbis feature from vcpkg package list

### LINKED ISSUE #5827: Another dependency change on w*dows?
**Describe the bug**
Building vcpkg deps fail in the workflow with error message:
```
sdl2-mixer has no feature named libvorbis.
```

**To reproduce**
See https://github.com/widelands/widelands/actions/runs/4541812472/jobs/8006464995?pr=5826
and https://github.com/widelands/widelands/actions/runs/4543971478/jobs/8009387502?pr=5824

matthiakl:
**Type of change**
Bugfix 

**Issue(s) closed**
Fixes #5827 

**New behavior**
libvorbis is now installed as dependency of sdl2-mixer

### DIFF
--- a/install-dependencies.sh
+++ b/install-dependencies.sh
@@ -204,7 +204,7 @@ elif [ "$DISTRO" = "void" ]; then
 elif [ "$DISTRO" = "vcpkg" ]; then
    echo "Installing dependencies for vcpkg..."
    vcpkg install --disable-metrics $@ asio gettext libpng icu glbinding sdl2 sdl2-ttf \
-     sdl2-mixer[libvorbis,libflac,mpg123] sdl2-image[libjpeg-turbo,tiff] graphite2 \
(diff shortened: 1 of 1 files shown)


tothxa (approved):
I guess it's OK… seems to work.

Will it not be a problem if some commit will be built on an old runner? OTOH, at least they'll go green (amd64 deps were first built on old, then rebuilt on new one for the real build), so build failures will be gone (for another few days/hours? :/ ), and hopefully all runners will get updated soon anyway.

tothxa:
> … so build failures will be gone (for another few days/hours? :/ )

Or not even a second… Now this is really getting ridiculous: https://github.com/widelands/widelands/actions/runs/4568934895/jobs/8064593486 (OK, it's just a download error, but the timing of it…) :angry:

PR #5841, opened 2023-04-03
Fix Window::move_out_of_the_way()

### LINKED ISSUE #5840: Pop-up messages steal button clicks
**Describe the bug**

A pop-up message can steal interactions on the world and button clicks on windows which are near the bottom-middle of the screen. This causes the view to zoom to the pop-up location instead of the click going to a button on a window or a click on the world.

**To reproduce**

1. Attempt to click in lower-middle area of world just as a pop-up shows. Particularly evident with several consecutive pop-ups, as they consume more screen space
2. Click is stolen by pop-up and instead, view is moved to location

When dealing with many building windows (e.g. upgrading many mines) this can become very annoying. Building windows always appear to spawn in a random place, and when they spawn near the bottom-middle, the pop-ups also steal clicks from them.

**Expected behavior**

- Pop-ups have minimal window ordering, or...
- Windows do not open in any area where a pop-up can appear, or...
- Pop-ups are disabled when the Inbox is open.

**Version:**
 - OS: Ubuntu 20.04
 - Widelands Version: 1.2~git26093 (ed99eb9@master) Release

tothxa:
**Type of change**
Bugfix / Refactoring

**Issue(s) closed**
Fixes #5840
Fixes fastclick problem when new building or field action windows open overlapping the toolbar

Also fixes positioning of building and field action windows when the clicked building or field is close to the edge of the main window and the building or field action window is too wide or tall.

**To reproduce**
*If it's a bugfix:*
1. Have the toolbar pinned at the bottom of the screen (the default mode)
2. Position the map view so that the HQ is near the bottom of the game window. (HQ has very tall window, so it's easier to present the error with it)
3. Click on the HQ to open its window
4. in master: (you may see the HQ window first positioned partly outside the game window, then jump to not cover the toolbar and) the mouse pointer is moved below the game window; with this branch the HQ window appears at the final position at once, and the mouse pointer is positioned on the "show workarea" button as when the window is opened in the middle of the game window
5. also try with HQ at the top of the game window and with e.g. a lumberjack at the left or right edge of the game window

**Possible regressions**
hopefully none?

**Additional context**
The errors were introduced by #5785.

I refactored the handling of the toolbar configuration to make it available in `ui_basic` without having to make it depend on `wui`. I also converted toolbar size to use a styled size instead of a `constexpr` for the same reason (as it would have to be done sooner or later anyway).


Noordfrees (commented):
The problem still also happens when the toolbar is at the top:
grafik

tothxa:
> The problem still also happens when the toolbar is at the top

Oops. Fixed now.

I'm wondering about the rest of #5840. Should I add more checks to keep windows out of the area that may be occupied by message pop-ups? I think that would be overkill. Or they could be opened above or below the clicked position depending on whether it's below or above the middle of the window, but I actually think the positioning is slightly more predictable this way, and I personally prefer having them below. I believe the main problem (to me anyway) were the messed up fastclick positions.

Noordfrees (approved):

tothxa:
@simply-peachy Please check whether you accept this as a fix for #5840.

tothxa:
@Noordfrees Sorry for the late changes. Please check whether they're OK. With them this becomes a proper fix for #5840 as much as it's possible.

Noordfrees (approved):
LGTM :)

tothxa:
Thank you for the reviews.

PR #5849, opened 2023-04-10
added size to tooltip

### LINKED ISSUE #2099: Add a representation of 'size' to the menu of immovables
There are immovables with size = none and other. Whereas immovables with size = 'non' have no impact on the gameplay (e.g. they can be removed by placing a flag or a road) the other immovables couldn't be removed during game play. Such immovables can used as a blocker, e.g. preventing the player to build something there.

Whereas one can see the impact if he places an immovable on the map because the size of an immovable is reflected by adjusting the buildhelp, but this is like try and error.

So my suggestion is to reflect the size directly in the immovable menu, maybe as addition to the Tooltip?

if size == 'none'
   "Snowman: Removable"

if size == 'small'
   "Debris: Size small"

if size == 'medium'
   "Ruin: Size medium"

if size == 'big'
   "Ruin: Size big"

 * date created: 2018-01-30T13:51:58Z
 * owner: franku
 * the launchpad url was https://bugs.launchpad.net/bugs/1746248

frankystone:
**Type of change**
Bugfix

**Issue(s) closed**
Fixes #2099 

*If it's a bugfix:*
1. In editor open the immovables 
2. Tooltip has no representation of "size"

**New behavior**
The representation of size is now in the tooltip

**How it works**
Adds some string to the tooltip of immovables

**Possible regressions**
I guess no

**Screenshots**
immovablesize_01
immovablesize_02

For trees this may have to be changed:
immovablesize_03

**Additional context**
Don't know if the texts needs some changes for translation

### DIFF
--- a/src/editor/ui_menus/tool_place_immovable_options_menu.cc
+++ b/src/editor/ui_menus/tool_place_immovable_options_menu.cc
@@ -33,17 +33,9 @@ namespace {
 UI::Checkbox* create_immovable_checkbox(UI::Panel* parent,
                                         LuaInterface* lua,
                                         const Widelands::ImmovableDescr& immovable_descr) {
-	const Image* pic = immovable_descr.representative_image();
+      const Image* pic = immovable_descr.representative_image();
 
-	std::string tooltip = immovable_descr.descname();
-	std::string size = Widelands::BaseImmovable::size_to_string(immovable_descr.get_size());
-	std::string size_str;
-	if (size == "none") {
-		size_str = _(" (removable by e.g. placing roads)");
-		} else {
-		size_str = format(_(" (needs a %1$s building plot)"), size);
-	}
-	tooltip.append(size_str);
+   std::string tooltip;
 
 	// Get information about preferred terrains
 	if (immovable_descr.has_terrain_affinity()) {
@@ -51,7 +43,14 @@ UI::Checkbox* create_immovable_checkbox(UI::Panel* parent,
 		std::unique_ptr<LuaCoroutine> cr(table->get_coroutine("func"));
 		cr->push_arg(immovable_descr.name());
 		cr->resume();
-		tooltip.append(cr->pop_table()->get_string("text"));
+		tooltip = format(_("%1$s %2$s"), immovable_descr.descname(), cr->pop_table()->get_string("text"));
+	} else {
+		std::string size = Widelands::BaseImmovable::size_to_string(immovable_descr.get_size());
+		if (size == "none") {
+			tooltip = format(_("%1$s (%2$s)"), immovable_descr.descname(), _("removable by e.g. placing roads"));
+		} else {
+			tooltip = format(_("%1$s (%2$s %3$s %4$s)"), immovable_descr.descname(), _("needs a"), size, _("building plot"));
+		}
 	}
 
 	UI::Checkbox* cb =
--- a/src/editor/ui_menus/tool_place_immovable_options_menu.cc
+++ b/src/editor/ui_menus/tool_place_immovable_options_menu.cc
@@ -37,12 +37,12 @@ UI::Checkbox* create_immovable_checkbox(UI::Panel* parent,
 
 	std::string tooltip = immovable_descr.descname();
 	std::string size = Widelands::BaseImmovable::size_to_string(immovable_descr.get_size());
-	std::string size_str = "";
+	std::string size_str;
 	if (size == "none") {
 		size_str = _(" (removable by e.g. placing roads)");
 		} else {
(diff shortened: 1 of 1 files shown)


Noordfrees (changes requested):
Please do not concatenate translatable strings since languages may want to change the word order or whitespacing. Use `format()` instead:
`format(_("%1$s (%2$s)"), tooltip, _("removable by placing roads"))`

frankystone:
I guess i did wrong again :smile: 

tooltip is already the localized string (immmovable.descname) and is defined in line 38

hessenfarmer:
> I guess i did wrong again 😄

unfortunately yes. What @Noordfrees meant was not to use the append function (or simple concatenation by + or similar), but rather use format instead to concatenate the strings. Reason is that if you have 2 already translatable strings with the append function they are conctenated from left to right. This is always the case even for Right to Left Languages (RTL) where this makes the string not properly translatable. While with format you can switch the order. 
I tend to make such mistake often as well, if this helps. 

> 
> tooltip is already the localized string (immmovable.descname) and is defined in line 38

frankystone:
Ok, thanks for the additional explanation. Hopefully this is correct now.

Do we need a translators comment for `size`? Something like _"size is one of [small|medium|big]"_?

src/editor/ui_menus/tool_place_immovable_options_menu.cc
```diff
@@ -33,17 +33,24 @@ namespace {
...
 		cr->push_arg(immovable_descr.name());
 		cr->resume();
-		tooltip.append(cr->pop_table()->get_string("text"));
+		tooltip = format(_("%1$s %2$s"), immovable_descr.descname(), cr->pop_table()->get_string("text"));
+	} else {
+		std::string size = Widelands::BaseImmovable::size_to_string(immovable_descr.get_size());
+		if (size == "none") {
+			tooltip = format(_("%1$s (%2$s)"), immovable_descr.descname(), _("removable by e.g. placing roads"));
+		} else {
+			tooltip = format(_("%1$s (%2$s %3$s %4$s)"), immovable_descr.descname(), _("needs a"), size, _("building plot"));
```
Noordfrees: Better, but one abstraction too much now :) 
```c++
			tooltip = format(_("%1$s (needs a %2$s building plot)"), immovable_descr.descname(), size);
```

However I just checked and it turns out that `size_to_string` is not localized. So this would result in e.g. "Braucht einen small Bauplatz". Instead of calling `size_to_string`, better use a `switch` statement over `immovable_descr.get_size()` to create a different hardcoded string for each of the four size constants `Widelands::BaseImmovable::NONE|SMALL|MEDIUM|BIG`.

frankystone: > ... better use a switch statement over immovable_descr.get_size() to create a different hardcoded string for each of the four size constants

Not sure if i understand correctly: Do you mean overloading this method?

Noordfrees: No:
```c++
switch (immovable_descr.get_size()) {
case Widelands::BaseImmovable::NONE:
	tooltip = format(_("%1$s (removable by e.g. placing roads)"), immovable_descr.descname());
	break;
case Widelands::BaseImmovable::SMALL:
	tooltip = format(_("%1$s (needs a small building plot)"), immovable_descr.descname());
	break;
... // similar for MEDIUM and BIG
```

frankystone: Thanks for your advice!

src/editor/ui_menus/tool_place_immovable_options_menu.cc
```diff
@@ -33,17 +33,39 @@ namespace {
...
+				break;
+		case Widelands::BaseImmovable::MEDIUM:
+				tooltip = format(_("%1$s (needs a medium building plot)"), immovable_descr.descname());
+				break;
+		case Widelands::BaseImmovable::BIG:
+				tooltip = format(_("%1$s (needs a big building plot)"), immovable_descr.descname());
+				break;
+		}
+
+		std::string size = Widelands::BaseImmovable::size_to_string(immovable_descr.get_size());
```
hessenfarmer: these following lines are superfluous

hessenfarmer (changes requested):

src/editor/ui_menus/tool_place_immovable_options_menu.cc
frankystone: oh, sorry.

hessenfarmer (approved):
LGTM now.

Noordfrees (approved):

PR #5857, opened 2023-04-19
fix constructionsite images in lua place_building

### LINKED ISSUE #3356: Player.place_building places upgrading site for enhancable buildings
If you place a "empire_sentry" or "empire_barrier" constructionsite using lua like this:
   player:place_building(building, field, true)
Then the constructionsite always looks as if you had enhanced an empire blockhouse/outpost. It should instead look like a normal constructionsite for buildings that can be built directly.

 * date created: 2018-11-20T12:18:57Z
 * owner: nordfriese
 * the launchpad url was https://bugs.launchpad.net/bugs/1804221

### LINKED ISSUE #4659: Forcing a construction site uses wrong images
When placing a new building per [place_building](https://www.widelands.org/documentation/autogen_wl_bases/#wl.bases.PlayerBase.place_building) and forcing a construction site some construction sites are shown with a buildings image. This happens when placing construction sites for _upgraded_ buildings.

E.g.: Placing a construction site of an empire sentry will immediately show the image of an empire barracks on the map.

Attached a small scenario where construction sites of most military buildings were set. This produces the following:

cs_bug

On the left are construction sites for upgraded buildings, on the right are construction sites for base buildings. The construction sites on the left side should be shown like the construction sites on the right side.

**To Reproduce**
1. Start the attached scenario

**Expected behavior**
All construction sites should use the same image.

**Version:**
 - OS: archlinux
 - Widelands Version: [3031c3f@3849_win_conditions] (Debug)
 
[test_cs_bug.zip](https://github.com/widelands/widelands/files/5936695/test_cs_bug.zip)

hessenfarmer:
**Type of change**
Bugfix 

**Issue(s) closed**
Fixes #4659
fixes #3356

**New behavior**
for buildable buildings a direct constructionsite is created. for non buildabale buildings (i.e. pure enhancements) a csite for the base building is created and enhanced until the desired building. 

**Possible regressions**
lua placement of buildings in scenarios.


frankystone:
Works as expected now :-)

I tested this intensively with the scenario from https://github.com/widelands/widelands/issues/4659#issue-802674032

hessenfarmer:
> Works as expected now :-)
> 
> I tested this intensively with the scenario from [#4659 (comment)](https://github.com/widelands/widelands/issues/4659#issue-802674032)

if it is ok for you don't hesitate to approve it. ;-)

src/scripting/lua_bases.cc
```diff
@@ -933,6 +940,11 @@ int LuaPlayerBase::place_building(lua_State* L) {  // NOLINT - can not be made c
 			report_error(L, "Couldn't %s '%s' at (%d, %d)!", tempname.c_str(), name.c_str(),
 			             c->coords().x, c->coords().y);
 		}
+		if (constructionsite) {
+			for (uint8_t i = 0; i < steps; i++) {
+				dynamic_cast<Widelands::ConstructionSite*>(b)->enhance(get_game(L));
```
Noordfrees:
```suggestion
			Widelands::ConstructionSite* cs = dynamic_cast<Widelands::ConstructionSite*>(b);
			Game& game = get_game(L);
			for (uint8_t i = 0; i < steps; ++i) {
				cs->enhance(game);
```

Noordfrees (commented):
LGTM, one nit for performance

Noordfrees (dismissed):
LGTM. Not tested, but if the testsuite likes it it's good to go :)

hessenfarmer:
thanks for the quick review

hessenfarmer:
testsuite was green before the performance tweak now it is red. 
might be due to the test fails in the editor where we dont have a game. I don't know yet, cause I can't run the testsuite locally.
@Noordfrees any idea?

Edit: it is the missing game in the editor. tested it in a debug build with the console. In the enhance function game is only needed for logging so I'd propose to not pass game and do a simple log instead of log_warn_time

Noordfrees (commented):
Instead of removing the `Game` parameter, please instead generalize to `EditorGameBase`. Then we can keep the timestamp, and in Lua simply us `get_egbase` instead of `get_game`.

Noordfrees (approved):

Issue #5654, opened 2022-11-13
Highlight specific buildings on the minimap


frankystone:
Sometimes it would useful to highlight specific buildings on the minimap.

See: https://www.widelands.org/forum/topic/5685/

**Describe the solution you'd like**
Add an additional row below the minimap-button-row, containing dropdowns, each for one size of building, similar to the buildings menus. The dropdowns contain the buildings (name and/or picture) corresponding to the size. Once a building get highlighted the dropdown button should indicate that at least one building from this size is highlighted.

**Describe alternatives you've considered**
In the buildings statistics window add a checkbox "Show on minimap". After enabling the minimap shows the highlighted buildings.

**Additional context**
Highlighting can be done by using a square of size x, optionally with a modified playercolor.

A mockup showing highlighted (light bluish squares of size 5x5) Warehouses and ports and a row of additional buttons for dropdowns.

![warehouses_on_minimap_mockup](https://user-images.githubusercontent.com/6730556/201516752-6f1a36ca-ed93-46f7-ac8b-572ef48c9f24.jpg)

**Implementation steps**
- [ ] 
- [ ] 
- [ ]

saulkim:
What about something like this? Maybe the buildings statistics menu and expand slider code can be reused to keep it collapsible. 

![wl](https://user-images.githubusercontent.com/21152260/201549139-e413656c-30d5-41b7-bd75-3dd21ee049a9.png)

The building highlight can blink to help stand out.
![wlanimatedBlink](https://user-images.githubusercontent.com/21152260/201549537-f251968d-af3e-47a5-8307-b92aa2e65ede.gif)

frankystone:
Blinking is good idea! Then the square on the minimap hasn't to be so big. Probably also coloring isn't needed as it would be difficult to find suitable colors for each player.

Your proposal would make the window cover a lot of the map though. An i am not sure if this would be possible, code wise.

tothxa:
> Maybe the buildings statistics menu […] code can be reused

Why not just use the Building Statistics menu itself? (https://github.com/widelands/widelands/pull/5876#issuecomment-1533255130)

PR #5876, opened 2023-04-26
Minimap improvements

### LINKED ISSUE #1781: add minimap information layer for traffic jam
Add a minimap information layer for traffic jams (suggested by Tibor at [http://wl.widelands.org/forum/topic/42/]). This could be done by making the pixel red if the map node has a flag with at least 4 wares waiting. It is also worth considering to make the pixel redder the more wares are waiting at the flag. (To be implemented in calc_minimap_color, render.cc.)

 * date created: 2009-03-24T21:54:00Z
 * owner: sigra
 * the launchpad url was https://bugs.launchpad.net/bugs/536541

### LINKED ISSUE #2074: Highlight buildings under attack on the minimap
Highlight buildings under attack on the minimap, maybe with flashing dots.

 * date created: 2017-04-19T16:56:04Z
 * owner: gunchleoc
 * the launchpad url was https://bugs.launchpad.net/bugs/1684227

### LINKED ISSUE #2144: Highlight player starting positions and artifacts in the editor
It would be nice to highlight player starting positions and artifacts in the editor. Possible solutions

- Use the same technique as for the workarea overlays
- Mark them on the minimap

Forum discussion: https://wl.widelands.org/forum/topic/4569/

 * date created: 2019-06-19T04:41:38Z
 * owner: gunchleoc
 * the launchpad url was https://bugs.launchpad.net/bugs/1833345

### LINKED ISSUE #4466: Minimap: Terrain and owner layers lack contrast
The terrain layer can make it impossible to distinguish a field's owner on the minimap. Best seen with blue player and desert_water:
shot0007
The fields owned by the blue player are depicted in almost exactly the same colour regardless of whether the owner layer is switched on (left) or off (middle). The difference is not visible to the naked eye. The contrast needs to be increased.

hessenfarmer:
**Type of change**
New feature

**Issue(s) closed**
fixes #1781, fixes #2144, fixes #4466, fixes #2074

**New behavior**
Signal busy roads and flags with more then 5 wares on the minimap in the inverse color

**How it works**
Determine whether a flag or road is busy and inverse the display color

**Possible regressions**
Minimap visibilty

**Screenshots**
grafik

**Additional context**
things that still need implementation
- [x] make traffic overlay switchable
- [x] fix contrast between layers and playercolors (#4466)
- [x] highlight Buildings under attack (#2074)
- [ ] eventually highlight specific buildings (#5654)


src/graphic/minimap_renderer.cc
```diff
@@ -40,6 +41,25 @@ inline RGBColor blend_color(const RGBColor& c1, const RGBColor& c2) {
...
+inline RGBColor invert_color(const RGBColor& c) {
+	// assuming overflow and underflow
+
+	uint8_t r;
+	uint8_t g;
+	uint8_t b;
+
+	if (c.r > c.g && c.r > c.b && c.g < 200 && c.g + 50 < c.r) {
+		r = 0;
+		g = c.g;  // + (255 - c.g) / 2;
```
Noordfrees: What about this comment?

Noordfrees (commented):
Looks good :) 
Code LGTM, one comment inline

Noordfrees:
But maybe we should not draw this information about roads and flags belonging to enemy players?

src/graphic/minimap_renderer.cc
hessenfarmer: leftover from the old branch this branch needs cleaning anyway and I planned to fix some other Minimap issues here as well. I just created a draft PR to get the old thing into a compilable format and collect some feedback.

hessenfarmer:
> But maybe we should not draw this information about roads and flags belonging to enemy players?

We draw only what is in our visibility, imho. So if we would see it in real, near the border we could safely draw it in the minimap as well. But I am open for others opinions on this.

hessenfarmer:
ok, first problem where I might need some help:

how to identify a building which is attacked

Noordfrees:
Buildings do not currently have a concept of "under attack". When a soldier approaches an enemy militarysite, he notifies the site's attack target, which may choose to send out some defenders early and perhaps send a message. But the fact of the attack is not cached anywhere. This would need some non-trivial changes to militarysite and soldier code to implement properly.

---

Alternatively, how about instead highlighting the positions of soldiers who are currently in the process of attacking or defending something?

```c++
for (Widelands::Bob* bob = field->get_first_bob(); bob != nullptr; bob = bob->get_next_bob()) {
	if (bob->descr()->type() == Widelands::MapObjectType::SOLDIER && dynamic_cast<Widelands::Soldier*>(bob)->is_on_battlefield()) {
		color = ...;
		break;
	}
}
```

hessenfarmer:
> Buildings do not currently have a concept of "under attack". When a soldier approaches an enemy militarysite, he notifies the site's attack target, which may choose to send out some defenders early and perhaps send a message. But the fact of the attack is not cached anywhere. This would need some non-trivial changes to militarysite and soldier code to implement properly.
> 
> Alternatively, how about instead highlighting the positions of soldiers who are currently in the process of attacking or defending something?
> 
> ```c++
> for (Widelands::Bob* bob = field->get_first_bob(); bob != nullptr; bob = bob->get_next_bob()) {
> 	if (bob->descr()->type() == Widelands::MapObjectType::SOLDIER && dynamic_cast<Widelands::Soldier*>(bob)->is_on_battlefield()) {
> 		color = ...;
> 		break;
> 	}
> }
> ```

nice idea will try to use that

src/graphic/minimap_renderer.h
```diff
@@ -39,6 +39,8 @@ enum class MiniMapLayer {
 	ViewWindow = 64,
 	StartingPositions = 128,
 	Ship = 256,
+	Traffic = 512,
+	Artifacts = 1024,
```
Noordfrees: Those are becoming unreadable ;)
When you have so many constants please use bitshift notation:
```c++
 1 = 1 << 0
 2 = 1 << 1
 4 = 1 << 2
 8 = 1 << 3
16 = 1 << 4
...
```

hessenfarmer: ok good idea

hessenfarmer:
@Noordfrees could you please check whether the fix for contrasting terrain and playercolor would be sufficient for you? I would appreciate any tweak for brightening and darkening in any case.

src/graphic/minimap_renderer.cc
```diff
@@ -34,12 +35,70 @@ namespace {
...
+
+inline RGBColor brighten_color(const RGBColor& c, uint8_t percent) {
+	// we need higher range variables to avoid overflow in multiplying
+	uint16_t r;
+	uint16_t g;
+	uint16_t b;
+
+	uint16_t R = c.r;
+	uint16_t G = c.g;
+	uint16_t B = c.b;
```
Noordfrees: You only need one set of these variables – remove r,g,b and reassign to R,G,B instead

src/graphic/minimap_renderer.cc
```diff
@@ -51,32 +110,72 @@ inline RGBColor calc_minimap_color(const Widelands::EditorGameBase& egbase,
...
+				map.get_tln(f, &coord[2]);
+				map.get_trn(f, &coord[3]);
+				map.get_rn(f, &coord[4]);
+				map.get_brn(f, &coord[5]);
+				map.get_bln(f, &coord[6]);
+				for (const Widelands::FCoords& fc : coord) {
+					if (fc.field->get_immovable() != nullptr &&
+					    fc.field->get_immovable()->descr().has_attribute(
+					       Widelands::MapObjectDescr::get_attribute_id("artifact"))) {
+						color = kRed;
```
Noordfrees, line 169: add a `break;` here

src/wui/minimap.h
```diff
@@ -122,13 +123,16 @@ class MiniMap : public UI::UniqueWindow {
...
+	MiniMapLayer additional_button_impl_;
 	View view_;
 	UI::Button button_terrn;
 	UI::Button button_owner;
 	UI::Button button_flags;
 	UI::Button button_roads;
 	UI::Button button_bldns;
 	UI::Button button_ships;
+	UI::Button button_addtn;
+	UI::Button button_attck;
```
Noordfrees: Although it's nice if all variables are the same number of characters – please use more intuitive names ;)

Noordfrees (commented):
Thanks :) 
I played around with the colours and I think darken=40 and brighten=30 provides excellent contrast without distorting the terrain and owner colourings.

src/wui/minimap.cc
```diff
@@ -329,7 +359,7 @@ bool MiniMap::handle_mousewheel(int32_t x, int32_t y, uint16_t modstate) {
 	return true;
 }
 
-constexpr uint16_t kCurrentPacketVersion = 1;
+constexpr uint16_t kCurrentPacketVersion = 2;
```
tothxa: Do we need to increase this? It looks like the ships layer was added without bumping the packet version and it didn't cause any problem. But if it's increased, please add compatibility code for loading.

tothxa:
There seems to be an off by one error in the calculation of the button widths, because the zoom button is only a few pixels strip at the right edge of the window.

src/wui/minimap.cc
hessenfarmer: not sure I'll decide ig this is ready at the end

Noordfrees: There is no need to increase the version since older clients are also able to load the new-style format. Bits that are set here but not present in older versions will just be silently ignored.

hessenfarmer: ok I came to the same conclusion

hessenfarmer:
I tried to implement a highlighting of selected buildings. But this is beyond my abilities. I did not even manage to add the UI for selecting the buildings to the minimap. So I give up for now on this part

Noordfrees (commented):
Code LGTM

Will test this again tomorrow evening

Noordfrees:
Tested again – LGTM, just two more nits:
- Please also enable the traffic layer by default.
- Please don't show high-traffic areas for other players. I find this very distracting as this is completely unimportant information but is highlighted as if it were very significant.

Once those are addressed this is good to go :)

tothxa:
> I tried to implement a highlighting of selected buildings. But this is beyond my abilities. I did not even manage to add the UI for selecting the buildings to the minimap. So I give up for now on this part

But maybe we don't even need any new UI for this. Just highlight the buildings of the type selected in the Building Statistics window. Even a new button isn't needed: do it always when buildings are shown.

hessenfarmer:
> Tested again – LGTM, just two more nits:
> 
>     * Please also enable the traffic layer by default.

I disabled the layer on purpose as I thought the traffic information is not relevant yet in a new game. 
What is your rational to enable it?     

> 
>     * Please don't show high-traffic areas for other players. I find this very distracting as this is completely unimportant information but is highlighted as if it were very significant.

Ok. good point will do this. 
 
> 
> 
> Once those are addressed this is good to go :)

hessenfarmer:
> > I tried to implement a highlighting of selected buildings. But this is beyond my abilities. I did not even manage to add the UI for selecting the buildings to the minimap. So I give up for now on this part
> 
> But maybe we don't even need any new UI for this. Just highlight the buildings of the type selected in the Building Statistics window. Even a new button isn't needed: do it always when buildings are shown.       
    
we could do so but this would be the first time where we have a coupling between different windows for one single task. 
I will think about this and I may provide some prototyping in a seperate branch. In any case I think a buttton is needed to potentially declutter the display while not having to use the second window. 
What I envisaged in the beginning would have been a very lightweight building grid with only the building image and the name as tooltip that expands ovefrom the top in the minimap window to enabel / disable this and which closes either on selecting a building or on a special button if we would allow multiselection.

tothxa:
> > > [Re: UI for selecting the buildings to the minimap]
> > But maybe we don't even need any new UI for this. Just highlight the buildings of the type selected in the Building Statistics window. Even a new button isn't needed: do it always when buildings are shown.
> we could do so but this would be the first time where we have a coupling between different windows for one single task.

Well, it all depends on how you look at it. This can be considered an extension of the existing feature in the Building Statistics window that allows jumping between buildings of the selected type.

Adding the building selection to the minimap would just duplicate a lot of the b. stats window and be completely foreign to the rest of the functionality of the minimap.

As for coupling, at worst the selection could be stored in the `InteractiveBase` class to be accessible by both windows.

frankystone:
Like @hessenfarmer i am also against coupling of different windows. UI-wise this is bad style, imho. Just because the UI will be mixed up and the player doesn't know which (other) window might have an effect on other windows too. This is also difficult to explain.

#5654 hasn't much importance because with the buildings statistics window one can also move to the buildings of interest. Maybe just add a button to the place building menu which opens the buildings statistic window. This could also solve a feature which i really miss: Show the number of buildings a player has already of this type in the place building window (thought there was a bug report about this, but can't find it). With a button to show the buildings statistic this information is just one click away.

hessenfarmer:
well.
I have an idea relatively similar to what @frankystone supposes for the building menu. I will try to provide a proposal in a different branch after this got merged. 
Basic idea will be to open the  building statistics from the minimap when selecting the Highlight overlay. and probably close it when the layer gets deselected.   
last thing for this branch is if @Noordfrees accepts my rational for not activating the traffic by default or has a strong argument against it (or at least another reviewer who deem this better)

Noordfrees:
My reason why I want traffic default-enabled is that giving the player an overview of high-traffic areas is always useful and never harmful. In the early game, jams might not be a problem, but would it really *hurt* to have the traffic layer enabled anyway? Whereas leaving it of by default would make it easy for players never to notice that the layer exists, or simply to forget to enable it. If it is default-enabled then a player who would glance at the minimap for some different reason would also be alerted to his traffic jams, because the information is just there in a way where it is easily spotted but not intrusive, instead of the player actively having to request it with a small button that is easily overlooked.

hessenfarmer:
> My reason why I want traffic default-enabled is that giving the player an overview of high-traffic areas is always useful and never harmful. In the early game, jams might not be a problem, but would it really _hurt_ to have the traffic layer enabled anyway? Whereas leaving it of by default would make it easy for players never to notice that the layer exists, or simply to forget to enable it. If it is default-enabled then a player who would glance at the minimap for some different reason would also be alerted to his traffic jams, because the information is just there in a way where it is easily spotted but not intrusive, instead of the player actively having to request it with a small button that is easily overlooked.

ok valid argument. will implement it.

Noordfrees (commented):
The high traffic suppression for other players isn't working
grafik

`owner` is the owner of the current field, you instead want to compare with the player number of the interactive player. If there is no interactive player (i.e. spectator) then showing it for all players is fine.

hessenfarmer:
> The high traffic suppression for other players isn't working
> `owner` is the owner of the current field, you instead want to compare with the player number of the interactive player. If there is no interactive player (i.e. spectator) then showing it for all players is fine.

stupid mistake. Thanks for the help. should be fixed now

Noordfrees:
This fixed it for flags but the fix needs to be applied to roads as well

hessenfarmer:
sorry I cant concentrate well today

Noordfrees (approved):
Thanks, working great now :)

PR #5879, opened 2023-04-28
Disallow add-ons which are incompatible with the current Widelands version

### LINKED ISSUE #5801: Required WL version is not checked when add-ons are loaded
**Describe the bug**
Incompatible add-ons can be enabled, and this may cause Widelands to crash.

**To reproduce**
1. Unzip [this add-on](https://github.com/widelands/widelands/files/10880235/larger_fonts_updated.zip) in your add-ons directory.
2. Start Widelands 1.1
3. Try to enable the add-on

also:
1. Enable the add-on using current master
2. Start Widelands 1.1
3. See error

**Expected behavior**
1. It shouldn't be possible to enable incompatible add-ons.
2. On starting up, only compatible add-ons should be loaded
3. In general (much harder), more error protection is needed to prevent problems in add-ons crashing the game

**Version:**
 - Widelands Version: v1.1

**Additional context**
See https://www.widelands.org/forum/post/40215/

The `matches_widelands_version()` function is only ever used to set the style of the version string in the list of add-ons. Incompatible add-ons shouldn't even be listed by default, and even if shown, the enable checkbox and the download button should be disabled for them.

`preload_addons()` should also check version compatibility.

"under discussion" whether (and how much) it should be fixed in a v1.1 update. For consideration, here's a minimal patch for v1.1 to prevent incompatible add-ons: [addon_ver_check.patch.txt](https://github.com/widelands/widelands/files/10883584/addon_ver_check.patch.txt)

tothxa:
**Type of change**
Bugfix

**Issue(s) closed**
Fixes #5801

**New behavior**
Incompatible add-ons are prevented from being installed / upgraded / enabled / loaded.

**Possible regressions**
Add-on loading and management

**Additional context**
This is just the patch from the bug report.


src/ui_fsmenu/addons/rows_ui.cc
```diff
@@ -401,6 +402,7 @@ RemoteAddOnRow::RemoteAddOnRow(Panel* parent,
 		ctrl->install_or_upgrade(info_, false);
 		ctrl->rebuild(true);
 	});
+	install_.set_enabled(info->matches_widelands_version());
```
Noordfrees, line 406: I think developers should still be able to download the add-on, so maybe only disable the Install and Upgrade buttons in release builds

Noordfrees:
LGTM, tested and working. However now that we strictly enforce this, please modify the version checking logic for development version so that WL version `1.x~something` is acceptable for add-ons with a minimum version of `1.x` (e.g. master should be considered compatible with simple_map_generator which requires 1.2).

tothxa:
> please modify the version checking logic for development version so that WL version `1.x~something` is acceptable for add-ons with a minimum version of `1.x`

Done.

Noordfrees (approved):
Great, thanks :)

tothxa:
Thanks for the review.

Now I suppose we won't release a fix for v1.1. Should I still upload the [updated large fonts theme](https://www.widelands.org/forum/post/40215/) to the add-ons server knowing it makes 1.1 crash without an easy way to get rid of it?

Noordfrees:
I don't know to how much trouble you're willing to go with this add-on, but it's not too hard to serve a different theme definition depending on Widelands version. Pseudocode:
```lua
if
	(get_build_id():find("1.1") ~= 1) and
	(get_build_id():find("1.0") ~= 1)
then
	include(path.dirname(__file__) .. "init_for_v1.2_and_newer.lua")
else
	include(path.dirname(__file__) .. "init_for_v1.1_and_older.lua")
end
```

Note also that even though Widelands itself doesn't enforce compatibility in v1.1, the *server* on the other hand only lists add-ons that are actually compatible, unless the client explicitly asks for the complete list (using Ctrl+click). This is currently broken due to a typo which I'll fix right away.
Update: Fixed. Now you should not even see min=1.2 add-ons unless you Ctrl-Click the Refresh Remotes button. This also works in v1.1.

tothxa:
> I don't know to how much trouble you're willing to go with this add-on, but it's not too hard to serve a different theme definition depending on Widelands version. Pseudocode:

Thanks, that's a good trick indeed. But I guess my answer is "None.", because I hope there'll be more to do before v1.2 (or ideally v1.2 shouldn't even require such an addon anymore). So with:

> the _server_ on the other hand only lists add-ons that are actually compatible, unless the client explicitly asks for the complete list (using Ctrl+click).

I think I'll just upload it.

Issue #5549, opened 2022-09-07
Translation update script should push updated templates before pulling translations


tothxa:
**Describe the bug**
It takes  2 rounds of `utils/merge_and_push_translations.sh` runs to properly propagate source string changes to `.po` files. `.pot` and `*.po` are inconsistent between the 2 runs.

**To Reproduce**
1. Check diff of `po/widelands/widelands.pot` in 9a4ed5d (4 Sep): strings `src/ui_fsmenu/addons/packager_box.cc:372`, and `:373` are removed, but still present in `po/widelands/<any lang>.po`.
2. Check diff of `po/widelands/<any lang>.po` in a1e798f (7 Sep): the above strings are only removed by this commit

**Expected behavior**
`.po` files should be updated at the same time as the `.pot`.

**Additional context**
I think the reason is that local updates of both `.pot` and `.po` files are handled by `utils/buildcat.py`, so solving this will need heavy refactoring (ideally merging?) of the 2 scripts.

SimonHeimberg:
If I read correctly, utils/buildcat.py only updates the *pot* files (see also https://github.com/widelands/widelands/actions/runs/5538564887/jobs/10108629088#step:5:5502, it only lists pot files)
As I see it, *po* files are updated by transifex only. (See my understanding of utils/merge_and_push_translations.sh below.)
But we first download (the po files) from transifex and later upload the new templates (the pot files). This results in a roundtrip (see below).

So we should adapt utils/merge_and_push_translations.sh to
- first generate the translation templates
- then push the template to transifex (maybe only if it has changed, not date only)
- wait for transifex to update the translation files
  (no idea how long this takes, either wait in the script or start a job later)
- download the translations from transifex
- generate all other files
- commit the templates, the translations and the other generated files

There are other strategies to avoid a round trip.
(v2: downloading the template files from transifex as well (`tx -fas`), update the other files and committing them all, and finally updateing the template files from source and push them to transifex without committing to git)
(v3: ??)

***more details***

****I understand utils/merge_and_push_translations.sh like this****
- [tx pull -fa](https://github.com/widelands/widelands/blob/70f49b9/utils/merge_and_push_translations.sh#L63) # pull **a**ll translation files (*.po) from transifex
- python3 utils/update_authors.py # update authors in data/i18n/locales/*
- python3 utils/update_appdata.py # update appdata in xdg/
[...]

PR #5904, opened 2023-05-06
Build vcpkg packages weekly

matthiakl:
**Type of change**
Refactoring

**Issue(s) closed**
Re https://github.com/widelands/widelands/pull/5837#issuecomment-1493064388 
Re https://github.com/widelands/widelands/discussions/5880

**New behavior**
vcpkg packages are built and cached once per week on master to get more stability in other builds.

**How it works**
When the packages are built successfully the corresponding version is pushed into `vcpkg_ref` file and cached. All other builds will use the stored version and should always find a valid cache for it.
Instead of caching the packages, the install directory is cached to reduce cache size from ~340MB to ~270MB.

Before merging there are some TODOs to resolve. In my fork the scheduled builds did not start, maybe because it is not on master?

**Additional context**
See the builds here: https://github.com/matthiakl/widelands/actions?query=branch%3Avcpkg_cache_key


.github/workflows/check_vcpkg.yaml
```diff
@@ -0,0 +1,86 @@
+name: vcpkg Update
+concurrency:
+  group: vcpkg-${{ github.ref }}
+  cancel-in-progress: true
+on:
+  workflow_dispatch:
+  schedule:
+    # Weekly on Friday at 2:00 a.m.
+    - cron: '0 2 * * 5'
```
tothxa, line 9: This clashes with the translations update if it happens on Friday. It looks like translations usually take about 16 minutes, but there was one 3 weeks ago that went over 2 hours. vcpkg deps need about 1 hour.

I think the trivial solution would be to schedule translations update on fixed weekdays instead of every 3rd day.

tothxa:
> In my fork the scheduled builds did not start, maybe because it is not on master?

I gather you could make this your default branch temporarily to test the cron job.

Is there any other issue why this is still draft?

matthiakl:
> I think the trivial solution would be to schedule translations update on fixed weekdays instead of every 3rd day.

I specified explicit days for the update, twice a week.

> Is there any other issue why this is still draft?

No real issue but some other reasons:
- I wanted to check on the link failure and have the CI green before changing it (now fixed in #5914)
- to wait for one or two updates for testing (should be good now)
- to prevent merging before the adapting the push urls to master

.github/workflows/i18n.yaml
```diff
@@ -5,8 +5,8 @@ concurrency:
 on:
   workflow_dispatch:
   schedule:
-    # Daily at 2:00 a.m.
-    - cron: '0 2 */3 * *'
+    # Monday and Thursday at 2:00 a.m.
+    - cron: '0 2 * * 1,4'
```
Noordfrees: I'd prefer to have the translations update running more often than only twice a week to get translation changes into the game for testing faster, especially considering #5549 
```suggestion
    # Monday, Thursday, Saturday at 2:00 a.m.
    - cron: '0 2 * * 1,4,6'
```

.github/workflows/build.yaml
```diff
@@ -286,9 +243,9 @@ jobs:
     - name: Restore vcpkg and its artifacts.
       uses: actions/cache@v3
       with:
-        path: ${{ env.VCPKG_DEFAULT_BINARY_CACHE }}
+        path: ${{ env.VCPKG_ROOT }}/installed
         key: |
-          ${{needs.vcpkg-packages.outputs.vcpkg_key}}-${{ matrix.arch }}
+          ${{ hashFiles( './install-dependencies.sh', '.github/scripts/vcpkg_ref' ) }}-${{ matrix.arch }}
```
tothxa: Maybe instead of hashing the stored vcpkg_ref file, the commit ID from it could be included in the key directly, so the caches could be easily identified.

tothxa (approved):

.github/workflows/build.yaml
matthiakl: I restored the previous key format

Issue #5041, opened 2021-08-26
Capitalisation of Tribe names


alasiax:
Tribe names are not capitalised within the game consistently. I noticed this with the Amazons, but hessenfarmer mentioned it should be the case for all tribes (see [comment 37928 in the forum](https://www.widelands.org/forum/topic/2897/?page=40#post-37928)).

I'll start with Amazons and then go through the other tribes.

alasiax:
I tried to label and self-assign this, but I can't figure out how, or perhaps I don't have permissions?

aummishra:
I would like to contribute to this issue

alasiax:
Hi Aummishra, the more the merrier! Did you have a preferred tribe to start on?

Vamsikrishnapac:
Can I take this.

hessenfarmer:
anybody's help is welcome. But if 2 persons will do the same this might be a waste.  So it would be good if you agree on which tribe is done by whom.

aummishra:
I'm currently working on another issue, I'd not be able to work on this issue. You can assign it to anyone else. Thank you!

alasiax:
Thanks for letting us know Aummishra.

I'm happy to start with Amazons, since I'm working with those files anyway.

alasiax:
@hessenfarmer : we should create a new branch here for this one, right? Call it 5041_Caps_fixes ?

hessenfarmer:
> 
> 
> @hessenfarmer : we should create a new branch here for this one, right? Call it 5041_Caps_fixes ?

yes it is recomended to create a new branch in your fork and create a pul request from this branch. don't forget to pull in master frequently to stay in sync with it.  
the name of the branch is your free choice. however mentioning the bugnumber which should be fixed is good practise. What is more important then the branch name is the title of the PR. Cause this title is what appears in the changelog in the end.

jgsogo:
I've went through the code and I think this issue has already been fixed.

hessenfarmer:
you need to go through the data folder as well, and search for translatable texts ( they are denoted by a leading un derscore like _"text" or inside a format function. in a quick check for barabarians I found some not capitalised Instances in fri02 campaign in the texts.lua for example

PR #5910, opened 2023-05-10
Capitalisation of tribe names - Barbarians

jgsogo:
**Type of change**
Refactoring

**Issue(s) closed**
#5041 (Barbarians)

**New behavior**

I started with #5041, but found myself _fixing_ the name of the tribes everywhere following some criteria I wanted to share with you before submitting so many changes:

 * I'm modifying:
    * string constants, 
    * comments,
 * I'm NOT changing:
    * `.po` files. I assume they will be regenerated from the sources
    * `Changelog`

I'm not native english speaker, so I have some doubts regarding some of the changes. Help/comments on this PR would be greatly appreciated, I should be able to learn the rule and apply it everywhere (I'll try my best, I promise 😊 )

PR #5913, opened 2023-05-11
Capitalisation of tribe names - Empire

jgsogo:
**Type of change**
Refactoring

**Issue(s) closed**
#5041 (Empire)

**New behavior**

 * I'm modifying:
    * string constants, 
    * comments,
 * I'm NOT changing:
    * `.po` files. I assume they will be regenerated from the sources
    * `Changelog`


data/campaigns/fri02.wmf/scripting/mission_thread.lua
```diff
@@ -260,7 +260,7 @@ function supply_murilius()
    local wh = p1:get_buildings("frisians_warehouse_empire")[1]
    local hq = p2:get_buildings("empire_headquarters")[1]
 
-   -- transfer all wares that frisians and empire have in common
+   -- transfer all wares that frisians and the Empire have in common
```
tothxa, line 263:
```suggestion
   -- transfer all wares that Frisians and the Empire have in common
```

data/tribes/initialization/frisians/units.lua
```diff
@@ -1385,7 +1385,7 @@ wl.Descriptions():new_tribe {
       {
          name = "wheatfield_medium",
          helptexts = {
-            -- TRANSLATORS: Helptext for an empire/barbarian immovable usable by frisians: Wheat field
+            -- TRANSLATORS: Helptext for an Empire/barbarian immovable usable by frisians: Wheat field
```
tothxa:
```suggestion
            -- TRANSLATORS: Helptext for an Empire/Barbarian immovable usable by Frisians: Wheat field
```

data/campaigns/fri02.wmf/scripting/mission_thread.lua
jgsogo: This PR is for the Empire, the one for the Frisians will go next... 😄

data/tribes/initialization/frisians/units.lua
jgsogo: This PR is for the Empire, the one for the Frisians will go next... 😄 , and the change for Barbarian is already included [here](https://github.com/widelands/widelands/pull/5910/files#diff-2a1dac2825a3dda7258dc30da855da9073e5db4858d48bf825ccf3facf3eecd1)

hessenfarmer: I understand the intention, but for drive by fixes in the same line or block it might be better to do them on the fly, because otherwise you get conflicts as it is the case now.
we are not that pedantic on such drive by fixes

hessenfarmer (approved):

hessenfarmer:
@jgsogo please fix the conflict and update with master then this can go in

PR #5914, opened 2023-05-12
Resolve static libraries during configuration

matthiakl:
**Type of change**
Refactoring

**Issue(s) closed**
Fixes MSVC linking issues

**New behavior**
Instead of relying on the linker to find the libraries, cmake searches for them during configuration and stores the full paths.

**Additional context**
It also works for mingw, maybe it could even be used to remove the hardcoded library paths?


Noordfrees (approved):

hessenfarmer:
after this change the x86 builds in appveyor and my environment fail

matthiakl:
In appveyor it compiles and links fine, but a test fails, so I am not sure that it is related to this change. Do you have the same error in your environment?

hessenfarmer:
I saw the appveyor fail so I did a build before merging this in my local repo, which went fine and after the change it fails.

PR #5917, opened 2023-05-15
Disable ignored clang-tidy checks in the config file

tothxa:
**Type of change**
Refactoring / Bugfix

**New behavior**
All checks that are suppressed in `check_clang_tidy_results.py` are now disabled in `.clang-tidy` for some speed-up of the workflow run and to reduce (practically eliminate on clean runs) the amount of output (thus cache size for a follow-up PR). They can still be enabled on the command line when needed.

I also did some clean-up in `check_clang_tidy_results.py` to fix reporting failed checks with multiple names.

**Possible regressions**
clang-tidy checks and reporting

**Additional context**
This is a side project of my work to enable caching of the clang-tidy workflow, re #5880.


hessenfarmer:
well this is kind of what I suggested here:     
  
https://github.com/widelands/widelands/discussions/5262#discussioncomment-2211726

tothxa:
> well this is kind of what I suggested here:
> 
> [#5262 (comment)](https://github.com/widelands/widelands/discussions/5262#discussioncomment-2211726)

Yes, that too, but to me that discussion is more about which tests to clean up and which to ignore. I'm not sure I want to debate here which additional tests we can disable. But if @Noordfrees agrees, I'd be happy to add `-abseil*,-altera*,-linuxkernel*,-zircon`. (though only abseil has a meaningful number of checks in the workflow according to the log)

hessenfarmer:
> > well this is kind of what I suggested here:
> > [#5262 (comment)](https://github.com/widelands/widelands/discussions/5262#discussioncomment-2211726)
> 
> Yes, that too, but to me that discussion is more about which tests to clean up and which to ignore. I'm not sure I want to debate here which additional tests we can disable. But if @Noordfrees agrees, I'd be happy to add `-abseil*,-altera*,-linuxkernel*,-zircon`. (though only abseil has a meaningful number of checks in the workflow according to the log)

I would be happy about this and prefer to activate more checks relevant for us.

tothxa:
Well, clearing some more desired but currently disabled checks is certainly not the goal of this PR.

hessenfarmer:
> Well, clearing some more desired but currently disabled checks is certainly not the goal of this PR.      
    
that was not my intention. I just wanted to state that every cycle spend on more useful checks would be a good idea in the future.
so if this is green in the end it should be good to go from my side. will do a formal review then

hessenfarmer (approved):
lgtm. and we already saved approx half an hour per run. 
However there is still potential in disabling the alias for checks were we already check the original check. (i.e. hicpp has only 5 genuine checks from which we do not perform 3 already so excluding hicpp completely and only activating the 2 checks we perform (exception-baseclass and no-assembler) wozuld result in the same level of checking with less checks to perform by removing redundant checks imho.

But your decision whether to go this way.

tothxa:
> lgtm. and we already saved approx half an hour per run. However there is still potential in disabling the alias for checks were we already check the original check. (i.e. hicpp has only 5 genuine checks from which we do not perform 3 already so excluding hicpp completely and only activating the 2 checks we perform (exception-baseclass and no-assembler) wozuld result in the same level of checking with less checks to perform by removing redundant checks imho.

I'm not sure they're actually done separately, because they're reported together. So maybe this doesn't actually matter. I'll try to test it.
*edit: Did the test, and it matters. I didn't expect that `clang-tidy` would be this stupid…*

Is there a way to list such aliases?
*edit2: I've found https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/list.html I'll disable all aliases then, but then that'll have to be maintained manually on upgrades…*

tothxa:
Please check the last change.

Don't be misled by the time. It ran parallel with the matrix phase of another workflow, so this is pretty much the worst case time now.

.clang-tidy
```diff
@@ -100,8 +84,128 @@ Checks: >-
...
+  -clang-analyzer-unix.MismatchedDeallocator,
+  -clang-analyzer-unix.Vfork,
+  -clang-analyzer-unix.cstring.BadSizeArg,
+  -clang-analyzer-unix.cstring.NullArg,
+  -cppcoreguidelines-avoid-c-arrays,
+  -cppcoreguidelines-avoid-magic-numbers,
+  -cppcoreguidelines-c-copy-assignment-signature,
+  -cppcoreguidelines-explicit-virtual-functions,
+  -cppcoreguidelines-non-private-member-variables-in-classes,
+  -fuchsia-header-anon-namespaces,
```
hessenfarmer: redundant as we already excluded them all, but  might be easier to maintain

.clang-tidy
```diff
@@ -100,8 +84,128 @@ Checks: >-
...
+  -cert-flp37-c,
+  -cert-msc30-c,
+  -cert-msc32-c,
+  -cert-oop11-cpp,
+  -cert-oop54-cpp,
+  -cert-pos44-c,
+  -cert-pos47-c,
+  -cert-sig30-c,
+  -cert-str34-c,
+  -clang-analyzer-core.CallAndMessage,
```
hessenfarmer: I am pretty much sure we are not running the Clang Static Analyzer as it is a separate tool and we never got any errors for the checks of clang-analyzer* we had disabled before from the static analyzer. 
So please revert all changes for the clang-analyzer* checks. 
We just need to ensure not to enable the static Analyzer checks

.clang-tidy
```diff
@@ -100,8 +84,128 @@ Checks: >-
...
+  -cppcoreguidelines-avoid-c-arrays,
+  -cppcoreguidelines-avoid-magic-numbers,
+  -cppcoreguidelines-c-copy-assignment-signature,
+  -cppcoreguidelines-explicit-virtual-functions,
+  -cppcoreguidelines-non-private-member-variables-in-classes,
+  -fuchsia-header-anon-namespaces,
+  -google-readability-braces-around-statements,
+  -google-readability-function-size,
+  -google-readability-namespace-comments,
+  -hicpp-avoid-c-arrays,
```
hessenfarmer, line 130: for the hicpp checks it would be easier to use:
-hicpp*,
hicpp-no-assembler,
hicpp-exception-baseclass,

(disable all of them and only enable the 2 we are currently checking)

hessenfarmer (changes requested):
I beleive it is not that easy as there are differences.
at least the clang analyzer issue needs to be fixed to not loose checks imho

.clang-tidy
tothxa: Actually the log says that they were still enabled, so it looks like they can't be disabled? So I'll take them out, to avoid misleading readers.

(but `-*` disables them, so I don't understand… OK, looks like they go together with the non-alias clang core checks?)

.clang-tidy
tothxa: I think it's better to not disable a class that may have interesting checks. This way we may get new useful tests with a new clang-tidy version even before the config is updated. The performance loss of some new superfluous aliases is less important IMO.

src/scripting/lua_bases.h
```diff
@@ -108,7 +108,9 @@ class LuaPlayerBase : public LuaBasesModuleClass {
 	 * Lua methods
 	 */
 	CLANG_DIAG_RESERVED_IDENTIFIER_OFF
```
hessenfarmer, line 110: it seems like this Macro does not work so can it be deleted. maybe replacew by NOLINTBEGIN and NOLINTEND

src/scripting/lua_bases.h
```diff
@@ -108,7 +108,9 @@ class LuaPlayerBase : public LuaBasesModuleClass {
...
+	// NOLINTNEXTLINE(bugprone-reserved-identifier,cert-dcl37-c,cert-dcl51-cpp)
```
hessenfarmer, line 111: as we disabled the redundant cert checks we can safely only list the bugprone check here

src/scripting/lua_bases.h
tothxa: I don't know. It looks like it sets a compiler flag, so it looks different from NOLINT.

src/scripting/lua_bases.h
tothxa: Users may provide their own list of checks, don't they?

hessenfarmer: ok good argument

hessenfarmer:
ok should be good to go

hessenfarmer (approved):

tothxa:
It's still 48 more checks than 6340b24, so let's see how it turns out...

```
clang-analyzer-cplusplus.Move
clang-analyzer-cplusplus.NewDelete
clang-analyzer-cplusplus.NewDeleteLeaks
clang-analyzer-deadcode.DeadStores
clang-analyzer-nullability.NullPassedToNonnull
clang-analyzer-nullability.NullReturnedFromNonnull
clang-analyzer-nullability.NullableDereferenced
clang-analyzer-nullability.NullablePassedToNonnull
clang-analyzer-osx.API
clang-analyzer-osx.SecKeychainAPI
clang-analyzer-osx.cocoa.AtSync
clang-analyzer-osx.cocoa.ClassRelease
clang-analyzer-osx.cocoa.Dealloc
clang-analyzer-osx.cocoa.IncompatibleMethodTypes
clang-analyzer-osx.cocoa.NSAutoreleasePool
clang-analyzer-osx.cocoa.NSError
clang-analyzer-osx.cocoa.NilArg
clang-analyzer-osx.cocoa.ObjCGenerics
clang-analyzer-osx.cocoa.RetainCount
clang-analyzer-osx.cocoa.SelfInit
clang-analyzer-osx.cocoa.SuperDealloc
clang-analyzer-osx.cocoa.UnusedIvars
clang-analyzer-osx.cocoa.VariadicMethodTypes
clang-analyzer-osx.coreFoundation.CFError
clang-analyzer-osx.coreFoundation.CFNumber
clang-analyzer-osx.coreFoundation.CFRetainRelease
clang-analyzer-osx.coreFoundation.containers.OutOfBounds
clang-analyzer-osx.coreFoundation.containers.PointerSizedValues
clang-analyzer-security.FloatLoopCounter
clang-analyzer-security.insecureAPI.DeprecatedOrUnsafeBufferHandling
clang-analyzer-security.insecureAPI.UncheckedReturn
clang-analyzer-security.insecureAPI.bcmp
clang-analyzer-security.insecureAPI.bcopy
clang-analyzer-security.insecureAPI.bzero
clang-analyzer-security.insecureAPI.getpw
clang-analyzer-security.insecureAPI.gets
clang-analyzer-security.insecureAPI.mkstemp
clang-analyzer-security.insecureAPI.mktemp
clang-analyzer-security.insecureAPI.rand
clang-analyzer-security.insecureAPI.strcpy
clang-analyzer-security.insecureAPI.vfork
clang-analyzer-unix.API
clang-analyzer-unix.Malloc
clang-analyzer-unix.MallocSizeof
clang-analyzer-unix.MismatchedDeallocator
clang-analyzer-unix.Vfork
clang-analyzer-unix.cstring.BadSizeArg
clang-analyzer-unix.cstring.NullArg
```

Of these, NewDelete and NewDeleteLeaks is filtered by check_clang_tidy_results.py.

tothxa:
Sorry for this last one, but you made me look deeper into all this. ;)

hessenfarmer:
thanks for looking into it a little bit deeper. this result was what I always had in mind. and a reduction of 40 to 50 minutes to our longest CI task is definitly a big improvement

tothxa:
Thank you for the reviews.

> and a reduction of 40 to 50 minutes to our longest CI task is definitly a big improvement

The zipped log archive download size is also reduced from ~190MB to ~27MB. These checks produced over 700MB garbage per run that noone ever looked at (included twice in the zip). In contrast, one testsuite instance ~with full debug logging~ produces 11MB output. *(edit: sorry, stdout is not actually stored for successful tests, this is only the build/install/uninstall log)*

PR #5921, opened 2023-05-21
Remove hard-coded library paths for MinGW builds

### LINKED ISSUE #5920: msys2 x86 builds broken
**Describe the bug**
Since #5914 the x86 (32bit) builds from msys2 are broken. In appveyor (and my own environment) the economy tests fail. 
Furthermore the downloadable daily is not running after installation,

**To reproduce**
1. download x86 daily msys2 build and try to run
2. or see log in appveyor

**Version:**
 - OS: [windows 10 64-bit"]
 - Widelands Version: [dev 1.2]

matthiakl:
**Type of change**
Bugfix

**Issue(s) closed**
Fixes #5920

**New behavior**
The new wl_add_static_libs function was not constrained to static libraries, so it linked mingw build dynamically.
By setting CMAKE_FIND_LIBRARY_SUFFIXES, we can instruct CMake to only accept static libraries. This also allows the removal of hard-coded library paths.

**How it works**
Instruct CMake to only search for static libraries based on their suffixes (.a for mingw, .lib for msvc).


CMakeLists.txt
```diff
@@ -122,30 +122,11 @@ if (WIN32)
...
-      set(SDL2MAIN_LIBRARY "C:/msys64/mingw32/lib/libSDL2main.a" CACHE INTERNAL "")
-      set(ICU_UC_LIBRARY_RELEASE "C:/msys64/mingw32/lib/libicuuc.a" CACHE INTERNAL "")
-      set(ICU_DT_LIBRARY_RELEASE "C:/msys64/mingw32/lib/libicudt.a" CACHE INTERNAL "")
-      set(Intl_LIBRARY "C:/msys64/mingw32/lib/libintl.a" CACHE INTERNAL "")
-    endif()
+  if(MSVC)
+    set(CMAKE_FIND_LIBRARY_SUFFIXES ".lib")
+  else()
+    set(CMAKE_FIND_LIBRARY_PREFIXES "lib")
+    set(CMAKE_FIND_LIBRARY_SUFFIXES ".a" ".dll.a")
```
hessenfarmer: the dll.a will find the dynamic library linking files. afaik.    
How do we ensure the correct library for a 64bit build or 32 bit build will be found and used

CMakeLists.txt
```diff
@@ -433,6 +414,9 @@ endif(USE_FLTO_IF_AVAILABLE STREQUAL  "yes")
 
 IF (WIN32)
   add_definitions(-D_WIN32_WINNT=0x0601)
+  wl_add_static_libs(SDL_IMG_STATIC_LIBS jpeg tiff lzma)
```
hessenfarmer: I don't see the logic when to add the libs, as all relvant add instructions are cluttered around in this code and not all of them are dependent on the BUILD_WINSTATIC Option

hessenfarmer (approved):
although I have some questions and don't really undertand every bit of it it seems to work. 
So I approve

CMakeLists.txt
matthiakl: Thanks for the review, I tried to clean it up and get it organized.

CMakeLists.txt
matthiakl: > the dll.a will find the dynamic library linking files. afaik.

Oh, you are right, we can remove it.

> How do we ensure the correct library for a 64bit build or 32 bit build will be found and used

CMake uses CMAKE_SIZEOF_VOID_P as we did to distinguish the architecture (cf. [find_library documentation](https://cmake.org/cmake/help/latest/command/find_library.html)). I checked the correct resolution in a previous build [here](https://github.com/matthiakl/widelands/actions/runs/5037018561/jobs/9034014353#step:4:58) and [here](https://github.com/matthiakl/widelands/actions/runs/5037018561/jobs/9034014382#step:4:56).
> Query the registry matching the architecture specified by the CMAKE_SIZEOF_VOID_P variable. If not defined, fall back to HOST view.

PR #5922, opened 2023-05-23
Cache clang-tidy

tothxa:
**Type of change**
New feature / Refactoring

**Issue(s) closed**
Re #5880 

**New behavior**
clang-tidy output is saved in a cache, so only changed files have to be re-checked. The cache size is very small as long as the default config is used (uncleared checks are disabled), so it keeps results from previous commits as well with 2 months retention time. If the configuration changes (e.g. enabled checks or clang-tidy version), the entire cache is wiped.

Local runs can also be cached. If command line options change clang-tidy behaviour, then the corresponding cache will be stored in a subdirectory of the main cache separately for repeated runs with the same options.

**Possible regressions**
clang-tidy, workflows

**Additional context**
Test runs can be seen at https://github.com/tothxa/widelands/actions/workflows/build.yaml (they're shown as failed because I included an always failing job to disable other tests)


utils/run-clang-tidy.py
```diff
@@ -122,33 +131,302 @@ def make_absolute(f, directory):
...
+        # Custom subdirs have to be cleaned up manually on upgrades.
+        if custom:
+            self.dir = os.path.join(self.dir, '__custom_args', self.config_hash)
+            if not os.path.isdir(self.dir):
+                os.mkdir(self.dir)
+
+        self.passed_file = os.path.join(self.dir, 'passed')
+        self.failed_dir = os.path.join(self.dir, 'failed')
+        if not os.path.isdir(self.dir):
+            os.mkdir(self.dir)
```
matthiakl: This is already done above.
```suggestion
```

utils/run-clang-tidy.py
```diff
@@ -122,33 +131,302 @@ def make_absolute(f, directory):
...
+
+
+def get_hash(text_to_hash):
+    return hashlib.sha256(text_to_hash.encode('utf-8')).hexdigest()
+
+
+extra_before = []
+extra_after = []
+
+def pp_hash(file, command, dir):
```
matthiakl, line 340: Can you add a short comment, what this function does? It seems to be the core of the cache key generation.

utils/run-clang-tidy.py
```diff
@@ -199,36 +477,110 @@ def apply_fixes(args, tmpdir):
...
-            sys.stdout.write(' '.join(invocation) +
-                             '\n' + output.decode('utf-8'))
-            if len(err) > 0:
-                sys.stdout.flush()
-                sys.stderr.write(err.decode('utf-8'))
+            if tidy_error:
+                failed_files.append(name)
+            sys.stdout.write('\n' + ' '.join(invocation) +
+                             '\n' + output)
+            sys.stdout.flush()
```
matthiakl: It looks like the flushing does not work as expected here, some of the outputs are inter mixed, e.g. [here](https://github.com/tothxa/widelands/actions/runs/5051191348/jobs/9062765463#step:6:980). Just a notice, I do not think that this is a problem.
```
clang-tidy -p=/home/runner/work/widelands/widelands/build /home/runner/work/widelands/widelands/src/logic/game_settings.cc

clang-tidy -p=/home/runner/work/widelands/widelands/build /home/runner/work/widelands/widelands/src/logic/map_objects/map_object_type.cc
Cache hit as OK
Cache hit as OK
```

matthiakl (commented):
Looks very promising. I started reviewing with some comments, will do it more in depth over the weekend.

utils/run-clang-tidy.py
tothxa: Oops, must have been left there when I added the custom dir handling code.

utils/run-clang-tidy.py
tothxa: I know. Commit [6ec20a4](https://github.com/widelands/widelands/pull/5922/commits/6ec20a4bd3eeebfa41551b2777e23a534e39e8ea) tried to address it, but didn't help.

I think the problem is that the cache hit line is actually printed on stderr, which is only captured by the runner, while stdout goes through `tee` (for later passing it to `check_clang_tidy_results.py`). Maybe `tee` even re-adds buffering?

Maybe we could try `2>&1` it to tee? I think the check script wouldn't mind the extra lines, it already has a regex for the important lines of the report.

tothxa: I've checked the outputs. Actually stdout doesn't really need the file name headers. Each reported error starts with the file name and line number, this is what `check_clang_tidy_results.py` uses. Although this means that errors in included headers are there too, but the clang-tidy command lines don't add much value, because they get lost in the "noise".

So I propose that only the stdout from the clang-tidy checks should be printed to the stdout of the script, and the filenames and the version and enabled checks info would go on stderr. The workflow would then don't even `tee` stdout, only redirect it to the log file. So the clang-tidy step would not show the found problems on the fly, which is the price of having a well organised displayed output.

utils/run-clang-tidy.py
```diff
@@ -199,36 +497,110 @@ def apply_fixes(args, tmpdir):
...
+
+def filter_config(config):
+    filtered_config = ''
+    for line in config.splitlines(keepends = True):
+        if not line.startswith(('Checks:', 'User:', 'FormatStyle:')):
+            filtered_config += line
+    return filtered_config
+
+def filter_version(tidy_version):
+    return tidy_version.splitlines()[0].strip()
```
matthiakl: This is not OS-agnostic, on my system (Arch Linux) it prints
```
LLVM (http://llvm.org/):
  LLVM version 15.0.7
  Optimized build.
  Default target: x86_64-pc-linux-gnu
```
So the version file only contains "LLVM (http://llvm.org/):" which will not change with a newer version, so you might want to check for 'version' keyword or the version regex.

utils/run-clang-tidy.py
```diff
@@ -122,33 +131,322 @@ def make_absolute(f, directory):
...
+        # 2 months (we don't expect the cache to grow big)
+        retention = 62 * 24 * 3600
+
+        old = int(time.time()) - retention
+        keep = dict()
+        for hash in self.passed:
+            mtime = self.passed[hash]
+            if mtime > old:
+                keep[hash] = mtime
+        self.passed = keep
```
matthiakl: A slightly more compact way of filtering the dictionary:
```suggestion
        self.passed = dict((hash, mtime) for hash, mtime in self.passed.items() if mtime > old)
```

utils/run-clang-tidy.py
matthiakl: > So the clang-tidy step would not show the found problems on the fly, which is the price of having a well organised displayed output.

But the check script only displays the matching line, the full output often includes the code snippet and position indicator to easily find the problematic word. So I guess, you could just leave it as is.

utils/run-clang-tidy.py
```diff
@@ -284,60 +660,104 @@ def main():
         # Find our database
         build_path = find_compilation_database(db_path)
 
+    tmpdir = None
+    if args.fix or (yaml and args.export_fixes):
+        check_clang_apply_replacements_binary(args)
+        tmpdir = tempfile.mkdtemp()
+        if args.cache:
+            print('Caching is disabled because of -fix\n', file=sys.stderr)
+        args.cache = False
```
matthiakl:
```suggestion
        if args.cache:
            print('Caching is disabled because of -fix\n', file=sys.stderr)
            args.cache = False
```

utils/run-clang-tidy.py
```diff
@@ -122,33 +131,322 @@ def make_absolute(f, directory):
...
+
+        self.config_hash = get_hash(tidy_version + config + checks)
+
+        # We only put caches of custom runs in separate subdirs, so we can clean
+        # up the main directory on upgrades or when the default config changes.
+        # Custom subdirs have to be cleaned up manually on upgrades.
+        if custom:
+            self.dir = os.path.join(self.dir, '__custom_args', self.config_hash)
+            if not os.path.isdir(self.dir):
+                os.mkdir(self.dir)
```
matthiakl: This fails for the first custom invocation when `__custom_args` does not yet exists.
```suggestion
            if not os.path.isdir(self.dir):
                os.makedirs(self.dir)
```

utils/run-clang-tidy.py
tothxa: This was actually intentional. Disabling it is important, reporting is less so.

matthiakl: It is just a minor optimization: if it does not report, args.cache is already False, so no need in setting it again.

utils/run-clang-tidy.py
tothxa: I changed it to print the invocation on stderr, so at least the cache reports will always come after it. Let's see how the output looks this way, whether we should redirect stderr to `tee` as well.

I can't decide about the initial version and checks report. I'd move them to stderr too, but then it looks somewhat strange that both stdout and stderr of `clang-tidy -version / -list-checks` goes to stderr. And checks were always part of the log, so I'm not sure about removing them for this reason either.

tothxa:
Should I reintroduce my testing error to see how findings look with the latest changes?

I've also realised that the cache can actually be very big if all checks are enabled and all files are checked. So I was thinking about shortening the retention time if the cache size is big. That however should need configurability probably, so a lot of hassle. (or I could make it delete all untouched entries (I'm a bit afraid about mtimes WRT timezones) when cache is over 10MB, and let users who want to experiment locally edit the script… not very nice, I guess) (for configurability I have a per directory json file in mind…)

matthiakl (approved):
> Should I reintroduce my testing error to see how findings look with the latest changes?

I tried it locally and it looks cleaner now.

> I can't decide about the initial version and checks report.

For me it's fine to have them in the log. Then you can also see which tests on which version are reported in this file.

> I've also realised that the cache can actually be very big if all checks are enabled and all files are checked. 

Yes, this could be some future improvement. My idea would be to have a cmdline argument for maximum cache size and then calculate a shorter retention time to reduce the overall size. We might also need a cleanup strategy for the Github caches since every commit will create a new one.

For now I think it is good to go in, to speed up most of our workflows and you can add some fine tuning later.

tothxa:
Thank you for the review. :)

PR #5934, opened 2023-06-04
Enforce use of global vcpkg in CI

### LINKED ISSUE #5932: msvc builds failing again - multiple problems
**Describe the bug**
msvc builds fail in the workflow. First problem was that vcpkg path was moved under `"Program Files"`, but was unquoted in `build.yaml`. Trying to fix that (#5391) revealed next problem of libpng detection failing again, which I can't fix.

**To reproduce**
See:
1. https://github.com/widelands/widelands/actions/runs/5157957934/jobs/9292265222
2. https://github.com/widelands/widelands/actions/runs/5159357223/jobs/9294560159

**Expected behavior**
Well, just about this by now… :/ msvc has been broken more often than not for quite a while… (see this week's failed vcpkg update for more proof)

**Version:**
 - OS: w*ndows workflow runners
 - Widelands Version: any recent PR (probably master too, if we tried to rebuild)

matthiakl:
**Type of change**
Bugfix 

**Issue(s) closed**
Fixes #5932

**How it works**
New Visual Studio version seems to ship a vcpkg version and overrides VCPKG_ROOT variable.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -277,6 +277,8 @@ jobs:
       run: |
         mkdir $env:GITHUB_WORKSPACE\build
         cd $env:GITHUB_WORKSPACE\build
+        echo "VCPKG_INSTALLATION_ROOT is ${{ env.VCPKG_INSTALLATION_ROOT }}"
+        $env:VCPKG_ROOT = "C:\vcpkg"
         cmake.exe -G "NMake Makefiles" .. -DVCPKG_TARGET_TRIPLET=${{ env.VCPKG_TARGET_TRIPLET }} -DCMAKE_TOOLCHAIN_FILE=${{ env.VCPKG_ROOT }}/scripts/buildsystems/vcpkg.cmake -DCMAKE_BUILD_TYPE=${{ matrix.config }} -DOPTION_BUILD_WEBSITE_TOOLS=OFF -DOPTION_BUILD_TRANSLATIONS=ON -DOPTION_BUILD_TESTS=ON -DOPTION_ASAN=OFF -DOPTION_BUILD_CODECHECK=OFF -DOPTION_BUILD_WINSTATIC=ON -DOPTION_USE_GLBINDING=ON -DOPTION_FORCE_EMBEDDED_MINIZIP=ON
         nmake
         if ("${{ matrix.config }}" -Match "Release") {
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -277,6 +277,8 @@ jobs:
       run: |
         mkdir $env:GITHUB_WORKSPACE\build
         cd $env:GITHUB_WORKSPACE\build
+        echo "VCPKG_INSTALLATION_ROOT is ${{ env.VCPKG_INSTALLATION_ROOT }}"
+        $env:VCPKG_ROOT = "C:\vcpkg"
         cmake.exe -G "NMake Makefiles" .. -DVCPKG_TARGET_TRIPLET=${{ env.VCPKG_TARGET_TRIPLET }} -DCMAKE_TOOLCHAIN_FILE=${{ env.VCPKG_ROOT }}/scripts/buildsystems/vcpkg.cmake -DCMAKE_BUILD_TYPE=${{ matrix.config }} -DOPTION_BUILD_WEBSITE_TOOLS=OFF -DOPTION_BUILD_TRANSLATIONS=ON -DOPTION_BUILD_TESTS=ON -DOPTION_ASAN=OFF -DOPTION_BUILD_CODECHECK=OFF -DOPTION_BUILD_WINSTATIC=ON -DOPTION_USE_GLBINDING=ON -DOPTION_FORCE_EMBEDDED_MINIZIP=ON
```
tothxa: I think quoting the path would be safer anyway.

```suggestion
        cmake.exe -G "NMake Makefiles" .. -DVCPKG_TARGET_TRIPLET=${{ env.VCPKG_TARGET_TRIPLET }} -DCMAKE_TOOLCHAIN_FILE="${{ env.VCPKG_ROOT }}/scripts/buildsystems/vcpkg.cmake" -DCMAKE_BUILD_TYPE=${{ matrix.config }} -DOPTION_BUILD_WEBSITE_TOOLS=OFF -DOPTION_BUILD_TRANSLATIONS=ON -DOPTION_BUILD_TESTS=ON -DOPTION_ASAN=OFF -DOPTION_BUILD_CODECHECK=OFF -DOPTION_BUILD_WINSTATIC=ON -DOPTION_USE_GLBINDING=ON -DOPTION_FORCE_EMBEDDED_MINIZIP=ON
```

tothxa:
Thank you for picking this up. :)

tothxa (approved):
It worked. I can only verify that the change is trivial, and it makes sense with your explanation, but otherwise I don't know this environment at all.

PR #5909, opened 2023-05-10
Tutorial 01 fixes

tothxa:
**Type of change**
Bugfix

**Issue(s) closed**
Fixes #5899
Fixes #2149

**New behavior**
 - Adjusted window sizes
 - Wording and continuity improvements
 - Teach about touchpad settings
 - Make website link clickable

**Possible regressions**
Tutorial 01

Issue #5899, opened 2023-05-02
Buildhelp image names need updating in Tutorial 01


tothxa:
**Describe the bug**
#5868 replaced the buildhelp images with mipmapped ones, but the tutorial referred to them by direct path and filename.

**To reproduce**
1. Start Tutorial 01
2. Get to the "How to Build a Quarry" message which expalins the buildhelp icons
3. See error

**Expected behavior**
The correct buildhelp icons should be shown.

**Screenshots**
![tutorial1_buildhelp](https://user-images.githubusercontent.com/80712255/235807088-55a31103-6cf2-4fb4-bdb8-c8a48e3959c2.png)

**Version:**
 - Widelands Version: [82cf848](https://github.com/widelands/widelands/commit/82cf848537cab6232255c9f4a4b8b9485a725723)

**Additional information**
The easy solution is to update the hardcoded names to the scale 1 names. The correct one would be to allow look-up of mipmapped images from Lua?

hessenfarmer:
I think loking up mipmaps would require to have the possibility to show animations in campaign message boxes. So for the moment being the hardcoded way to a single scale 1 picture would be sufficient.

Noordfrees:
Since this code is separate from animations and entirely handled by `ImageCache::get` we should definitely implement the proper fix there, so it still serves the 1x mipmap even when no scale is provided. Sorry for overlooking this :(

tothxa:
> Since this code is separate from animations and entirely handled by `ImageCache::get` we should definitely implement the proper fix there, so it still serves the 1x mipmap even when no scale is provided.

That will be easy indeed, thanks for the tip. Though this will then be separate from the other tweaks I planned for the tutorial.

tothxa:
OK, not so simple after all. There are 3 places involved:

1. The campaign texts are rendered by the richtext engine, which just looks up the image with the default options in
2. `g_image_cache`. The default options are to look up without scaling and
3. allow themes to override by the style manager.

The fallback image we get is provided by the style manager.

I'm not sure whether it would be OK to skip theme lookup in the richtext engine, but scaling should be implemented, because the markup allows it. But that's a separate problem.

I don't think `resolve_template_filename()` should add the "_1" on its own, and try all the way through with both names.

But `ImageCache::get()` is heavily recursive, so it's not trivial to try both ways in the case of a lookup failure.

[...]

PR #5935, opened 2023-06-04
Allow lookup of richtext inline images without specifying scale suffix

tothxa:
**Type of change**
Bugfix

**Issue(s) closed**
Missed occurences of the problem described in #5899 in tutorial 03 and in the Encyclopedia

**To reproduce**
*If it's a bugfix:*
1. Open the help for a building
2. Scroll down to 'Building Requirements'
3. See error

**Possible regressions**
Richtext image lookup

**Additional context**
Possible other changes:
 - The code in `rt_render.cc` only applies a scale if the image is scaled down. I added this to the documentation, but I had a plan how to look up mipmap entries before I realised this. So I could implement it if we want to allow scaling up.
 - I wonder whether I should revert the filenames in tutorial 01 (re #5909)


Noordfrees (approved):

tothxa:
Thank you for all reviews. :)

PR #5943, opened 2023-06-17
CMakeLists.txt cleanups

### LINKED ISSUE #5628: Please don't activate ASan nor TSan in RelWithDebInfo cmake mode, to support the official Debian packaging guidelines
Hello,

I'm part of the official packaging team for Debian, and I would like to drop one of the patches that we have locally:
https://salsa.debian.org/games-team/widelands/-/blob/master/debian/patches/dbg_symbols

This patch simply adds `-g` to the build args in Cmake in Release mode, because Debian wants to retain the debug symbols in the release. These symbols are then automatically stripped and moved to a widelands-dbg package, that can be installed separately by people wanting to get an informative backtrace of their segfault when reporting a bug (or other similar usecases).

 In my opinion, Debian would like to use [intend of the RelWithDebInfo](https://stackoverflow.com/a/59314670/1756334), but I seem to understand that this build type also activates AddressSanitizer or ThreadSanitizer, which is probably not something we would like to do in production, I guess.

Would it be possible to make RelWithDebInfo more in line with Debian expectations? I guess that not setting the sanitizers in this case would be enough. Alternatively, a new cmake configuration dedicated for that would be very nice.

Thanks in advance, Mt

tothxa:
Fix RelWithDebInfo build type and add note about documentation in README.md to cmake options

**Type of change**
Bugfix / Refactoring

**Issue(s) closed**
Fixes #5628
Re #5576

**Possible regressions**
RelWithDebInfo builds?

**Additional context**
CMake adds `-g` on its own for `RelWithDebInfo`. It also adds `-Werror -Wno-error=unused-variable -Wno-error=unused-but-set-parameter`, but I couldn't figure out how this is controlled. But it doesn't seem to hurt.


CMakeLists.txt
```diff
@@ -204,34 +206,21 @@ if(CMAKE_BUILD_TYPE STREQUAL "Debug")
       option(OPTION_ASAN "Build with AddressSanitizer" ON)
     endif()
   endif()
-elseif(CMAKE_BUILD_TYPE STREQUAL "Release")
+elseif(CMAKE_BUILD_TYPE STREQUAL "Release" OR CMAKE_BUILD_TYPE STREQUAL "RelWithDebInfo")
```
Noordfrees:
```suggestion
elseif((CMAKE_BUILD_TYPE STREQUAL "Release") OR (CMAKE_BUILD_TYPE STREQUAL "RelWithDebInfo"))
```

Noordfrees (commented):
Looks safe AFAICT, one nit

Noordfrees (approved):

Noordfrees (approved):

PR #5944, opened 2023-06-18
Debugconsole and input history improvements

### LINKED ISSUE #1668: Move chat history from EditBox to GameChatPanel
At the moment, the chat history is implemented in EditBox. It logical place is the GameChatPanel though. We can use boost:signals2 to notify the GameChatPanel of relevant key presses.

 * date created: 2017-02-16T07:40:30Z
 * owner: gunchleoc
 * the launchpad url was https://bugs.launchpad.net/bugs/1665254

### LINKED ISSUE #2066: Chat window should only handle keys when the input field has focus
When the chat window is open or minimzed, keyboard shortcuts for everything else won't work. This could be solved by only handling the keys when the input field is focused.

Also, if the user left-clicks anywhere inside the chat window, the input field should be focused.

 * date created: 2016-12-04T20:49:10Z
 * owner: gunchleoc
 * the launchpad url was https://bugs.launchpad.net/bugs/1647208

### LINKED ISSUE #5793: Keep command history of lua console after map reload
When testing lua scripts in the editor it is sometimes needed to reload the map to get a "clean" map without any modifications from a lua script. E.g. i am working on a function `mountain_auto_height()` which should apply heights to mountainous areas automatically. During testing i have to reload the map several times. If i then open the lua script console i see the output from the previous calls, but the entered commands are not available by pressing `↑`. The previously used commands have to be copied again and pasted into the script console.

Just reloading a map should keep the previously used commands in the history, so they will available when pressing  `↑`.

**Additional context**
Just two questions:
1. Why is the command history restricted to ~5 commands?
2. Writing long commands in the script console isn't possible because there is a limit of character length. I guess this is because the input field is restricted?

tothxa:
**Type of change**
New feature / Refactoring

**Issue(s) closed**
Fixes #5793
Fixes #1668
Fixes #2066

**New behavior**
 - Save history to file
 - Also adds new command line option to allow enabling the developer features in release builds. More extensive reporting of usage of these features in network games.
 - Pass textinput to the input field when any part of the chat or script console window is focused and switch focus to input field on textinput

**Possible regressions**
EditBox and chat/script console behaviour and history handling

**Additional context**
Postponed:
 - UI file selection for `dofile()`
 - autosave for history? when?

Scrapped:
 - Configurable history size


Noordfrees:
Code LGTM, tested and working.
However before we can merge this we need to check that all advantages conferred upon players who use this are communicated to others in network games. Opening a field/object debug window and using the chat console's builtin functions (`switchplayer` and `mapobject`) needs a system chat message like we have for running Lua commands and entering cheat mode. There may be others as well, we'll need to check this carefully now that just checking for "release" as part of the version string is no longer a reliable indicator that cheats are disabled.

tothxa:
> However before we can merge this we need to check that all advantages conferred upon players who use this are communicated to others in network games.

Yes, of course, that was my intention too.

tothxa:
@Noordfrees Regarding #1668, IMO this PR already does what's reasonable. Any more separation would require jumping through hoops back and forth between `EditBox` and `GameChatPanel`. Do you agree we can close #1668 with this PR without any further changes?

src/ui_basic/textinput.cc
```diff
@@ -1097,4 +1106,49 @@ void AbstractTextInputPanel::Data::refresh_ww() {
...
+			   "Loading %s, line %" PRIuS ": %s", filename.c_str(), entries_.size() + 1, e.what());
+		}
+	}
+}
+
+void EditBoxHistory::save(const std::string& filename) const {
+	try {
+		FileWrite fw;
+		for (auto it = entries_.rbegin(); it != entries_.rend(); ++it) {
+			fw.print_f("%s\n", it->c_str());
```
Noordfrees, line 1167: This format looks a bit fragile, could you please use the `Profile` class instead? E.g. a key `"n_entries"` and then entries `"entry_0"`, `"entry_1"` etc, and a key/value pair `version=1`.

tothxa: Well, I'm going for the simplicity of `.bash_history`. The biggest problem I see is if the history contains binary garbage. So I tested it with some data from `/dev/urandom`, and it doesn't seem to do any harm.

OTOH `Profile`'s format would just bring in extra complexity e.g. for string constants and probably more.

What other protection do you hope to achieve by it?

tothxa:
@Noordfrees BTW do you think the cheat reporting is sufficient now? May I make it a command-line option instead of the compile flag?

src/ui_basic/textinput.cc
tothxa: Does the reference in https://github.com/widelands/widelands/pull/6056#issue-1837689145 mean that you're OK with it then?

tothxa:
I believe the last commit solves the remaining part of #2066.

tothxa:
OK, this is big enough now, and the important parts are in place. No more new features for this PR.

src/ui_basic/textinput.cc
```diff
@@ -1111,4 +1121,64 @@ void AbstractTextInputPanel::Data::refresh_ww() {
...
+		entries_.erase(entries_.begin() + max_size_, entries_.end());
+	}
+}
+
+void EditBoxHistory::load(const std::string& filename) {
+	FileRead fr;
+	if (fr.try_open(*g_fs, filename)) {
+		entries_.clear();
+		try {
+			while (char* line = fr.read_line()) {
```
Noordfrees: implicit bool conversion

Noordfrees (commented):
Tested this again. I was a bit surprised that saving the script console history to file had to be explicitly enabled, IMO we could remove the option and do that always.

I also don't see the value of letting the user configure the history size, why not just hardcode this with a size of 64 of both (or even more, maybe 256, this shouldn't take up that much memory?).

Other than that, working well, great feature :) Code LGTM, one nit.

tothxa:
> Tested this again. I was a bit surprised that saving the script console history to file had to be explicitly enabled, IMO we could remove the option and do that always.

OK, though I was afraid that it might be a privacy issue for some, like the chat history.

> I also don't see the value of letting the user configure the history size, why not just hardcode this with a size of 64 of both (or even more, maybe 256, this shouldn't take up that much memory?).

OK. Then I'll revert it to 64. If somebody needs more, then they should probably create helper scripts that they can use with `do_file()`.

tothxa:
(sorry for the incomplete pushes)

There's one more thing I'm thinking about: should the histories be saved regularly to keep more of them on crashes? (the script console in particular is supposed to be used in conditions where crashes are more likely)

One obvious place for saving would be after each autosave, but it would bring in a new dependency on a file from `wui` for `save_handler.cc`, while there's already a TODO comment in CMakeLists to eliminate even the existing one. Also maybe on exiting the editor or a game, but that's usually too late WRT crashes.

Or I could keep a counter of new entries, and save e.g. on every 10th, but that's probably too much overhead.

frankystone:
Mostly i get Lua errors and no complete crashes when working with the lua console. So i am not sure if a regular file save is needed.

Noordfrees (approved):
Thanks :)

tothxa:
Thanks for the reviews. :)

Issue #3129, opened 2019-09-09
Games can't be a draw


bunnybot:
E.g. in Collectors, if 2 players or teams have the same highest score, only one of them shall win. This should be a draw instead.

------------------------------------
Imported from Launchpad using lp2gh.

 * date created: 2016-05-14T10:40:53Z
 * owner: gunchleoc
 * the launchpad url was https://bugs.launchpad.net/bugs/1581788

SimonHeimberg:
This is fixed for win condition *collectors* since d695399a4169e4a32f8bb39b2fd5aeb839d92edb

I looked over the other win conditions. Most look fine, but ...

This still only has one winner:
- *[wood_gnome](/widelands/widelands/blob/6deb23c/data/scripting/win_conditions/wood_gnome.lua#L98)*
  The issue could happen when time limit reached only. Chance that the number of trees is equal then is low. Could be easily fixed similar to *collectors*.

This one is probably fine:
- *territorial_time*
   ([territorial_functions.lua](/widelands/widelands/blob/6deb23c/data/scripting/win_conditions/territorial_functions.lua) is not totally clear for me. But the comparison to see winners in territory_game_over() is `if territory_points.points[i][2] >= territory_points.points[1][2] then`, which allows several parties to have the same points.)
  The issue could happen (if it really can happend) when time is reached only. The probability that the conquered area is totally equal then is low.

PR #5952, opened 2023-06-22
win condition wood_gnome can have several winners

SimonHeimberg:
In win condition wood_gnome, all players with maximum points win the game. (In case they have the same points at the end by chance.) The message still only lists one of them.
Teams are still not supported.

**Type of change**
partial Bugfix

**Related issue(s)**
improves #3129


data/scripting/win_conditions/wood_gnome.lua
```diff
@@ -98,18 +98,20 @@ local r = {
...
+         local end_msgs = lost_game_over
+         if points[i][2] >= win_points then
+            end_msgs = won_game_over
+            win_lost = 1
+         end
+         privmsg = end_msgs.body
          privmsg = privmsg .. msg
-         points[i][1]:send_to_inbox(lost_game_over.title, privmsg)
-         wl.game.report_result(points[i][1], 0, make_extra_data(points[i][1], wc_descname, wc_version, {score=points[i][2]}))
+         points[i][1]:send_to_inbox(end_msg.title, privmsg)
```
hessenfarmer:
```suggestion
         points[i][1]:send_to_inbox(end_msgs.title, privmsg)
```

hessenfarmer (changes requested):
one typo inline.
please fix the messages as well, as currently only the first entry of the sorted table will be mentioned in the message.  
I was not aware that we do not have a teams option for woodgnome so it would be good to fix that as well for this (but this is not mandatory for this PR)

SimonHeimberg:
> please fix the messages as well, as currently only the first entry of the sorted table will be mentioned in the message. I was not aware that we do not have a teams option for woodgnome so it would be good to fix that as well for this (but this is not mandatory for this PR)

How do you suggest to to fix the message? Remove mentioning the winner since wood_gnome is the only win condition doing this?

hessenfarmer:
skipping the message would be the easy fix, collecting all winning players in a table in another iteration of the points table and concatenating their names in the message (use of ngettext would be required) would be the cherry on the cake

hessenfarmer (approved):

hessenfarmer:
I can live with that remaining TODO.  
So I will merge when CI is green, unless you tell me you will fix the TODO right here in this PR.    
How do you want to be credited (name, nickname, both)?

SimonHeimberg:
> I can live with that remaining TODO. So I will merge when CI is green, unless you tell me you will fix the TODO right here in this PR. How do you want to be credited (name, nickname, both)?

only nickname aDiscoverer is fine

PR #5954, opened 2023-06-23
Continuous Bob Animations

### LINKED ISSUE #5951: Ship animations are not played continuously
**Describe the bug**
Ship animations are played with jumps when any kind of ships are moving and when expedition- or warships are waiting. Only transport ships in idle state play the idle animation continuously. The problem is not in the animation frames, the jumps happen with animations that are continuous when the frames are looped in an image viewer. It looks like the animations are started over more often than a full cycle would take.

**To reproduce**
1. Open a game with ships
2. Watch the ships as they travel or wait around
3. Compare idle animations of transport ships (smooth) and expedition ships (bad jumps)

**Expected behavior**
When the ship state and travel direction don't change, the animation should be played continuously.

**Version:**
 - Widelands Version: 1.2~git26177 (eebf356@master)

Noordfrees:
**Type of change**
Bugfix

**Issue(s) closed**
Fixes #5951

**New behavior**
Don't start the animation from the beginning if the new anim is the same as the current anim

**Possible regressions**
Bob animations

### DIFF
--- a/src/logic/map_objects/bob.cc
+++ b/src/logic/map_objects/bob.cc
@@ -824,8 +824,10 @@ void Bob::draw(const EditorGameBase& egbase,
  * Set a looping animation, starting now.
  */
 void Bob::set_animation(const EditorGameBase& egbase, uint32_t const anim) {
-	anim_ = anim;
-	animstart_ = egbase.get_gametime();
+	if (anim_ != anim) {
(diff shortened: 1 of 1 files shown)


tothxa (approved):
Wow, I didn't expect it would be this simple. Thank you. Tested and works.

PR #5956, opened 2023-06-25
Z-Order Map Options Window Above Save Map Window

### LINKED ISSUE #5791: Map options not in foreground of save map menu
**To reproduce**
1. Open a map in editor
2. Click on 'Main menu → Save map'
3. Click on 'Show Map Options'
4. The map options window is behind the 'Save map' window. It is only accessible by clicking on 'Close/Cancel' in the 'Save Map' window

**Expected behavior**
The Map options menu should be in the foreground, not covered by the 'Save Map' menu.

**Version:**
 - OS: archlinmux
 - Widelands Version: 1.2~git26063 (abe0827@master) Debug

Noordfrees:
**Type of change**
Bugfix

**Issue(s) closed**
Fixes #5791 

**New behavior**
The map options window gets the same Z order as a fullscreen window so it opens above the save map window.


src/editor/ui_menus/main_menu_map_options.cc
```diff
@@ -508,6 +508,9 @@ MainMenuMapOptions::MainMenuMapOptions(EditorInteractive& parent, Registry& regi
 	ok_.sigclicked.connect([this]() { clicked_ok(); });
 	cancel_.sigclicked.connect([this]() { clicked_cancel(); });
 
+	// This window is accessible from the Save Map fullscreen window, so it needs a higher Z order.
+	set_z(UI::Panel::ZOrder::kFullscreenWindow);
```
tothxa: This is not right when the window is opened from the editor menu or by the keyboard shortcut. Could it be <z_order_of_parent> + 1? But it still needs to work, if the options window is open when the save window is opened, then the options button is clicked…

*edit: Also the options window should probably be modal when opened from the save map window. So an easy solution may be to always make it modal?*

tothxa (approved):
Thank you. Tested and works. Code looks OK too.

(BTW the Map Options window stays on top of the toolbar if the Save Map window is closed beneath it, but that's quite a corner case and isn't much of a problem anyway)

PR #5964, opened 2023-06-30
count trees for start conditions like poor_hamlet

SimonHeimberg:
... instead of just checking for one tree.

**Type of change**
Enhancement

**Additional context**
Starting with poor hamlet and then realising there are not enough trees is demotivating and needs to start a new game.

Counting trees is easy, and each one produces one log. So do this. The minimum of logs needed is determined from waretable_trees. (This lists the ware you get if there are not enough trees.)
Exactly seen, the logs in the wares you get could be reduced by the number of trees. It is a harder start anyway if you have to grow the trees for getting logs. So do not reduce the logs.

The rocks are still not counted. But it is unlikely that there is only one rock around which has the size one. And counting rocks (with the size) is more difficult.

**To reproduce**

Steps to reproduce the behavior which cause the problem in master but not in your branch:
1. Select map "Atoll"
2. Choose to play at yellow position, select "poor hamlet" as starting condition (win condition does not matter)
3. Start the game
4. (You get the message that there are no rocks nearby but you get some wares instead.)
5. Problem: You only can build your first building. But you can not cut enough trees to build any other.
 
**New behavior**
...
4. You get two messages, one for no rocks, one for not enough trees.
5. With the additional wares, you can start your economy. (Of course you still must follow the correct order and never send a builder too early.)


data/scripting/infrastructure.lua
```diff
@@ -259,19 +259,24 @@ end
...
-function check_trees_rocks_poor_hamlet(player, sf, warehouse, waretable_rocks, waretable_trees)
-   -- NOTE: pessimistically, this could be a single rock and a single tree
+function check_trees_rocks_poor_hamlet(player, sf, warehouse, waretable_rocks, waretable_trees, min_trees)
+   if not min_trees then -- pass argument if this default does not match
+      min_trees = waretable_trees["log"] + (waretable_trees["planks"] or 0) * 2
+   end
+   -- NOTE: pessimistically, this could be a single rock
    local has_rocks = false
-   local has_trees = false
+   local n_trees = 0
```
hessenfarmer, line 271: as the bool variable here is now an integer with a changed name please adapt the condition in line 276 to reflect this

hessenfarmer (changes requested):
one thing inline.
As this is meant to be the bare minimum start I would like it better to reduce the wares you get for each tree present.

data/scripting/infrastructure.lua
SimonHeimberg: Oops, not using an undefined variable makes sense of course.
It worked anyway. My code was continuing to count the trees until enough rocks were found. The final conditions were correct, so the result was as expected.

SimonHeimberg:
> As this is meant to be the bare minimum start I would like it better to reduce the wares you get for each tree present.

OK, I will reduce the logs.

data/scripting/infrastructure.lua
```diff
@@ -281,8 +286,9 @@ function check_trees_rocks_poor_hamlet(player, sf, warehouse, waretable_rocks, w
       player:send_to_inbox(_("No rocks nearby"), _("There are no rocks near to your starting position. Therefore, you receive extra resources for bootstrapping your economy."))
    end
    -- adding exactly one forester
-   if not has_trees then
+   if n_trees < min_trees then
+      waretable_trees.logs = math.max(0, waretable_trees.logs - n_trees) -- reduce additional logs if there are trees
```
hessenfarmer:
```suggestion
      waretable_trees["log"] = math.max(0, waretable_trees["log"] - n_trees) -- reduce additional logs if there are trees
```

hessenfarmer (changes requested):

hessenfarmer:
Although it works with your new nomenclature I would prefer a consistent acces to lua tables.  (i.e. using the [] )

data/scripting/infrastructure.lua
```diff
@@ -259,19 +259,24 @@ end
...
 --    :arg waretable_trees: a table of pairs {:const:`"name"` = value} to add if no trees were found nearby
 --    :type waretable_trees: :const:`table`
+--    :arg min_trees: minimal number of trees (optional)
+--    :type min_trees: :class:`integer`
 
-function check_trees_rocks_poor_hamlet(player, sf, warehouse, waretable_rocks, waretable_trees)
-   -- NOTE: pessimistically, this could be a single rock and a single tree
+function check_trees_rocks_poor_hamlet(player, sf, warehouse, waretable_rocks, waretable_trees, min_trees)
+   if not min_trees then -- pass argument if this default does not match
+      min_trees = waretable_trees["log"] + (waretable_trees["planks"] or 0) * 2
```
SimonHeimberg: @hessenfarmer 
I see. Here I do `waretable_trees["log"]` ...

data/scripting/infrastructure.lua
```diff
@@ -281,8 +286,9 @@ function check_trees_rocks_poor_hamlet(player, sf, warehouse, waretable_rocks, w
       player:send_to_inbox(_("No rocks nearby"), _("There are no rocks near to your starting position. Therefore, you receive extra resources for bootstrapping your economy."))
    end
    -- adding exactly one forester
-   if not has_trees then
+   if n_trees < min_trees then
+      waretable_trees.log = math.max(0, waretable_trees.log - n_trees) -- reduce additional logs if there are trees
```
SimonHeimberg, line 290: ... but here `waretable_trees.log`

I try to use `a_table[variable]` or `a_table[3]` but `a_table.label` (not `a_table["label"]`), but I mixed here :no_mouth:

hessenfarmer: I prefer using the [] always as basically there is no difference in a key to the table whether it being a number, a letter or a string like in this case. Lua allows all of them in the same table afaik.

hessenfarmer (approved):
forget about my preferences we have done it like this on various other locations.  so shall be good to go

data/scripting/infrastructure.lua
SimonHeimberg: You are also free to write `f["immovable"]:has_attribute("tree")`, for lua a `wl.map.Field` is just a special table. But I guess we agree this would not look nice.

hessenfarmer: correct I realized that myself just seconds before your post.    
So I only miss how to credit your work in our credit list?

SimonHeimberg:
By the way, I saw that
[data/tribes/initialization/atlanteans/starting_conditions/poor_hamlet.lua](https://github.com/widelands/widelands/blob/master/data/tribes/initialization/atlanteans/starting_conditions/poor_hamlet.lua#L52)
has `-1` as a value. As it is in `waretable_rocks`, it will not interfere with my change. In case somebody does this for `waretable_logs` (for any new tribe), my minimum of 0 would probably not be correct.

data/scripting/infrastructure.lua
SimonHeimberg: only nickname "aDiscoverer" is fine

PR #5979, opened 2023-07-02
define PRIuS using PRIuPTR if available

madroach:
**Type of change**
Bugfix

**Issue(s) closed**
On OpenBSD amd64 a warning is emitted during compilation about `__WORDSIZE` being undefined. `PRIuS` is then wrongly defined as `u`.

**Fix**
Define `PRIuS` as `PRIuPTR` if the latter is available.

**Possible regressions**
Since it falls back to the old behavior in case `PRIuPTR` is undefined I don't expect any regressions. It still won't help platforms having neither `__WORDSIZE` nor `PRIuPTR`.

### DIFF
--- a/src/base/macros.h
+++ b/src/base/macros.h
@@ -128,6 +128,9 @@
 #define PRINTF_FORMAT(b, c)
 #endif
 
+#ifdef PRIuPTR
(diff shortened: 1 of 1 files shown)


Noordfrees (approved):

PR #5982, opened 2023-07-03
Fix Segfault in `FieldActionWindow`

Noordfrees:
**Type of change**
Bugfix

**Issue(s) closed**
When you click on an unowned field that has a shipyard interface but no buildcaps the buildbox has not been created yet, so it segfaults.

**To reproduce**
Triggered by clicking on a blocked field on unowned territory in range of an own shipyard, like this shipconstructionsite here:
grafik

**Additional context**
Found in yesterday's naval battle royale


tothxa (approved):
Code looks OK. Tested and fix confirmed.

PR #5995, opened 2023-07-05
Revert preventing move_inside_parent() to move UniqueWindows with saved position

### LINKED ISSUE #5984: FieldActionWindow no longer moves inside parent
**Describe the bug**
The field action window no longer moves inside the Widelands window when changing its tab makes it bigger.

**To reproduce**
1. Click on a big building plot to open the field action window. May need out trying until you get a position where the window opens near the bottom of the screen.
2. Switch to the Medium tab.
3. The lower part of the window is cut off, rendering buildings therein inaccessible until manually dragging and repositioning the window.

**Expected behavior**
The window should jump to be fully inside the parent, like it used to in v1.1.

**Screenshots**
After opening the window:
shot0003
After using the mousewheel to select a different tab:
shot0004

**Version:**
 - OS: Debian Unstable
 - Widelands Version: Tournament2023

**Additional context**
Other tabbed windows such as the stock menu are also affected. Seems to be a general problem with `Window::move_inside_parent`.

tothxa:
**Type of change**
Bugfix

**Issue(s) closed**
Fixes #5984

**New behavior**
`UniqueWindow`s now always honour `move_inside_parent()`

**Possible regressions**
Windows intentionally moved partially outside by the user get moved back in

**Additional context**
I wanted to make window size changes work similar to when the main window is resized, but I'll be away for 2 weeks starting on Saturday, so this is just the quick fix for now.

### DIFF
--- a/src/ui_basic/unique_window.cc
+++ b/src/ui_basic/unique_window.cc
@@ -156,7 +156,9 @@ void UniqueWindow::move_inside_parent() {
 		return;
 	}
 
-	if (registry_->valid_pos) {
+	// TODO(tothxa): needs refining to handle window size changes (e.g on tab switching)
+	/*
+	if (moved_by_user()) {
 		constexpr int32_t kMinVisible = 100;
 
 		const bool top_visible = registry_->y < parent->get_inner_h() - kMinVisible;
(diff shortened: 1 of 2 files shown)


Noordfrees (approved):
Thanks :) Works well

PR #5996, opened 2023-07-05
move lunit.lua in test/scripting/

SimonHeimberg:
**Type of change**
Cleanup

**New behavior**
lunit.lua is not delivered with every normal installation. It is not used there.


Noordfrees (commented):
Please keep the includes sorted alphabetically. Otherwise LGTM

Noordfrees (approved):

Issue #5991, opened 2023-07-04
gamehost translates the name of computer players in its locale


SimonHeimberg:
**Describe the bug**
The gamehost translates the player name of computer players. This name is transmitted to the clients.

:+1: This makes sense because then all players talk about the same players (with the same name and numbers).
:-1: But this does not make sense because this mixes languages when host and client do not have the same locale.
bad case for :-1: : Your host has Arabic (or any other language with special letters) configured as language. You join from a computer showing English. Now you see Arabic letters as player name for computer players. Doing diplomacy like this is quite confusing. As developer: The host has set a player name with non-ascii letters, which is not allowed as typed player names.

**To reproduce**
Steps to reproduce the behaviour:
1. Start widelands with arabic language setting (`LANG=ar widelands` in a shell or ...)
2. Type p for a local network game
3. Prepare a game to host (select a map)

5. Start another instance of widelands with normal language setting (can be on the same computer)
6. Type p for a local network game
7.  Choose the game on the left and confirm with OK (or double click)

8.  Go Back on the first game instance (the Arabic one)
9. Start the game

10. Go to the second game instance (the english one)
11. Open the diplomacy dialogue
12. It shows Latin and Arabic letters :astonished:

**Expected behavior**
Not sure what is best to happen. Translating results in text with mixed language if it is done on the host. If it is done on the client it results in different player names (plus different numbers in case of name conflicts in some locales).
Maybe not translating computer player names ("Computer 1" for all languages)?

**Version:**
 - OS: Debian, flatpak
 - Widelands Version: "v1.1"

**Additional context**
The gamehost sets the name here:
[`src/network/gamehost.cc`](https://github.com/widelands/widelands/blob/a6e19e126fbc18c95a75ee90a075e765537099af/src/network/gamehost.cc#L1616)

no problem, as long as random maps do not work for network games (in case this is the dialogue for editing only, no problem probably anyway):
[`src/editor/ui_menus/main_menu_random_map.cc`](https://github.com/widelands/widelands/blob/a6e19e126fbc18c95a75ee90a075e765537099af/src/editor/ui_menus/main_menu_random_map.cc#L699)

Issue #5985, opened 2023-07-03
choosing a player name with a space only partially prevented


SimonHeimberg:
**Describe the bug**
In the dialogue to join a multiplayer game (LAN / direct ip), the button to join is disabled when the human chooses a name with a space in it. But with a **double click** on the game or **click** and **enter**, joining continues anyway.

**To reproduce**
1. Start widelands
2. Type "p" for a multiplayer game "LAN / direct IP"
3. Enter a valid player name here (like "testHost")
4. Click the button to "Host a new game"
5. Click the button to Choose a map
6. Choose any map
7. Confirm with double-click or ok button

8. Start a 2nd instance of widelands
    (from a terminal/"new window" in context menu in gnome/on another computer in the lan/...)
10. Type "p" for a multiplayer game "LAN / direct IP"
11. Enter a name for your player containing a space (like "join 2 win")
12. The buttons to "Host new a game" and "Join a game" are disabled
   (This is expected. The frame of the player name is in red and the reason is written in its tooltip)
14. Double-click on the game listed in the list => you join with the illegal name
15. click the button back
16. click on the game in the list and type enter => you join with the illegal name

**Expected behavior**
Choosing a player name containing a space (like "me too") is not possible in any way. (Or possible in all ways if it does not matter.)

**Version:**
 - OS: Debian 6.1.27, flatpak
 - Widelands Version: v1.1

**Additional info:**
Was first mentioned in the forum in https://www.widelands.org/forum/topic/5797/?page=1#post-40394 somewhere below the title *add possibility to detect human players in multiplayer game*

PR #6001, opened 2023-07-07
Lua attribute player.ai_type

SimonHeimberg:
**Type of change**
New feature

**Before**
In a multiplayer game, it is not possible to to detect reliably if a player is controlled by an ai.
The property `wl.Game().interactive_player` only shows the human player on the current game instance.
Computer players are named "Computer N", but this is not a reliable source as it is translated in the hosts locale (#5991) and human players can choose such a name (partially prevented, see #5985).

**New behavior**
With the property `player.ai_type`, determining ai players is reliable. Making the exact type available to a script should not be a problem.

**Additional context**
Be aware the type of the ai player can change, probably only when a human player looses the connection and is replaced by an ai.
It is possible that a player is controlled by a human and an ai (by using `playerX.switchplayer()`). This can not be detected.

A use case is in [forum post ..48](https://www.widelands.org/forum/topic/5797/?page=1#post-40448)


src/scripting/lua_game.cc
```diff
@@ -338,6 +339,19 @@ int LuaPlayer::set_hidden_from_general_statistics(lua_State* L) {
 	return 0;
 }
 
+/* RST
+   .. attribute:: ai_type
+
+      (RO) type of the ai controlling this player
+      ("normal", "weak", "very_weak", "empty", ..., "" if human controlled only)
+*/
```
```suggestion
/* RST
   .. attribute:: ai_type

      .. versionadded:: 1.2

      (RO) Type of the AI controlling this player
      ("normal", "weak", "very_weak", "empty"; "" if human controlled only)
*/
```

SimonHeimberg:
Now the tests pass.
How can I verify that the netclient has this attribute correct?

SimonHeimberg:
I could not yet practically test if the attribute is set in the game.

The setting `player.ai` is received and saved in netclient (see below, probably writes in player from game_settings.cc). So it hopefully is also available in game (on the netclient). But maybe `player.ai` (of src/logic/player.cc) stays empty.
https://github.com/widelands/widelands/blob/6ba442a57ac5a57d409488388e2d276c4eb3f5dc/src/network/gameclient.cc#L643

The type of ai is visible in the new game dialogue for the netclient. I guess it is still available in game.

SimonHeimberg:
I finally managed to compile and have everything so I can start the executable.

The ai_type works as expected in multiplayer mode. Random ai players get a valid value in the game.
In single player mode random players seem to get the value "" (empty string), which is used for human players normally. As it is known that there is only one human player, it is still obvious, but unhandy.

SimonHeimberg:
This could be included in 1.2 as is. It already adds helpful functionality. And it warns about what is not reliable.

The fixing of single player must be in a complete different place, probably in single player startup code. It could be investigated later.

hessenfarmer:
@SimonHeimberg 
will try to test this in the next days

hessenfarmer (approved):
code LGTM. 
tested and working for:
- SP
- Local MP (host)
- Local MP (client)
 therefore lets have it

hessenfarmer:
@bunnybot merge

PR #6003, opened 2023-07-09
Follow links in building help window

### LINKED ISSUE #5662: Make links work in help windows
Re https://github.com/widelands/widelands/pull/5661#issuecomment-1320216912 and https://github.com/widelands/widelands/pull/5661#issuecomment-1320446488

Encyclopedia hyperlinks should work from item help windows too. I guess the easier approach might be to open the encyclopedia right for the help if that is acceptable instead of the current standalone window.

matthiakl:
**Type of change**
New feature

**Issue(s) closed**
Fixes #5662 

**New behavior**
Clicking on a link inside the help window goes to the corresponding page and provides a history to go back to the starting point.

**How it works**
The help window subscribes to the Hyperlink notification based on the panel name (starting with "encyclopedia").

**Possible regressions**
Since there are multiple subscribers now, the link is opened in the currently focused panel. This may result in unexpected behavior because you can click on a link without giving focus to the panel, but I currently have no idea how to fix it.


src/ui_basic/panel.cc
```diff
@@ -1700,7 +1700,7 @@ NamedPanel::NamedPanel(Panel* const nparent,
      name_(name),
      hyperlink_subscriber_(
         Notifications::subscribe<NoteHyperlink>([this](const NoteHyperlink& note) {
-	        if (name_ == note.target) {
+	        if (name_.find(note.target, 0) == 0) {
```
Noordfrees:
```suggestion
	        if (starts_with(name_, note.target)) {
```

Noordfrees (commented):
I tried it out but the help window comes up empty?

grafik

src/ui_basic/panel.cc
matthiakl: Thanks, that was the function I searched for :). The content should show up now, I cleaned up one line too much before pushing.

Noordfrees (commented):
Works fine now :) Code LGTM (clang-tidy has some nits).
One suggestion: Ctrl-Clicking the Back button should go back to the beginning.

src/wui/helpwindow.cc
```diff
@@ -37,28 +41,111 @@ BuildingHelpWindow::BuildingHelpWindow(Panel* const parent,
...
-		cr->push_arg(building_description.name());
+		cr->push_arg(tribe_.name());
+		cr->push_arg(item);
 		cr->resume();
 		std::unique_ptr<LuaTable> return_table = cr->pop_table();
-		return_table->do_not_warn_about_unaccessed_keys();  // We won't display the title here
 		textarea_->set_text(as_richtext(return_table->get_string("text")));
+		titlearea_->set_text(return_table->get_string("title"));
+		if (forward) {
+			history_.emplace_back(std::make_pair(type, item));
```
Noordfrees:
```suggestion
			history_.emplace_back(type, item);
```

Noordfrees (approved):
LGTM :)

One small nit inline

PR #6012, opened 2023-07-12
Show error message immediately when host leaves metaserver game

Noordfrees:
**Type of change**
Bug mitigation

**Issue(s) closed**
Fixes the very annoying bug that in a multiplayer game played over the metaserver, clients do not receive a notification when the host leaves the game. Instead the game just pauses indefinitely and it can take a while for the client to notice.

**To reproduce**
1. Open two Widelands clients.
2. Start a multiplayer game with these two instances (one as host, one as client) over the **lobby**. The bug does **not** happen with LAN games.
3. Close the hosting instance.
4. Observe client behaviour.

**New behavior**
We immediately fall to the main menu with a "Connection lost – please report as a bug – do you want an emergency savegame" message.

**Additional context**
Ideally we would show a more accurately worded message and not ask for a bug report, but this is already an improvement over the current situation.


hessenfarmer (approved):
LGTM
Tested and working

PR #6016, opened 2023-07-13
Remove hardcoded panel sizes in launch game window

### LINKED ISSUE #5631: Launch Game window does not resize contents properly

matthiakl:
**Type of change**
Bugfix

**Issue(s) closed**
Fixes #5631 

**How it works**
If the desired size of a panel overflows the box there is currently no check to prevent this. The short term fix for the bug is to reduce the desired sizes of the problematic panels. The long term fix would be to sanitize box layouts to not overflow in general.

**Additional context**
I already tested a bit of a generic overflow detection, but this is no trivial task because of the `set_size() -> layout() -> set_size()` cycles.


Noordfrees:
🎉 

The bug still happens though with the MP setup window at 1024×768 when clicking the "Show More Options" button:
grafik


Also, the label for the win condition duration spinner is still cut off if the window is narrow. That's not necessarily in scope of this PR but #5631 isn't truly fixed before this is also solved somehow.

matthiakl:
Should be all fixed now:
- The spinbox label is hidden when the space is not sufficient
- The advanced options are in a scrollable box, it is a bit cramped, but does not overflow anymore.

src/ui_basic/box.cc
```diff
@@ -291,6 +285,18 @@ void Box::update_positions() {
...
 		}
 	}
+
+#ifndef NDEBUG
+	// Subtract last spacing + some leeway from infspace calculation
+	totaldepth = totaldepth - items_.size() - inner_spacing_;
+	int max_depths = orientation_ == Horizontal ? get_w() : get_h();
+	if (!force_scrolling_ && !scrolling_ && max_depths > 0 && totaldepth > max_depths) {
+		// TODO(matthiakl): one day this should be an error
+		verb_log_warn("Overflowing %s box %s: %d > %d ",
```
Noordfrees:
```suggestion
		verb_log_dbg("Overflowing %s box %s: %d > %d ",
```

Noordfrees (commented):
LGTM, works well :)
One request, the label for the spinbox shouldn't disappear completely, could you make it a tooltip instead when not shown (or maybe always)?

Noordfrees (approved):
Great, thanks :)

PR #6021, opened 2023-07-15
Resolve compiler warnings

matthiakl:
**Type of change**
Refactoring

**Issue(s) closed**
Re #5641 (still one warning left)

**New behavior**
- Re-enable -Werror for gcc 13.1.1+
- Update clang version
- Resolved most gcc warnings, release builds have a warning with atomic_load, but this is an open issue with gcc: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=106297
- Resolved warnings in mingw builds

**Possible regressions**
I only checked gcc 13.1.1, not 13.1.0 or 13.0


src/third_party/eris/loslib.c
```diff
@@ -102,14 +102,11 @@ static time_t l_checktime (lua_State *L, int arg) {
...
 ** {==================================================================
 ** Configuration for 'tmpnam':
-** By default, Lua uses tmpnam except when POSIX is available, where
-** it uses mkstemp.
+** By default, Lua uses POSIX mkstemp.
 ** ===================================================================
 */
 #if !defined(lua_tmpnam)	/* { */
 
-#if defined(LUA_USE_POSIX)	/* { */
```
Noordfrees: Please document this change in `third_party/README`. We will need to reapply it every time we want to upgrade to a new Eris version.

Noordfrees (approved):
Great :)

matthiakl:
Since MSVC is not POSIX, the build failed. I changed the code back to original and use the defines to select an implementation. Hopefully this does not cause any issues in other places.

Noordfrees (approved):

PR #6027, opened 2023-07-18
5549 keep po and pot files in sync

### LINKED ISSUE #5549: Translation update script should push updated templates before pulling translations

SimonHeimberg:
**Type of change**
Bugfix / New feature / Refactoring / ...

**Issue(s) closed**
Fixes #5549

**How it works**
- push new translation template files to Transifex
- pull the updated translation files after

**Possible regressions**
Not sure if the updated translations are available immediately or there needs more waiting time. Could potentially fetch po files with different files. (I did not find any documentation about this.)

### DIFF
--- a/utils/merge_and_push_translations.sh
+++ b/utils/merge_and_push_translations.sh
@@ -121,6 +121,11 @@ python3 utils/fix_formatting.py --lua --dir data/txts
 # Undo one-liner diffs of pure timestamps with no other content
 undo_oneliner_diffs
 
+if [ -z "$(git status -s)" ]; then
+  echo "Run completed, nothing to commit."
+  exit 0
+fi
+
 # Stage changes
 # - Translations
 git add po/*/*.po po/*/*.pot data/i18n/locales/*.json xdg/translations/*.json || true
@@ -132,11 +137,6 @@ git add data/i18n/*.lua || true
 git add xdg/org.widelands.Widelands.appdata.xml xdg/org.widelands.Widelands.desktop || true
 # - Statistics
 git add data/i18n/translation_stats.conf || true
-
-if [ -z "$(git status -s)" ]; then
-  echo "Run completed, nothing to commit."
-  exit 0
-fi
 
 # Commit and push.
 git commit -m "Fetched translations and updated catalogs."
--- a/utils/merge_and_push_translations.sh
+++ b/utils/merge_and_push_translations.sh
@@ -74,9 +74,8 @@ undo_oneliner_diffs() {
 }
 undo_oneliner_diffs
 
-if [ -z "$(git status -s)" ]; then
-  true # nothing changed, do not upload to transifex
-else
+if [ -n "$(git status -s)" ]; then
+  # Only upload to Transifex if anything changed
   # Push source catalogs to Transifex
   tx push -s
   sleep 5 # ?? # wait for translation files to be updated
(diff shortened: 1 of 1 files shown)


utils/merge_and_push_translations.sh
```diff
@@ -57,7 +57,32 @@ echo "Working tree is clean, continuing"
...
+      git checkout $entry
+    fi
+  done
+  set -x
+}
+undo_oneliner_diffs
+
+if [ -z "$(git status -s)" ]; then
+  true # nothing changed, do not upload to transifex
+else
```
Noordfrees:
```suggestion
if [ -n "$(git status -s)" ]; then
  # Only upload to Transifex if anything changed
```

utils/merge_and_push_translations.sh
```diff
@@ -57,7 +57,32 @@ echo "Working tree is clean, continuing"
...
+  # Push source catalogs to Transifex
+  tx push -s
+  sleep 5 # ?? # wait for translation files to be updated
```
Noordfrees: Transifex should be pretty quick with updating this, but maybe wait one minute to be sure. This workflow doesn't need to be very fast.

utils/merge_and_push_translations.sh
```diff
@@ -97,15 +118,12 @@ python3 utils/fix_formatting.py --lua --dir data/i18n
...
-    git checkout $entry
-  fi
-done
-set -x
+undo_oneliner_diffs
+
+if [ -z "$(git status -s)" ]; then
+  echo "Run completed, nothing to commit."
+  exit 0
+fi
```
tothxa, line 126: This block shouldn't be moved. The next steps may also introduce changes from translations not in .po format.

SimonHeimberg: I do not understand, sorry.
The files are modified before. The next commands are "git add XXX" only and do not modify a file.
"git status -s" reports one line for every modified file, independent if the change is added (`M  fileName`) or only changed on disk (` M fileName`) or partially added (`MM fileName`). There is no filter for "*.po".

tothxa: Oops, sorry. You're right.

SimonHeimberg: No problem at all.

tothxa (approved):
I've looked more into `buildcat.py`, and indeed, `do_buildpo()`, `do_tunepo()` and `do_update_po()` are not actually used. I'm not sure whether they should be removed or just some comments added to this effect. Looks like they were disabled in 2010, so they've probably gone untested since then, so maybe removing them would be the right thing to do.

PR #6029, opened 2023-07-18
5880 less ci noise

SimonHeimberg:
continuation of base pull request

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -96,7 +96,7 @@ jobs:
       uses: dorny/paths-filter@v2
       id: changed_files
       with:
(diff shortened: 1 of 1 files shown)


SimonHeimberg:
checked:
- :heavy_check_mark:  codecheck skipped steps
- :heavy_check_mark: clang_tidy skipped steps
- :heavy_check_mark: testsuite (whole matrix) is skipped
- :heavy_check_mark: installers are skipped
- :grey_question: (can not check if pre-release works, because this is not on master)

The total duration was still 1h 29min, but only because there was no runner available. The sum of the run times is only 2m 38s.
It seems to work

Issue #5550, opened 2022-09-07
Too much churn in translation templates


tothxa:
**Describe the bug**
Most changes in translation updates are just reordering translator comments (IIRC also back references). This not only pollutes the commit history, but it often makes it very hard to see the actual translation updates.

**To Reproduce**
1. See the diff of almost any translation update. (eg. a1e798f `po/tribes_encyclopedia/<any lang>.po`)

**Expected behavior**
Ordering of comments for a string should not change when there is no actual change.

**Additional context**
Probably a gettext bug, but can we find or write a workaround?

Noordfrees:
buildcats.py does not pass the files to xgettext in any particular order. Some alphabetical sorting on the filenames might do the trick.
```diff
diff --git a/utils/buildcat.py b/utils/buildcat.py
index 51be558393..0156f613fe 100755
--- a/utils/buildcat.py
+++ b/utils/buildcat.py
@@ -242,7 +242,8 @@ def run_xgettext(infiles, outfile, opts):
     xgettext = subprocess.Popen("xgettext %s --files-from=- --output=\"%s\"" %
                                 (opts, outfile), shell=True, stdin=subprocess.PIPE, universal_newlines=True)
     try:
-        for fname in infiles:
+        for fname in sorted(infiles):
             xgettext.stdin.write(os.path.normpath(fname) + '\n')
         xgettext.stdin.close()
     except IOError as err_msg:
```

tothxa:
Oh, I thought xgettext was supposed to order them. Will try to test this.

tothxa:
I tested it and it works.

I was able to reliably reproduce the problem by running the script with python3 in different branches, but with the patch I get the same order in all of them.

With python2 (default on my Debian oldstable system) there was no problem (running it in different branches resulted in the same order), however the python2 stable order is different from the python3 sorted order.

tothxa:
@Noordfrees Can we have this before the next translation update? I don't see any risk, and it would be most useful during the freeze period. (it'll still require 2 runs after merging to settle down, due to #5549)

tothxa:
Thank you!

PR #6040, opened 2023-07-29
smaller changes in po files by script removing line numbers

SimonHeimberg:

**Type of change**
Enhancement

**Issue(s) closed**
none -

Looking at the history of po files or a commit of translation (see [ar.po in 41ecfa93](/widelands/widelands/commit/3b56cc4094a1d89e18a9c02c97b089dfa4fa1516#diff-86aeb00a79ac1d1a7abdb53dd1abfeec09f716970145055e6032ce1d9964e43f)) shows many files only changing line numbers (plus the pot-date).

I suggest to remove the line numbers from po files (but keep them in pot files).
Translators still see the location on transifex, no change for them.
If somebody wants to translate locally, the po files with location can be downloaded from transifex or generated with msgmerge xx.pot

`tc pull ...` has no argument to fetch the files without locations.

The file path could be deleted as well (as with `msgcat --no-location`).

**Additional context**
`msgcat --add-location=file` also keeps the file only (and msgXXX --add-location), but each file is only listed once

Change of line numbers is still visible in pot files (so 1x).
Translators on transifex will not see any difference.

removes noise from translation, loosely related to #5550

**Implementation steps**
- [x] adapt utils/merge_and_push_translations.sh
- [ ] do the duplicate lines hurt? (When there are several occurrences in one file, the file is listed several times.)
       This is different to `msgcat --add-location=file`.
- [ ] probably remove translation lines from po files in one commit (for less noise later)

### DIFF
--- a/utils/merge_and_push_translations.sh
+++ b/utils/merge_and_push_translations.sh
@@ -62,6 +62,8 @@ set -x
 # Force-pull translations because some would get skipped accidentally
 tx pull -fa
 
+git diff --diff-filter=MA --name-only '*.po' | xargs -r -d '\n' -- sed -i '/^#: / s/:[0-9]*$//' # remove source line numbers from po files
+
 # Update authors file
 python3 utils/update_authors.py
 if [ $? -eq 0 ] ; then


tothxa:
:-1: 

I've considered this in the past too, but finally I came to the decision that it's better to put up with this part of the churn. I translate most stuff locally, but debugging is easier too if the .po files have this info.

tothxa:
@SimonHeimberg Can we close this as replaced by #6064?

PR #6049, opened 2023-07-31
Add missing `richtext_escape` for add-ons error message

Noordfrees:
**Type of change**
Bugfix

**Issue(s) closed**
If the add-ons server sends an error message like `java.lang.NullPointerException: Cannot read the array length because "<parameter2>" is null`, the `<` symbol needs a richtext escape.

Happened a moment ago due to a cornercase bug in the server which I have already fixed there directly.


tothxa (approved):
Looks safe.

PR #6055, opened 2023-08-03
Fix `Window::move_inside_parent` bottom docking

Noordfrees:
**Type of change**
Bugfix

**To reproduce**
1. Configuration: Toolbar at the top, dock windows to edges enabled with 8px snap distance.
2. In a game, open some saveloaded window (e.g. stock menu) and dock it to the bottom edge.
3. Save and reload (without changing the size of the Widelands window).
4. The window floats some distance above the bottom edge.

**New behavior**
Don't subtract the toolbar height in the calculation of the new Y position. Now the window is bottom-edge-snapped in the described configuration.

**Possible regressions**
Vertical repositioning of windows in lots of *other* configurations, with other toolbar positions and edge docking on/off. I have no concept of how exactly this magic formula really works anymore…


tothxa:
I think only lines 387 and 388 need re-adding `toolbar_top_h` to `ph`.

src/ui_basic/window.cc
```diff
@@ -384,8 +384,8 @@ void Window::move_inside_parent() {
 		} else {
 			if (py < toolbar_top_h) {
 				py = toolbar_top_h;
-			} else if (py + get_h() > ph) {
-				py = ph - get_h();
+			} else if (py + get_h() > ph + toolbar_top_h + toolbar_bottom_h) {
+				py = ph - get_h() + toolbar_top_h + toolbar_bottom_h;
```
tothxa: Only top. This is actually a coordinate transformation for `py`: The toolbar height is subtracted from `ph` to make it the effective usable height for comparisons with the window height. But the toolbar is also inside the parent, so the "effective" y coordinate (distance from the bottom of the toolbar in case of toolbar on top) is `toolbar_top_h` less than `py`. (Lines 385-386 handle the same issue for the "top of window above the top of the effective usable area" case.)

```suggestion
			} else if (py + get_h() > ph + toolbar_top_h) {
				py = ph - get_h() + toolbar_top_h;
```

tothxa: Or maybe this one's better:

```suggestion
			} else if (const int32_t p_bottom = parent->get_inner_h() - toolbar_bottom_h; py + get_h() > p_bottom) {
				py = p_bottom - get_h();
```

tothxa (approved):
Tested and works.

tothxa:
>     2. In a game, open some saveloaded window (e.g. stock menu) and dock it to the bottom edge.
> 
>     3. Save and reload (without changing the size of the Widelands window).

You don't need save/load for testing.

2. Open some statistics window and dock it to the bottom edge
3. Close and reopen the window

Issue #6036, opened 2023-07-26
Improve box layout logic


matthiakl:
**Is your feature request related to a problem? Please describe.**
The UI layout using boxes is convenient but has some quirks that can cause unexpected layout issues (mostly during resizing).

**Describe the solution you'd like**
The box layout logic should be cleaned up to provide a more robust and intuitive layout behavior. 

**Additional context**
Re #5631, #6016 and associated regressions

**Implementation steps**
- [ ] Add item type ItemBox to handle nested boxes
  - [ ] Propagate `set_max_size` calls to child boxes to stay in bounds 
    (currently the bottom most child box has to set the correct size, otherwise it overflows)
  - [ ] Do not assign variable size larger than childs max size 
    (this is what max size should limit, right?)
- [ ] Do not always update desired size on `add` or `set_max_size` to prevent resizing issues 
   (the parents desired layout is implicitly overridden and might break)
- [ ] Sanitize box content size to fit into defined dimensions
  - [ ]  Reduce recursive calls of `set_desired_size` → `update_desired_size` → `layout`
  - [ ] Fix overflowing layouts, the sum of all item sizes in a box should never exceed the actual box size

PR #6059, opened 2023-08-06
Options window cleanup

### LINKED ISSUE #4733: Thoughts on options menu
> The Interface and Window tabs of the options menu were merged recently because the tab row got too wide, so it > seems rather cluttered now. Suggestions how to unclutter it without adding more tabs welcome…

_Originally posted by @Noordfrees in https://github.com/widelands/wl_addons_server/issues/23#issuecomment-786811833_

Edit: Will ad some of my thoughts in the next days

tothxa:
**Type of change**
Cleanup

**Issue(s) closed**
Fixes #4733

Also fixed a bug in `Statebox::layout()` that prevented setting the width of checkboxes to exactly fit the text.

**Possible regressions**
 * Setting options, especially writing syncstreams and mouse grabbing issues.
 * Checkbox layouting.

**Additional context**
The translation stats space fix has some ugly workarounds for issues like #6036

### DIFF
--- a/src/wlapplication.cc
+++ b/src/wlapplication.cc
@@ -465,17 +465,6 @@ WLApplication::WLApplication(int const argc, char const* const* const argv)
 	// register it once.
 	UI::Panel::register_click();
 
-	// Disable mouse grabbing
-	// TODO(tothxa): keep around until tested without it
-	/*
-	if (g_gr != nullptr) {
-	   SDL_Window* sdl_window = g_gr->get_sdlwindow();
-	   if (sdl_window != nullptr) {
-	      SDL_SetWindowGrab(sdl_window, SDL_FALSE);
-	   }
-	}
-	*/
-
 	// Make sure we didn't forget to read any global option
 	check_config_used();
 
--- a/src/wlapplication.cc
+++ b/src/wlapplication.cc
@@ -465,7 +465,16 @@ WLApplication::WLApplication(int const argc, char const* const* const argv)
 	// register it once.
 	UI::Panel::register_click();
 
-	set_input_grab(get_config_bool("inputgrab", false));
+	// Disable mouse grabbing
+	// TODO(tothxa): keep around until tested without it
+	/*
+	if (g_gr != nullptr) {
+		SDL_Window* sdl_window = g_gr->get_sdlwindow();
+		if (sdl_window != nullptr) {
+			SDL_SetWindowGrab(sdl_window, SDL_FALSE);
+		}
+	}
+	*/
 
 	// Make sure we didn't forget to read any global option
 	check_config_used();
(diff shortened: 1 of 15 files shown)


Noordfrees:
I don't like the changed sound options. With the new design it is no longer intuitive which slider belongs to which checkbox because the distance is so huge. The sliders just seem to be hanging in the middle of nothing with no visual connection to the labels or checkboxes.

grafik

Other than that, tested and working, mouse grab and syncstreams behaviour as expected. I like the cleaner General options tab :)

tothxa:
> I don't like the changed sound options. With the new design it is no longer intuitive which slider belongs to which checkbox because the distance is so huge. The sliders just seem to be hanging in the middle of nothing with no visual connection to the labels or checkboxes.

I made the label checkboxes adapt to the biggest actual text width. Turns out checkboxes had a bug in the `layout()` function that prevented fitting them closely, so I fixed that too. Please check if it's better for you this way.

It still doesn't work right for RTL languages, but I think that's a bug in the `text_width()` function (in `graphic/text_layout`), probably due to using LTR fonts for the missing translations. So out of scope here.

> Other than that, tested and working, mouse grab and syncstreams behaviour as expected. I like the cleaner General options tab :)

Well, not for long if I finish #5944 ;)

Noordfrees:
That's better but still looks somewhat untidy. Since the tab has a lot of unused space, perhaps you could increase the spacing between the rows to set them more apart, like this:
grafik

Also, another idea: Instead of disabling the sliders for deactivated sound how about hiding them completely and showing sliders only for enabled sound types?

tothxa:
> That's better but still looks somewhat untidy. Since the tab has a lot of unused space, perhaps you could increase the spacing between the rows to set them more apart
> 
> Also, another idea: Instead of disabling the sliders for deactivated sound how about hiding them completely and showing sliders only for enabled sound types?

Both done. Though hiding sliders is even worse for RTL, because the checkbox jumps to the left when a type is disabled. Again, I tried to solve it, but gave up. (I guess I could work it around again with extra padding panels, but I didn't want to.)

src/wlapplication.cc
```diff
@@ -465,7 +465,16 @@ WLApplication::WLApplication(int const argc, char const* const* const argv)
 	// register it once.
 	UI::Panel::register_click();
 
-	set_input_grab(get_config_bool("inputgrab", false));
+	// Disable mouse grabbing
+	// TODO(tothxa): keep around until tested without it
```
Noordfrees: tested, this can go

Noordfrees (commented):
Sound options look good to me now :)

Code LGTM

Noordfrees (approved):

tothxa:
Thank you. :)

PR #6060, opened 2023-08-08
Fix editbox horizontal scrolling mouse clicks regression

Noordfrees:
**Type of change**
Bugfix

**Issue(s) closed**
Fixes a regression with clicking into editboxes that scroll horizontally.

**To reproduce**
1. In any single-line editbox, enter a very long text so it scrolls horizontally.
2. Move the caret to the right.
3. Click into the editbox with the mouse to move the caret or to select text.

**New behavior**
Add the scrolling offset in the cursorpos-to-caretpos calculation.


tothxa (approved):
Code looks OK. Tested and working.

PR #6062, opened 2023-08-09
Fix documentation CI workflow

Noordfrees:
`sphinxcontrib-serializinghtml` v1.1.6 breaks the workflow, force version 1.1.5


tothxa (approved):
We'll have to watch when it gets fixed upstream, but it made it work.

PR #6063, opened 2023-08-09
WIP: Cleanup shellcheck

SimonHeimberg:
**Type of change**
Cleanup

**Issue(s) closed**
Fixes -

**Rationale**
Why checking shell scripts makes sense is shown in many examples on https://shellcheck.net (load a random example)

suggestions for github actions:
- shellcheck: https://github.com/marketplace/actions/differential-shellcheck
- actionlint: https://github.com/raven-actions/actionlint

**Possible regressions**
I fixed carefully and tested many scripts, so hopefully none.


tothxa:
Just one more thing against it: It already needed 6 overrides in 15 scripts. That's a pretty high false positive rate IMO.

tothxa:
> I suggest to do a small, much more moderate cleanup.

Cleanup is fine. :) Just don't try to "fix" what isn't broken. :)

> I vote to do [8df0cd8](https://github.com/widelands/widelands/commit/8df0cd8e60de0520770252f33133f68ccbe679a9) because this not as planned quite sure. ( [ -f a/b -a c/d ] should have -f before the 2nd path)

I even missed the actual bugfix in those. Certainly, `g++`/`clang++` needs the `-f`. But please don't replace `[ expr1 -a expr2 ]` with `[ expr1 ] && [ expr2 ]`. This use is most certainly not ambiguous, so using 2 tests is just wrong and ugly as far as I'm concerned.

> And some things which could do harm: `rm -Rf $xxx/` # in case the var is unset, this would try to delete all files.

But if there's no code path in the script where it isn't set, then why not?

> and maybe some which makes code more readable: `$(cmd x)` # instead of `` `cmd x` ``, at least I do hardly see the optical difference between ` and '

I don't quite get the zealous crusade against backticks myself (like bash completion refusing to work within them), but have no opposition for that one.

> That there is no simple way to check if a directory is empty (or the opposite) I see as a shortcoming of bash (https://www.cyberciti.biz/faq/linux-unix-shell-check-if-directory-empty/, ...)

That page seems to describe many things, except for just testing. I guess I'd do `[ -d "$DIR" -a $(find $DIR | wc -l) -eq 1 ]`. (is that `-a` ambiguous? I don't know, but if the parser is halfway sane, it shouldn't be. If it is, it's a bug in the parser, not a limitation of `-a`. No quotes just in spite. :) ) And the `-d` test may be omitted, if the script already knows that the dir exists.

hessenfarmer:
since this has been lying around now for over a year I believe we could savely close it

PR #5937, opened 2023-06-05
Skip timestamp-only POT changes during translations update

tothxa:
**Type of change**
Bugfix / Refactoring

**Issue(s) closed**
Fixes #5936

**Possible regressions**
Translation updates

**Additional context**
I made it only check the templates. PO files shouldn't have timestamp-only updates of their own, but if something slips through, they should be in sync with the template anyway.

PR #6064, opened 2023-08-09
commit pure translations separately

SimonHeimberg:
**Type of change**
Enhancement

**Issue(s) closed**
replaces #6040 

**To reproduce**
Try to review the translations in 0c2abcc5b6. Quite difficult because there are many line changes as well. 

**New behavior**
Edits of translations are committed separately from updating the files from the code.
This allows to review the translations much simpler.

**How it works**
- Fetch the translations from transifex
- commit those if there are any (with a identifiable commit message)
- update the translation template files (.pot)
- push them to transifex
- fetch the updated translations
- (update other files like statistics and authors)
- commit translations files and their templates (with changes related to each other)

**Possible regressions**
- a bit more noise because the same translation file might be committed twice in a row
- no other known

**Additional context**
not sure about the commit messages: `Fetched translations.` and `Updated translations catalogs.`
(before `Fetched translations and updated catalogs.`)

I included some clean-up as well (related to #6063) because having edits and cleanup of one file in separate pr results in merge conflicts.

### DIFF
--- a/utils/merge_and_push_translations.sh
+++ b/utils/merge_and_push_translations.sh
@@ -106,11 +106,12 @@ set -x
 # use force to make sure really all files get pulled
 tx pull -a -f
 
-runUpdateScripts=1 # run them at least once
+run_update_scripts=at_least_once # run them at least once
 if [ -n "$(git status -s)" ]; then
   update_authors
   update_appdata
   update_statistics
+  run_update_scripts= # they have run now
 
   # Stage translations
   git add 'po/*/*.po' 'data/i18n/locales/*.json' 'xdg/translations/*.json'
@@ -119,8 +120,6 @@ if [ -n "$(git status -s)" ]; then
 
   # commit translations
   git commit -m "Fetched translations and updated data."
-
-  runUpdateScripts= # has run now
 fi
 
 # -----------------------
@@ -154,10 +153,13 @@ if [ -n "$(git status -s)" ]; then
 
   # Undo one-liner diffs of pure timestamps with no other content
   undo_oneliner_diffs
-  runUpdateScripts=2 # run, files have changed
+
+  if [ -z "$run_update_scripts" ] && [ -n "$(git status -s)" ]; then
+    run_update_scripts=for_pot # run again, pot files have changed
+  fi
 fi
 
-if [ -n "$runUpdateScripts" ]; then # if not run yet or pot file has changed
+if [ -n "$run_update_scripts" ]; then # if not run yet or pot files have changed
   update_authors # in case something has changed unexpectedly
   update_appdata # in case something has changed unexpectedly
   update_statistics
--- a/utils/merge_and_push_translations.sh
+++ b/utils/merge_and_push_translations.sh
@@ -45,7 +45,7 @@ git checkout master
 git pull "$push_target" master
 
 # Double-check that it's clean
-STATUS="$(git status)"
+STATUS="$(LANG=C git status)"
 echo "${STATUS}"
 if [[ "${STATUS}" != *"nothing to commit, working tree clean"* ]]; then
   echo "git status must be empty to prevent accidental commits etc."
--- a/utils/merge_and_push_translations.sh
+++ b/utils/merge_and_push_translations.sh
@@ -43,6 +43,7 @@ fi
 # Checkout master and pull latest version
 git checkout master
 git pull "$push_target" master
+had_commit=''
 
 # Double-check that it's clean
 STATUS="$(LANG=C git status)"
--- a/utils/merge_and_push_translations.sh
+++ b/utils/merge_and_push_translations.sh
@@ -171,15 +171,13 @@ if [ -n "$(git status -s)" ]; then
   # Undo one-liner diffs of pure timestamps with no other content
   # for fetched translations (*.po) (in case something plays tricks, see #5937)
   undo_oneliner_diffs
-fi
 
-# in case translations were edited while this script was running
-update_authors
-update_appdata
-# and this also changes by a catalog change
-update_statistics
+  # in case translations were edited while this script was running
+  update_authors
+  update_appdata
+  # and this also changes by a catalog change
+  update_statistics
 
-if [ -n "$(git status -s)" ]; then
   # Stage changes
   # - Translations and templates
   git add 'po/*/*.po' 'po/*/*.pot' 'data/i18n/locales/*.json' 'xdg/translations/*.json' || true
--- a/utils/merge_and_push_translations.sh
+++ b/utils/merge_and_push_translations.sh
@@ -57,6 +57,18 @@ echo "Working tree is clean, continuing"
 # Print all commands.
 set -x
 
+# Pull All translations from Transifex
+# use git timestamps because the timestamps on disk are too new
+tx pull -a --use-git-timestamps
+
+if [ -n "$(git status -s)" ]; then
+  # Stage and commit translations
+  git add po/*/*.po data/i18n/locales/*.json xdg/translations/*.json || true
+  git commit -m "Fetched translations."
+fi
+
+# -----------------------
+
 # Update source catalogs
 python3 utils/buildcat.py
 
@@ -82,7 +94,8 @@ if [ -n "$(git status -s)" ]; then
 fi
 
 # Pull All translations from Transifex
-tx pull -a --use-git-timestamps
+# (timestamps on disk are old enough this time)
+tx pull -a
 
 # Update authors file
 if python3 utils/update_authors.py; then
@@ -117,21 +130,22 @@ undo_oneliner_diffs
 
 if [ -z "$(git status -s)" ]; then
   echo "Run completed, nothing to commit."
-  exit 0
+else
+  # Stage changes
+  # - Translations
+  git add po/*/*.po po/*/*.pot data/i18n/locales/*.json xdg/translations/*.json || true
+  # - Authors
+  git add data/txts/*.lua || true
+  # - Locale data
+  git add data/i18n/*.lua || true
+  # - Appdata
+  git add xdg/org.widelands.Widelands.appdata.xml xdg/org.widelands.Widelands.desktop || true
+  # - Statistics
+  git add data/i18n/translation_stats.conf || true
+
+  # Commit
+  git commit -m "Updated translations catalogs."
 fi
 
-# Stage changes
-# - Translations
-git add po/*/*.po po/*/*.pot data/i18n/locales/*.json xdg/translations/*.json || true
-# - Authors
-git add data/txts/*.lua || true
-# - Locale data
-git add data/i18n/*.lua || true
-# - Appdata
-git add xdg/org.widelands.Widelands.appdata.xml xdg/org.widelands.Widelands.desktop || true
-# - Statistics
-git add data/i18n/translation_stats.conf || true
-
-# Commit and push.
-git commit -m "Fetched translations and updated catalogs."
+# push fetched translations and updated catalogs
 git push "$push_target" master
--- a/utils/merge_and_push_translations.sh
+++ b/utils/merge_and_push_translations.sh
@@ -137,6 +137,7 @@ if [ -n "$(git status -s)" ]; then
 
   # commit translations
   git commit -m "Fetched translations and updated data."
+  had_commit=yes
 fi
 
 # -----------------------
@@ -184,9 +185,10 @@ else
 
   # Commit
   git commit -m "Updated translations catalogs and statistics."
+  had_commit=yes
 fi
 
-if [ "$(git show --no-patch --format=format:_ FETCH_HEAD HEAD)" != "_" ]; then # check if it is the same commit
+if [ -n "$had_commit" ]; then # if a commit was created
   # push fetched translations and updated catalogs
   git push "$push_target" master
 fi
--- a/utils/merge_and_push_translations.sh
+++ b/utils/merge_and_push_translations.sh
@@ -147,5 +147,7 @@ else
   git commit -m "Updated translations catalogs."
 fi
 
-# push fetched translations and updated catalogs
-git push "$push_target" master
+if [ "$(git show --no-patch --format=format:_ FETCH_HEAD HEAD)" != "_" ]; then # check if it is the same commit
+  # push fetched translations and updated catalogs
+  git push "$push_target" master
+fi
--- a/utils/merge_and_push_translations.sh
+++ b/utils/merge_and_push_translations.sh
@@ -82,9 +82,7 @@ if [ -n "$(git status -s)" ]; then
 fi
 
 # Pull All translations from Transifex
-# tx pull -a
-# Force-pull translations because some would get skipped accidentally
-tx pull -fa
+tx pull -a --use-git-timestamps
 
 # Update authors file
 if python3 utils/update_authors.py; then
--- a/utils/merge_and_push_translations.sh
+++ b/utils/merge_and_push_translations.sh
@@ -20,14 +20,14 @@ fi
 
 # Move up if we're not in the base directory.
 if [ -d "../utils" ]; then
-	pushd ..
+  pushd ..
 fi
 
 # Make sure 'utils/buildcat.py' is there.
 if [ ! -f "utils/buildcat.py" ]; then
-	echo "Unable to find 'utils/buildcat.py'."
-	echo "Make sure you start this script from Widelands' base or utils directory.";
-	exit 1
+  echo "Unable to find 'utils/buildcat.py'."
+  echo "Make sure you start this script from Widelands' base or utils directory.";
+  exit 1
 fi
 
 # Ensure that our git is clean
--- a/utils/merge_and_push_translations.sh
+++ b/utils/merge_and_push_translations.sh
@@ -90,9 +90,9 @@ update_statistics() {
 gitAddGeneratedFiles() {
   # Stage changes
   # - Authors
-  git add data/txts/*.lua || true
+  git add 'data/txts/*.lua' || true
   # - Locale data
-  git add data/i18n/*.lua || true
+  git add 'data/i18n/*.lua' || true
   # - Appdata
   git add xdg/org.widelands.Widelands.appdata.xml xdg/org.widelands.Widelands.desktop || true
   # - Statistics
@@ -113,7 +113,7 @@ if [ -n "$(git status -s)" ]; then
   update_statistics
 
   # Stage translations
-  git add po/*/*.po data/i18n/locales/*.json xdg/translations/*.json
+  git add 'po/*/*.po' 'data/i18n/locales/*.json' 'xdg/translations/*.json'
   # and generated files
   gitAddGeneratedFiles
 
@@ -168,7 +168,7 @@ if [ -z "$(git status -s)" ]; then
 else
   # Stage changes
   # - Translations and templates
-  git add po/*/*.po po/*/*.pot data/i18n/locales/*.json xdg/translations/*.json || true
+  git add 'po/*/*.po' 'po/*/*.pot' 'data/i18n/locales/*.json' 'xdg/translations/*.json' || true
   # - generated files
   gitAddGeneratedFiles
 
--- a/utils/merge_and_push_translations.sh
+++ b/utils/merge_and_push_translations.sh
@@ -131,8 +131,8 @@ python3 utils/buildcat.py
 undo_oneliner_diffs() {
   # Undo one-liner diffs of pure timestamps with no other content
   set +x
-  git diff --numstat po/ | sed -En 's/^1\t1\t//p' | while IFS= read -r entry; do
-    if ! git diff "$entry" | grep '^[+-][^+-]' | grep -qv '^[+-]"POT-Creation-Date:'
+  for entry in $(git diff --numstat po/ | sed -En 's/^1\t1\t//p'); do
+    if [ -z "$(git diff "$entry" | grep '^[+-][^+-]' | grep -v '^[+-]"POT-Creation-Date:')" ]
     then # no other diff line remaining
       echo "Skipping changes to $entry"
       git checkout "$entry"
--- a/utils/merge_and_push_translations.sh
+++ b/utils/merge_and_push_translations.sh
@@ -63,11 +63,11 @@ python3 utils/buildcat.py
 undo_oneliner_diffs() {
   # Undo one-liner diffs of pure timestamps with no other content
   set +x
-  for entry in $(git diff --numstat po/ | sed -En 's/^1\t1\t//p'); do
-    if [ -z "$(git diff "$entry" | grep '^[+-][^+-]' | grep -v '^[+-]"POT-Creation-Date:')" ]
-    then
+  git diff --numstat po/ | sed -En 's/^1\t1\t//p' | while IFS= read -r entry; do
+    if ! git diff "$entry" | grep '^[+-][^+-]' | grep -qv '^[+-]"POT-Creation-Date:'
+    then # no other diff line remaining
       echo "Skipping changes to $entry"
-      git checkout $entry
+      git checkout "$entry"
     fi
   done
   set -x
@@ -87,8 +87,7 @@ fi
 tx pull -fa
 
 # Update authors file
-python3 utils/update_authors.py
-if [ $? -eq 0 ] ; then
+if python3 utils/update_authors.py; then
   echo "Updated authors";
 else
   echo "Failed updating authors";
@@ -96,8 +95,7 @@ else
 fi
 
 # Update appdata
-python3 utils/update_appdata.py
-if [ $? -eq 0 ] ; then
+if python3 utils/update_appdata.py; then
   echo "Updated appdata";
 else
   echo "Failed updating appdata";
@@ -105,8 +103,7 @@ else
 fi
 
 # Update statistics
-python3 utils/update_translation_stats.py
-if [ $? -eq 0 ] ; then
+if python3 utils/update_translation_stats.py; then
   echo "Updated translation stats";
 else
   echo "Failed to update translation stats";
--- a/utils/merge_and_push_translations.sh
+++ b/utils/merge_and_push_translations.sh
@@ -55,6 +55,9 @@ fi
 
 echo "Working tree is clean, continuing"
 
+# -----------------------
+# functions:
+
 update_authors() {
   # Update authors file
   if python3 utils/update_authors.py; then
@@ -116,6 +119,10 @@ undo_oneliner_diffs() {
 # Print all commands.
 set -x
 
+# ------------------------------------------------
+# translations from transifex (*.po, *.json)
+# and changes from git (since last script run)
+
 # Pull All translations from Transifex
 # use force to make sure really all files get pulled
 tx pull -a -f
@@ -142,6 +149,7 @@ if [ -n "$(git status -s)" ]; then
 fi
 
 # -----------------------
+# Source catalogs:
 
 # Update source catalogs
 python3 utils/buildcat.py
@@ -163,16 +171,15 @@ if [ -n "$(git status -s)" ]; then
   # Undo one-liner diffs of pure timestamps with no other content
   # for fetched translations (*.po) (in case something plays tricks, see #5937)
   undo_oneliner_diffs
-
 fi
 
-update_authors # in case something has changed unexpectedly
-update_appdata # in case something has changed unexpectedly
+# in case translations were edited while this script was running
+update_authors
+update_appdata
+# and this also changes by a catalog change
 update_statistics
 
-if [ -z "$(git status -s)" ]; then
-  echo "Run completed, nothing to commit."
-else
+if [ -n "$(git status -s)" ]; then
   # Stage changes
   # - Translations and templates
   git add 'po/*/*.po' 'po/*/*.pot' 'data/i18n/locales/*.json' 'xdg/translations/*.json' || true
--- a/utils/merge_and_push_translations.sh
+++ b/utils/merge_and_push_translations.sh
@@ -54,6 +54,51 @@ fi
 
 echo "Working tree is clean, continuing"
 
+update_authors() {
+  # Update authors file
+  if python3 utils/update_authors.py; then
+    echo "Updated authors"
+  else
+    echo "Failed updating authors"
+    exit 1
+  fi
+  # Fix formatting for Lua
+  python3 utils/fix_formatting.py --lua --dir data/i18n
+  python3 utils/fix_formatting.py --lua --dir data/txts
+}
+
+update_appdata() {
+  # Update appdata
+  if python3 utils/update_appdata.py; then
+    echo "Updated appdata"
+  else
+    echo "Failed updating appdata"
+    exit 1
+  fi
+}
+
+update_statistics() {
+  # Update statistics
+  if python3 utils/update_translation_stats.py; then
+    echo "Updated translation stats"
+  else
+    echo "Failed to update translation stats"
+    exit 1
+  fi
+}
+
+gitAddGeneratedFiles() {
+  # Stage changes
+  # - Authors
+  git add data/txts/*.lua || true
+  # - Locale data
+  git add data/i18n/*.lua || true
+  # - Appdata
+  git add xdg/org.widelands.Widelands.appdata.xml xdg/org.widelands.Widelands.desktop || true
+  # - Statistics
+  git add data/i18n/translation_stats.conf || true
+}
+
 # Print all commands.
 set -x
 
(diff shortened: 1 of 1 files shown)


utils/merge_and_push_translations.sh
```diff
@@ -57,17 +57,29 @@ echo "Working tree is clean, continuing"
 # Print all commands.
 set -x
 
+# Pull All translations from Transifex
+# use git timestamps because the timestamps on disk are too new
+tx pull -a --use-git-timestamps
```
tothxa: Wouldn't it be safer to still force pull and let git figure out whether there were changes? (see also https://github.com/widelands/widelands/pull/6064#discussion_r1289806795)

utils/merge_and_push_translations.sh
```diff
@@ -57,17 +57,29 @@ echo "Working tree is clean, continuing"
...
 # Update source catalogs
 python3 utils/buildcat.py
 
 undo_oneliner_diffs() {
   # Undo one-liner diffs of pure timestamps with no other content
   set +x
-  for entry in $(git diff --numstat po/ | sed -En 's/^1\t1\t//p'); do
-    if [ -z "$(git diff "$entry" | grep '^[+-][^+-]' | grep -v '^[+-]"POT-Creation-Date:')" ]
-    then
+  git diff --numstat po/ | sed -En 's/^1\t1\t//p' | while IFS= read -r entry; do
```
tothxa: I can understand the rationale for this, but still think it's much less readable than `for … in $(…)`.

utils/merge_and_push_translations.sh
```diff
@@ -57,17 +57,29 @@ echo "Working tree is clean, continuing"
...
+    if ! git diff "$entry" | grep '^[+-][^+-]' | grep -qv '^[+-]"POT-Creation-Date:'
```
tothxa: Probably it's only me (and I understand this is the fault of the shell syntax), but it feels icky (or at least hard to read) to depend on the return value of the last element of a long pipeline in if.

utils/merge_and_push_translations.sh
```diff
@@ -82,31 +94,27 @@ if [ -n "$(git status -s)" ]; then
 fi
 
 # Pull All translations from Transifex
-# tx pull -a
-# Force-pull translations because some would get skipped accidentally
-tx pull -fa
+# (timestamps on disk are old enough this time)
+tx pull -a
```
tothxa: Please keep force pull. I've seen the tx client miss changes many times even with old local files.

utils/merge_and_push_translations.sh
```diff
@@ -122,21 +130,24 @@ undo_oneliner_diffs
...
+  git add data/txts/*.lua || true
+  # - Locale data
+  git add data/i18n/*.lua || true
+  # - Appdata
+  git add xdg/org.widelands.Widelands.appdata.xml xdg/org.widelands.Widelands.desktop || true
+  # - Statistics
+  git add data/i18n/translation_stats.conf || true
+
+  # Commit
+  git commit -m "Updated translations catalogs."
```
tothxa:
```suggestion
  git commit -m "Updated translations catalogs and statistics."
```

or so…

Maybe appdata and authors should be done for first commit. I didn't check, but don't think they're affected by `.pot` changes.

tothxa (commented):
I think I like the idea, but I'd feel safer if we got other opinions whether the double commits are OK.

utils/merge_and_push_translations.sh
Noordfrees: Yes, please always force-pull, `tx`'s "detection" of "recent" files is extremely unreliable

Noordfrees:
I'm not really sure about this. On the one hand it's a good point, the noise from line number updates can make it impossible to filter out the "real" changes. On the other hand that does not happen all the time, and having 2 commits every time is also a different kind of noise that makes changelogs and commit ranges harder to review. I'm not sure how this will look in practice, but since this change should not impact anything, IMHO there's no harm in having it now, and if it turns out to create too many useless commits we can still reconsider.

tothxa:
There's one more risk: IIRC tx has some kind of rate limiting, so pulling all translations twice in a row may be too much for them.

Noordfrees:
The rate limit should not be a problem for us:

> To maintain optimal performance and availability of the API, we limit the request rate to 500 requests/min or 30,000 requests / 5 minutes / IP. Polling endpoints like `resource_translations_async_downloads/{resource_translations_async_download_id}` and `resource_strings_async_downloads/{resource_strings_async_download_id}` have a rate limit of 1,200 requests/min. Once past the respective limit, requests will be throttled and you will get a 429 Throttled response.
>
> https://developers.transifex.com/reference/rate-limit

utils/merge_and_push_translations.sh
SimonHeimberg: The changed pot files will be uploaded to transifex and then the po files downloaded again. So I am totally unsure about the effects.

tothxa:
> > Except for statistics, which are affected by both… maybe update them in each step?
> 
> 🤔 This could be helpful. Unfortunately I do not really understand the statistics. Why do the number for most languages go down in [5befd11](https://github.com/widelands/widelands/commit/5befd110003feaf80ccc889590b450272be2bdfc)? (At least "lt" and general go up.)

I think general is the total number of words in translatable strings in all .pot files. So it goes up when there are new translatable strings. While I presume languages get the total number of words in the original translatable strings that have a translation in the given lang.

And lt really had a handful of new translations. :)

> We might leave it as now and a translation change triggers a 2nd commit for sure. 🤔 🤔

What I mean is to run `update_translation_stats.py` after both pulls and merge the new values in git along with the other files both times. This way stat changes due to new translations and due to pot changes are separated too in the 2 separate commits, and they don't trigger a "catalog" commit when there are only new translations but no pot changes.

utils/merge_and_push_translations.sh
tothxa: We have the appdata translations and the list of translators in JSON format, which are not affected by gettext. (I'm not sure how the appdata source can get updated.) `tx` is configured to pull these along with .po files, and `update_appdata.py` and `update_authors.py` merges them. That's why they should be run with the first .po pull, not with the .pot push and consequent pull.

SimonHeimberg:
Just thinking "loud":

As the commits are made now:
1. The commit "Fetched translations." just contains translation changes. It makes sense to review those.
2. When the number of translations changes, this triggers a 2nd commit "Updated translations catalogs and ..." just for statistics (by update_translation_stats.py). I do not like a commit for statistic changes only either.
3. When a new translator worked on transifex, the 2nd commit contains changes to `data/txts/translators_data.lua` (by update_authors.py)
4. When a new locale is added (on transifex which generates po files), the 2nd commit contains `data/i18n/locales.lua` (by update_authors.py) ...
5. ... and `xdg/org.widelands.Widelands.appdata.xml`  and `xdg/org.widelands.Widelands.desktop` (at least when translated) (by update_appdata.py)
6. After a developer was added (manually to `data/txts/developers.json`), the 2nd commit contains `data/txts/developers.lua` (by update_authors.py)
7. When a pot file is changed, the 2nd commit contains the changed pot files and the resulting change in the po files.  And the statistics as well.
8. In case a translation is edited between downloading po files for the 1st and the 2nd time, those translations are included in the 2nd commit.
9. On release `xdg/org.widelands.Widelands.appdata.xml.stub` is updated, resulting in the 2nd commit (by update_appdata.py).

My result:

We want to keep 1. and 7., 8. can not be avoided (that I know of). We want to avoid 2., the rest just happens seldom but could be moved to the 1st commit ().
To avoid 2., `update_translation_stats.py` needs to run twice (as tothxa writes).
In case 7. happens, update_appdata.py and update_authors.py should run for the 2nd commit to not delay this changes to the next workflow run (where they would be totally unrelated).

Resulting in questions:

Which update_* calls should run before which commit?
Do you agree to my results?

utils/merge_and_push_translations.sh
SimonHeimberg: Will do. (I am a bit surprised to not find an issue about in https://github.com/transifex/cli/issues)

utils/merge_and_push_translations.sh
SimonHeimberg: Looking at it now I find it hard to read as well. Trying ...

tothxa:
> I agree to have update_appdata.py and update_authors.py in the first commit. Should we repeat them for the 2nd one, in case of 7.?

(I suppose you mean case 8) I guess it doesn't hurt to make extra sure everything's in sync, and they don't take long.

utils/merge_and_push_translations.sh
SimonHeimberg: I reverted it. An idea might be in a new pr.

utils/merge_and_push_translations.sh
SimonHeimberg: I reverted it. An idea might be in a new pr.

utils/merge_and_push_translations.sh
SimonHeimberg: Now `update_appdata.py` and `update_authors.py` run on the 1st commit (translations). And on the 2nd (catalogs) if a pot file has changed (or no po file had changed).

SimonHeimberg:
Updated the pr.

Now (from 0d6290f):
When translations are edited, only commit "translation" is created.
When translation sources change, only commit "catalog" is created.
When both change, both commits are created.
When a file was changed (on release, ...) and no translation changes,
  the commit "catalog" is created.
No commit triggers the other one.

utils/merge_and_push_translations.sh
```diff
@@ -54,9 +54,77 @@ fi
...
+}
+
 # Print all commands.
 set -x
 
+# Pull All translations from Transifex
+# use force to make sure really all files get pulled
+tx pull -a -f
+
+runUpdateScripts=1 # run them at least once
```
tothxa: I don't think we need this. We can probably trust git.

utils/merge_and_push_translations.sh
```diff
@@ -79,64 +147,36 @@ if [ -n "$(git status -s)" ]; then
...
 
-# Update appdata
-python3 utils/update_appdata.py
-if [ $? -eq 0 ] ; then
-  echo "Updated appdata";
-else
-  echo "Failed updating appdata";
-  exit 1;
+  # Undo one-liner diffs of pure timestamps with no other content
+  undo_oneliner_diffs
```
tothxa, line 175: Hmm, I believe this shouldn't be necessary any more, but tx tricked me last time too, so let's see it in action.

edit: OTOH, tx may dump such belated updates on us at any time, so if we keep it, then it should be done after the first tx pull as well.

utils/merge_and_push_translations.sh
```diff
@@ -79,64 +147,36 @@ if [ -n "$(git status -s)" ]; then
...
-python3 utils/update_translation_stats.py
-if [ $? -eq 0 ] ; then
-  echo "Updated translation stats";
-else
-  echo "Failed to update translation stats";
-  exit 1;
+if [ -n "$runUpdateScripts" ]; then # if not run yet or pot file has changed
+  update_authors # in case something has changed unexpectedly
+  update_appdata # in case something has changed unexpectedly
+  update_statistics
```
tothxa: Yes, you should do these right after `tx pull` again, and leave it to git to see if there were changes.

But the functions could include checking the git status whether their source files have changed before running anything.

utils/merge_and_push_translations.sh
```diff
@@ -79,64 +147,36 @@ if [ -n "$(git status -s)" ]; then
...
-git add data/i18n/*.lua || true
-# - Appdata
-git add xdg/org.widelands.Widelands.appdata.xml xdg/org.widelands.Widelands.desktop || true
-# - Statistics
-git add data/i18n/translation_stats.conf || true
-
-# Commit and push.
-git commit -m "Fetched translations and updated catalogs."
-git push "$push_target" master
+if [ "$(git show --no-patch --format=format:_ FETCH_HEAD HEAD)" != "_" ]; then # check if it is the same commit
```
tothxa: You could set e.g. `HAD_COMMIT=yes` on commits to make this test simpler.

utils/merge_and_push_translations.sh
```diff
@@ -54,9 +54,77 @@ fi
...
+tx pull -a -f
+
+runUpdateScripts=1 # run them at least once
+if [ -n "$(git status -s)" ]; then
+  update_authors
+  update_appdata
+  update_statistics
+
+  # Stage translations
+  git add 'po/*/*.po' 'data/i18n/locales/*.json' 'xdg/translations/*.json'
```
tothxa, line 142: Cool! I didn't know git did glob expansion. :)

utils/merge_and_push_translations.sh
SimonHeimberg: I do trust git. But I do not want to run update_XXX twice when files only changed once.
What I want is:
a) translations have changed and catalogs have changed: run update_XXX for "translation commit" and again for "catalog commit"
b) only translations have changed: run update_XXX for "translation commit" only
c) only catalogs have changed: run update_XXX for "catalog commit"
d) nothing has changed: run update_XXX for (potential) "catalog commit" in case a author/... file was changed since last run of this script. If it does not update anything, no commit is created.

utils/merge_and_push_translations.sh
SimonHeimberg: I kept it because of https://github.com/widelands/widelands/pull/5937#issuecomment-1585531380

After the 1st tx pull ... so [there](https://github.com/widelands/widelands/pull/6064/files#diff-3712968fb6773d1263ad66dbf48a0eff788ce700d64e936497b2f64c5c7bb680R107) ... makes sense of course

utils/merge_and_push_translations.sh
SimonHeimberg: I think this would not cover [case d) (from above)](#r1304847903)

edit:
Or do you mean to check in the case before, after `undo_oneliner_diffs`? This would save an unnecessary run.

utils/merge_and_push_translations.sh
SimonHeimberg: Shouldn't I trust git? :wink: I will adapt, makes it more obvious.

tothxa: I know. :) I meant it as "if you like", for readability. But it's OK this way too.

utils/merge_and_push_translations.sh
tothxa: Ah, I see.

(You should have included the json changes too in the descriptions. :)

So if I get it right, you want to skip `update*` on the first pull if there are no po changes, so in that case json changes will be committed on the second run, even if there are no catalog changes.

OTOH, json changes logically go together with po changes, ie. they are changes made on tx. So I think in the rare case when we get json changes, but no po changes, we should still do the updates and a "translation commit".

And regardless of the first commit, due to the extremely unlikely case when json files are edited on tx between the 2 pulls, the updates should still always be run the second time too (with `git status --short -- x.json` checks if you want to spare some cycles, see [below](https://github.com/widelands/widelands/pull/6064#discussion_r1300743261)) for consistency between the json inputs and the generated files.

utils/merge_and_push_translations.sh
SimonHeimberg: I did :+1: for readability.

utils/merge_and_push_translations.sh
SimonHeimberg: > Ah, I see.
> 
> (You should have included the json changes too in the descriptions. :)

In which description? :thinking: probably commit.
> 
> So if I get it right, you want to skip `update*` on the first pull if there are no po changes, so in that case json changes will be committed on the second run, even if there are no catalog changes.
> 

Almost (or maybe).

I want to run update_* (and do a commit) only if anything was changed by transifex (any translation in *.po or *.json). `a)` (I just double-checked I do `git status -s` without any path. So it should be for all changes by transifex.) And I want to have all resulting changes (generated) in this commit.

> OTOH, json changes logically go together with po changes, ie. they are changes made on tx. So I think in the rare case when we get json changes, but no po changes, we should still do the updates and a "translation commit".
> 
> And regardless of the first commit, due to the extremely unlikely case when json files are edited on tx between the 2 pulls, the updates should still always be run the second time too (with `git status --short -- x.json` checks if you want to spare some cycles, see [below](https://github.com/widelands/widelands/pull/6064#discussion_r1300743261)) for consistency between the json inputs and the generated files.

I want to have unexpected changes in the "translation commit" (the 2nd) `d)` 
The update_* scripts for this commit are run if they have not yet been run for "translation commit".
And they were run again when new pot files were pushed to transifex . Now this still is done, but only when really something has change. (Well, obviously something has changed. Else there would have been no push to transifex. I probably should undo this change.)

reference a) to e) reference [my comment above](#issuecomment-1685073340)

utils/merge_and_push_translations.sh
```diff
@@ -54,89 +54,141 @@ fi
 
 echo "Working tree is clean, continuing"
 
-# Print all commands.
-set -x
+update_authors() {
```
tothxa, line 61:
```suggestion
update_authors() {
  if [ -z "$(git status -s -- data/i18n/locales/)" ]; then
    return  # no changes
  fi
```

BTW the paths in the comments in `update_authors.py` should be updated…

utils/merge_and_push_translations.sh
```diff
@@ -54,89 +54,141 @@ fi
...
+    exit 1
+  fi
+  # Fix formatting for Lua
+  python3 utils/fix_formatting.py --lua --dir data/i18n
+  python3 utils/fix_formatting.py --lua --dir data/txts
+}
 
-# Update source catalogs
-python3 utils/buildcat.py
+update_appdata() {
```
tothxa, line 74:
```suggestion
update_appdata() {
  if [ -z "$(git status -s -- xdg/translations/)" ]; then
    return  # no changes
  fi
```

utils/merge_and_push_translations.sh
```diff
@@ -54,89 +54,141 @@ fi
...
+  # Update appdata
+  if python3 utils/update_appdata.py; then
+    echo "Updated appdata"
+  else
+    echo "Failed updating appdata"
+    exit 1
+  fi
+}
+
+update_statistics() {
```
tothxa, line 84: Not sure… probably on any changes in `po`?

utils/merge_and_push_translations.sh
```diff
@@ -45,7 +45,7 @@ git checkout master
 git pull "$push_target" master
```
tothxa, line 45:
```suggestion
git pull "$push_target" master
had_commit=''
```

utils/merge_and_push_translations.sh
```diff
@@ -54,89 +54,141 @@ fi
...
-# Update statistics
-python3 utils/update_translation_stats.py
-if [ $? -eq 0 ] ; then
-  echo "Updated translation stats";
-else
-  echo "Failed to update translation stats";
-  exit 1;
+  if [ -z "$run_update_scripts" ] && [ -n "$(git status -s)" ]; then
+    run_update_scripts=for_pot # run again, pot files have changed
+  fi
```
tothxa:
```suggestion
  if [ -n "$(git status -s)" ]; then
    update_authors # in case something has changed unexpectedly
    update_appdata # in case something has changed unexpectedly
    update_statistics
  fi
```

So you can remove the `run_update_scripts` variable.

utils/merge_and_push_translations.sh
tothxa: Yes.

utils/merge_and_push_translations.sh
tothxa: > > (You should have included the json changes too in the descriptions. :)
> 
> In which description? 🤔 probably commit.

No, here in the comments for the cases.

utils/merge_and_push_translations.sh
SimonHeimberg: The source files for update_authors.py are data/i18n/locales AND data/txts/developers.json

The 2nd is not changed by transifex, so your test is fine for the 1st ("translation") commit.
But the script has to run (before the 2nd) commit in case a new developer was added (last time in 13632b60). It takes less than a second to run it, so this could run for the 2nd commit anyway. (see `d)` in [my comment](#issuecomment-1685073340))

--
Ah, data/ is missing in the paths in the comments in `update_authors.py`, I can update this soon.

utils/merge_and_push_translations.sh
SimonHeimberg: This would be fine for the 1st ("translation") commit.
But it does not catch the change on release, like in 5665009f6. And other potential manual changes to the template files. (see `d)` in [my comment](#issuecomment-1685073340))

utils/merge_and_push_translations.sh
SimonHeimberg: This would probably work. At least I can find no real corner case.
(If somebody edits a po file outside transifex, it is overwritten by transifex pull.)
(If somebody edits update_translation_statistics.py, the changes to the statistics file are hopefully committed as well. If not they will come on a later 

I want to make sure all is in sync after the script has run. But here it can not really get out of sync.

utils/merge_and_push_translations.sh
SimonHeimberg: Looks like you prefer to have the variable initialised.
This is not necessary as we did not run `set -u` (or `set -o nounset`). But I like the style.

utils/merge_and_push_translations.sh
SimonHeimberg: The variable run_update_scripts is for case `d)` of [my comment](#issuecomment-1685073340). 
I want the update_* scripts to run once at least. As this has happened before this pr. So I run them before the second commit if they were not run for the 1st commit, even without change of a po file.
This is necessary for update_authors and update_appdata, for update_statistics it is just for 

To read the variable before setting it is unnecessary (and probably less readable). This just saves a call to git, but git will run fast.

An other option to achive this without a variable is to run update_authors and update_appdata in the beginning. If this changes any files, a 0th commit is created ("updated appdata and authors").
This has the advantage that such changes do not end up either in "translation commit" or "catalog commit" depending on what else changes.
And it probably is more readable.
(Normally, only one or two of the three commits will be triggered.)
(I think update_appdata and update_authors do not need to run for the "translation commit" with this idea.)

utils/merge_and_push_translations.sh
tothxa: Actually I'm OK without these checks (if potential sources are scattered, then it's even better to not do them – that's part of what I meant by trusting git), I just wanted to show what I meant in earlier comments.

utils/merge_and_push_translations.sh
tothxa: Well, if we want to be sure, then it's best to always run these updates, in both turns. :)

utils/merge_and_push_translations.sh
SimonHeimberg: Now we have to trust the programmer to set has_commit=yes after a commit was done.
We could save the the hash of HEAD in the beginning and check in the end if it is still the same. This should be more readable than my previous git command.
But there should not be many new commits created, and the programmers are trustworthy.

utils/merge_and_push_translations.sh
tothxa: > The variable run_update_scripts is for case `d)` of [my comment](#issuecomment-1685073340). I want the update_* scripts to run once at least.

OK, then you can remove the `if` in line 171.

> An other option to achive this without a variable is to run update_authors and update_appdata in the beginning. If this changes any files, a 0th commit is created ("updated appdata and authors").

Oh, I see what you mean. Well, for purity's sake we could do that, and they don't change often, so they certainly won't spam the commit history. (and I believe appdata is usually updated right when the stub is changed – mostly on releases)

But still -1 from me, because it's getting unreadable. They can go in the "translation commit", so you don't need to put comments explaining why you run them 3 times.

> This has the advantage that such changes do not end up either in "translation commit" or "catalog commit" depending on what else changes.

I think they can only get updated in the "catalog" commit if there are edits on tx between the 2 pulls, so that's not really an issue. *edit: OK, I get it now. Sorry, I'm not used to the situation when there are no changes in po files. OK, I still vote readability: run them unconditionally in both turns.*

> And it probably is more readable.

Yes.

tothxa: OK, so if we run them unconditionally below, then this block should be deleted completely.

utils/merge_and_push_translations.sh
tothxa: > But there should not be many new commits created, and the programmers are trustworthy.

This is only inside this script. Not too hard to verify correctness.

utils/merge_and_push_translations.sh
```diff
@@ -54,89 +55,141 @@ fi
...
+# Undo one-liner diffs of pure timestamps with no other content
+# for fetched translations (*.po) (in case something plays tricks, see #5937)
+undo_oneliner_diffs
+
+run_update_scripts=at_least_once # run them at least once
+if [ -n "$(git status -s)" ]; then
+  update_authors
+  update_appdata
+  update_statistics
+  run_update_scripts= # they have run now
```
tothxa:
```suggestion
# These may have changes either from git or from transifex
update_authors
update_appdata

if [ -n "$(git status -s)" ]; then
  update_statistics
```

utils/merge_and_push_translations.sh
SimonHeimberg: > > An other option to achive this without a variable is to run update_authors and update_appdata in the beginning. If this changes any files, a 0th commit is created ("updated appdata and authors").
> 
> Oh, I see what you mean. Well, for purity's sake we could do that, and they don't change often, so they certainly won't spam the commit history. (and I believe appdata is usually updated right when the stub is changed – mostly on releases)
> 
Right, it was on last release in https://github.com/widelands/widelands/commit/5665009f6b2606ea38fdac58afb3a745a7145848, probably utils/update_appdata.py was run manually. :+1:

utils/merge_and_push_translations.sh
```diff
@@ -54,89 +55,141 @@ fi
...
-python3 utils/fix_formatting.py --lua --dir data/i18n
-python3 utils/fix_formatting.py --lua --dir data/txts
-
-# Undo one-liner diffs of pure timestamps with no other content
-undo_oneliner_diffs
+if [ -n "$run_update_scripts" ]; then # if not run yet or pot files have changed
+  update_authors # in case something has changed unexpectedly
+  update_appdata # in case something has changed unexpectedly
+  update_statistics
+fi
```
tothxa:
```suggestion
update_authors # in case something has changed unexpectedly
update_appdata # in case something has changed unexpectedly
update_statistics
```

utils/merge_and_push_translations.sh
```diff
@@ -54,89 +55,141 @@ fi
...
-# Update statistics
-python3 utils/update_translation_stats.py
-if [ $? -eq 0 ] ; then
-  echo "Updated translation stats";
-else
-  echo "Failed to update translation stats";
-  exit 1;
+  if [ -z "$run_update_scripts" ] && [ -n "$(git status -s)" ]; then
+    run_update_scripts=for_pot # run again, pot files have changed
+  fi
```
tothxa:
```suggestion
```

tothxa:
Thank you. This is now almost ready, and I even did some limited testing.

Some last suggestions that would have been harder inline in https://github.com/SimonHeimberg/widelands/pull/1

SimonHeimberg:
> Some last suggestions that would have been harder inline in [SimonHeimberg#1](https://github.com/SimonHeimberg/widelands/pull/1)

Thank you for the link. I would not have seen it else.

Took me some minutes to see that the two if cases for the catalog can be merged. But of course.

SimonHeimberg:
Updated the pr.

Now (with 6e89aa9) functionality is only a bit different than before:
```diff
 a) When translations are edited, only commit "translation" is created.
  b) When translation sources change, only commit "catalog" is created.
  c) When both change, both commits are created.
-d) When a file was changed (on release, ...) and no translation changes, the commit "catalog" is created.
+d) When a file was changed (in git by ...), the commit "translation" is created.
 e) No commit triggers the other one.
```
The functionality and the code are more obvious now.

tothxa (approved):

Noordfrees (approved):

SimonHeimberg:
The first commits by the changed script are done:

Updated translations catalogs and statistics. b8630e53b2f424cd284042c16381302be3424625
Fetched translations and updated data. 991ddcecd12755959e825dae91316483a105c2ea

Thank you for the team work

tothxa:
> The first commits by the changed script are done:

It made me wonder though whether we got the right granularity. I'd much prefer having line reference updates separate from actual translatable string changes (which I wouldn't mind that much going with translation changes to keep it max 2 commits), but I don't think any tool could help us with that, and making our own would be too error prone. So I guess the current solution is the best we can do.

SimonHeimberg:
> > The first commits by the changed script are done:
> 
> It made me wonder though whether we got the right granularity. I'd much prefer having line reference updates separate from actual translatable string changes (which I wouldn't mind that much going with translation changes to keep it max 2 commits), but I don't think any tool could help us with that, and making our own would be too error prone. So I guess the current solution is the best we can do.

I do not get how you would like to have separation.

"Fetched translation ..." 991ddcecd12755959e825dae91316483a105c2ea
- contains changed **translation strings** (including newly translated) only

 "... catalogs ...." b8630e53b2f424cd284042c16381302be3424625
- contains translations strings only only **removed** or *added* (untranslated)
- contains **Line references** change
- (The only possibility a translation change ends up here is [case 8)](#issuecomment-1675977588), editing on transifex while the tool is running. Not very likely. We could try to shorten the time between the two `tx pull` calls to make it even less likely. I have seen no option to tell tx to pause including translation changes.)

Both commits
- contain the files which change because of its changes to *.po or *.pot.

Where would you like to have different granularity?

tothxa:
I said I knew this PR did what we could easily do to improve the situation. :)

The problem is that catalog updates are messy in themselves with the line reference updates, so ideally the split should be between line reference updates and all other changes. The problem is that it's hard to filter out line reference changes, because they may belong to new or deleted translatable strings. But if we can do it, then I think the best approach would be to

1. update and upload catalogs
2. download translations
3. separate and commit line reference changes (.pot and .po together)
4. update data and commit everything

Now I see this would also need a lot of copying files around, to keep the final files during step 3., and restore them for step 4.

Let's hope most catalog updates will be lighter, so this won't be an issue. This one was a double update (during a busy period), because the previous one failed because of #6119.

SimonHeimberg:
I understand that you are pleased with this pr. But I would still like to understand.

I still do not get which part of the catalog update you would like to see.

Let's take some parts of [the diff of po/widelands/bg.po](https://github.com/widelands/widelands/commit/b8630e53b2f424cd284042c16381302be3424625#diff-6551a9be471e93bd58f438bbd9810cb10d262832b7c6ae865a6d56200ea633d1) as example.

* Removed msgid (with empty msgstr) [L9111-R9133](https://github.com/widelands/widelands/commit/b8630e53b2f424cd284042c16381302be3424625#diff-6551a9be471e93bd58f438bbd9810cb10d262832b7c6ae865a6d56200ea633d1L9111-R9133)
```diff
 
-#. * TRANSLATORS: Tooltip for Script Console in the game's main menu
-#: ../../src/wui/interactive_gamebase.cc:139
-msgctxt "hotkey"
-msgid "Ctrl+Shift+Space"
-msgstr ""
-
 #. * TRANSLATORS: Tooltip for Sound Options in the game's main menu
```
  uninteresting for me

* Removed msgid (with existing msgstr) [L9111-L9118](https://github.com/widelands/widelands/commit/b8630e53b2f424cd284042c16381302be3424625#diff-6551a9be471e93bd58f438bbd9810cb10d262832b7c6ae865a6d56200ea633d1L9111-L9118)
```diff
 
-#. * TRANSLATORS: Tooltip for Script Console in the game's main menu
-#: ../../src/wui/interactive_gamebase.cc:139
-msgctxt "hotkey"
-msgid "Ctrl+Shift+Space"
-msgstr "Ctrl+Shift+Space"
-
 #. * TRANSLATORS: Tooltip for Sound Options in the game's main menu
```
maybe interesting, but I doubt

* Added msgid (with empty msgstr) [L3554-L3555](https://github.com/widelands/widelands/commit/b8630e53b2f424cd284042c16381302be3424625#diff-6551a9be471e93bd58f438bbd9810cb10d262832b7c6ae865a6d56200ea633d1L3554-L3555)
```diff
 #, c-format
+msgid "Client %s has cheating features enabled!"
+msgstr ""
+
+#: ../../src/network/network_gaming_messages.cc:78
+#, c-format
 msgid "Client %s is cheating!"
```
maybe interesting, but I can filter on transifex for new untranslated strings. I can see from the [diff of translation_stats.conf](https://github.com/widelands/widelands/commit/b8630e53b2f424cd284042c16381302be3424625#diff-3290789172ef510924cb48731d577cf45f0fc49fe21147b941b34c132a50a375L8) if there are translations missing now

* line reference change [L7379-L7382](https://github.com/widelands/widelands/commit/b8630e53b2f424cd284042c16381302be3424625#diff-6551a9be471e93bd58f438bbd9810cb10d262832b7c6ae865a6d56200ea633d1L7379-L7382)
```diff
 
-#: ../../src/wlapplication_options.cc:713 ../../src/wui/game_chat_menu.cc:55
-#: ../../src/wui/interactive_gamebase.cc:410
+#: ../../src/wlapplication_options.cc:714 ../../src/wui/game_chat_menu.cc:68
+#: ../../src/wui/interactive_gamebase.cc:408
 msgid "Chat"
```
uninteresting for you and me

* part of the message changed (from other commit 58044d94554c299d5c991e and other file po/scenario_atl01.wmf/pt_BR.po)
```diff
 "fish breeder makes sure that a school of fish does not go extinct by "
-"breeding more. Make sure that there are always fish left, otherwise the fish"
-" breeder won’t be able to breed new ones. The smoking happens in a smokery, "
-"you will need at least two of those."
-msgstr "As outras duas importantes fontes alimentares são peixe e carne defumada. Carne crua é fornecida pelos caçadores. Os pescadores pescam peixes no mar enquanto o piscicultor garante que
 os peixes não serão extintos. Garanta que sempre haja peixe sobrando, do contrário o piscicultor não po- derá procriá-los. A defumação ocorre na defumaria, você precisará de ao menos duas del
as."
+"breeding more. The smoking happens in a smokery, you will need at least two "
+"of those."
+msgstr ""
 
```
interesting because I see I can reuse part of the old message

&nbsp;

Did I miss any diff (maybe not in the example commit) which could be interesting?

PR #5861, opened 2023-04-20
Update performance helptext for buildings producing more wares

gpalino:
Continuation of https://github.com/widelands/widelands/pull/5778
My goal is now to add/update help texts, so that it is clear how fast are wares produced under two common conditions:
- economy needs all wares produced by building
- economy need only one ware produced by building

Second case is usual for tool buildings and unevently generated wares, like barbarian inns, crystal mine or honey bread bakery.

Suggestion from [tothxa](https://github.com/tothxa) for full cycle:
"If all wares are needed by the economy and it is fully supplied, then this building can produce on average one <ware1> in <time1>, then one <ware2> in <time2>, … and finally one <ware[n]> in <time[n]>, for a total of <total_time>."
I'm not sure how you calculate with the sleep time between cycles, but in this case it should probably not be spread over the wares, but included directly after ware[n], like "then it sleeps for <sleeptime>, for a total…"

Suggestion from [hessenfarmer](https://github.com/hessenfarmer):
"When this building is fully supplied, production of <ware_a> takes <time_1> and <ware_b> takes <time_b>. If all wares are needed this results in a total time of <time_tot> for <wares_a> and <wares_b>."

My suggestion:
"When this building is fully supplied and all wares are needed by the economy, production of <ware_a> takes <time_a> and <ware_b> takes <time_b> on average. If only one ware is needed by the economy, production of <ware_a> takes <time_a1> and <ware_b> takes <time_b1> on average."

I don't think specifying total time or sleep times is helpful, the time of ware production when economy needs only one ware can't be calculated precisely from full cycle parameters.

Example for atlantean crystal mine:
Total time = 387.4 sec
Basic production time for 1 granite = 15+3.6 = 18.6 sec
Basic production time for 1 quartz = 10+3.6 = 13.6 sec
Basic production time for 1 diamond = 25+3.6 = 28.6 sec
All sleep times = 38.4+30+40 = 108.4 sec
Full cycle produces 9 granite, 4 quartz and 2 diamonds

When e.g. only diamonds are needed:
2 diamonds take 57.2 sec
sleep time is just 40 sec
but also one granite is produced, which delays diamonds production by another 18.6 sec
So if we would like to reflect everything what is happening, it would lead to complex help text.

When I am considering what I expect from performance help text, I can imagine these 3 cases:
- getting comparison of the production speeds between buildings in production chain, e.g.:
  If I have 2 iron mines, how many smelting works should I build to have full speed of iron production?
[...]

Issue #5268, opened 2022-02-19
Missing lore and production info in Empire/Atlantean building help


Wuzzy2:
**Describe the bug**
If you play as the Empire or Atlanteans and open the building help for any of the building, there is a section called "Lore" and "Production" that is currently empty.
For the lore it shows "Text needed" and for the "Production" section is shows "Calculation needed".

Other tribes seems t

**To Reproduce**
1. Play as Empire or Atlanteans
2. Click on any building to open the building help
3. Scroll through the window to see if my claims above are true

**Expected behavior**
The sections "Lore" and "Production" have actual contents instead of dummy texts.

**Version:**
 - Widelands Version: 1.0

#### Lore
 - [ ] Amazons lore: @alasiax had some very nice and imaginative work on this (#5037), but unfortunately she's been inactive since then
 - [x] Atlantean lore (#5271)
 - [x] Empire lore (#5285)
 - [ ] Frisian lore for rest of buildings

#### Production times
 - [x] Sites needing experience (#5768)
 - [x] Fix sleep times (#5778)
 - [x] Sites producing multiple wares (#5861)
 - [x] Small sites comparable between tribes (#6068)
 - [x] Medium sites comparable between tribes
 - [x] Big sites comparable between tribes
 - [x] Farm-like buildings
 - [x] Rest of amazons buildings
 - [x] Rest of atlanteans buildings
 - [x] Rest of barbarians buildings
 - [x] Rest of empire buildings
 - [x] Rest of frisians buildings

klaus-halfmann:
Well this is not really a Bug but some missing texts, to be added . @Noordfrees  please tell us where we can add such texts, I would go ahead and invente some Lore and such.

Lore is "just" a bit of imagination describing the building in a funny way.

Production is something different, this should be done mathematically correct. One could derive some estimations by watching regular gameplay, which would be ok for the most players.

Floy288:
Duplicate of #

Noordfrees:
> Well this is not really a Bug but some missing texts, to be added . @Noordfrees please tell us where we can add such texts, I would go ahead and invente some Lore and such.

Lore texts are defined in the building's entry in `data/tribes/initialization/<tribename>/units.lua` by adding helptexts with the keys `lore` and `lore_author`, like this:

https://github.com/widelands/widelands/blob/eda47cde863b64577473b181950f5900bc4dee92/data/tribes/initialization/barbarians/units.lua#L1082-L1094

[...]

Issue #3437, opened 2019-09-09
Calculate durations when reading production programs


bunnybot:
We should have production programs' durations available, so that we can use them in the encyclopedia and the AI.
For worker calculations, see wuatek.no-ip.org/~rak/widelands/docs/ForestersWoodcutters/ForestersAndWoodcutters1.3.pdf

Forum discussion:

https://wl.widelands.org/forum/topic/4523/

------------------------------------
Imported from Launchpad using lp2gh.

 * date created: 2019-05-20T07:40:25Z
 * owner: gunchleoc
 * the launchpad url was https://bugs.launchpad.net/bugs/1829714

bunnybot:
(by gunchleoc)

[ForestersAndWoodcutters1.3.pdf](https://github.com/widelands/widelands/files/3615163/ForestersAndWoodcutters1.3.pdf)

PR #6068, opened 2023-08-11
Update performance helptext for comparable buildings between tribes, small sites

gpalino:
**Type of change**
Fix and add missing performance help texts in lua scripts.

Now I want to focus on common production buildings, found in multiple tribes.

Proposed pattern for "outdoor" production sites:
```
performance = {
   pgettext("aaa_building", "The worker needs %s to do bbb, not counting the time he needs to reach the destination and go home again."):bformat(format_seconds(ccc)),
   pgettext("aaa_building", "Afterwards he rests in the building for %s."):bformat(format_seconds(ddd))
}
```


gpalino:
Few questions:

### Strange delay for wells
When checking timing for atlantean well production, I measured time differences between receiving 2 water buckets at flag, for depleted water resource case:
- 2 successful water pumpings after each other: 44 secons (as expected: 20 + 20 + 3.6 sec)
- 1 failed water pumping: 94 seconds (my expectation: 43.6 + 40)
- 2 failed water pumpings: 144 seconds (my expectation: 43.6 + 2 * 40)

Is this additional 10 seconds delay caused by https://github.com/widelands/widelands/blob/9024dc81321f6f23d225a4b693a4efa4a8776617/src/logic/map_objects/tribes/productionsite.cc#L1123 , or anything else?

### Timing of PR finalization
I expect I can finish this PR completely in more than 1 month. Not sure when you plan 1.2 release. Is it better for you (and translators) to split work into more smaller PRs?

### First pgettext parameter
I've noticed that sometimes pgettext has "building" instead of "tribe_building", e.g. https://github.com/widelands/widelands/blob/9024dc81321f6f23d225a4b693a4efa4a8776617/data/tribes/initialization/atlanteans/units.lua#L1571
Is this expected for repetitive strings (e.g. the same string of second parameter for multiple tribes)?

### Rounding
How much should I care for updating of timings, to have result in whole seconds? E.g. this
https://github.com/widelands/widelands/blob/9024dc81321f6f23d225a4b693a4efa4a8776617/data/tribes/buildings/productionsites/barbarians/quarry/init.lua#L50
can be left as is, or updated to 18 sec.

tothxa:
> "The worker needs %s to do bbb, not counting the time he needs to reach the destination and go home again.")

I think this should be shortened, but I have no alternate suggestion yet.

In addition, I suggest we come up with some way to also give an estimation for the walking time, based on the walking speed (AFAIK constant for all workers and animals) and the work area radius. Maybe approximate minimum and maximum on level ground without obstacles, instead of a mean.

> ### Timing of PR finalization
> 
> I expect I can finish this PR completely in more than 1 month. Not sure when you plan 1.2 release. Is it better for you (and translators) to split work into more smaller PRs?

I don't think it's likely we can start the freeze in a month, but if you can only work on it intermittently then probably smaller PRs are better.

BTW you should mention #5268 in these PRs. After all they will eventually solve a major part of it.

And of course thank you very much for doing all this tedious work! :)

gpalino:
> > "The worker needs %s to do bbb, not counting the time he needs to reach the destination and go home again.")
> 
> I think this should be shortened, but I have no alternate suggestion yet.
> 
> In addition, I suggest we come up with some way to also give an estimation for the walking time, based on the walking speed (AFAIK constant for all workers and animals) and the work area radius. Maybe approximate minimum and maximum on level ground without obstacles, instead of a mean.

Good idea with minimum and maximum, because mean could be complicated to calculate. If we have walking time, then we can just show summary time, like for indoor buildings.

Example for atlantean quarry (radius 6):
q2
stone-cutting duration: 18.5 sec
sleeping duration: 18.5 sec
walking duration in one direction: 3.6 ... 16.2 sec (if I understand walking logic correctly)
shortest case: 2 * 3.6 + 18.5 + 18.5 = 44.2 sec
longest case: 2 * 16.2 + 18.5 + 18.5 = 69.4 sec

My proposal for helptext:
```
performance = {
   pgettext("atlanteans_building", "This building can produce one granite between %1$s and %2$s."):bformat(format_seconds(44), format_seconds(69)),
   pgettext("atlanteans_building", "In case of slopes or obstacles in the way, the time is increased.")
}
```

Then there are more complicated buildings, like farms. Here I propose long-term in-game simulation, to get estimated duration in ideal case (flat and free terrain).

tothxa:
> Then there are more complicated buildings, like farms. Here I propose long-term in-game simulation, to get estimated duration in ideal case (flat and free terrain).

Well, AFAICS the farms always call "plant" then "harvest", so unless there are too many farms close to each other, the farmer always has a spot for planting, but may have to wait for a harvest in the beginning and if it has too few fields. So we can give the (min and max) time for the steady state (1 plant and 1 harvest producing 1 crop), and a sum of the growing times of the plant as reference. Maybe even an estimate for the number of required fields for 100% productivity.

gpalino:
I am finished with common small buildings, but I have still 3 points not clear for me
Strange delay for wells and First pgettext parameter - see my first comment
### Calculation of scout performance
I was measuring time between leaving atlantean scout house and entering scout house again in flat terrain, the time was between 100 and 151 seconds.
When using formula for other outdoor buildings, max time should be 2 * 18 * 1.8 + 75 = 140 seconds.
`
  programs = {
      scout = {
         "scout=radius:15 duration:1m15s",
         "return"
      ...
`
So probably I don't understand the algorithm of scout walking. Could you help me with this?

tothxa:
> ### Calculation of scout performance
> 
> I was measuring time between leaving atlantean scout house and entering scout house again in flat terrain, the time was between 100 and 151 seconds. When using formula for other outdoor buildings, max time should be 2 * 18 * 1.8 + 75 = 140 seconds. `programs = { scout = { "scout=radius:15 duration:1m15s", "return" ...` So probably I don't understand the algorithm of scout walking. Could you help me with this?

Just a guess, but by a quick reading of the scouting functions in `logic/map_objects/tribes/worker.cc`, I can't see any limit for where the scout can go, so he/she may end up farther than the starting position radius at the end of the scouting time. If that's so, then the worst case time is (2 * time_to_worst_spot_within_radius) + (2 * duration), though this is extremely unlikely, as it would mean he/she walked in a straight line away from his/her house from the worst possible spot.

---

I had doubts about hunters too, as they often have to chase the game, but having watched and measured a Barbarian hunter, I think this is negligible, as he often ended up closer to his hut by the end of the chase. For other tribes, the limiting factor is the lack of nearby game anyway.

gpalino:
> Just a guess, but by a quick reading of the scouting functions in `logic/map_objects/tribes/worker.cc`, I can't see any limit for where the scout can go, so he/she may end up farther than the starting position radius at the end of the scouting time. If that's so, then the worst case time is (2 * time_to_worst_spot_within_radius) + (2 * duration), though this is extremely unlikely, as it would mean he/she walked in a straight line away from his/her house from the worst possible spot.

Thanks, I added also scout houses' performance in latest commit.

data/tribes/workers/amazons/hunter_gatherer/init.lua
```diff
@@ -16,6 +16,7 @@ wl.Descriptions():new_worker_type {
 
    programs = {
       hunt = {
+         -- time: 1 sec
```
tothxa: I think you should document the walking time calculation here. After all, its parameters are all here, and it's part of the actual time spent within the worker program. Of course it has the drawback that then you need to take 2 values from here for min and max, so there are still 2 calculations in the building too.

data/tribes/buildings/productionsites/amazons/hunter_gatherers_hut/init.lua
```diff
@@ -47,6 +47,11 @@ wl.Descriptions():new_productionsite_type {
          -- TRANSLATORS: Completed/Skipped/Did not start hunting because ...
          descname = _("gathering meat"),
          actions = {
+            -- time of worker: 1 sec
+            -- radius: 14
+            -- steps from building to animal: 2...17
+            -- min. time: 2 * 2 * 1.8 + 1 + 34 = 42.2 sec
+            -- max. time: 2 * 17 * 1.8 + 1 + 34 = 96.2 sec
```
tothxa: So in the buildings it could be:
```
-- time of worker: <min_w>...<max_w>
-- min time: <min_w> + <sleep, etc.> = <min>
-- max time: <max_w> + <sleep, etc.> = <max>
```

or

```
-- time of worker: <min_w>...<max_w>
-- total time: (<min_w>...<max_w>) + <sleep, etc.> = <min>...<max>
```

data/tribes/buildings/productionsites/atlanteans/well/init.lua
```diff
@@ -43,6 +43,9 @@ wl.Descriptions():new_productionsite_type {
          -- TRANSLATORS: Completed/Skipped/Did not start working because ...
          descname = _("working"),
          actions = {
+            -- time total: 20 + 20 + 3.6 = 43.6 sec
```
tothxa, line 46: Why no rounding here? :)

data/tribes/workers/amazons/hunter_gatherer/init.lua
gpalino: I agree, I'll rework these calculation comments for all outdoor buildings.

data/tribes/buildings/productionsites/atlanteans/well/init.lua
gpalino: If there are multiple output times (e.g. here 43.6 and 70.523 sec), I tend to not round it at all.
For wells, the second time is more interesting IMO, because wells will sooner or later degrade to slower speed.
And in the formula, number 10 is still questionable, so I would personally round it only after this is resolved.

hessenfarmer:
sorry for not being able to comment for a while.   
I think calculating walk distances even for max and min is dependent on surrounding immovables. So every value we can provide else then pausing time might be wrong. So providing a (potentially not exact) average might not be more false then anay other value but could help the player more. So my suggestion would be to calculate the average in an ideal scenario without any blocking for the buildings where the worker leaves the building.

tothxa:
> sorry for not being able to comment for a while. I think calculating walk distances even for max and min is dependent on surrounding immovables. So every value we can provide else then pausing time might be wrong. So providing a (potentially not exact) average might not be more false then anay other value but could help the player more.

The problem is that there are at least 2 ways in which you could calculate the "average": By simply averaging the current min and max values, or by averaging over all possible fields (this will give more weight to the farther fields (except for radius 1), because there are more of them). But the actual times in a well organised buildup can be actually less even than the simple (min/max) "average", if only the closest possible fields are left free for the given building. Then of course you'd still have to add the obstacles-and-slopes disclaimer.

IMO, giving the min and max values emphasises more to the players that they can affect the performance of the building by careful planning.

gpalino:
I have one proposal for the help text - to include average time in the beginning, and min+max in brackets, as additional info. E.g. for atlantean quarry:
`This building can produce one granite in 57 seconds on average (between 44 seconds and 1 minute and 9 seconds). In case of slopes or obstacles in the way, the time is increased.`
I personally prefer this, but give your opinion, which text option you prefer (min+max only, avg only or both).

I've included average calculation to the scripts in latest commit, so that we have all numbers visible here. I've used following logic for average formula decision:

### walk=coords
This kind of walking is choosing random positions inside hexagonal area. I used formulas from this document, chapter 2:
https://github.com/widelands/widelands/issues/3437#issuecomment-529550739
for radius > 7: linear model

### walk=object
Worker is choosing nearest positions, not random. There seems to be min+max average more precise.

### exception 1: water workers
Fishers, fish breeders and amazon carriers are not moving inside hexagonal area, but are choosing random positions the shore line only.
I observed average shore line distances in 9 cases from the games, where fisher buildings were in good positions near water.
Shortest distance was 4.571, longest distance 5.667 - so both closer to min+max average (6), which I've chosen for this case.
Screenshots of these two cases:
lesserRing3
crater3


### exception 2: hunters without gamekeepers
For 3 tribes without gamekeeper, the hunter will eventually hunt every accessible animal, so even if walk=object, the overall hunting positions will be probably more close to random distances inside hexagonal area, so I've chosen mathematical mean formula.

### scouts
no idea what is mathematically more correct, I used min+max average.

tothxa:
> give your opinion, which text option you prefer (min+max only, avg only or both).

Just for the record: I still vote min-max only.

```
can produce one <ware> between <min> seconds and <max> seconds, depending on how far <he/she> has <needed resource>. In case of slopes or obstacles in the way, the time is increased.
```

data/tribes/initialization/frisians/units.lua
```diff
@@ -1588,8 +1598,12 @@ wl.Descriptions():new_tribe {
             purpose = pgettext("building", "Plants trees in the surrounding area."),
             -- TRANSLATORS: Note helptext for a Frisian production site: Forester's House
             note = pgettext("frisians_building", "The forester’s house needs free space within the work area to plant the trees."),
-            -- TRANSLATORS: Performance helptext for a Frisian production site: Forester's House
-            performance = pgettext("frisians_building", "The forester pauses %s before going to work again."):bformat(format_seconds(12))
+            performance = {
+               -- TRANSLATORS: Performance helptext for a Frisian production site: Forester's House, part 1
+               pgettext("frisians_building", "The forester plants one tree between %1$s and %2$s."):bformat(format_seconds(24), format_seconds(46)),
```
tothxa: depending on how far

data/tribes/initialization/frisians/units.lua
```diff
@@ -1742,8 +1770,12 @@ wl.Descriptions():new_tribe {
             no_scouting_building_connected = pgettext("frisians_building", "You need to connect this flag to a scout’s house before you can send a scout here."),
             -- TRANSLATORS: Purpose helptext for a Frisian production site: Scout's House
             purpose = pgettext("building", "Explores unknown territory."),
-            -- TRANSLATORS: Performance helptext for a Frisian production site: Scout's House
-            performance = pgettext("frisians_building", "The scout pauses %s before going to work again."):bformat(format_seconds(30))
+            performance = {
+               -- TRANSLATORS: Performance helptext for a Frisian production site: Scout's House, part 1
+               pgettext("frisians_building", "The scout makes exploration between %1$s and %2$s."):bformat(format_minutes_seconds(1, 52), format_minutes_seconds(4, 5)),
```
tothxa: consumes (or needs) a ration?

data/tribes/workers/amazons/carrier/init.lua
```diff
@@ -13,6 +13,11 @@ wl.Descriptions():new_carrier_type {
 
    programs = {
       fetch_water = {
+         -- first bucket takes 1 sec, second bucket (from Water Gatherer’s Hut) takes 3.6 sec, so 2 buckets take 4.6 sec
```
tothxa: I think the other bucket of water shouldn't be mentioned here.

data/tribes/buildings/productionsites/amazons/water_gatherers_hut/init.lua
```diff
@@ -37,6 +37,10 @@ wl.Descriptions():new_productionsite_type {
          -- TRANSLATORS: Completed/Skipped/Did not start working because ...
          descname = _("working"),
          actions = {
+            -- time of worker: 11.8-44.2 sec, min+max average 28 sec
+            -- min. time total: 11.8 + 20 = 31.8 sec
+            -- max. time total: 44.2 + 20 = 64.2 sec
+            -- avg. time total: 28 + 20 = 48 sec
```
tothxa: I think the 3.6s for the extra bucket of water below should be added here.

data/tribes/buildings/productionsites/amazons/water_gatherers_hut/init.lua
```diff
@@ -37,6 +37,10 @@ wl.Descriptions():new_productionsite_type {
...
             "sleep=duration:20s",
             "callworker=fetch_water",
             -- he carries 2 buckets so we need to create one now
```
tothxa: "she", while we're at it.

data/tribes/workers/amazons/woodcutter/init.lua
```diff
@@ -19,6 +19,11 @@ wl.Descriptions():new_worker_type {
 
    programs = {
       harvest = {
+         -- steps from building to tree: 2-13, min+max average 7.5
+         -- min. worker time: 2 * 2 * 1.8 + 30 + 2 = 39.2 sec
+         -- max. worker time: 2 * 13 * 1.8 + 30 + 2 = 78.8 sec
+         -- avg. worker time: 2 * 7.5 * 1.8 + 30 + 2 = 59 sec
+         -- some of trees delay woodcutting by 1-1.4 sec (aspen, balsa, ironwood, oak, rubber)
```
tothxa: Only aspen and oak of world trees. Plain woodcutters of Amazons can't cut rare trees.

data/tribes/workers/amazons/woodcutter_master/init.lua
```diff
@@ -44,6 +59,11 @@ wl.Descriptions():new_worker_type {
          "return"
       },
       harvest = {
+         -- steps from building to tree: 2-13, min+max average 7.5
+         -- min. worker time: 2 * 2 * 1.8 + 25 + 2 = 34.2 sec
+         -- max. worker time: 2 * 13 * 1.8 + 25 + 2 = 73.8 sec
+         -- avg. worker time: 2 * 7.5 * 1.8 + 25 + + 2 = 54 sec
+         -- some of trees delay woodcutting by 1-1.4 sec (aspen, balsa, ironwood, oak, rubber)
```
tothxa: only aspen and oak here

data/tribes/initialization/amazons/units.lua
```diff
@@ -1311,7 +1323,13 @@ wl.Descriptions():new_tribe {
             -- TRANSLATORS: Purpose helptext for an Amazon production site: Jungle Preserver's Hut
             purpose = pgettext("amazons_building", "Plants trees in the surrounding area."),
             -- TRANSLATORS: Note helptext for an Amazon production site: Jungle Preserver's Hut
-            note = pgettext("amazons_building", "The jungle preserver’s hut needs free space within the work area to plant the trees.")
+            note = pgettext("amazons_building", "The jungle preserver’s hut needs free space within the work area to plant the trees."),
+            performance = {
+               -- TRANSLATORS: Performance helptext for an Amazon production site: Jungle Preserver's Hut, part 1
+               pgettext("amazons_building", "The jungle preserver or jungle master plants one tree between %1$s and %2$s."):bformat(format_seconds(20), format_seconds(45)),
```
tothxa: "depending on how far she has free spots"

data/tribes/initialization/amazons/units.lua
```diff
@@ -1380,8 +1407,12 @@ wl.Descriptions():new_tribe {
             no_scouting_building_connected = pgettext("amazons_building", "You need to connect this flag to a scout’s hut before you can send a scout here."),
             -- TRANSLATORS: Purpose helptext for an Amazon production site: Scout's Hut
             purpose = pgettext("amazons_building", "Explores unknown territory."),
-            -- TRANSLATORS: Performance helptext for an Amazon production site: Scout's Hut
-            performance = pgettext("amazons_building", "The scout pauses %s before going to work again."):bformat(format_seconds(30))
+            performance = {
+               -- TRANSLATORS: Performance helptext for an Amazon production site: Scout's Hut, part 1
+               pgettext("amazons_building", "The scout makes exploration between %1$s and %2$s."):bformat(format_minutes_seconds(1, 52), format_minutes_seconds(4, 5)),
```
tothxa: "consumes a ration"?

tothxa (commented):
I've checked Frisians and Amazons so far, with the below comments.

You did the calculation for the worker time of the master woodcutter, but not the rare tree cutter's hut, and it's also missing from `units.lua`.

data/tribes/initialization/atlanteans/units.lua
```diff
@@ -1509,7 +1521,13 @@ wl.Descriptions():new_tribe {
                                           [[Only after the last piece of green has become bare soil<br>]] ..
                                           [[Then will you find that nature needs to be cared for.’]]),
             -- TRANSLATORS: Note lore_autor for an Atlantean production site: Forester's House
-            lore_author = pgettext("atlanteans_building", "Prophecy of the foresters’ guild")
+            lore_author = pgettext("atlanteans_building", "Prophecy of the foresters’ guild"),
+            performance = {
+               -- TRANSLATORS: Performance helptext for an Atlantean production site: Forester's House, part 1
+               pgettext("atlanteans_building", "The forester plants one tree between %1$s and %2$s."):bformat(format_seconds(23), format_seconds(45)),
```
tothxa: "depending on how far..."

data/tribes/initialization/atlanteans/units.lua
```diff
@@ -1602,7 +1638,13 @@ wl.Descriptions():new_tribe {
             -- TRANSLATORS: Lore helptext for an Atlantean production site:  Scout's House
             lore = pgettext("atlanteans_building", "You do not see everything from your high towers. Sometimes you need someone to find out in person, that’s me!"),
             -- TRANSLATORS: Lore author helptext for an Atlantean production site:  Scout's House
-            lore_author = pgettext("atlanteans_building", "A scout answering a soldier")
+            lore_author = pgettext("atlanteans_building", "A scout answering a soldier"),
+            performance = {
+               -- TRANSLATORS: Performance helptext for an Atlantean production site: Scout's House, part 1
+               pgettext("atlanteans_building", "The scout makes exploration between %1$s and %2$s."):bformat(format_minutes_seconds(1, 52), format_minutes_seconds(4, 5)),
```
tothxa: "consumes/needs one portion of food"?

data/tribes/initialization/barbarians/units.lua
```diff
@@ -1173,8 +1177,12 @@ wl.Descriptions():new_tribe {
             purpose = pgettext("building", "Plants trees in the surrounding area."),
             -- TRANSLATORS: Note helptext for a Barbarian production site: Ranger's Hut
             note = pgettext("barbarians_building", "The ranger’s hut needs free space within the work area to plant the trees."),
-            -- TRANSLATORS: Performance helptext for a Barbarian production site: Ranger's Hut
-            performance = pgettext("barbarians_building", "The ranger needs %s to plant a tree, not counting the time he needs to reach the destination and go home again."):bformat(format_seconds(5))
+            performance = {
+               -- TRANSLATORS: Performance helptext for a Barbarian production site: Ranger's Hut, part 1
+               pgettext("barbarians_building", "The ranger plants one tree between %1$s and %2$s."):bformat(format_seconds(28), format_seconds(50)),
```
tothxa: "depending on how far..."

data/tribes/initialization/barbarians/units.lua
```diff
@@ -1246,7 +1266,13 @@ wl.Descriptions():new_tribe {
             -- TRANSLATORS: Lore author helptext for a Barbarian production site: Scout's Hut
             lore_author = pgettext("barbarians_building", "Chat’Karuth<br>at the oath taking ceremony of the first scout troupe"),
             -- TRANSLATORS: Purpose helptext for a Barbarian production site: Scout's Hut
-            purpose = pgettext("building", "Explores unknown territory.")
+            purpose = pgettext("building", "Explores unknown territory."),
+            performance = {
+               -- TRANSLATORS: Performance helptext for a Barbarian production site: Scout's Hut, part 1
+               pgettext("barbarians_building", "The scout makes exploration between %1$s and %2$s."):bformat(format_minutes_seconds(1, 52), format_minutes_seconds(4, 5)),
```
tothxa: "consumes/needs a ration"?

data/tribes/initialization/empire/units.lua
```diff
@@ -1246,8 +1246,18 @@ wl.Descriptions():new_tribe {
...
-            note = pgettext("empire_building", "The quarry needs rocks to cut within the work area.")
+            note = {
+               -- TRANSLATORS: Note helptext for an Empire production site: Quarry, part 1
+               pgettext("empire_building", "The quarry needs rocks to cut within the work area."),
+               -- TRANSLATORS: Note helptext for an Empire production site: Quarry, part 2
+               pgettext("empire_building", "The stonemason finds 5 blocks of granite and 2 blocks of marble out of 7 rocks.")
+            },
+            performance = {
+               -- TRANSLATORS: Performance helptext for an Empire production site: Quarry, part 1
+               pgettext("empire_building", "This building can produce one granite or one marble between %1$s and %2$s, depending on how far the stonemason has stones."):bformat(format_seconds(42), format_minutes_seconds(1, 7)),
```
tothxa: "one block of stone"?

Shouldn't we give the values for granite and marble separately? (59s-1m34s, 2m28s-3m56s)

data/tribes/initialization/empire/units.lua
```diff
@@ -1289,7 +1305,13 @@ wl.Descriptions():new_tribe {
             -- TRANSLATORS: Purpose helptext for an Empire production site: Forester's House
             purpose = pgettext("building", "Plants trees in the surrounding area."),
             -- TRANSLATORS: Note helptext for an Empire production site: Forester's House
-            note = pgettext("empire_building", "The forester’s house needs free space within the work area to plant the trees.")
+            note = pgettext("empire_building", "The forester’s house needs free space within the work area to plant the trees."),
+            performance = {
+               -- TRANSLATORS: Performance helptext for an Empire production site: Forester's House, part 1
+               pgettext("empire_building", "The forester plants one tree between %1$s and %2$s."):bformat(format_seconds(23), format_seconds(45)),
```
tothxa: "depending on how far..."

data/tribes/initialization/empire/units.lua
```diff
@@ -1354,7 +1388,13 @@ wl.Descriptions():new_tribe {
             -- TRANSLATORS: Lore author helptext for an Empire production site: Well
             lore_author = pgettext("empire_building", "Imperial proverb"),
             -- TRANSLATORS: Purpose helptext for an Empire production site: Well
-            purpose = pgettext("building", "Draws water out of the deep.")
+            purpose = pgettext("building", "Draws water out of the deep."),
+            performance = {
+               -- TRANSLATORS: Performance helptext for an Empire production site: Well, part 1
+               pgettext("empire_building", "The carrier needs %s to get one bucket full of water."):bformat(format_seconds(44)),
+               -- TRANSLATORS: Performance helptext for an Empire production site: Well, part 2
+               pgettext("empire_building", "If the water resource is depleted, the carrier needs %s on average."):bformat(format_seconds(71))
```
tothxa: 1m11s? (all tribes)

data/tribes/initialization/empire/units.lua
```diff
@@ -1374,7 +1414,13 @@ wl.Descriptions():new_tribe {
             lore_author = pgettext("empire_building", "A belated scout’s report"),
             no_scouting_building_connected = pgettext("empire_building", "You need to connect this flag to a scout’s house before you can send a scout here."),
             -- TRANSLATORS: Purpose helptext for an Empire production site: Scout's House
-            purpose = pgettext("building", "Explores unknown territory.")
+            purpose = pgettext("building", "Explores unknown territory."),
+            performance = {
+               -- TRANSLATORS: Performance helptext for an Empire production site: Scout's House, part 1
+               pgettext("empire_building", "The scout makes exploration between %1$s and %2$s."):bformat(format_minutes_seconds(1, 52), format_minutes_seconds(4, 5)),
```
tothxa: "consumes/needs a ration"?

tothxa (commented):
I've checked all now.

Apart from the below comments, if @hessenfarmer is OK with min and max only, then the "averages" should be removed from the program comments.

data/tribes/initialization/amazons/units.lua
hessenfarmer: consumation is already included in a different chapter oif the units help. so we do not need to add this in the performance

data/tribes/initialization/barbarians/units.lua
hessenfarmer: see above consumation is already contained in the help (automatically determined from the lua files)

data/tribes/initialization/empire/units.lua
hessenfarmer: as the ratio is given above in the note I'd vote to use one block of granite or marble and only give the time for one cycle-.

data/tribes/workers/amazons/woodcutter/init.lua
```diff
@@ -19,6 +19,11 @@ wl.Descriptions():new_worker_type {
 
    programs = {
       harvest = {
+         -- steps from building to tree: 2-13, min+max average 7.5
+         -- min. worker time: 2 * 2 * 1.8 + 30 + 2 = 39.2 sec
+         -- max. worker time: 2 * 13 * 1.8 + 30 + 2 = 78.8 sec
+         -- avg. worker time: 2 * 7.5 * 1.8 + 30 + 2 = 59 sec
+         -- some of trees delay woodcutting by 1.4 sec (aspen, oak)
```
hessenfarmer, line 26: this should probably be fixed for all trees to have the same cycle values 
(i.e. wee need to include a delay of 1.4 for all of them (until we may define proper falling animations) and turn all trees into fallen trees), but not in this PR thanks for finding this I'll raise an issue

data/tribes/initialization/amazons/units.lua
```diff
@@ -1293,7 +1293,13 @@ wl.Descriptions():new_tribe {
             -- TRANSLATORS: Purpose helptext for an Amazon production site: Stonecutter's Hut
             purpose = pgettext("amazons_building", "Cuts raw pieces of granite out of rocks in the vicinity."),
             -- TRANSLATORS: Note helptext for an Amazon production site: Stonecutter's Hut
-            note = pgettext("amazons_building", "The stonecutter’s hut needs rocks to cut within the work area.")
+            note = pgettext("amazons_building", "The stonecutter’s hut needs rocks to cut within the work area."),
+            performance = {
+               -- TRANSLATORS: Performance helptext for an Amazon production site: Stonecutter's Hut, part 1
+               pgettext("amazons_building", "This building can produce one granite between %1$s and %2$s, depending on how far the stonecutter has stones."):bformat(format_seconds(52), format_minutes_seconds(1, 17)),
```
hessenfarmer:
```suggestion
               pgettext("amazons_building", "This building can produce one granite between %1$s and %2$s, depending on how far the stonecutter has to walk for stones."):bformat(format_seconds(52), format_minutes_seconds(1, 17)),
```
I think this is more appropriate in english language. so we should correct all instances like that, including the ones where @tothxa requested this part of wording.

hessenfarmer (commented):
I think we can live with the current min max texts. 
I was shortly thinking about just giving the time for walking 1 field but this would not be helpful for the player either. 

I have one wording suggestion inline and responded to some comments of @tothxa.

after my suggestion and his comments have been included I am fine to have this.      

A very big Thank You to @gpalino for doing all these calculations and providing this very valuable piece of work

data/tribes/initialization/frisians/units.lua
tothxa: OK, how about "One exploration trip of the scout takes between…"?

data/tribes/initialization/amazons/units.lua
tothxa: Should've posted this here: OK, how about "One exploration trip of the scout takes between…"?

data/tribes/initialization/empire/units.lua
tothxa: In the previous PR, when a full cycle had more of some items than of others, we always gave the averages separately. This would be harder for users to calculate for themselves than those were.

data/tribes/initialization/amazons/units.lua
tothxa: "depending on how far the stonecutter can find rocks"?

tothxa: Except "has" implies "within her/his work area", while "can find" doesn't…

hessenfarmer: as the difference is made by walking time I felt like emphasizing on walk so we could simply skip the "for resource" part and write "has to walk." in all instances.

tothxa: OTOH simply "has to walk" de-emphasizes that the player can affect this with careful placement of buildings and roads.

But it's indeed the simplest form, so I don't mind it if both of you prefer it.

data/tribes/initialization/empire/units.lua
hessenfarmer: well I believe giving the time of one walk might be easier to understand, but for consistency we can do it like you suggested.

hessenfarmer:
> I've checked all now.
> 
> Apart from the below comments, if @hessenfarmer is OK with min and max only, then the "averages" should be removed from the program comments.

I think we should keep them in the code as it does not do any harm there, @gpalino invested some energy on calcualting them and we knever where and when we might need them.

tothxa:
> > the "averages" should be removed from the program comments.
> 
> I think we should keep them in the code as it does not do any harm there, @gpalino invested some energy on calcualting them and we knever where and when we might need them.

Here's the calculation of actual mathematical averages to show why the min-max averages are misleading (the big building calculation does not consider the increased number of steps to NW positions from the building):
[wl_walking_distances.ods](https://github.com/widelands/widelands/files/13313923/wl_walking_distances.ods)
walking_distances

And a map to help verify the calculation: 
workarea_vs_steps
(immovables show steps, terrains show work areas, standing stone marks the building, liana tree sapling the flag position)

Also another map with a farm's work area:
farm_steps
Snowy terrain shows a 4x4 rhombus with the E-W - NE-SW road system, dark meadow the best possible working positions with a single road. The actual averages are 4.8 steps vs. 2.33 steps (or 2.2 if only the 5 best nodes are left free).

hessenfarmer:
well the clacualted averages are not always max -min/s +min. as described above the averages in the code already reflect the building types like @gpalino identified them above. maybe we could update the farm value but I am against throwing them away.

tothxa:
> maybe we could update the farm value

Farms are not in this PR yet, because this one's only for small buildings. And in any case, I don't propose these averages for inclusion in the performance text. On the contrary, I meant them as an illustration that providing any kind of averages is futile.

> If we want to be 100% correct we can't provide any calculation for these buildings as we do not reflect roads and blocking immovables and mixed terrains and so on. so we would only end up with one field takes 3.6 secs to walk there and back. And even then we would miss slopes.

Yes, that's why we added the disclaimers for obstacles and slopes. Min and max for level terrain are 100% correct, and IMO give a good guideline to the player.

data/tribes/initialization/amazons/units.lua
gpalino: I like the "has to walk." idea. New players can quickly learn where the workers have to walk, simply by watching game. The generic form is also appropriate for workers which go to random positions (like foresters, fishers).

tothxa:
Thank you for the updates. From my side only the Empire granite/marble times are missing.

hessenfarmer (approved):
ok from my side.
I can live with @tothxa suggestion on empire stonecutter so you might consider adding the values as suggested and this is fine to go

data/tribes/buildings/productionsites/atlanteans/well/init.lua
```diff
@@ -43,6 +43,10 @@ wl.Descriptions():new_productionsite_type {
...
+            -- average time if water resource is depleted:
+            --   (65 * (20 + 20 + 3.6) + 35 * (20 + 20 + 10)) / 65 = 70.523 sec
+            -- number 10 is measured experimentally
```
tothxa: I think this is the usual skipped program delay. That's why I hadn't commented on it earlier.

data/tribes/initialization/amazons/units.lua
```diff
@@ -1325,9 +1343,12 @@ wl.Descriptions():new_tribe {
                -- TRANSLATORS: Note helptext for an Amazon production site: Hunter-Gatherer's Hut
                pgettext("amazons_building", "Roads and trees along the shoreline block fishing."),
             },
-            -- TRANSLATORS: Performance helptext for an Amazon production site: Hunter-Gatherer's Hut
-            performance = pgettext("amazons_building", "The hunter-gatherer pauses %s before going to work again."):bformat(format_seconds(34))
-
+            performance = {
+               -- TRANSLATORS: Performance helptext for an Amazon production site: Hunter-Gatherer's Hut, part 1
+               pgettext("amazons_building", "This building can produce one meat or one fish between %1$s and %2$s, depending on how far the hunter-gatherer has to walk."):bformat(format_minutes_seconds(1, 24), format_minutes_seconds(3, 12)),
```
tothxa: Doesn't it produce one of each in that time if it has both? I believe the previous text was correct in any case, regardless of whether she can only hunt, only fish, or both hunt and fish.

gpalino:
> Here's the calculation of actual mathematical averages to show why the min-max averages are misleading (the big building calculation does not consider the increased number of steps to NW positions from the building): [wl_walking_distances.ods](https://github.com/widelands/widelands/files/13313923/wl_walking_distances.ods) 

If you could decrease values of blue cells (B5, C6, D7, ...) by one, this would be even more precise. The positions of shortest road to flag can be excluded. In my libreoffice, such change is damaging org.gnumeric.sumproduct() results.

data/tribes/initialization/amazons/units.lua
```diff
@@ -1355,14 +1376,26 @@ wl.Descriptions():new_tribe {
...
       },
       {
          name = "amazons_rare_tree_cutters_hut",
          helptexts = {
             -- TRANSLATORS: Purpose helptext for an Amazon production site: Rare Tree Cutter's Hut
-            purpose = pgettext("amazons_building", "Harvests rare trees in the surrounding area. Delivers ironwood, balsa and rubber.")
+            purpose = pgettext("amazons_building", "Harvests rare trees in the surrounding area. Delivers ironwood, balsa and rubber."),
+            performance = {
+               -- TRANSLATORS: Performance helptext for an Amazon production site: Rare Tree Cutter's Hut, part 1
+               pgettext("amazons_building", "This building can produce one of each rare wood between %1$s and %2$s, depending on how far the woodcutter has to walk."):bformat(format_minutes_seconds(2, 22), format_minutes_seconds(4, 20)),
```
tothxa: This holds when all 3 are available, but if only 1 or 2, then those are faster. I suggest something like:
"… can produce a rare wood between 47s and 1m27s depending on how far the master woodcutter has to walk. If all kinds of rare tree are available within her work area, then one of each kind is produced between 2m22s and 4m20s."

tothxa (commented):
Thank you for updating the Empire quarry.

Some comments for the new changes.

tothxa:
> > [wl_walking_distances.ods](https://github.com/widelands/widelands/files/13313923/wl_walking_distances.ods)
> 
> If you could decrease values of blue cells (B5, C6, D7, ...) by one, this would be even more precise. The positions of shortest road to flag can be excluded.

Hmmm, indeed, that subtracts one for each distance, which is correct for a single connecting road. However we also have some buildings (foresters, fish breeder, game keeper, berry farm) that can work without any road, and for others with bigger work areas, having only the single connecting road within it is a rather rare arrangement. So we really don't need to go down this path with the calculations. The very consideration just further proves that we shouldn't even try to cover averages.

> In my libreoffice, such change is damaging org.gnumeric.sumproduct() results.

(Oops... Yes, I actually made it with Gnumeric (much prefer that over LOcalc), I wasn't aware of this incompatibility. And it even loads fine (I actually checked what the graph looks like), only errors out when it needs recalculating. Strange design decisions...

Just to make the problem of universal sharing unsolveable, the gnumeric help says that gn. sumproduct is compatible with excel, but not with ODS. So I can use ods.sumproduct, but then that's incompatible with excel. I wonder what happens when LO saves its own sumproduct to xls...)

I've just tried it, and you can fix it by just removing the `org.gnumeric.` prefix. The actual incompatibility is supposed to be with bool values, so doesn't affect this calculation. The formula can be safely copied to the whole column after fixing the first one.

data/tribes/initialization/amazons/units.lua
gpalino: This is good idea - similar style like in https://github.com/widelands/widelands/pull/5861
I used similar wording also for Hunter-Gatherer's Hut, to prevent confusion.

data/tribes/buildings/productionsites/atlanteans/well/init.lua
gpalino: OK, I didn't know this. Now I've removed these additional comments. From my side all is finished in this PR.

Just one question: I've measured time of atlantean gold mine, when there are no gold resources around. Full cycle took about 126.5 sec, measured by food consumption. I expected this time: 40.2 + 3 * (22 + 10) = 136.2 sec. Why are there only two skipped program delays for three mining attempts?

tothxa (approved):
Thank you. It's all fine now.

@hessenfarmer Are you OK with the latest changes too?

data/tribes/buildings/productionsites/atlanteans/well/init.lua
tothxa: I looked up the code, and can confirm it must be the failed/skipped program delay.

So I think I kind of understand now how the delay works in `ProductionSite::program_start()` (and `…::program_end()`). It only prevents the same program to run within 10s after a failure, so if other programs run that take longer than that, then next time the failed program can be tried again immediately.

In the case of the well, the main program itself fails, so it has to wait 10s.

OTOH, in the case of the gold mine, when a subprogram fails, it only gets skipped (according to the documentation of `on failure` in `production_program.cc`), and the main program continues with the next step. So the delay applies after the first and second failed tries, but not after the last one, because of the sleep before the first try in the next run of the main program. AFAICT the main program never returns a failure because of a subprogram, see `ProductionProgram::ActCall::execute()`, so there's no delay for the main program here.

(BTW this also means that it will be rather painful to calculate the average times of empty mines with the same programs (hmmm, `avg_delay = 10*(n_calls - 1)/n_calls`?)… and even more so for the atl. crystal mine and emp. marble mine…)

hessenfarmer (approved):
i am more then happy

hessenfarmer:
finally
@bunnybot merge

PR #6072, opened 2023-08-14
Force all textures to 32bit - probably required for Mesa 23.x

tothxa:
**Type of change**
Portability fix

**Issue(s) closed**
Fixes garbled soldier level icons and playercolour images in flatpak builds with current runtimes

**To reproduce**
*If it's a bugfix:*
1. Build Widelands with the freedesktop.org 23.08beta runtime from flathub-beta (maybe other hosts with Mesa 23.x too?)
2. Start or load a game
3. Open the soldier statistics (or look at a Frisian shipyard that belongs to a player)

**New behavior**
All textures are converted to 32 bits for passing to OpenGL.

**Possible regressions**
Texture display

**Screenshots**
Without this fix:
shot0000

**Additional context**
This bug report is the reason why I suspect Mesa 23.x is at least part of the problem (the other part is current SDL):
 - https://discourse.flathub.org/t/how-to-use-a-specific-mesa-version-in-a-flatpak-manifest/4236
 - https://gitlab.freedesktop.org/mesa/mesa/-/issues/9007

(I first got similar weird colours when I naively converted the SDL surface to RGBA (big endian) instead of ABGR after I identified the 24 vs. 32 bit problem.)

### DIFF
--- a/src/graphic/texture.cc
+++ b/src/graphic/texture.cc
@@ -117,9 +117,11 @@ Texture::Texture(SDL_Surface* surface, bool intensity) : owns_texture_(false) {
 	uint8_t bpp = surface->format->BytesPerPixel;
 
 	if ((surface->format->palette != nullptr) || width() != surface->w || height() != surface->h ||
-	    (bpp != 3 && bpp != 4) || is_bgr_surface(*surface->format)) {
+	    (bpp != 4) || is_bgr_surface(*surface->format)) {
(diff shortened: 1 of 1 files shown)


Noordfrees (approved):
Code looks safe. I didn't observe any problems with this change (though I have Mesa 23.1.4-1 and do not see the original bug)

PR #6079, opened 2023-08-15
Fix LaunchMPG Layouting Segfault

Noordfrees:
**Type of change**
Bugfix

**Issue(s) closed**
Fixes a new regression where resizing the Widelands window during an MP game segfaults.

**To reproduce**
1. Play a multiplayer game
2. Resize the Widelands main window
3. Segfault

**New behavior**
By design, the launch game window is only destroyed after the game ends, whereas its chat panel is destroyed earlier, as soon as the game launches. So during layout the chat may no longer exist. This was already checked for in line 135, now we check it in line 123 and 129 as well. 

**Possible regressions**
none


tothxa (approved):
Looks safe. Tested and works.

PR #6081, opened 2023-08-16
Fix Food Production objective in first atlantean campaign

gpalino:
**Type of change**
Bugfix of text description for Food Production objective, in first atlantean campaign.

**New behavior**
Removed sentence which was claiming the fish breeder won’t be able to breed new fishes if fisher catches all. This is not true.


hessenfarmer (approved):
LGTM. 
Thanks for spotting and fixing

PR #6082, opened 2023-08-16
Fix selection cursor setting after picking custom starting position

### LINKED ISSUE #5664: Custom starting position and mouse cursor
After placing a headquarters the mouse cursor remains to "click to place headquarters". Only after placing a road or a building it changes to the normal cursor.

**To Reproduce**
1. Start game with custom starting position
2. click on the map to place your hadquarters
3. mouse cursor remains to "place headquarters"

**Expected behavior**
After placing the headquarters the mouse should turn into the normal cursor.

**Screenshots**

csp_mouse_cursor

**Version:**
 - OS: archlinux
 - Widelands Version: 1.1 Release

tothxa:
**Type of change**
Bugfix / Refactoring

**Issue(s) closed**
Fixes #5664

**New behavior**
Picking the starting position now has an additional intermediate state between sending the playercommand and actually executing it.

**Possible regressions**
Saving/loading games while some players are picking starting position? (Can a save actually happen between sending and executing the command? I'm not sure whether that state would be reloaded correctly, regardless of this change.)


tothxa:
@Noordfrees May I propose to include this in the tournament version for the next run now that it will have custom starting positions? Though this is only a minor visual glitch, much less important than the other PR.

Noordfrees (approved):

tothxa:
Thanks. :)

Noordfrees:
This change causes a reproducible desync when using custom starting positions in a networking game. Please advise how to proceed for the tournament…

tothxa:
Possible fix is compiling… I reverted this and use a separate bool variable for the intermediate state for the local player only.

Issue #5895, opened 2023-05-01
Naval Warfare: implement AI handling


hessenfarmer:
Once the basic feature is pretty much stable we need to implement (at least basic) handling of this feature for the AI

PR #6083, opened 2023-08-16
Navalwarfare ai

hessenfarmer:
**Type of change**
New feature 

**New behavior**
First attempt to implement some rudimentary handling of warships by the AI

**How it works**
AI should refit ships to warships. set the capacity to zero. and park an amount of warships in front of each port it has.    
AI should garrison some soldiers in Ports as well.

re #5895


hessenfarmer:
@Noordfrees 
this branch works now and does what is described above. It just implements the basic mechanics without any genetics as a proof of concept. Shall we have it (as it is better then nothing) or shall we wait until I have implemented genetics and more features?.

tothxa:
It's not in the description and I couldn't see it in the code either at a quick glance, but I believe it would be easy: The AI should set some number of soldiers (can be fixed 5 or even only 3 for now) for port garrisons.

This may be harder, but maybe just some adjustment for the existing checks: Could you also make the AI treat ports and port spaces like enemy borders, so that it keeps some military buildings nearby?

hessenfarmer:
@Noordfrees 
unfortunately I am too dumb to get the container iteration over portsites correct to satisfy all of our CI requirements 
please help

src/ai/defaultai_seafaring.cc
```diff
@@ -545,6 +585,27 @@ void DefaultAI::expedition_management(ShipObserver& so) {
...
+
+	const Time& gametime = game().get_gametime();
+
+	game().send_player_warship_command(*so.ship, Widelands::WarshipCommand::kSetCapacity, {0u});
+
+	for (auto i : portsites) {
+		if (i.guard_ship_needed) {
+			game().send_player_ship_set_destination(*so.ship, i.site->get_portdock());
+			i.guard_ship_needed = false;
+			i.ships_assigned = ++i.ships_assigned;
```
tothxa: This looks like it should be:

```suggestion
			++i.ships_assigned;
```

but I didn't check the wider context whether that's so, or the value should be assigned to some other variable.

tothxa: Or not... I never know the precedences... Maybe some parentheses are needed?

hessenfarmer: omg you are right. what a stupid mistake from me

frankystone:
I am testing this with a local LAN-Game now:
* Map: The Nile
* All tribes with me as a spectator
* Disabled diplomacy
* Enabled Naval warefare

I get an assertion close after starting the game:
~~~~
widelands: /home/kaputtnik/Quellcode/widelands-repo/widelands/src/ai/defaultai_seafaring.cc:176: bool AI::DefaultAI::marine_main_decisions(const Time&): Zusicherung »tradeships_count >= expeditions_in_progress« nicht erfüllt.
~~~~

hessenfarmer:
> I am testing this with a local LAN-Game now:
> 
>     * Map: The Nile
> 
>     * All tribes with me as a spectator
> 
>     * Disabled diplomacy
> 
>     * Enabled Naval warefare
> 
> 
> I get an assertion close after starting the game:
> 
> ```
> widelands: /home/kaputtnik/Quellcode/widelands-repo/widelands/src/ai/defaultai_seafaring.cc:176: bool AI::DefaultAI::marine_main_decisions(const Time&): Zusicherung »tradeships_count >= expeditions_in_progress« nicht erfüllt.
> ```
> 
> Version is: 1.2~git26569 ([aac77ba](https://github.com/widelands/widelands/commit/aac77bab479dbcb7f5dd6dc50f5bee1daf25a4cc)@navalwarfare-ai) Debug

which startconditions did you use?

hessenfarmer:
fixed it (hopefully) thanks for testing

frankystone:
> which startconditions did you use?

The only thing which i very seldom change: The normal pre-selected startcondition. Win condition was Autocrat (or HQ-Hunter).

Yes, now it works. The game is running...

frankystone:
After about 4 hrs game time some empty Warships are created and positioned at a port.

From my point of view the ratio Ship/Warship/Ports is not good. E.g. in this game:
* Yellow has 3 ports, 4 Transport ships and only 1 Warship (should have 3 warships, one for each port)
* White has 4 Ports, 5 Transportation ships and only 1 warship (should have 4 warships, one for each port)
* Orange has 1 Port, 3 Transportation ships and 1 expedition, no warship (should have 1 warship for the port)

If feasible there should be at least one warship per Port and the amount of warships should be something like 1/2 of Transportships+Expeditionships. So for Orange this would mean (3 Transportships +1 Expedition) / 2 = 2 Warships. If the division not equal then prefer warships. At least building warships should be somehow preferred over building normal ships imho.

Save Game (multiplayer LAN): [test_naval_ai_2.zip](https://github.com/widelands/widelands/files/13692194/test_naval_ai_2.zip)

frankystone:
Thanks!

In this game the gray player has 5 ships and 2 warships. But one of this warship is not empty and is not stationed at a port.

[test_naval_ai.zip](https://github.com/widelands/widelands/files/13702178/test_naval_ai.zip)

hessenfarmer:
Thanks for the testing. Whiloe running a test on my own I already realized that the stationi8ng code is not Working as intended (anymore). I had one player with 3 ports and 4 warships and 1 port had 3 ships one port had 1 ship and the last had none.
Which should be 2, 1 and 1.
will need to debug and add some logging to find and correct the issue

hessenfarmer:
As I am lacking time and inspiration currently. I would like to have this in its current state. If I will find either of the 2 (time or inspiration I might prepare a second PR based on this one. But better to have this in the Release in its current state rather then wait for me.   
So @widelands/developers coud anybody do a code review to get this merged?

src/ai/ai_help_structs.h
```diff
@@ -580,7 +580,15 @@ struct WarehouseSiteObserver {
 
 	Widelands::Warehouse* site;
 	BuildingObserver* bo;
-	uint32_t flag_distances_last_update;
+	// uint32_t flag_distances_last_update;
```
bunnybot: Nordfriese*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4441#issuecomment-1423419)*
*On Sun Jan 07 12:16:42 CET 2024, **Benedikt Straub (Nordfriese)** wrote:*

---

should be deleted if it's unused

src/ai/defaultai.h
```diff
@@ -79,7 +79,7 @@ struct DefaultAI : ComputerPlayer {
 	enum class WalkSearch : uint8_t { kAnyPlayer, kOtherPlayers, kEnemy };
 	enum class WoodPolicy : uint8_t { kDismantleRangers, kStopRangers, kAllowRangers };
 	enum class NewShip : uint8_t { kBuilt, kFoundOnLoad };
-	enum class FleetStatus : uint8_t { kNeedShip = 0, kEnoughShips = 1, kDoNothing = 2 };
+	// enum class FleetStatus : uint8_t { kNeedShip = 0, kEnoughShips = 1, kDoNothing = 2 };
```
bunnybot: Nordfriese*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4441#issuecomment-1423422)*
*On Sun Jan 07 12:18:49 CET 2024, **Benedikt Straub (Nordfriese)** wrote:*

---

should be deleted if it's unused

src/ai/defaultai_seafaring.cc
```diff
@@ -171,15 +173,8 @@ bool DefaultAI::marine_main_decisions(const Time& gametime) {
...
-	for (const ShipObserver& observer : allships) {
-		if (observer.ship->state_is_expedition()) {
-			++expeditions_in_progress;
-		}
-	}
-
-	assert(allships.size() >= expeditions_in_progress);
-	bool ship_free = allships.size() - expeditions_in_progress > 0;
+	assert(tradeships_count >= expeditions_in_progress);
+	bool ship_free = tradeships_count - expeditions_in_progress - ports_count / 3 > 0;
```
bunnybot: Nordfriese*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4441#issuecomment-1423439)*
*On Sun Jan 07 12:21:01 CET 2024, **Benedikt Straub (Nordfriese)** wrote:*

---

This looks suspicious, there might be an overflow bug with unsigned arithmetics. Please explicitly cast all these values to signed int before subtracting them just to be safe.

src/ai/defaultai_seafaring.cc
```diff
@@ -189,7 +184,8 @@ bool DefaultAI::marine_main_decisions(const Time& gametime) {
 	const bool need_ship =
 	   ports_count > 0 && shipyards_count > 0 && basic_economy_established &&
 	   (!ship_free || persistent_data->ships_utilization > 5000 ||
-	    static_cast<int>(allships.size()) - ports_count - expeditions_in_progress < 0);
+	    static_cast<int>(tradeships_count) - ports_count - expeditions_in_progress < 0 ||
+	    static_cast<int>(ports_finished_count) * 2 - warships_count > 0);
```
bunnybot: Nordfriese*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4441#issuecomment-1423440)*
*On Sun Jan 07 12:21:13 CET 2024, **Benedikt Straub (Nordfriese)** wrote:*

---

same here

src/ai/defaultai_seafaring.cc
```diff
@@ -227,21 +224,38 @@ bool DefaultAI::marine_main_decisions(const Time& gametime) {
 		}
 	}
 
+	if ((static_cast<int>(ports_finished_count) * 2 - warships_count) > 0 &&
+	    (tradeships_count - expeditions_in_progress - ports_count / 3) > 1) {
```
bunnybot: Nordfriese*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4441#issuecomment-1423441)*
*On Sun Jan 07 12:21:25 CET 2024, **Benedikt Straub (Nordfriese)** wrote:*

---

and here

bunnybot (commented):
Nordfriese*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4441#issuecomment-1423450)*
*On Sun Jan 07 12:16:42 CET 2024, **Benedikt Straub (Nordfriese)** **requested changes** on this pull request:*

Code looks good, a few comment inline. I'm not very familiar with the AI code though.

src/ai/defaultai_seafaring.cc
tothxa: bunnybot didn't pick this up from CB. I think you could convert these like this to avoid ugly casts:

```suggestion
	bool ship_free = tradeships_count > expeditions_in_progress + ports_count / 3;
```

hessenfarmer:
implemented all suggestions except garrison. will implement this in this PR as I overlooked my own PR description.

src/ai/defaultai_seafaring.cc
```diff
@@ -227,21 +225,38 @@ bool DefaultAI::marine_main_decisions(const Time& gametime) {
...
+		warship_needed = true;
+	}
+	if (!ship_free && warships_count > 0) {
+		tradeship_refit_needed = true;
+	}
+
 	// starting an expedition? if yes, find a port and order it to start an expedition
+	bool start_expedition = false;
 	if (ports_count > 0 && expeditions_in_progress == 0 && expeditions_in_prep == 0 &&
 	    !persistent_data->no_more_expeditions && ship_free && basic_economy_established) {
```
bunnybot: tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4441#issuecomment-1424402)*
*On Sun Jan 07 21:49:01 CET 2024, **Tóth András (tothxa)** wrote:*

---

I think this would better use `!need_ship` instead of `ship_free` (if I got things right, but maybe some different logic with warship vs. ports count is needed here), to only start colonising when the existing ports are protected. In my test game I saw AIs too eager to colonise and defend ports too late.

bunnybot (commented):
tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4441#issuecomment-1424403)*
*On Sun Jan 07 21:49:01 CET 2024, **Tóth András (tothxa)** commented.*

src/ai/defaultai_seafaring.cc
hessenfarmer: correct but I believe that is currently out of scope counting my time and the upcoming release. So as for diplomacy in the first release when it was implemented this is just a very rudimentary solution.      
I am not much in favor to do much more then getting the ports garrison done as we still might face big changes in naval warfare.

tothxa: I think this will already go a long way while also allowing escape from the first island where needed:

```suggestion
	    !persistent_data->no_more_expeditions && (ports_count < 2 ? ship_free : !need_ship) && basic_economy_established) {
```

src/ai/defaultai_seafaring.cc
```diff
@@ -227,21 +225,38 @@ bool DefaultAI::marine_main_decisions(const Time& gametime) {
...
+	for (PortSiteObserver& p_obs : portsites) {
+		if (start_expedition) {
+			verb_log_dbg_time(game().get_gametime(),
+			                  "  %1d: Starting preparation for expedition in port at %3dx%3d\n",
+			                  player_number(), p_obs.site->get_position().x,
+			                  p_obs.site->get_position().y);
+			game().send_player_start_or_cancel_expedition(*p_obs.site);
+			start_expedition = false;
 		}
+		p_obs.guard_ship_needed = warships_count / num_ports > p_obs.ships_assigned;
```
tothxa: I think this will distribute warships more evenly: (yes, C integer math is tricky :)

```suggestion
		p_obs.guard_ship_needed = warships_count > num_ports * p_obs.ships_assigned;
```

Though I think there's also more accounting needed for warships, like when ports are lost then any assigned warships should be sent elsewhere, and when a warship is lost, its port's `ships_assigned` should be decreased.

src/ai/defaultai_seafaring.cc
```diff
@@ -254,24 +269,44 @@ bool DefaultAI::check_ships(const Time& gametime) {
...
+			if (so.ship->get_ship_type() == Widelands::ShipType::kWarship) {
+				++warships_count;
+				if (tradeship_refit_needed) {
+					game().send_player_refit_ship(*so.ship, Widelands::ShipType::kTransport);
+					tradeship_refit_needed = false;
+					--warships_count;
+				}
+			} else {
+				++tradeships_count;
+				if (warship_needed) {
```
tothxa: Not observed, but I suspect expedition ships may need protection:

```suggestion
				if (warship_needed && !so.ship->state_is_expedition()) {
```

hessenfarmer: good idea thanks

src/ai/defaultai_seafaring.cc
hessenfarmer: well I beleive this would lead to a lot of warships shipping around as it would trigger the guard_ship needed much more often. I assume that integer division is always floored, while in your suggestion it would be ceiled. 
if a port is lost we reduce num_ports and erase the port site observer (defaultai.cc line 6872 by this the ships gets send to an existing port.
the ship does not know to which port it has been send, we only count ships sent to a port, however we might need to check that no ship got send away.

src/ai/defaultai_seafaring.cc
hessenfarmer: good idea

src/ai/defaultai_seafaring.cc
tothxa: > well I beleive this would lead to a lot of warships shipping around as it would trigger the guard_ship needed much more often.

Actually I suspect warships that already have a destination should not look for new assignments. I'm planning to do that in the branch I forked from this, but I hit a hard to debug `std::exception`, so I need to solve that before I can continue testing.

> I assume that integer division is always floored, while in your suggestion it would be ceiled.

Let's see with 3 ports. Warships are ordered until we have 6.

1 ship: ships/ports = 1/3 → 0, no port requests it, so it just lingers around where it was refitted (at least that port gets one ;)  
   with my suggestion all ports request a ship, lucky one gets it  
2 ships: 2/3 → 0: same: if you're lucky, you have 2 ports with 1 ship each, but you may get both ships at the same port  
   my: previous port is happy, another one gets the new ship  
3 ships: 3/3 = 1: finally ports request ships, everything's fine  
   my: 3rd ports gets new ship, everything's fine

Now, if we continue from the same state, we get the same pattern, so let's examine the state where at least one port has 2 ships and at least another (e.g. a newly built one) has 0:

4 ships: 4/3 → 1: port with 2 ships is more than happy, port with 1 ship is happy, though it could request the ship, port without ships requests a ship, 2-1-1 guaranteed  
   my: ports with 1 ship and without ship request ship: this is actually worse, may end up 2-2-0  
5 ships: 5/3 → 1: ports with 2 ships are happy, ports with 1 ship are happy, but shouldn't be, only new port requests ship: may end up 2-1-1 with one ship in limbo, so it may be effectively 3-1-1  
   my: 2-2-1 is guaranteed
6 ships: everything's fine again with both calculations

Extending that to more ports: Your calculation only starts to distribute ships when there are more warships than ports, and after that, ports only request 2nd ship when all ships are ready. With my calculation new ports may lose out if unlucky, but there will always be requests for ready ships. The new port thing could be checked by the ship when choosing a port: if it finds one without ships then go there, otherwise just remember the requesting port, and fulfil the request at the end of the loop (if not broken out by a port without ships).

hessenfarmer: the problem is that ports do not know how many if any are actually there actually. And we iterate over all warships all the time. so we might easily send away a ship in your suggestion and do this much seldom in my current state. 
Again I only intended to implement a real dumb workaround to give the players an impression of the feature.   
however as I am still struggling with time you might easily take over here and just commit to this branch as well.

tothxa: Thank you, I'll do that when I have sorted out the above mentioned exception problem.

I also implemented garrisons in my local branch, but that involves more places in the code, so I couldn't make suggestions, and also I'm afraid that the source of the exceptions may be in those changes.

src/ai/defaultai_warfare.cc
```diff
@@ -647,23 +647,32 @@ void DefaultAI::count_military_vacant_positions() {
...
+		} else {
+			on_stock_ += wh.site->soldier_control()->stationed_soldiers().size() -
+			             wh.site->get_desired_soldier_count();
+		}
 	}
 
-	vacant_mil_positions_ += on_stock_;
+	vacant_mil_positions_ -= on_stock_;
 
+	// TODO(tothxa): This doesn't seem to be used for anything.
```
hessenfarmer: correct. Could you please remove this as it is not used.

src/ai/defaultai_seafaring.cc
```diff
@@ -546,6 +606,83 @@ void DefaultAI::expedition_management(ShipObserver& so) {
...
+				picked_port = &p_obs;
+				picked_port_guard_ships = p_obs.ships_assigned;
+				if (picked_port_guard_ships == 0) {
+					break;
+				}
+			}
+		}
+
+		if (picked_port != nullptr && picked_port_guard_ships < kWarshipsPerPort) {
+			assert(picked_port_guard_ships <= kWarshipsPerPort);
```
hessenfarmer: this assert seems to be useless as it can not fail anytime ( we simply wouldn't rreach this point if this would fail

src/ai/defaultai_seafaring.cc
```diff
@@ -546,6 +606,83 @@ void DefaultAI::expedition_management(ShipObserver& so) {
...
+		} else {
+#ifndef NDEBUG
+			if (picked_port == nullptr) {
+				assert(portsites.empty());
+			} else {
+				assert(picked_port_guard_ships == kWarshipsPerPort);
+				assert(picked_port_guard_ships == picked_port->ships_assigned);
+			}
+#endif
+			log_dbg_time(gametime, "AI %d: all ports have enough guard ships, warship %s remains idle",
```
hessenfarmer: I always thought about how to increase the max number per port. maybe we could just increase this by 1 if this is the case.

src/ai/defaultai_seafaring.cc
tothxa: Oops. Thanks!

src/ai/defaultai_seafaring.cc
tothxa: That would mess up the refit needed calculations.

But I've just thought of a quick and dirty workaround...

hessenfarmer:
thanks for the help provided. I'll test this tomorrow.

hessenfarmer:
> I think this should be ready for v1.2 now. I haven't looked into the shipyard thing directly, but while doing the above, I've seen code assuming there's only one shipyard, so allowing even 2 will not be a simple task, so we'll have to do without it in v1.2 unfortunately.
> 

I will have a look, if we can do something about this.     

> As a very ugly workaround, how about gifting the AI a free guardship for every newly built port (or the first 3 or 4, _[edit:] and if it has less than e.g. 5 ships_)? That should be simple enough to be done either in this PR, or (probably better in) a quick follow-up.

-1. We always had the philosophy that our AI is not using any cheats that a player can't use. (e.g. knowledge of where to find minerals can be obtained from the editor and good players do use this information as well. )

tothxa:
I've just got a segfault in `DefaultAI::expedition_management()` testing this branch:
```
##############################
FATAL ERROR: Received signal 11 (Segmentation fault)
Backtrace:
src/widelands-naval/widelands-naval-ai(+0x7876db)[0x5612740416db]
/lib/x86_64-linux-gnu/libc.so.6(+0x37970)[0x7f2c49874970]
src/widelands-naval/widelands-naval-ai(_ZN2AI9DefaultAI21expedition_managementERNS_12ShipObserverE+0x10c)[0x561273c21c9c]
src/widelands-naval/widelands-naval-ai(_ZN2AI9DefaultAI11check_shipsERK4Time+0x305)[0x561273c22235]
src/widelands-naval/widelands-naval-ai(_ZN2AI9DefaultAI5thinkEv+0xbab)[0x561273c653fb]
src/widelands-naval/widelands-naval-ai(_ZN26SinglePlayerGameController5thinkEv+0x113)[0x561273f769e3]
src/widelands-naval/widelands-naval-ai(_ZN9Widelands4Game5thinkEv+0xdd)[0x561273fbb87d]
src/widelands-naval/widelands-naval-ai(_ZN15InteractiveBase16game_logic_thinkEv+0x163)[0x561273e89763]
src/widelands-naval/widelands-naval-ai(_ZN2UI5Panel12logic_threadEv+0x8b)[0x561273f27c6b]
/lib/x86_64-linux-gnu/libstdc++.so.6(+0xbbb2f)[0x7f2c49c55b2f]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x7fa3)[0x7f2c49d8cfa3]
/lib/x86_64-linux-gnu/libc.so.6(clone+0x3f)[0x7f2c4993606f]

Please report this problem to help us improve Widelands, and provide the complete output.
##############################
```

It was about 3 minutes after the last autosave, so I'm not sure if I will be able to reproduce.

hessenfarmer:
just guessing about the segfault: could it be possible that we send a ship that is refiiting into expeditions management

hessenfarmer:
> tothxa_[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4441#issuecomment-1531856)_ _On Tue Jan 23 22:53:15 CET 2024, **Tóth András (tothxa)** wrote:_
> 
> Unfortunately I wasn't able to reproduce the segfault from the last autosave.
> 
> Later on I saw another different bug: All ships of one AI player disappeared at once (by the ships graph in general stats), shortly after it finally started colonising. I suppose the conditions of the wrong shipyard position check must have triggered. So that needs revising too.

yes I can think of different situations where this might happen. Which map were you testing on and do you have a replay or a savegame of this game.

src/ai/defaultai_seafaring.cc
```diff
@@ -171,77 +173,129 @@ bool DefaultAI::marine_main_decisions(const Time& gametime) {
...
+		if (ps_obs.site != nullptr && ps_obs.bo->is(BuildingAttribute::kShipyard)) {
+
+			// Verify whether only the correct fleet is reachable
+			bool dismantle = ps_obs.site->get_ship_fleet_interfaces().empty();  // wrong place?
+			// TODO(tothxa): When the AI is made to handle multiple fleets, this needs to be
+			//    changed to only dismantle if it has interfaces to more than one fleet.
+			if (Widelands::Serial fleet = get_main_fleet(); fleet != Widelands::kInvalidSerial) {
+				for (const Widelands::ShipFleetYardInterface* yard_if :
+				     ps_obs.site->get_ship_fleet_interfaces()) {
+					if (yard_if->get_fleet()->serial() != fleet) {
```
bunnybot: hessenfarmer*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4441#issuecomment-1535340)*
*On Fri Jan 26 09:43:02 CET 2024, **Stephan Lutz (hessenfarmer)** wrote:*

---

I beleive we might need inverse logic here.
currently we dismantle the shipyard if it has one wrong ship interface.  
But there might be a case where we might need a shipoyard to operate if it has at least 1 correct interface.
(think of a narrow strip of land where we have the ocean on one side and an inner lake in reachability of the shipyard. 
I will try to test this on Europe in the baltic area (this was the map to test this feature in the first place)

bunnybot (commented):
hessenfarmer*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4441#issuecomment-1535341)*
*On Fri Jan 26 09:43:02 CET 2024, **Stephan Lutz (hessenfarmer)** commented.*

tothxa:
I merged your shipyards deque changes to my branch, using my target setting and start-stop changes. It compiles for me, and my test game loads and AIs behave reasonably, build additional shipyards, etc. Not ran it for long though, neither started new game from scratch.

I don't want to create a PR to this one, because that would pollute the project PRs.

src/ai/defaultai.cc
```diff
@@ -6137,15 +6138,16 @@ BuildingNecessity DefaultAI::check_building_necessity(BuildingObserver& bo,
 			return BuildingNecessity::kForbidden;
 		}
 		if (bo.is(BuildingAttribute::kShipyard)) {
-			if (bo.total_count() > 0 ||
+			if ((num_ports < 3 ? bo.total_count() > num_ports : bo.total_count() > num_ports / 2) ||
```
tothxa:
```suggestion
			if (bo.total_count() > num_ports / 2 + 1) ||
```

Not only simpler, but also:

np | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
if | 0 | 1 | 2 | 1 | 2 | 2 | 3 | 3 | 4 | 4 | 5
n/2+1 | 1 | 1 | 2 | 2 | 3 | 3 | 4 | 4 | 5 | 5 | 6

tothxa: Sorry, this is post fact, so this may be better:

```suggestion
			if (bo.total_count() > (num_ports + 1) / 2) ||
```

np | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
if | 0 | 1 | 2 | 1 | 2 | 2 | 3 | 3 | 4 | 4 | 5
n/2+1 | 0 |1 | 1 | 2 | 2 | 3 | 3 | 4 | 4 | 5 | 5

hessenfarmer:
ok I took over your changes.   
I will start testing now again

src/ai/defaultai_seafaring.cc
```diff
@@ -116,133 +115,252 @@ bool DefaultAI::marine_main_decisions(const Time& gametime) {
...
+			// TODO(tothxa): player commands are asynchronous, need to return here if we want the
+			//               below extra check
+			// return false;
 		}
-	}
 
-	// If non-seafaring economy, no sense to go on with this function
-	if (!map_allows_seafaring_) {
+		// if (count_buildings_with_attribute(BuildingAttribute::kShipyard) == 0) {
+		// TODO(tothxa): bring back old code going over all buildings, or trust above?
```
bunnybot: tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4441#issuecomment-1539102)*
*On Sun Jan 28 15:51:53 CET 2024, **Tóth András (tothxa)** wrote:*

---

We can probably trust it, and remove this and the other one inside the loop.

src/ai/defaultai_seafaring.cc
```diff
@@ -116,133 +115,252 @@
...
+		// }
+
+		// if (!allships.empty()) {
+		// TODO(tothxa): delete ships?
```
bunnybot: tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4441#issuecomment-1539103)*
*On Sun Jan 28 15:52:12 CET 2024, **Tóth András (tothxa)** wrote:*

---

Leave it for now?

src/ai/defaultai_seafaring.cc
```diff
@@ -116,133 +115,252 @@
...
 	}
 
-	// iterates over all ships
-	for (const ShipObserver& observer : allships) {
-		if (observer.ship->state_is_expedition()) {
-			++expeditions_in_progress;
+	// Evaluate fleet, decide needs
+
+	{  // scope for temporary variables
+		// TODO(tothxa): turn this and the shipyard and port handling parts into separate functions
```
bunnybot: tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4441#issuecomment-1539104)*
*On Sun Jan 28 15:53:07 CET 2024, **Tóth András (tothxa)** wrote:*

---

I'd do it in this PR.

src/ai/defaultai_seafaring.cc
```diff
@@ -116,133 +115,252 @@
...
-	// getting some base statistics
-	player_ = game().get_player(player_number());
-	uint16_t ports_count = 0;
-	uint16_t shipyards_count = 0;
-	uint16_t expeditions_in_prep = 0;
-	uint16_t expeditions_ready = 0;
-	uint16_t expeditions_in_progress = 0;
-	bool idle_shipyard_stocked = false;
 
 	// goes over productionsites and gets status of shipyards
```
bunnybot: tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4441#issuecomment-1539109)*
*On Sun Jan 28 15:54:53 CET 2024, **Tóth András (tothxa)** wrote:*

---

No longer true, needs updating

src/ai/defaultai_seafaring.cc
```diff
@@ -116,133 +115,252 @@
...
+			continue;
+		}
+
+		if (fleet != nullptr && p_obs.site->get_portdock() != nullptr &&
+		    p_obs.site->get_portdock()->get_fleet() != nullptr &&
+		    p_obs.site->get_portdock()->get_fleet()->serial() != fleet->serial()) {
+			// Only allow one fleet for now.
+			// TODO(tothxa): v1.3: Make the AI handle multiple fleets.
+			// TODO(tothxa): Immediate: When evaluating portspaces for building additional ports,
+			//               check whether portsites.front() is reachable.
```
bunnybot: tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4441#issuecomment-1539116)*
*On Sun Jan 28 15:58:05 CET 2024, **Tóth András (tothxa)** wrote:*

---

I wrote it should be fixed here, but I'm not even sure I could find where it's done...

src/ai/defaultai_seafaring.cc
```diff
@@ -546,6 +700,108 @@
...
+			so.last_command_time = gametime;
+			so.waiting_for_command_ = false;
+			return;
+		}
+
+		verb_log_dbg_time(gametime, "AI %d: Warship %s is looking for assignment.",
+		                  player_->player_number(), so.ship->get_shipname().c_str());
+		PortSiteObserver* picked_port = nullptr;
+		Widelands::Quantity picked_port_guard_ships = kWarshipsPerPort + 1;
+		// TODO(tothxa): Should also consider distance to avoid unnecessary travel and
```
bunnybot, line 776: tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4441#issuecomment-1539118)*
*On Sun Jan 28 16:00:38 CET 2024, **Tóth András (tothxa)** wrote:*

---

Not in this PR.

bunnybot (commented):
tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4441#issuecomment-1539126)*
*On Sun Jan 28 15:51:53 CET 2024, **Tóth András (tothxa)** commented:*

I've annotated my TODO-s to mark the ones I think need attention in this PR.

src/ai/defaultai_seafaring.cc
```diff
@@ -116,133 +115,252 @@ bool DefaultAI::marine_main_decisions(const Time& gametime) {
...
+
+		const bool ship_free = tradeships_count > min_tradeships;
+		const bool ships_full = tradeships_count >= tradeships_target;
+
+		const uint32_t warships_target = ports_finished_count * kWarshipsPerPort;
+		const bool warship_shortage =
+		   game().naval_warfare_allowed() && (warships_target > warships_count);
+
+		const bool consider_expedition =
+		   ports_count > 0 && expeditions_in_progress == 0 && expeditions_in_prep == 0 &&
```
bunnybot: tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4441#issuecomment-1539174)*
*On Sun Jan 28 16:21:10 CET 2024, **Tóth András (tothxa)** wrote:*

---

also `expeditions_ready == 0`

bunnybot (commented):
tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4441#issuecomment-1539175)*
*On Sun Jan 28 16:21:10 CET 2024, **Tóth András (tothxa)** commented.*

hessenfarmer:
I have seen your changes.  Will try to incorporate them here.    
However I chose a different way for coping with shipyards while not having a port yet.   
I think this might delay starting the first ship until we might really need it, while the first ship in your case will get started earlier.

hessenfarmer:
> tothxa_[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4441#issuecomment-1539802)_ _On Sun Jan 28 21:24:35 CET 2024, **Tóth András (tothxa)** wrote:_
> 
> https://codeberg.org/wl/widelands/commit/31af330d2c3044aceacc2cf2d76b901f68e2b5be is not needed AFAICT. I believe `static` meant they were shared by all AIs, which was fine, as whether a map allows seafaring is common to all players, isn't it? It's already quite paranoid to check it every 20s, we don't need all AIs doing it separately.

well everything in our ai code is performed by every Ai all the time so we do check it exactly like you described already

hessenfarmer:
I will not add anything today. 
regarding ships production:
either we waist 1 ship without a port or have one ship ready if we manage to build a port (your solution)   
or we need to wait a little for the first ship after the port is finished and may have multiple ships if we finally mange to finish our port ( my solution)     
both have pros and cons and for both of them there are cornercases where they are not working well. 
So it is a matter of preference and subjective evaluation,     
I still believe the chance for not building a port is higher (due to the weakness in AI code you already stated ) so I prefer my solution as it is hard to prevent this situation.      
So if you could live with my solution I would really appreciate not to change anything here.     
   
Only changes possible but not necessary before merging this are:    
- adopt ports garrison to soldiersStatus (lower if we are short, higher if we have enouzgh of them)
- find a measure to increase guarding ships per port. (e.g. ships full and more then 20 stock of all needed wares. )

tothxa:
> regarding ships production: either we waist 1 ship without a port or have one ship ready if we manage to build a port (your solution) or we need to wait a little for the first ship after the port is finished and may have multiple ships if we finally mange to finish our port ( my solution) both have pros and cons and for both of them there are cornercases where they are not working well. So it is a matter of preference and subjective evaluation, I still believe the chance for not building a port is higher (due to the weakness in AI code you already stated ) so I prefer my solution as it is hard to prevent this situation. So if you could live with my solution I would really appreciate not to change anything here.

For now I just did a merge, so both are there. This is the safest anyway, from both points of view.

> Only changes possible but not necessary before merging this are:
> 
> * adopt ports garrison to soldiersStatus (lower if we are short, higher if we have enouzgh of them)

This is up to you. I can only work on new aspects of AI code if I get a starting point. :)

I would never decrease it though, because naval invasions can be too powerful, and only increase it (max it out if we can afford?) when a guarding ship gets an attack notification. So the guarding ships and the initial garrison are there to get time for the reinforcements to arrive.

> * find a measure to increase guarding ships per port. (e.g. ships full and more then 20 stock of all needed wares. )

I don't think we should spend more effort on this until the further direction of naval warfare is decided, and real AI is implemented for it. The 2 ships should be fine for the above role.

src/ai/defaultai_seafaring.cc
```diff
@@ -105,22 +105,12 @@ uint8_t DefaultAI::spot_scoring(Widelands::Coords candidate_spot) {
...
-	return map_allows_seafaring_;
-}
-
 // This function scans current situation with shipyards, ports, ships, ongoing expeditions
 // and makes two decisions:
 // - build a ship
 // - start preparation for expedition
 bool DefaultAI::marine_main_decisions(const Time& gametime) {
-	if (!check_seafaring_allowed()) {
+	if (game().map().allows_seafaring()) {
```
hessenfarmer: here we miss the negation

tothxa: Already fixed in last commit.

hessenfarmer:
> tothxa_[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4441#issuecomment-1540692)_ _On Mon Jan 29 14:33:00 CET 2024, **Tóth András (tothxa)** wrote:_
> 
> > I would never decrease it though, because naval invasions can be too powerful, and only increase it (max it out if we can afford?) when a guarding ship gets an attack notification.
> 
> Though even this is not so simple, because then the AI could still be easily tricked into sending all free soldiers to a port, then launch the real invasion or a ground attack elsewhere. So I'm leaning towards leaving it constant for this PR, and probably also for v1.2.

I just thought on handling them like the other military sites which isn't  worsening the situation imho.

src/ai/defaultai_seafaring.cc
```diff
@@ -546,6 +730,109 @@ void DefaultAI::expedition_management(ShipObserver& so) {
...
+		if (tradeship_refit_needed) {
+			// Let's keep this simple. (for now?)
+			// We only backfit if we find an idle warship. Trade ships shouldn't disappear anyway, so
+			// backfit requests should be extremely rare.
+
+			verb_log_dbg_time(gametime, "AI %d: Refit warship %s to trade ship",
+			                  player_->player_number(), so.ship->get_shipname().c_str());
+			game().send_player_refit_ship(*so.ship, Widelands::ShipType::kTransport);
+			tradeship_refit_needed = false;
+			--warships_count;
```
bunnybot, line 765: tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4441#issuecomment-1541565)*
*On Tue Jan 30 00:01:11 CET 2024, **Tóth András (tothxa)** wrote:*

---

Looks like a

```
++tradeships_count;
```

is missing here.

bunnybot (commented):
tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4441#issuecomment-1541566)*
*On Tue Jan 30 00:01:11 CET 2024, **Tóth András (tothxa)** commented.*

src/ai/defaultai_seafaring.cc
```diff
@@ -110,168 +109,363 @@ uint8_t DefaultAI::spot_scoring(Widelands::Coords candidate_spot) {
...
+
 		// False indicates that we can postpone next call of this function
 		return false;
 	}
 
+	expeditions_in_progress = 0;
+	warships_count = 0;
+	tradeships_count = 0;
+
 	if (!allships.empty()) {
```
bunnybot: tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4441#issuecomment-1541582)*
*On Tue Jan 30 00:14:42 CET 2024, **Tóth András (tothxa)** wrote:*

---

This can be removed, it only containst the `for()` below. But that means reindenting ~125 lines.

Should I do it?

bunnybot (commented):
tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4441#issuecomment-1541583)*
*On Tue Jan 30 00:14:42 CET 2024, **Tóth András (tothxa)** commented.*

src/ai/defaultai_seafaring.cc
```diff
@@ -110,255 +109,438 @@ uint8_t DefaultAI::spot_scoring(Widelands::Coords candidate_spot) {
...
-						game().send_player_set_ware_priority(*observer.site, Widelands::wwWARE,
-						                                     observer.bo->inputs.at(k),
-						                                     Widelands::WarePriority::kHigh);
-					}
-				}
-			}
+			// Good utilization is 10 pieces of ware onboard, to track utilization we use range
+			// 0-10000
+			// to avoid float or rounding errors if integers in range 0-100
+			const int16_t tmp_util =
```
bunnybot, line 517: tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4441#issuecomment-1541758)*
*On Tue Jan 30 02:53:15 CET 2024, **Tóth András (tothxa)** wrote:*

---

This type is too narrow for the calculation. 10'000*20 = 200'000 > 2^15-1 = 32'767. But something fishy is going on in the background, because it actually works... I guess the purpose of the below assert is to make sure that that fishy thing keeps going on...

But it's just wrong to use `int16_t` here, and should be switched to 32bit (maybe also uint) in v1.3. Needs saveloading compatibility code, because of `persistent_data`.

bunnybot (commented):
tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4441#issuecomment-1541759)*
*On Tue Jan 30 02:53:15 CET 2024, **Tóth András (tothxa)** commented.*

bunnybot (commented):
tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4441#issuecomment-1541780)*
*On Tue Jan 30 03:27:55 CET 2024, **Tóth András (tothxa)** **approved** this pull request:*

Anyway, that's old code and works for now. Leave it alone in this PR.

bunnybot (commented):
Nordfriese*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4441#issuecomment-1541933)*
*On Tue Jan 30 09:28:14 CET 2024, **Benedikt Straub (Nordfriese)** commented:*

Thanks to you both for this big branch and the huge amount of effort you put into this :)

<@>bunnybot merge

bunnybot (commented):
Nordfriese*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4441#issuecomment-1541941)*
*On Tue Jan 30 09:34:23 CET 2024, **Benedikt Straub (Nordfriese)** **approved** this pull request:*

Codeberg sometimes seems to get its review states mixed up…

src/ai/defaultai_seafaring.cc
hessenfarmer: it is not working. It keeps overflowing and utilization has a max theoretical value of 1638 so checking for utilization hoigher then 5000 is useless.   
I had these issues in the past and I bet it is not the last one. I really would love to get this fixed asap.
Furthermore in persistent we already save it as uint so this is a complete mess

PR #6089, opened 2023-08-17
Fix documentation CI dependencies, again

Noordfrees:
Sphinx 7.2 seems to break stuff again, so add a `requirements.txt` with versions that were previously known to work.


tothxa (approved):
Let's make the CI work again. Dealing with breakage in non-critical toolchain can wait. :)

PR #6097, opened 2023-08-18
Fix `PlayerMessageCommand` Saveloading

Noordfrees:
**Type of change**
Bugfix

**Issue(s) closed**
During the tournament we encountered a corrupt savegame that failed to load with "FATAL EXCEPTION: command queue: player message: (player 1): message id is null". I could not reproduce the race condition that triggers it; but since it is valid for a message to disappear between sending and executing the command, and existence is also checked by `MessageQueue::set_message_status`, I think `null` is actually valid here.


tothxa (approved):
Looks OK. AFAICT, the ID will eventually be used by `MessageQueue::set_message_status()`, where its validity is checked, so it should be safe.

PR #6099, opened 2023-08-18
Stop ship battles when ship is destroyed

Noordfrees:
**Type of change**
Bugfix

**Issue(s) closed**
Fixes the "mirror battle not found" bug from the tournament

**New behavior**
When a ship is destroyed, it now stops fighting immediately even if it has battles left.

**Possible regressions**
The ending of complex ship battles


tothxa (approved):
Code looks OK. Is there a simple way to test this?

PR #6107, opened 2023-08-20
Fix custom naming lists for spectators

Noordfrees:
**Type of change**
Bugfix

**Issue(s) closed**
Fix a bug introduced by #6056 

**To reproduce**
1. Create custom ship/warehouse naming lists
2. Start an MP game where the host is a spectator
3. Assertion error

**New behavior**
Properly ignore naming lists for participants who don't control a player. This is a bug with bad usage of signed/unsigned/fixed-width-integer arithmetic.


tothxa (approved):
I think I get now what happened here. (spectators get uint8 max as playernumber, but here we use int, so it can be increased without problems, which was the problem :) But the different numbering systems of players (and here spectators) is very confusing.

Anyway, it doesn't crash any more. :)

PR #6116, opened 2023-08-25
update_appdata.py reports validation failure for all files

SimonHeimberg:

**Type of change**
Enhancement
**Issue(s) closed**
Fixes -

**Old behavior**
update_appdata.py reports validation failure only for `desktop-file-validate`

**New behavior**
update_appdata.py reports validation failure for all files (by return code)

### DIFF
--- a/utils/update_appdata.py
+++ b/utils/update_appdata.py
@@ -185,17 +185,17 @@ print('Done!')
 
 
 # Validate Appdata
-skip_screenshot_check = len(sys.argv) > 1 and sys.argv[1] == "--nonet"
+skip_screenshot_check = len(sys.argv) > 1 and sys.argv[1] == '--nonet'
 
 if shutil.which('appstreamcli'):
     validate_cmd = ['appstreamcli', 'validate', appdata_filepath]
     if skip_screenshot_check:
-        validate_cmd.insert(2, "--no-net")
+        validate_cmd.insert(2, '--no-net')
     appdata_result = subprocess.run(validate_cmd)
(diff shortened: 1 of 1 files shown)


tothxa:
Uh-oh… I so knew one day this would come bite me. :) While perfectly correct behaviour, this would make the flatpak beta builds fail.

Could you please add this option to allow fixing them:
https://github.com/SimonHeimberg/widelands/pull/2

(feel free to correct my python :) … except I deliberately didn't use `argparse`, because it would be too heavy for this simple task)

SimonHeimberg:
I am fine to add this option. (only one suggestion about a command in the pr)

I partially remembered to have seen the manifest and found it again. I see the call on [github.com/flathub/.../org.widelands.Widelands.yaml#L63](https://github.com/flathub/org.widelands.Widelands/blob/fa06bd147923841352d906ddcea36130f99ebe40/org.widelands.Widelands.yaml#L63)
A thought: What about modifying xdg/org.widelands.Widelands.appdata.xml instead of the stub (in the manifest)? This file is validated already. If you want to be sure it is still valid after, you can run `appstreamcli  validate --no-net xdg/...appdata.xml`

tothxa:
> A thought: What about modifying xdg/org.widelands.Widelands.appdata.xml instead of the stub (in the manifest)?

I don't want them to be out of sync, so I think it's best to use the "official" script for changes. Validation is a bonus.

tothxa (approved):
Thank you. :)

PR #6117, opened 2023-08-25
Update_authors, minor enhancement

SimonHeimberg:

**Type of change**
Enhancement (minor)

**Issue(s) closed**
Fixes -

**New behavior**
No change of generated content.
Generated files look a bit nicer and clearer.
- all have a header comment
- all are on multiple lines (and not on a very long single line)


tothxa:
Could you please break the lines even more? Like all entries on separate lines and all nested braces on an additional indentation level.

SimonHeimberg:
> Could you please break the lines even more? Like all entries on separate lines and all nested braces on an additional indentation level.

The lines in :mag:  data/txts/developers.lua and data/txts/translators_data.lua obviously.

data/txts/developers.lua, current state in pr:
```
function developers() return {
   {heading = _("Chieftains"), image = "images/players/genstats_player.png",entries = {{members = {"Benedikt Straub (Nordfriese) (Since 2022)","GunChleoc (2016 – 2022)","Holger Rapp (SirVer) (2001 – 2016)",},},},},
   {heading = ...
```
should become this:
```
function developers() return {
   {
      heading = _("Chieftains"),
      image = "images/players/genstats_player.png",
      entries = {
         {
            members = {
               "Benedikt Straub (Nordfriese) (Since 2022)",
               "GunChleoc (2016 – 2022)",
               "Holger Rapp (SirVer) (2001 – 2016)",
            },
         },
      },
   },
   {
      heading = ...
```
Did I get this correct? Or some fewer line breaks?

Sure, this is doable. If we really want to. Will take some time to properly implement (all spaces counted properly). The files will become bigger because they get quite some spaces.

tothxa:
Yes, that's what I meant. It would be much more readable this way.

utils/update_authors.py
```diff
@@ -63,10 +63,10 @@
         if translators['translator-list'] != 'translator-credits':
             locale_message += ' translators and'
             lua_translators += '\t{\n'  # entry
-            lua_translators += f'\t\theading = "{ translators["your-language-name"] }'
+            lua_translators += f'\t\theading = "{ translators['your-language-name'] }'
```
SimonHeimberg: This automatic formatting (and all others in this commit) break the code. (Pulled it locally, running the script reports an error on this line.)
How can I prevent this?

Else the code would be ready.

tothxa: Looks like a bug in the formatter. I don't know python much, so I may be wrong, but you could try:

`f'''..."{ t['i'] }'''`

tothxa:
BTW I've just had a look at `developers.json`, and realised that a simple `tr` would do 90% of the job for that file. :)

```
tr '[]' '{}' <developers.json | sed -E '/:/s/"([^"]*)" ?: ?/\1 = /'
```

would do 99.9%, only without checking the syntax. :)

utils/update_authors.py
SimonHeimberg: The try seems to be successful. Thanks.

tothxa (approved):
Thank you. Looks good now.

tothxa:
@bunnybot merge

tothxa:
@bunnybot merge force

SimonHeimberg:
> BTW I've just had a look at `developers.json`, and realised that a simple `tr` would do 90% of the job for that file. :)
> 
> ```
> tr '[]' '{}' <developers.json | sed -E '/:/s/"([^"]*)" ?: ?/\1 = /'
> ```
> 
> would do 99.9%, only without checking the syntax. :) – _edit: No, that was wrong. Headings need `_()` wrappers too (easy), and `translate` sections can not be handled by `sed`, so that's a show stopper._

I am quite sure this could be done more efficiently.

PR #6120, opened 2023-08-26
Improve update translations stats

SimonHeimberg:
**Type of change**
Enhancement

**Issue(s) closed**
Fixes -

**New behavior**
- data/i18n/translation_stats.conf has a comment header: 6469c426c05dd0392066e850f489619b7dd246f6
- on error, no other error is raises: 14c9c3abc235d4141448e25eb875798b3cc51edb

and code cleanup
- remove useless code: cf1936f6d4a2ea03feda2409c63d18d2fdf66aed
- line breaks in nicer places, : 6469c426c05dd0392066e850f489619b7dd246f6
- ...

**Additional context**
There will be a pr speeding up the script by calling pocount once with all directories: you can preview it at [branch update_translations_stats_speedup](https://github.com/SimonHeimberg/widelands/commits/update_translations_stats_speedup)
My intention of doing two steps is easier review, but maybe this makes it more complicated. Then I will adapt.


utils/update_translation_stats.py
```diff
@@ -58,18 +57,17 @@ def generate_translation_stats(po_dir, output_file):
...
+            stats_output = subprocess.check_output(
+                ['pocount', '--csv', subdir],
                 encoding='utf-8',
                 stderr=subprocess.STDOUT,
-                shell=True)
+                )
             if 'ERROR' in stats_output:
-                print('\nError running pocount:\n' + stats_output.split('\n', 0)
-                      [0]) + '\nAborted creating translation statistics.'
+                print('\nError running pocount:\n' + stats_output.split('\n', 1)[0] +
```
tothxa: Hmm, this is strange. I don't know what the original intent was, but printing only the first line of the output doesn't seem to make sense. At least print all lines containing "ERROR"?

klaus-halfmann:
@SimonHeimberg I will try to do a review, as I have some vacation right now. Can you answer/fix @tothxa s question?

utils/generate_glossary.py
```diff
@@ -202,11 +202,6 @@ def generate_glossary(po_dir, output_path, input_glossary, output_glossary, only
             # We need shell=True for the wildcards.
             poterminology_result = subprocess.run(
                 ['poterminology ' + input_path + ' -o ' + pot_path], capture_output=True, shell=True, text=True, check=True)
-            if poterminology_result.returncode != 0:
```
klaus-halfmann: Where/How is this generate_glossary function used?

klaus-halfmann (commented):
python utils/update_translation_stats.py works fine after installing 
`brew install translate-toolkit`

utils/generate_glossary.py
SimonHeimberg: This is unrelated. Somehow this slipped in the wrong commit, sorry.

utils/update_translation_stats.py
SimonHeimberg: My final suggestion is to use separate stderr (6f22d8178e566def9e34cc02bf3e013e489f5307 in [branch update_translations_stats_speedup](https://github.com/SimonHeimberg/widelands/commits/update_translations_stats_speedup)) Shall I include the commits of speedup pr in this one? Or make the 2nd pr later?

tothxa: I think it would be OK if you just printed the whole `stats_output` in case of error here, and did the stderr thing in the later PR.

utils/update_translation_stats.py
```diff
@@ -100,35 +98,33 @@ def generate_translation_stats(po_dir, output_file):
...
                     sys.exit(1)
                 locale_stats[locale] = entry
 
     print('\n\nLocale\tTotal\tTranslated')
     print('------\t-----\t----------')
 
     # The total goes in a [global] section and is identical for all locales
-    result = '[global]\n'
-    result = '{}total={}\n\n'.format(
+    result = '# This file is managed by utils/update_translations_stats.py\n\n[global]\n'
```
tothxa:
```suggestion
    result = '# This file is managed by utils/update_translations_stats.py\n\n'
    result += '[global]\n'
```

utils/update_translation_stats.py
```diff
@@ -100,35 +98,33 @@ def generate_translation_stats(po_dir, output_file):
...
+    result = '{}total={}\n'.format(
```
tothxa: I think `result += 'total={}\n'` could be used instead of this…

utils/update_translation_stats.py
```diff
@@ -100,35 +98,33 @@ def generate_translation_stats(po_dir, output_file):
...
         result, list(locale_stats.values())[0].total)
     # Write translation stats for all locales
-    for locale in (sorted(list(locale_stats.keys()), key=str.lower)):
+    for locale in sorted(locale_stats.keys(), key=str.lower):
         entry = locale_stats[locale]
         print(locale + '\t' + str(entry.total) + '\t' + str(entry.translated))
-        result = '{}[{}]\n'.format(result, locale)
-        result = '{}translated={}\n\n'.format(result, str(entry.translated))
+        result = '{}\n[{}]\n'.format(result, locale)
+        result = '{}translated={}\n'.format(result, str(entry.translated))
```
tothxa: `+=` for both?

tothxa (commented):
In addition to the inline comments:

 - `while` loop in line 89 should be replaced by `for column_counter in range(len(header_entries)):`
 - can't we drop the python version check from lines 148-150 now that we specify python3 in the `#!` line?

utils/update_translation_stats.py
SimonHeimberg: It was similar to this before 067d75e4015 (update from python2 to python3), but `+=` still works, no idea why it was changed. (There is a 3rd occurrence of this.)

SimonHeimberg:
> In addition to the inline comments:
> 
>     * `while` loop in line 89 should be replaced by `for column_counter in range(len(header_entries)):`

Agree.

>     * can't we drop the python version check from lines 148-150 now that we specify python3 in the `#!` line?

It could still be started with python2. (I do not have it on my system anymore, so it gets unlikely for others as well.)
I'd say the question is if we want to have a clear error message (as now) or just any error message for this rare case.

tothxa:
> > can't we drop the python version check from lines 148-150 now that we specify python3 in the `#!` line?
> 
> It could still be started with python2. (I do not have it on my system anymore, so it gets unlikely for others as well.) I'd say the question is if we want to have a clear error message (as now) or just any error message for this rare case.

I was wrong anyway. This part was actually added when the script was converted to python3. This script had already been `#!…python3` before #6187.

utils/update_translation_stats.py
```diff
@@ -86,12 +86,11 @@ def generate_translation_stats(po_dir, output_file):
         column_counter = 0
```
tothxa: This can be deleted too.

utils/update_translation_stats.py
```diff
@@ -88,47 +86,46 @@ def generate_translation_stats(po_dir, output_file):
...
     print('------\t-----\t----------')
 
     # The total goes in a [global] section and is identical for all locales
-    result = '[global]\n'
-    result = '{}total={}\n\n'.format(
-        result, list(locale_stats.values())[0].total)
+    result = '''# This file is managed by utils/update_translations_stats.py
+
+[global]
+'''
```
tothxa: (multiline `'''` strings just make python's ugly indenting even worse… but of course that's just my taste)

tothxa (commented):
Thank you.

utils/update_translation_stats.py
SimonHeimberg: I did.

utils/update_translation_stats.py
```diff
@@ -85,68 +82,63 @@ def generate_translation_stats(po_dir, output_file):
                 break
             header = header + ' ' + line
```
SimonHeimberg: A question: does pocount really "distributes its header over multiple lines"? I have never seen this. Not in its bug tracker, not in any output on my machine, not in any example online. Can we get rid of this? In case it really happens, the code will now (since 62addf1487373448defea9e5cfa7d13b6b762795) detect the missing columns and report it.

tothxa: Since we tell it to use the CSV format, it would actually be a major bug if it did… Maybe the terminal line wrapping confused the original author of this?

IMO you can try removing it.

utils/update_translation_stats.py
```diff
@@ -85,68 +82,63 @@ def generate_translation_stats(po_dir, output_file):
...
+        translated_column = None
+        for column_counter, cell in enumerate(header_entries):
+            if cell.strip() == 'Total Source Words':
                 total_column = column_counter
-            elif header_entries[column_counter].strip() == 'Translated Source Words':
+            elif cell.strip() == 'Translated Source Words':
                 translated_column = column_counter
-            column_counter = column_counter + 1
+        if not total_column or not translated_column:
+            for_column = 'total_column' if not total_column else translated_column
```
tothxa: I think you mean:

```suggestion
            for_column = 'total_column' if not total_column else 'translated_column'
```

But what if neither was found?

SimonHeimberg: Oops, thanks for catching

> But what if neither was found?

Then only "total_column" is reported.

tothxa: I know :) but I don't think that's right. Either print them separately as needed, or add one more condition for `'total_column and translated_column'`. But not as an inline `if`, please.

Also, this now reports "column for total_column not found", which is not nice, and `total_column` is internal to the script, so it doesn't belong in this message. The actual column names needed by the script should be reported instead.

utils/update_translation_stats.py
```diff
@@ -12,12 +12,11 @@
 """
 
 from collections import defaultdict
-from subprocess import call, check_output, CalledProcessError
+import csv
```
tothxa, line 15: Cool. :) Shows I don't know much about python. It's even a standard module.

utils/update_translation_stats.py
```diff
@@ -58,95 +57,78 @@ def generate_translation_stats(po_dir, output_file):
...
+        COLUMNS = {
+            'filename': 'Filename',
+            'total': 'Total Source Words',
+            'translated': 'Translated Source Words'
+        }
+        result = csv.DictReader(csv.StringIO(stats_output), dialect='unix', skipinitialspace=True)
+        missing_cols = set(COLUMNS.values()) - set(result.fieldnames)
+        if missing_cols:
+            sys.exit(
+                'Columns "{}" not found in output of pocount'.format('", "'.join(missing_cols)))
```
tothxa:
```suggestion
                'Column(s) "{}" not found in output of pocount'.format('", "'.join(missing_cols)))
```

utils/update_translation_stats.py
```diff
@@ -58,95 +57,78 @@ def generate_translation_stats(po_dir, output_file):
...
 
         # Now do the actual counting for the current textdomain
-        for line in result:
-            cells = line.split(',')
-            po_filename = cells[0]
+        for cells in result:
```
tothxa: `line` was better. Or could be `row` (to go with `COLUMNS`), or maybe `record` (database and `awk` jargon :). But I find `cells` misleading.

*edit: Ah, sorry, `cells` because `line` was broken into `cells` in the old code. But you don't do that any more, so still any of the above suggestions would be better.*

tothxa (approved):
Thank you. Using the csv module was clever. :)

I wonder whether we could drop some of the paranoid checks in the follow-up, but for now we can keep to the original functionality.

utils/update_translation_stats.py
SimonHeimberg: I changed it to `row`.

tothxa: Thank you. :)

utils/update_translation_stats.py
```diff
@@ -58,95 +57,77 @@ def generate_translation_stats(po_dir, output_file):
...
 
     print('\n\nLocale\tTotal\tTranslated')
     print('------\t-----\t----------')
 
     # The total goes in a [global] section and is identical for all locales
-    result = '[global]\n'
-    result = '{}total={}\n\n'.format(
-        result, list(locale_stats.values())[0].total)
+    result = ('# This file is managed by utils/update_translations_stats.py\n\n'
+              '[global]')
```
tothxa: Caught by testing:

```suggestion
              '[global]\n')
```

SimonHeimberg: Thanks. Looks like I have forgotten to run after this change.

tothxa (approved):
Thank you!

tothxa:
@bunnybot update

tothxa:
@bunnybot merge

PR #6146, opened 2023-09-09
Improve census and status lines of warehouses

tothxa:
**Type of change**
New feature / Refactoring

**New behavior**
 - Soldier stats are now shown for warehouses. HQ and ports are like military sites, plain warehouses only show number of soldiers
 - Display of warehouses is now similar to ships: Census shows name, icons are shown instead of type
 - Also similar to ships, the name editbox is shown at the top of the building window

**Possible regressions**
Soldier status display

**Screenshot**
warehouse_soldiers

**Additional context**
Discussion thread in the forum: https://www.widelands.org/forum/topic/5852/


src/logic/map_objects/tribes/soldiercontrol.cc
```diff
@@ -0,0 +1,95 @@
...
+	              StyleManager::color_tag(
+	                 as_string(has), g_style_manager->building_statistics_style().high_color()),
+	              wants);
+}
+
+}  // namespace
+
+namespace Widelands {
+
+#define CAPIDX TribeDescr::CapacityStringIndex
```
Noordfrees: Please don't do that, just spell it out below

Noordfrees (commented):
I tried to test it but loading any of my existing naval warfare saves with 95e2a81 results in an `std::bad_cast`. The same saves load fine with dde0280.

src/logic/map_objects/tribes/soldiercontrol.cc
tothxa: Sigh… I knew… But the long name makes it unreadable (beating the whole purpose of using an enum), and the confusing inconsistent line wrapping and indenting rules in our clang-format config will make it look even worse. Can't I use `using` or `namespace` somehow or something like that?

tothxa:
> I tried to test it but loading any of my existing naval warfare saves with [95e2a81](https://github.com/widelands/widelands/commit/95e2a81947332f4fb81d44a9344bacf27ce0227f) results in an `std::bad_cast`. The same saves load fine with [dde0280](https://github.com/widelands/widelands/commit/dde0280a24e6a3fdd58edee75439600b45c032f5).

Do you know where it happens? It works for me on old savegames and with games saved with the last tournament version too. And I don't think there's any unusual casting or type conversion magic here.

src/logic/map_objects/tribes/soldiercontrol.cc
```diff
@@ -0,0 +1,100 @@
...
+#include <memory>
+
+#include "base/i18n.h"
+#include "base/string.h"
+#include "graphic/style_manager.h"
+#include "logic/map_objects/tribes/tribe_descr.h"
+
+namespace {
+
+std::string format_extra_soldiers(int has, int wants) {
```
tothxa: I guess these could or even should be `Quantity`-es, but that shouldn't really matter for casting, should it?

tothxa:
Hmmm, https://en.cppreference.com/w/cpp/types/bad_cast says: "An exception of this type is thrown … also from [std::use_facet](https://en.cppreference.com/w/cpp/locale/use_facet) if the requested facet does not exist in the locale."

tothxa:
@Noordfrees I think https://github.com/widelands/widelands/pull/6146/commits/e499df01684afc7b887a43158fdf241021baaf4b corrects an overlook in my refactoring that involves locales and looks suspiciously plausible that it may have caused the `std::bad_cast` for you but not for me. Not that I understand how it works though…

Noordfrees:
Actually it turns out the bad_cast also happens with regular `protected/navalwarfare` branch, not related to this PR. I only didn't notice it there before because apparently it's a debug-build-only problem. I'll have to look into this.

I tested this briefly now, looks good, but
- Please add a whitespace between the symbol and the warehouse's name
- IMHO we shouldn't show this for warehouses that can't have a garrison setting and are not attackable

tothxa:
>     * Please add a whitespace between the symbol and the warehouse's name

Done.
 
>     * IMHO we shouldn't show this for warehouses that can't have a garrison setting and are not attackable

I added it because I missed it. And that's exactly why I put it in parentheses and draw any positive values in green as surplus.

Noordfrees:
I though about this some more and actually I think not only should we not show it for warehouses, it also should not be shown for ports and HQs with a garrison setting of 0. To me this emphasizes the military aspect too much by showing players all the time and right from the start that micromanaging your soldiers is such an important part of the game even your warehouses have to show you detailed soldier stats. And that totally misrepresents what Widelands is supposed to be about; in an ideal economy you would never have to worry about tracking individual soldiers.

So I'd like to hear some more opinions on this…

tothxa:
> I though about this some more and actually I think not only should we not show it for warehouses, it also should not be shown for ports and HQs with a garrison setting of 0.

Wouldn't that be too confusing?

I started this branch after watching some tournament replays and having to open the port windows at the soldiers tab to see the status of attacks on ports. But some players still don't set a garrison for their ports, and in that case we would still need to open the window to see the remaining soldiers.

Actually I only realised that I miss it from warehouses during testing (see the commit history), also for consistency with ports and the HQ.

tothxa:
I wonder if I may refactor the translatable strings even more, to only have "soldier"/"soldiers" in the tribe description, and have the rest as tribe independent, but still translatable formats here.

hessenfarmer:
I see the point of not over emphasize the military part. However with the long wanted naval warfare feature, defense of Ports (Portspaces) is now essential. As it adds a lot of tasks to manage as well, a quick possibility of keeping the essential defense part in the eyes is rreally appreciable imho.   
So @tothxa could you please rebase this feature on master?
I'd like to test it and really would vote for it.

PR #6017, opened 2023-07-13
different overlay image for valid start positions

SimonHeimberg:
**Type of change**
enhancement

**Issue(s) closed**
Fixes #4909

**New behavior**
While selecting start position, a different icon is shown on valid start positions. It is the icon which was already shown at the original start position, but with a grey flag.
The big green building icon is not shown, only the one for the harbour.

**Additional context**

_

***old, try1:***
It was easier to add an additional icon than to show the big overlay icon in another colour. (Latter was suggested in #4909.)
The existing image I have chosen was already loaded (in special_coords_marker_pic_). It should not conflict with the other cases as there is no building yet (so none has special work coords).

PR #6148, opened 2023-09-11
Different overlay image for valid start positions

tothxa:
**Type of change**
New feature

**Issue(s) closed**
Replaces #6017, with apologies to @SimonHeimberg. I still consider him the primary author of this change.

**New behavior**
While selecting the starting position, a different icon is shown on valid starting positions, replacing the green big building space icon, but portspaces are still shown normally even if they are valid starting positions.

The icon for the default starting position is used, with a white flag and smaller size.

The default starting positions (in playercolour) are shown with slight transparency if valid, more transparency if invalid,  until the corresponding player picks a starting position.

The default starting positions and the selection cursor are now shown in actual player colour instead of default.

**To reproduce**
*If it's a bugfix:*
1. Start a game with custom starting positions
2. Enable buildhelp
3. Old: Valid starting positions are indistinguishable from other big build spaces; New: Valid starting positions are well visible

**Possible regressions**
Buildhelp overlays?

### DIFF
--- a/src/wui/interactive_player.cc
+++ b/src/wui/interactive_player.cc
@@ -577,7 +577,8 @@ void InteractivePlayer::draw_map_view(MapView* given_map_view, RenderTarget* dst
 			}
 			const Image* player_image;
 			if (p == 0u) {
-				player_image = playercolor_image(color(220, 220, 220), "images/players/player_position.png");
+				player_image =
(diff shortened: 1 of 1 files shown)


Noordfrees (approved):
Tested and working, no regressions found; Code LGTM

tothxa:
Thank you.

PR #6150, opened 2023-09-12
Amazons balancing adjustment

tothxa:
**Type of change**
Balancing

**Issue(s) closed**
Re https://www.widelands.org/forum/post/40728/

Amazons are still easier than other tribes. (part of this PR is already merged in #6170)

**New behavior**
Make Amazons a little less easy to play by requiring more materials (most notably 3 quartz in total) for armours and some more fish/meat for training. Charcoal production is made a little more efficient to compensate the increased need. The total food requirement for a fully trained soldier now has similar amount of fish/meat as other tribes.

**Possible regressions**
Amazons economy

**Additional context**
Discussion thread in the forum: https://www.widelands.org/forum/topic/5847/


Noordfrees:
Actually I'm not sure this rare tree kiln is very useful at all. When I
play amazons then invariably I have a shortage of logs early but they
become abundant later as they are needed in notable quantities only for
construction and charcoal but almost nothing else. Rare wood and granite
remain precious the whole game since they are constantly consumed for
soldier equipment.

tothxa:
> Actually I'm not sure this rare tree kiln is very useful at all.

Yes, it's meant as somewhat of a gimmick. :) Though it may help a little by being faster, and it expands logically on the notion of Amazons knowing trees and charcoal burning better.

tothxa:
Here are the media source files (blender and gimp) for the media repo:

[rare_tree_kiln.zip](https://github.com/widelands/widelands/files/13245761/rare_tree_kiln.zip)
[charcoal_master.zip](https://github.com/widelands/widelands/files/13245770/charcoal_master.zip)
[charcoal_master_menu_gimp.zip](https://github.com/widelands/widelands/files/13245771/charcoal_master_menu_gimp.zip)

tothxa:
And the intermediate rendered images for the same:

[rare_tree_kiln_render.zip](https://github.com/widelands/widelands/files/13245778/rare_tree_kiln_render.zip)
(charcoal master zip file was not accepted by github due to size limit)

tothxa:
I think with the latest changes this is much better than current master (see https://www.widelands.org/forum/post/41401/), which isn't all that widely tested either. Could we have it for v1.2?

The posted on-demand rare-tree cutting patch should not be included though, because it can cause no longer required rare trees to block valuable growing places from the currently required ones. It could be considered for v1.3 with one or two extra program steps that cut any kind of rare tree randomly, but that would be best done by adding a `rare-tree` attribute for all ripe rare trees.

I'd still consider for v1.2 speeding up the rare tree kiln to the speed of the Frisian charcoal kiln now that I further increased the speed of the normal kiln, but I didn't want to push that yet in case somebody wants to watch my replays.

tothxa:
After some more testing, I realised that water is not much of a problem, because water gatherers are much faster than wells, so I reverted gold diggers back to master.

I committed the speed-up of the rare tree kiln mentioned in the previous comment, and I implemented on-demand rare tree cutting without changing the trees.

I also noticed during testing that the training glade often ran out of bread and cocoa, so I increased capacity of these.

So the latest adjustments are the result of my testing, also tested themselves.

hessenfarmer:
just a short feedback.

I will review this next week but this needs testing and I have to refamiliarize my self with the amazons.
so it will take some more time

hessenfarmer:
> [tothxa](https://codeberg.org/tothxa)_[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4468#issuecomment-13189371)_ _On Wed Apr 15 23:19:44 CEST 2026, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) wrote:_
> 
> > I don't think the current amount of quartz in master is enough to fix the balancing, but unfortunately it never got properly tested in a tournament.
> 
> The closest to that was <@>frankystone's [single player tournament](https://www.widelands.org/forum/post/42614/). I think the results support my assessment.

Well I still beleive the main critics were the rush capabilities, due to the Possibility to get to a level 7 without a food and mining economy. This is fixed by the quartz. 
For further balancing not only the needed wares do count but the overall time needed. Amazons really need more space for an efficient economy which does need some time. So in a normal multiplayer on e.g. Crater the other tribes are still faster in starting an economy.      
So I believe rather then kaputtniks tournament, where the the rally for space was critical anyway I beleive, that the Tempo training tournament of World Savior (https://www.widelands.org/forum/post/32071/) is more representative, where the results even for the version wiothout quartz were just in the midfield.       
So overall I am not in favor to the proposed changes (with the exception of the fix of the translator helptexts).      

However In am very busy in my profesional life and I have to argue a lot there so I simply do not have the energy to argue here as well. So although I am not in favor of it I have no strong feelings about it.

PR #6152, opened 2023-09-13
Revert and replace #6082, only use intermediate state for local player

tothxa:
Fix desync regression caused by #6082

### DIFF
--- a/src/logic/player.h
+++ b/src/logic/player.h
@@ -593,8 +593,8 @@ public:
 		return is_picking_custom_starting_position_;
 	}
 	bool get_starting_position_suitability(const Coords&) const;
-	bool is_setting_custom_starting_position() const {
-		return is_setting_custom_starting_position_;
+	bool local_player_starting_position_is_pending() const {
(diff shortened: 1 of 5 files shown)


src/logic/player.h
```diff
@@ -725,7 +725,9 @@ class Player {
 	uint8_t initialization_index_;
 	bool see_all_{false};
 	bool random_tribe_{false};
-	StartingPositionState starting_position_state_{StartingPositionState::kFinal};
+	bool is_picking_custom_starting_position_{false};
+	// Only for the local player
+	bool is_setting_custom_starting_position_{false};
```
Noordfrees: These names are too similar with only 4 out of 36 characters differing, so they can easily be confused with each other. Perhaps rename the second one `local_player_is_selecting_custom_starting_position_` or such.

Noordfrees (commented):
Tested and working, one comment inline.
Thanks for the quick patch

Noordfrees (approved):

PR #6153, opened 2023-09-16
Allow loading last savegame/replay/edited map from the command line

tothxa:
**Type of change**
Bugfix / Cleanup

**Issue(s) closed**
- Fixes #6138
- Fixes `ReplayLoader::is_valid_gametype()` and game type reporting for replay preloading

**Cleanups**
 - consistent use of filesystem directory name constants
 - use `as_font_tag()` shortcuts in `ui_fsmenu/main.cc`
 - use `switch()` in `WLApplication::run()`
 - use UI message boxes for more startup errors

**New behaviour**
Last replay can be started from main menu too. The Replays button is now a dropdown.

**Possible regressions**
File listing in load/save dialogs, file (pre)loading, startup

PR #6154, opened 2023-09-16
Refactor commandline handling

### LINKED ISSUE #6061: Command line options that don't use a value should warn when receiving one
**Describe the bug**
Some command line options can use a provided value (mostly to override config file options), but some don't. However, all options are placed in a `std::map` split on the "=" sign for processing, and options without a value are only `count()`ed, but it is not checked whether they received one. This can mislead users, e.g. the "--nozip" option overrides a config file option, but "--nozip=false" will actually disable zipping, instead of enabling it.

**To reproduce**
1. Open the Options Menu / Saving tab
2. Enable "Compress Widelands data files"
3. Choose "OK" to save settings
4. Quit Widelands
5. Start Widelands with "--nozip=false" (or actually any nonsense value)
6. Check the Options menu again
7. See error: compression is disabled

**Expected behavior**
Options that don't use the provided value should cause Widelands to quit with a warning like "Command line parameter 'x' does not accept a value", just like options that use values warn when they don't receive it.

**Version:**
 - Widelands Version: 1.2~git26254 (f0cd556@master)

tothxa:
**Type of change**
Bugfix / Refactoring

**Issue(s) closed**
Fixes #6061

It's no longer possible to use conflicting values for `--fullscreen` and `--maximized`. Old code only checked the presence of the options, not the actual values.

Non-option command line help entries are filtered in `wlapplication_messages.cc`. Didn't seem to cause any problems, but better to be safe.

Moved handling of settings that don't (seem to) affect localisation setup after it to allow localised error reporting for them.

**New behavior**
Helper functions are used to check whether a given option was given on the command line.

`--nozip` got a `--zip` counterpart, as the value is saved in the config.

`--fullscreen` and `--maximized` don't use a value any more.

**Possible regressions**
Command line handling, moved functions may affect / may be affected by localisation setup

**Additional context**
I'd also delegate datadir detection and window mode setting out to separate functions to make `handle_commandline_parameters()` more readable. But instead of adding even more one time use functions inside `WLApplication`, I think it would be better to move all command line handling into a separate new source file. The problem is that many values need passing to `WLApplication`, so this probably needs a new struct with those values and possibly as many globals as we can easily move there. But I'd rather not use member functions to set it up, because that wouldn't be much different from the current situation.

However that's another rather big change, so maybe it should be a separate PR after this one?


src/wlapplication.cc
```diff
@@ -1301,45 +1301,76 @@ void WLApplication::parse_commandline(int const argc, char const* const* const a
...
-	}
-	if (commandline_.count("nozip") != 0u) {
-		set_config_bool("nozip", true);
-		commandline_.erase("nozip");
 	}
-	if (commandline_.count("localedir") != 0u) {
-		localedir_ = commandline_["localedir"];
-		commandline_.erase("localedir");
+
+	if (auto localedir_option = get_commandline_option_value("localedir");
```
Noordfrees: Instead of so many `auto` types that all resolve to the same type maybe define `using OptionalParameter = std::optional<std::string>`

src/wlapplication.cc
```diff
@@ -1451,164 +1481,159 @@ void WLApplication::handle_commandline_parameters() {
...
-			filename_.erase(filename_.size() - 1);
+	const std::map<GameType, std::string> game_type_options = {
+		{ GameType::kEditor, "editor" },
+		{ GameType::kReplay, "replay" },
+		{ GameType::kFromTemplate, "new_game_from_template" },
+		{ GameType::kLoadGame, "loadgame" },
+		{ GameType::kScenario, "scenario" }
+	};
+
+	for (auto pair : game_type_options) {
```
Noordfrees: `const auto&`

also, `game_type_options` could be `static`

src/wlapplication.cc
```diff
@@ -1451,164 +1481,159 @@ void WLApplication::handle_commandline_parameters() {
...
 	}
-	if (commandline_.count("script") != 0u) {
-		script_to_run_ = commandline_["script"];
-		if (script_to_run_.empty()) {
-			throw_empty_value("--script");
-		}
+
+	if (auto val = get_commandline_option_value("script"); val.has_value()) {
+		script_to_run_ = *val;
 		if (*script_to_run_.rbegin() == '/') {
```
Noordfrees: Why `rbegin()` for read-only access? What's wrong with `back()`? :)

Noordfrees (commented):
a few comments, not tested

src/wlapplication.cc
tothxa: Oops. :) Though it comes from master actually.

klaus-halfmann:
@tothxa I could do a review but I see this is still a draft ...?

tothxa:
> I could do a review but I see this is still a draft ...?

This was meant to be merged after #6153, so it's a draft because there may be some merge conflicts.

bunnybot (commented):
Nordfriese*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4471#issuecomment-1557217)*
*On Fri Feb 09 18:36:24 CET 2024, **Benedikt Straub (Nordfriese)** **approved** this pull request:*

LGTM, tested and working

PR #6159, opened 2023-09-23
Do not handle_mousein in ware display

### LINKED ISSUE #6155: Multi Selection in Economy options window by dragging not working
**Describe the bug**
in previous versions it was possible to select multiple wares in the economy option window by dragging the mouse over them while mousebutton pressed. Even selection by spanning a virtual window was possible. 
currentl this does not work anymore and you need to select all wares by a single click.   

**To reproduce**
1. open economy options 
2. press mouse button over first weapon (top right corner)
3. pull mouse down while button pressed

**Expected behavior**
restore multiselection 
.

matthiakl:
**Type of change**
Bugfix 

**Issue(s) closed**
Fixes #6155

It turned out that moving the chat above all windows in #6127 messes with panel focusing

**New behavior**
The ware selection is not reset when the cursor leaves the window and so can handle the short loss of focus to the chat overlay. IMHO this is actually better than stopping the selection when outside the window. Only mouse release the selection is stopped.

**Possible regressions**
More windows having problems with focus loss.

### DIFF
--- a/src/wui/waresdisplay.cc
+++ b/src/wui/waresdisplay.cc
@@ -165,12 +165,6 @@ bool AbstractWaresDisplay::handle_mousemove(
 	return true;
 }
 
-void AbstractWaresDisplay::handle_mousein(bool inside) {
-	if (!inside) {
-		finalize_anchor_selection();
-	}
-}
-
 bool AbstractWaresDisplay::handle_mousepress(uint8_t btn, int32_t x, int32_t y) {
 	if (btn == SDL_BUTTON_LEFT) {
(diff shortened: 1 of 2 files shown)


bunnybot (commented):
Nordfriese*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4501#issuecomment-1240234)*
*On Sun Sep 24 09:38:19 CEST 2023, **Benedikt Straub (Nordfriese)** **approved** this pull request:*

<@>bunnybot merge

PR #6169, opened 2023-09-28
Fix heap-use-after-free in message preview drawing

### LINKED ISSUE #6168: ASan heap use after free in regression test suite run
**Describe the bug**
A single step in the testsuite run https://github.com/widelands/widelands/actions/runs/6267720185/job/17023177777?pr=6153 failed with an ASan heap use after free error. Does not look like it's related to the PR itself, it's more probably a race condition.

**To reproduce**
1. Look at the logs of https://github.com/widelands/widelands/actions/runs/6267720185/job/17023177777?pr=6153

**Expected behavior**
No random HUAF

**Additional context**
The log of the failed test run: [log.txt](https://github.com/widelands/widelands/files/12743757/log.txt)

Noordfrees:
**Type of change**
Bugfix

**Issue(s) closed**
Should fix #6168

**Additional context**
None-reproducible race condition

### DIFF
--- a/src/wui/info_panel.cc
+++ b/src/wui/info_panel.cc
@@ -66,12 +66,16 @@ inline bool MessagePreview::message_still_exists() const {
 }
 
 void MessagePreview::think() {
+	MutexLock m(MutexLock::ID::kMessages);
+
 	if (!message_still_exists() || SDL_GetTicks() - creation_time_ > kMessagePreviewMaxLifetime) {
 		owner_.pop_message(this);
 	}
 }
 
 void MessagePreview::draw(RenderTarget& r) {
+	MutexLock m(MutexLock::ID::kMessages);
+
 	if (!message_still_exists()) {
 		return;
 	}


tothxa (approved):
Looks reasonable.

BTW what happened to the mirror bot? Neither the bug, nor this PR got mirrored. Nor my updates to some of my PRs yesterday.

Noordfrees:
The bot occasionally falls over while checking for updated releases because GitHub likes to transiently send incomplete responses when querying the `/repos/{owner}/{repo}/releases` endpoint, so the bot sometimes thinks it has to mirror a newly created release when the target release in fact already exists, resulting in a 422 Unprocessable Entity error. I have some hope that I might finally have fixed the bug this morning, we'll see…

The bot sends me an e-mail with the error message whenever that happens so I know to restart it. This always takes 4½ hours though to read the cache file with state info for all issues, comments etc, which consists of more than a million lines of JSON (and that's already the shortened version; maybe I should pick a simpler format to speed up reading, and also writing which is 4 minutes per cycle). There are also some delays when there are new commits to mirror since the formatting script takes a few minutes. It's active again now.

@bunnybot merge

PR #6170, opened 2023-09-29
Add quartz to the Amazon economy

bunnybot:
[tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4512)*
*Created on Fri Sep 29 18:10:03 CEST 2023 by [**Tóth András (tothxa)**](https://codeberg.org/tothxa)*

---

**Type of change**
Balancing

**Issue(s) closed**
Lighter version of [CB#6150](https://codeberg.org/wl/widelands/pulls/4468) / [GH<#>6150](https://github.com/widelands/widelands/pull/6150)

**New behavior**
Amazons have quartz as a new mined ware that is used for many tools, advanced weapons and advanced buildings.

**Possible regressions**
Amazon economy

**Additional context**
Discussion thread in the forum: https://www.widelands.org/forum/topic/5847/

This will certainly make a difference, but according to my calculations it's far from enough to close the gap to the other tribes.


data/tribes/buildings/productionsites/amazons/stonemine/init.lua
```diff
@@ -46,38 +46,26 @@ wl.Descriptions():new_productionsite_type {
 
    programs = {
       main = {
-         -- TRANSLATORS: Completed/Skipped/Did not start mining granite because ...
-         descname = _("mining granite"),
+         -- TRANSLATORS: Completed/Skipped/Did not start mining granite and quartz because ...
+         descname = _("mining granite and quartz"),
```
hessenfarmer: please make these at least 2 subprograms again as the reason for having this structure was to ensure we get 2 attempts per ration which we do not have if the first attempt fails in the changed structure.

data/tribes/buildings/militarysites/amazons/fortress/init.lua
```diff
@@ -14,24 +14,28 @@ wl.Descriptions():new_militarysite_type {
...
+         granite = 1,
+         quartz = 1,
       }
    },
 
    buildcost = {
       ironwood = 6,
-      granite = 6,
+      granite = 4,
+      quartz = 2,
```
hessenfarmer: I would think we should have only 1 quartz max in the buildcost and may increase the enhancement costs:
buildcost 5 granite and 1 quartz; enhancement  no granite but 2 quartz     

Reason is not to overdo the requirment of quartz as it is really costly.   I just want to avoid to weaken the amazons too much.

data/tribes/buildings/militarysites/amazons/tower/init.lua
```diff
@@ -25,13 +26,13 @@ wl.Descriptions():new_militarysite_type {
 
    buildcost = {
       rope = 2,
-      granite = 2,
+      quartz = 2,
```
hessenfarmer:
```suggestion
     granite = 2,
```
we may require additional Ironwood but we should not require such a precious material for a tower, as we do not require iron for example for the other tribes towers.

data/tribes/buildings/productionsites/amazons/stone_workshop/init.lua
```diff
@@ -77,10 +78,12 @@ wl.Descriptions():new_productionsite_type {
             "call=produce_spear_stone_tipped",
             "call=produce_stone_bowl",
             "call=produce_spear_hardened",
+            "call=produce_firestones",
+            "call=produce_spear_wooden",
```
hessenfarmer: why did you add an additional wooden spear cycle?
additionally we need to adjust the time consumed comment as well.
many thanks for fixing the translator comments below; very good spot

data/tribes/buildings/productionsites/amazons/stone_workshop/init.lua
```diff
@@ -48,7 +48,8 @@ wl.Descriptions():new_productionsite_type {
    },
 
    inputs = {
-      { name = "granite", amount = 7 },
+      { name = "granite", amount = 5 },
```
hessenfarmer:
```suggestion
      { name = "granite", amount = 4 },
```
should be enough for the reduced needs

data/tribes/buildings/productionsites/amazons/stone_workshop/init.lua
```diff
@@ -77,10 +78,12 @@ wl.Descriptions():new_productionsite_type {
             "call=produce_spear_stone_tipped",
             "call=produce_stone_bowl",
             "call=produce_spear_hardened",
+            "call=produce_firestones",
+            "call=produce_spear_wooden",
          },
       },
       produce_shovel = {
-         -- TRANSLATORS: Completed/Skipped/Did not start forging a felling ax because ...
+         -- TRANSLATORS: Completed/Skipped/Did not start making a shovel because ...
```
hessenfarmer, line 85: we may require an additional log for the shovel so ironwood makes the blade and log makes the handle

data/tribes/initialization/amazons/units.lua
```diff
@@ -584,7 +608,7 @@ wl.Descriptions():new_tribe {
             name = "amazons_stonecutter",
             helptexts = {
                -- TRANSLATORS: Helptext for an Amazon worker: Stonecutter
-               purpose = pgettext("amazons_worker", "Cuts blocks of granite out of rocks in the vicinity.")
+               purpose = pgettext("amazons_worker", "Cuts blocks of granite and quartz out of rocks in the vicinity.")
```
hessenfarmer: needs to be differentiated for stonecutters and stonemines

data/tribes/initialization/amazons/units.lua
```diff
@@ -301,7 +314,7 @@ wl.Descriptions():new_tribe {
             helptexts = {
                purpose = {
                   -- TRANSLATORS: Helptext for an Amazon ware: Pick
-                  pgettext("amazons_ware", "Picks are used by stonecutters to cut granite from rocks or in mines.")
+                  pgettext("amazons_ware", "Picks are used by stonecutters to cut granite and quartz from rocks or in mines.")
```
hessenfarmer: as quartz can only be mined and should only be mined this teyt needs to reflect this

hessenfarmer (commented):
some inliners     
seems to be ok for me generally 
to do:
remaining start conditions
will do some testing for poor hamlet how much quartz is really needed
other missing conditions may be calculated from this then

data/tribes/buildings/productionsites/amazons/stonemine/init.lua
tothxa: Oops, didn't think of that. Fixed on CB. (mirrorbot's stopped again…)

data/tribes/buildings/militarysites/amazons/fortress/init.lua
tothxa: I don't think it's that expensive, but done on CB.

data/tribes/buildings/productionsites/amazons/stone_workshop/init.lua
tothxa: I didn't want to break the rythm. But the performance text would be too complicated, so removed the 4th wooden spear. (on CB)

data/tribes/buildings/productionsites/amazons/stone_workshop/init.lua
tothxa: It was for symmetry with ironwood. But I reviewed the actual needs of a full cycle and it turns out more ironwood is needed, so adjusted all.

data/tribes/initialization/amazons/units.lua
tothxa: Done on CB.

data/tribes/initialization/amazons/units.lua
tothxa: Done on CB.

data/tribes/buildings/productionsites/amazons/stonemine/init.lua
```diff
@@ -46,38 +47,49 @@ wl.Descriptions():new_productionsite_type {
 
    programs = {
       main = {
-         -- TRANSLATORS: Completed/Skipped/Did not start mining granite because ...
-         descname = _("mining granite"),
+         -- TRANSLATORS: Completed/Skipped/Did not start mining because ...
+         descname = _("mining"),
          actions = {
+            -- Total time: 28.8 + 17 + 21 + 2 * 3.6 = 74
```
hessenfarmer, line 53: the 28 do not match the 23 from the program

hessenfarmer (commented):
lgtm now.
only the start condition need to be fixed.   
as we now need quartz instead of granite for a lot of tools, we may reduce granite by one third I believe.

hessenfarmer:
ok I tried poor hamlet and we need to add just one tiny quartz tom the start condition for the charcoal kiln. 
However it is very tricky to start from this.

tothxa:
Thank you very much for the help with the starting conditions.

> ok I tried poor hamlet and we need to add just one tiny quartz tom the start condition for the charcoal kiln. However it is very tricky to start from this.

Doesn't it also need one rope for the jungle preserver? (or lots of trees for logs for the liana cutter and the rope weaver. And I suppose more rocks for granite are supposed to be conquered, otherwise a lot more would be necessary.)

But this is real evil indeed with the requirement to convert your wood economy to rare trees before really being able to proceed. And having to send the liana cutter to weave rope. I'm not sure about other tribes' poor hamlets, but I think it's almost impossible to not generate a request for some other tool that would steal that single quartz before the stone workshop and the charcoal kiln are ready.

hessenfarmer (approved):
can go in from my side

hessenfarmer:
@bunnybot merge

PR #6172, opened 2023-09-30
Fix lobby game name editbox horizontal scrolling

### LINKED ISSUE #6130: game title box in multiplayer is not correctly layouted
**Describe the bug**
in the Multiplayersetup screen the box with the game title to be published on the meta server is not showing the text from the beginning, if the previous and remembered title was long. see screenshot
grafik

jtorruellas22:

**Type of change**
Bug Fix

**Issue(s) closed**
Fixes  #6130 (game title box in multiplayer is not correctly layouted)

**To reproduce**
*If it's a bugfix:*
1. Go to config file
2. Edit servername to something long
3. Save config and load the game
4.  Open the multiplayer menu and view text

**New behavior**
Changed layout() in textinput.cc to include a new offset to make sure saved text is showed from the beginning regardless of window size.

**How it works**
Scroll offset is calculated based on the ratio of the length of text and the height of the window. 

**Screenshots**
image

### DIFF
--- a/src/ui_basic/textinput.cc
+++ b/src/ui_basic/textinput.cc
@@ -207,6 +207,8 @@ void AbstractTextInputPanel::Data::draw(RenderTarget& dst, bool with_caret) {
 }
 
 void AbstractTextInputPanel::layout() {
+	// Offset snaps the text panel to the start of input. Saved text always displays from beginning
+	d_->scrolloffset = get_text().length() / get_h();
 	Panel::layout();
 	d_->scrollbar.set_pos(Vector2i(get_w() - Scrollbar::kSize, 0));
 }
--- a/src/ui_basic/box.cc
+++ b/src/ui_basic/box.cc
@@ -178,7 +178,7 @@ void Box::layout() {
 	int totaldepth = 0;
 	int spacing = -1 * inner_spacing_;
 	int infspace_count = 0;
-
+	
 	for (size_t idx = 0; idx < items_.size(); ++idx) {
 		int depth;
 		int unused = 0;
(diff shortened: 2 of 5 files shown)


src/ui_basic/textinput.cc
```diff
@@ -207,6 +207,8 @@ void AbstractTextInputPanel::Data::draw(RenderTarget& dst, bool with_caret) {
 }
 
 void AbstractTextInputPanel::layout() {
+	// Offset snaps the text panel to the start of input. Saved text always displays from beginning
+	d_->scrolloffset = get_text().length() / get_h();
```
Noordfrees, line 211: Since `scrolloffset` is about horizontal scrolling I don't understand why you divide by height here, can you explain?

Noordfrees (commented):
Hi @jtorruellas22 and thanks for the patch :)

One question inline. 

It seems that you accidentally committed the binaries, please remove them from the branch again (`git rm *.exe`). That's an oversight in the `.gitignore`.

src/ui_basic/textinput.cc
jtorruellas22: Not entirely sure why but the height adjust helps the text appear consistently across all window sizes. Just using the length of the string seems to show the text in almost the midpoint of the string. Using the width dimension causes the game to crash when attempting to load the multiplayer menu. The goal was to use number logic that would remain consistent.

jtorruellas22: I believe the binaries have been remove correctly now

Noordfrees (approved):
LGTM, tested and working. Thanks for the patch :) 

@bunnybot merge

PR #6177, opened 2023-10-06
Rename brotli static dependency

### LINKED ISSUE #6175: Workflow msvc failure: cmake doesn't find brotlidec-static
**Describe the bug**
msvc build in today's workflow build of master failed because of cmake error on missing brotlidec-static.

**To reproduce**
See https://github.com/widelands/widelands/actions/runs/6427093985/job/17457822507

**Expected behavior**
Build succeeds.

**Version:**
 - OS: w*ndows runner image
 - Widelands Version: https://github.com/widelands/widelands/commit/fd8328649b85a3e00e586a370c1b23f13a3db258

**Additional context**
@matthiakl 
Caused by last vcpkg update? If so, we should include a test build of Widelands in the vcpkg update workflow.

Also, why is brotli even needed on w*ndows if not on other platforms?

matthiakl:
**Type of change**
Bugfix

**Issue(s) closed**
Fixes #6175 

**New behavior**
Building successfully again

**How it works**
Rename brotli library caused by update 1.0.9 to 1.1.0


tothxa:
Thank you.

As far as I'm concerned, we can have this if the workflow succeds. Probably most users also use the pre-built images, and those who build for themselves can be told to update their vcpkg setup (but I don't know how it is typically used).

However, if it were for a platform that I actually use, I'd much prefer making it able to detect both the old and the new version for some transitional period.

@hessenfarmer What do you think about that?

tothxa (approved):
Seems to work. :)

tothxa:
I guess it's most important to fix the workflow.

Hmm, let's see how this works…

@bunnybot merge

tothxa:
@Noordfrees The "update branch" button is gone. Is that part of the stronger approval requirement setting?

Maybe bunnybot should handle this on its own though: if the branch is outdated by the time of the merge command, it should try to update the PR branch first, wait for the workflow checks, and merge on success.

It should also get a standalone "update" command to get back that functionality.

tothxa (approved):

tothxa:
@bunnybot merge

Didn't it say it'll do it?

PR #6147, opened 2023-09-10
Update msvc toolset version

tothxa:
**Type of change**
Dependency update

**Issue(s) closed**
This months's (we're getting better ;) msvc build failure, hopefully

**Additional context**
Thanks to @matthiakl for hunting this down the last time and for documenting where to look it up when it needs updating.

Too bad we're stuck with workarounds requiring manual updates…

PR #6181, opened 2023-10-09
Improve vcpkg update check

matthiakl:
**Type of change**
CI improvement

**Issue(s) closed**
As requested in #6175 and #6147

**New behavior**
The weekly vcpkg update check runs a test compilation with the new package versions.
It now also checks for toolset updates and bumps the version if it works fine.

**How it works**
Parse the toolset json using jq and filter out the newset version number to test.

**Additional context**
Check a test run here: https://github.com/matthiakl/widelands/actions/runs/6452825677


.github/workflows/check_vcpkg.yaml
```diff
@@ -36,30 +37,50 @@ jobs:
...
+        toolset: ${{ steps.prepare.outputs.toolset }}
+    - name: Compiler
+      id: build
+      if: steps.lookup.outputs.cache-hit != 'true'
+      env:
+        VCPKG_ROOT: C:\vcpkg
+      run: |
+        mkdir $env:GITHUB_WORKSPACE\build
+        cd $env:GITHUB_WORKSPACE\build
+        cmake.exe -G "NMake Makefiles" .. -DVCPKG_TARGET_TRIPLET=${{ env.VCPKG_TARGET_TRIPLET }} -DCMAKE_TOOLCHAIN_FILE="${{ env.VCPKG_ROOT }}/scripts/buildsystems/vcpkg.cmake" -DCMAKE_BUILD_TYPE=${{ matrix.config }} -DOPTION_BUILD_WEBSITE_TOOLS=OFF -DOPTION_BUILD_TRANSLATIONS=ON -DOPTION_BUILD_TESTS=ON -DOPTION_ASAN=OFF -DOPTION_BUILD_CODECHECK=OFF -DOPTION_BUILD_WINSTATIC=ON -DOPTION_USE_GLBINDING=ON -DOPTION_FORCE_EMBEDDED_MINIZIP=ON
```
tothxa, line 76: Looks reasonable. I reviewed most of it (even the `jq` command :) except for this line. But instead of duplicating from the main workflow, it would be better to move this into a helper script. That would make sure it never gets out of sync.

tothxa:
Sorry, I forgot we're dealing with w*ndows here, so scripting is much harder...

Can't you put the `run` stage in a separate `yaml` file, and include it in both workflows?

matthiakl:
> Can't you put the compiler stage in a separate yaml file, and include it in both workflows?

I will try that. I don't know what is screwed with the script, even in the last commit the packages are not found, but the command should run directly in the workflows context...

tothxa:
Hmm, you're only running cmake from the script, and only want to pass 3 values. Can't they be passed as command line arguments to the script?

matthiakl:
No, I do not think the variables are the problem (see https://github.com/widelands/widelands/actions/runs/6460924253/job/17539657709#step:7:66).
I assume the environment setup done by the msvc dev console step is not properly passed to the called script.

tothxa:
OK then. I see it's not even possible to break up long command lines... Sigh...

BTW can we test the actual vcpkg update part?

tothxa:
OTOH this workflow also calls `install_dependencies.sh`, so `bash` is installed at one point, isn't it? Can't you put it in a `bash` script instead of powershell?

tothxa:
Whee! At least in the revert commit diff I could easily compare the 2 command lines. :)

matthiakl:
> BTW can we test the actual vcpkg update part?

I linked a test run in PR description, you can find the test runs here: https://github.com/matthiakl/widelands/actions/workflows/check_vcpkg.yaml
The workflow can be launched from the web UI by selecting the corresponding branch.

matthiakl:
> Can't you put it in a bash script instead of powershell?

No, because msvc requires the development console setup which is bound to powershell. Scripting on windows is no fun...

tothxa (approved):
Thank you and sorry for all the trouble I made.

matthiakl:
@bunnybot merge

Issue #6176, opened 2023-10-06
Lua Coroutine failed


spirillen:
Please fill out the relevant sections below and delete the rest.

**Describe the bug**
![image](https://github.com/widelands/widelands/assets/44526987/2a36be20-8bf5-4ff9-a0e2-75221c567927)

**To reproduce**
1. Start new random game
2. ![image](https://github.com/widelands/widelands/assets/44526987/c346301e-383f-4113-b018-a156d68717e3)
3. 
![image](https://github.com/widelands/widelands/assets/44526987/493dccfd-3d72-40a0-a92c-8948c738abe6)

**Expected behavior**
Hmm let's see... :thinking: 

**Screenshots**

**Version:**
 - OS: "Ubuntu 22.04"
 - widelands --version
This is Widelands version 1.2~bzr10781-202310060346 Release
[00:00:00.000 real] INFO: Set home directory: /home/spirillen/.local/share/widelands
[00:00:00.000 real] INFO: Set configuration file: /home/spirillen/.config/widelands/config
[00:00:00.000 real] WARNING: No locale translations found in /usr/share/games/widelands/data/locale
[00:00:00.000 real] INFO: selected language: (system language)
[00:00:00.000 real] INFO: using locale en_DK.UTF-8
=========================================================================
This is Widelands version 1.2~bzr10781-202310060346 Release

tothxa:
Oops. We should have properly tested all starting conditions for #6170…

spirillen:
> Oops. We should have properly tested all starting conditions for #6170…
> 
> Thanks for reporting. :)

Can see you added the `amazon` tag, this happened for two others as well, please just don't ask me to remember which from yesterday...

tothxa:
> Can see you added the `amazon` tag, this happened for two others as well, please just don't ask me to remember which from yesterday...

My guess is that one of the computer players must have had Amazons with one of the affected starting conditions (trading outpost, village, fortified village). I've tested it and I got the same message if it was another player. I also tested trading outpost and village of all other tribes (that's what fits on a 8 player map), and it started without any problem.

frankystone:
If i remember correctly there were similar issues with the starting conditions in the past. Maybe there should be a test for this kind of issues? Don't know if it's feasible though.

tothxa:
> If i remember correctly there were similar issues with the starting conditions in the past. Maybe there should be a test for this kind of issues? Don't know if it's feasible though.

[...]

PR #6182, opened 2023-10-10
Regression tests for start and win conditions, fix race regressions

### LINKED ISSUE #6327: Deadlock when expedition is created while port window is open and verbose logging enabled
tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/issues/4690)*
*Created on Sat Jan 27 00:13:13 CET 2024 by **Tóth András (tothxa)***

---

**Describe the bug**
Widelands hung when 2 ships arrived to a port where an expedition was ready to embark. IIRC the port window was open, and maybe a Ship Fleet Options window for a nearby shipyard too. The game was running at about 5x speed.

I sent `sigsegv` manually to obtain backtraces, and I was fortunate enough that the backtraces were created at different times. (see #6325) (I actually had to send the signal twice, that may have helped.)

As far as I can tell from them, the UI thread was trying to obtain a mutex lock while creating the soldier list for a warehouse window (maybe the port window wasn't open, but I was just opening it?), while the logic thread was creating a `NoteThreadSafeFunction` for a `BuildingNote` (presumably to annonce the expedition).

**To reproduce**
Don't know. Tried to re-create the above situation, but the expedition was successfully created, and the port window updated.

<details><summary>

**Console log**

</summary>

```
[03:26:02.645 game] DEBUG: MO(1692,empire_ship): Updating existing soldier request at Apulum
[03:26:02.645 game] DEBUG: MO(1497,empire_ship): Updating existing soldier request at Apulum
[03:26:02.690 game] DEBUG: MO(621,barbarians_ranger):   field no longer free
[03:26:02.717 game] DEBUG: MO(1217,barbarians_ship): Updating existing soldier request at Ul’nsaker
[03:26:02.729 game] DEBUG: MO(221,barbarians_soldier): [transfer]: starting task [movepath] and setting location to road 3916
[03:26:02.787 game] DEBUG: MO(1925,barbarians_ship): Updating existing soldier request at Al’ver
[03:26:02.788 game] DEBUG: MO(1646,empire_ship): Updating existing soldier request at Forum Hadriani
[03:26:02.810 game] DEBUG: MO(1235,empire_ship): Updating existing soldier request at Sostra
[03:26:02.810 game] DEBUG: MO(1355,empire_ship): Updating existing soldier request at Sostra
[03:26:02.892 game] DEBUG: MO(1084,barbarians_ship): Updating existing soldier request at L’dingen
[03:26:02.908 game] DEBUG: MO(1881,empire_ship): Updating existing soldier request at Sopiane
[03:26:02.908 game] DEBUG: MO(1211,empire_ship): Updating existing soldier request at Sopiane
[03:26:02.915 game] DEBUG: MO(1692,empire_ship): Updating existing soldier request at Apulum
[03:26:02.915 game] DEBUG: MO(1497,empire_ship): Updating existing soldier request at Apulum
[03:26:02.987 game] DEBUG: MO(1217,barbarians_ship): Updating existing soldier request at Ul’nsaker
[03:26:03.057 game] DEBUG: MO(1925,barbarians_ship): Updating existing soldier request at Al’ver
[03:26:03.058 game] DEBUG: MO(1646,empire_ship): Updating existing soldier request at Forum Hadriani
[03:26:03.080 game] DEBUG: MO(1235,empire_ship): Updating existing soldier request at Sostra
[03:26:03.080 game] DEBUG: MO(1355,empire_ship): Updating existing soldier request at Sostra
[00:08:13.135 real] DEBUG: MO(11794,log): cancel_moving
[03:26:03.144 game] DEBUG: Player: 5; Jobs: 1; delay: 268; gamespeed: 4000 
[03:26:03.144 game] DEBUG: [AI 5] flags count: 26, avg:   6
[03:26:03.144 game] DEBUG: Player: 6; Jobs: 1; delay: 496; gamespeed: 4000 
[03:26:03.162 game] DEBUG: MO(1084,barbarians_ship): Updating existing soldier request at L’dingen
[03:26:03.178 game] DEBUG: MO(1881,empire_ship): Updating existing soldier request at Sopiane
[03:26:03.178 game] DEBUG: MO(1211,empire_ship): Updating existing soldier request at Sopiane
[03:26:03.185 game] DEBUG: MO(1692,empire_ship): Updating existing soldier request at Apulum
[03:26:03.185 game] DEBUG: MO(1497,empire_ship): Updating existing soldier request at Apulum
[03:26:03.193 game] DEBUG: MO(1646,empire_ship): Updating existing soldier request at Forum Hadriani
[03:26:03.257 game] DEBUG: MO(1217,barbarians_ship): Updating existing soldier request at Ul’nsaker
[03:26:03.300 game] DEBUG: MO(297,empire_carrier): [waitforcapacity]: Wake up: flag capacity.
[03:26:03.327 game] DEBUG: MO(1925,barbarians_ship): Updating existing soldier request at Al’ver
[00:08:13.214 real] DEBUG: MO(11656,coal): cancel_moving
[03:26:03.350 game] DEBUG: MO(1235,empire_ship): Updating existing soldier request at Sostra
[03:26:03.350 game] DEBUG: MO(1355,empire_ship): Updating existing soldier request at Sostra
[00:08:13.265 real] DEBUG: MO(11649,fish): cancel_moving
[03:26:03.432 game] DEBUG: MO(1084,barbarians_ship): Updating existing soldier request at L’dingen
[03:26:03.448 game] DEBUG: MO(1881,empire_ship): Updating existing soldier request at Sopiane
[03:26:03.448 game] DEBUG: MO(1211,empire_ship): Updating existing soldier request at Sopiane
[03:26:03.455 game] DEBUG: MO(1692,empire_ship): Updating existing soldier request at Apulum
[03:26:03.455 game] DEBUG: MO(1497,empire_ship): Updating existing soldier request at Apulum
[03:26:03.463 game] DEBUG: MO(1646,empire_ship): Updating existing soldier request at Forum Hadriani
[03:26:03.527 game] DEBUG: MO(1217,barbarians_ship): Updating existing soldier request at Ul’nsaker
[03:26:03.528 game] DEBUG: MO(1340,empire_ship): ship_update: Arrived at dock 980
[03:26:03.528 game] DEBUG: [Shipping Schedule] 
[03:26:03.528 game] DEBUG: ShippingSchedule::ship_arrived(Azurea Sea at 980 Forum Hadriani)
[03:26:03.528 game] DEBUG: [Shipping Schedule] Loading expedition

##############################
FATAL ERROR: Received signal 11 (Segmentation fault)
...

```

</details>

**Backtraces**
( The game didn't crash, it hung. `sigsegv` was sent manually.)

Logic thread:
[2024-01-26T23.04.51.txt](https://codeberg.org/attachments/3a34e51a-66f9-43a4-af9d-4022689ed836)

UI thread:
[2024-01-26T23.05.05.txt](https://codeberg.org/attachments/ae8a6ae8-830a-46ef-8e5a-02cc7d9c0c16)

**Savegames**
(trying to recreate the situation from an earlier save of the hung game)
[deadlock.zip](https://codeberg.org/attachments/eb6c2f11-23ae-492f-95a3-2f27524f5957)

**Version:**
 - Widelands Version: naval-warfare? (navalwarfare-ai, but don't seem to be related)

tothxa:
**Type of change**
New feature / Bugfix

**Issue(s) closed**
Re https://github.com/widelands/widelands/issues/6176#issuecomment-1751738993

Found and fixed some race condition regressions in master

Fixes #6327

**New behavior**
There are now regression tests for all starting and win conditions, and also for the completeness of these.
Incidentally, `--new_game_from_template` is tested now, hopefully with all possible game options used at least once.

Most Lua calls are now protected by the Objects and Lua locks. The order is important, see added TODO comments and testsuite locking failures like https://github.com/widelands/widelands/actions/runs/7886026322/job/21518487518

Lock debug logging is done by direct output when acquiring a lock for the logger.

**Possible regressions**
Workflow failures (, locking?)


tothxa:
Hmm, didn't work the same as for me locally...

tothxa:
OK, so it was that simulated merge-into-master thing. Now it was only the expected failure.

tothxa:
It should be now good then.

tothxa:
It succeeded again, but I'm not sure how it will behave the next time under stress. I really don't even understand how lua, its coroutines, and particularly `sleep()` work when the game is paused by the game summary window. But the cause of many timeout failures seemed to be that the check coroutine didn't run after the summary window had been opened.

tothxa:
OK, so I have to solve the problem of reliably run a script while the game is paused or when the game summary window is opened. I'm not sure which would be cleaner.

frankystone:
Maybe a stupid question: Isn't it possible to have a switch `show_game_result` which defaults to true but for these tests it can be set to false? And if it is `false` then don't pause the game, show the result in the console and quit immediately?

Show result in console may can help to automate AI-training.

tothxa:
> Maybe a stupid question: Isn't it possible to have a switch `show_game_result` which defaults to true but for these tests it can be set to false? And if it is `false` then don't pause the game, show the result in the console and quit immediately?

(Verbose) Logging end result reporting and setting as well as the end of the game is definitely a good idea. It's currently only done in the game host, but it should be added to the client and singleplayer too.

As for the tests, since I've finally managed to make them work reliably, I'd stick to my current approach because I think it's better to have tests that don't need active co-operation from the tested functions.

And the realtime tasks may come in handy for other uses too. Most notably the new UI plugin scripts, because actually, contrary to my code comments, my testing showed that for scripts run from the debug console, it's the old (gametime) sleep that causes an immediate desync even on a `print("hello")` (I guess because resuming coroutines is written to the syncstream), while the new realtime sleep works just fine. (obviously as long as the script doesn't manipulate the game state)

> Show result in console may can help to automate AI-training.

Yes, for that use case we may need an option for game summary to exit instead of pause, in addition to the above mentioned logging.

utils/check_template_test_coverage.py
```diff
@@ -0,0 +1,143 @@
+#!/usr/bin/env python
+# encoding: utf-8
+
+from glob import glob
+import os.path
+import re
+import sys
+
+import iniparse
```
SimonHeimberg: `import configparser` from core python works. Iniparse is very handy for writing, which we do not do. (And some other things.)

utils/check_template_test_coverage.py
```diff
@@ -0,0 +1,143 @@
...
+    sys.exit(1)
+
+unknown_wc = []
+unknown_tribe = []
+unknown_sc = dict()
+
+# Collect uses
+for test in tests:
+    with open(test, 'r', encoding = 'utf-8') as fh:
+        cfg = iniparse.INIConfig(fh)
```
SimonHeimberg: instead of with open(...
```
cfg = configparser.ConfigParser()
cfg.read(test)
```

utils/check_template_test_coverage.py
```diff
@@ -0,0 +1,143 @@
...
+
+    for s in cfg:
+        section = cfg[s]
+
+        # Test's win condition
+        if s == 'global':
+            if not section.win_condition in win_conditions:
```
SimonHeimberg: The only change for `configparser` is that all instances of `section.xxx` must be changed to `section["xxx"]`. Should not harm much.

.github/workflows/build.yaml
```diff
@@ -31,6 +31,10 @@ jobs:
         pushd src
         ../utils/find_unused_includes.py
         popd
+    - name: Check regression test coverage of starting and win conditions
+      run: |
+        pip install iniparse
```
SimonHeimberg: iniparse can be replaced by configparser from core python. See comments in [utils/check_template_test_coverage.py](https://github.com/widelands/widelands/pull/6182/files#diff-8fd256175994905417783fda6ceb202031cea5eaca92ff8df52340ce29a388e1)

regression_test.py
```diff
@@ -269,6 +269,18 @@ def discover_scenario_tests(regexp, suite):
                     WidelandsTestCase(test_script,
                         scenario=wlmap, script=test_script))
 
+def discover_game_template_tests(regexp, suite):
+    """Add all tests using --new_game_from_template to the 'suite'."""
+    for templ in sorted(glob(os.path.join("test", "templates", "test*.wgt"))):
+        test_script = templ[:-3] + 'lua'
+        if not os.path.isfile(templ) or not os.path.isfile(test_script):
+            continue
```
SimonHeimberg, line 288: Maybe having a template but no test file should at least print a warning.

test/templates/test_struggling_outpost-wood_gnome.lua
```diff
@@ -0,0 +1,35 @@
...
+run(function()
+  sleep(1000)
+  sf = p:get_buildings("empire_headquarters")[1].fields[1]
+
+  local ff = {}
+  for d = 1,6 do
+    ff[d] = sf
+  end
+
+  for s = 1, 6 do
```
SimonHeimberg, line 19: :thinking: Ah, this goes 6 steps in all 6 directions. A comment would make this more obvious.

SimonHeimberg: or maybe
```
for i, dir in ipairs({"ln", "tln", "trn", "rn", "brn", "bln"} do
   f = sf[dir][dir][dir][dir][dir][dir]
   print(_("Placing forester ...
```

test/templates/test_discovery-territorial_lord.wgt
```diff
@@ -0,0 +1,59 @@
...
+win_condition="territorial_lord.lua"
+
+# Territorial lord is not time based, but we can use this in the script for the timeout.
+win_condition_duration="3"
+
+# Skip 20 minute delay
+peaceful="true"
+
+fogless="false"
+custom_starting_positions="false"
```
SimonHeimberg, line 15: above are default values. Should we include them in each file?

test/templates/test_discovery-territorial_lord.wgt
```diff
@@ -0,0 +1,59 @@
...
+
+[player_1]
+tribe="amazons"
+init="discovery.lua"
+team="0"
+ai="empty"
+closed="false"
```
SimonHeimberg, line 23: team="0" and closed="false" are default values. Should they be included in each file for each player?

test/templates/artifacts.wmf/elemental
```diff
@@ -0,0 +1,13 @@
+# Automatically created by Widelands 1.2~git26321 (4b2f93f@master) Release
+
+[global]
+packet_version="1"
+map_w="160"
+map_h="160"
+nr_players="8"
+name=_"Artifacts test"
+author="tothxa"
+descr=_"Test map for the Artifacts win condition"
```
SimonHeimberg:
```suggestion
descr=_"Test map for the Artifacts win condition for 8 players with 1 artifact"
```

test/templates/test_fortified_village-artifacts.wgt
```diff
@@ -0,0 +1,59 @@
...
+fogless="false"
+custom_starting_positions="false"
+
+[player_1]
+tribe="amazons"
+init="fortified_village.lua"
+team="0"
+ai="normal"
+closed="false"
+playercolor="0 255 255"
```
SimonHeimberg, line 21: Why is this set? Could be checked in test.

test/templates/test_village-hq_hunter.wgt
```diff
@@ -0,0 +1,61 @@
...
+closed="false"
+
+[player_2]
+tribe="atlanteans"
+init="village.lua"
+team="0"
+ai="empty"
+closed="false"
+
+# test reporting with closed player slots in the middle
```
SimonHeimberg, line 30: :+1:

SimonHeimberg:
Additional idea:
- one test could have `custom_starting_positions="false"` to check that the AI selects a position. Currently the AI just selects the default position, but the test should not care about this. test_headquarters-defeat_all.wgt looks like a good candidate.
- one test (or a few) could check if the number and settings of players is as expected. This would check functionality of  `--new_game_from_template`.
- one test could set random="true" (and ai=empty, ...) for an additional player and check if it is generated.

test/templates/test_fortified_village-artifacts.wgt
tothxa: It's set to test the parser code. :) But also testing it is a good idea. (Do we have lua interface for it?)

SimonHeimberg:
> > Why is the map not in test/maps/ ? As long as there is no test/maps/XXX.wmf/scripting/test_*.lua, no other test will be created for it.
> 
> I think that would be confusing. But I made the map initially in the hope that the AI would conquer the artifact on its own, but it turned out it fails more often than not, so now the test script conquers it. So I can look for a suitable standard map instead.
> 
I doubt there is an existing map with at least 5 players and one artifact only.

utils/check_template_test_coverage.py
tothxa: sigh... `iniparse` is very much nicer to work with in general too...

tothxa:
>     * one test could have `custom_starting_positions="false"` to check that the AI selects a position.

Trading outpost / collectors uses custom starting positions, the rest don't.

>     * one test (or a few) could check if the number and settings of players is as expected. This would check functionality of  `--new_game_from_template`.

I think the number is already tested when winners and losers are checked. I'll have a look into what settings we can check.
 
>     * one test could set random="true" (and ai=empty, ...) for an additional player and check if it is generated.

OK.

regression_test.py
tothxa: I added a check for this in the coverage test script, because without a script the test won't end. Is that OK for you?

test/templates/test_discovery-territorial_lord.wgt
tothxa: I think it's best to have a controlled test environment for the "real" tests. But I added a test that leaves everything to the defaults. Will that do?

test/templates/test_discovery-territorial_lord.wgt
tothxa: (see above)

test/templates/test_fortified_village-artifacts.wgt
tothxa: No, we don't. The current lua interface can only report the default playercolour.

tothxa:
> one test could set random="true" (and ai=empty, ...) for an additional player and check if it is generated.

Done.

SimonHeimberg:
> > one test could set random="true" (and ai=empty, ...) for an additional player and check if it is generated.
> 
> Done, except for the check. How should I check it?

I made a pr to this branch with a suggestion how to test the player with a random tribe.
And some more pr with other suggestions

tothxa:
@SimonHeimberg I took over from your suggestions what I agree with. Thank you for them, some of them were major improvements.

Sorry about the rest.

tothxa:
I believe the latest failures are due to a regression in master. Most likely introduced by #6153, because the failures only happen in the template tests and that has been the only PR touching the template startup code since the previous successes. (or actually in a very long time?)

tothxa:
> I believe the latest failures are due to a regression in master. Most likely introduced by #6153

I was completely wrong. The failures started after the naval warfare merge, but I didn't notice because I expected failure due to the macos issue. That's much harder to track then.

src/base/multithreading.cc
```diff
@@ -230,7 +232,7 @@ MutexLock::MutexLock(const ID i) : id_(i) {
 
 	if (record.current_owner != kNoThread) {
 		for (const auto& pair : acting_as_another_thread) {
-			if (pair.first == self && pair.second == record.current_owner) {
+			if (pair.first == self && pair.second == record.current_owner && id_ != ID::kLog) {
```
bunnybot: Nordfriese*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4524#issuecomment-1561030)*
*On Mon Feb 12 15:40:36 CET 2024, **Benedikt Straub (Nordfriese)** wrote:*

---

This one looks dangerous since the new condition skips more than just a log. The other multithreading changes look OK.

Skipping the log output is not great for debugging, but using printf directly to circumvent the mutexes will mess up the output which is just as bad. Perhaps we could instead remove `kLog` and instead use a regular `std::mutex` and `std::lock_guard` in the logging implementation function `do_log`?

bunnybot (commented):
Nordfriese*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4524#issuecomment-1561032)*
*On Mon Feb 12 15:35:31 CET 2024, **Benedikt Straub (Nordfriese)** commented.*

bunnybot (commented):
Nordfriese*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4524#issuecomment-1563388)*
*On Tue Feb 13 20:52:39 CET 2024, **Benedikt Straub (Nordfriese)** **requested changes** on this pull request:*

I don't like the change to the timers. With this change, timer frequency is affected by the logic frame rate, which can vary greatly, rather than the UI frame rate, which should be more or less constant.
I am currently developing a bigger plugin with continuous timer-based animation, and the contrast when the game is set to a high speed (which would also happen at normal speeds in a very big game) is jarring. 

Please find a better way to use timers for your use-case. Why do you need this change?

src/base/multithreading.cc
```diff
@@ -231,9 +233,11 @@ MutexLock::MutexLock(const ID i) : id_(i) {
 	if (record.current_owner != kNoThread) {
 		for (const auto& pair : acting_as_another_thread) {
 			if (pair.first == self && pair.second == record.current_owner) {
```
bunnybot, line 287: tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4524#issuecomment-1561063)*
*On Mon Feb 12 16:12:41 CET 2024, **Tóth András (tothxa)** wrote:*

---

Oops... Thanks! That probably explains the change in the failures.

src/base/multithreading.cc
```diff
@@ -231,9 +233,11 @@
...
```
bunnybot, line 287: tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4524#issuecomment-1562083)*
*On Tue Feb 13 00:17:55 CET 2024, **Tóth András (tothxa)** wrote:*

---

> Skipping the log output is not great for debugging, but using printf directly to circumvent the mutexes will mess up the output which is just as bad.

Well, this only disables debugging for the lock inside `do_log()`. Does that affect anything else than these logging calls themselves? OTOH direct output in this case shouldn't mess up the output, as `do_log()` can only proceed when this constructor returns, can't it?

> Perhaps we could instead remove kLog and instead use a regular std::mutex and std::lock_guard in the logging implementation function do_log?

Or have a lockless option for `do_log()`? I don't know what would be the most elegant way to do that though.

tothxa:
> `LuaPlayer::message_box` (and any other lua functions that use `NoteThreadSafeFunction::instantiate` too)

Good news(?), message box is the only one.

tothxa:
> I don't know if it's possible, but I think `LuaPlayer::message_box` (and any other lua functions that use `NoteThreadSafeFunction::instantiate` too) should make their coroutine yield, release locks for the duration of the UI thread call, then when the call returns, reacquire locks and resume the coroutine.

Oops, that doesn't seem to be possible... The lock is held at the original lua entry point, so a lua function can't do anything about it, can it? Yield is probably also impossible without returning to the original entry point...

tothxa:
My locking(?) test is here: https://github.com/tothxa/widelands/pull/11

If the testsuite passes, I'll add the locking related changes in this PR. The window internal name changes should probably be a separate PR for v1.3, and the new test itself too.

I'm not really sure what the new testcase does actually test though... I'm trying to create notifications by starting and stopping the buildings and changing input settings, and that should trigger cross thread calls, right?

tothxa:
I think the lock logging changes are now quite nicely refined thanks to testing in https://github.com/tothxa/widelands/pull/11

The same tests show that there are still possible memory violations in master (possibly due to lock borrowing) even when plugin timers are not used. With the use of plugin timers, those violations are much more frequent, but obviously, plugin timers are not meant to manipulate the game state (though they're not prevented from doing so). But none of this can be reasonably fixed in v1.2.

src/base/multithreading.cc
```diff
@@ -310,7 +382,11 @@ MutexLock::MutexLock(const ID i) : id_(i) {
 					}
 
 					s_mutex_.unlock();
-					log_err("%s", info.c_str());
+					if (id_ != ID::kLog) {
+						log_err("%s", info.c_str());
```
tothxa, line 401: Maybe this should be always `std::cout`? We are about to throw an exception for a deadlock, that's more important than not messing up the output if that risks a lesser deadlock.

src/scripting/lua_ui.cc
```diff
@@ -4081,10 +4083,12 @@ static int L_show_messagebox(lua_State* L) {
 	std::string text = luaL_checkstring(L, 2);
 	bool allow_cancel = nargs < 3 || luaL_checkboolean(L, 3);
 
-	UI::WLMessageBox m(
+	std::unique_ptr<UI::WLMessageBox> m(new UI::WLMessageBox(
```
bunnybot: Nordfriese*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4524#issuecomment-1579944)*
*On Fri Feb 23 12:06:33 CET 2024, **Benedikt Straub (Nordfriese)** wrote:*

---

I don't understand the advantage of using a pointer here since the scope is the same. Wouldn't this work just as well if the object is still allocated on the stack with just the run() moved into the NoteThreadSafeFunction?

src/base/multithreading.cc
```diff
@@ -212,15 +216,27 @@ void MutexLock::pop_stay_responsive_function() {
 	stay_responsive_.pop_back();
 }
 
+// Only used for verbose logging of borrowing
```
bunnybot: Nordfriese*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4524#issuecomment-1579956)*
*On Fri Feb 23 12:12:50 CET 2024, **Benedikt Straub (Nordfriese)** wrote:*

---

All of these variables should have a name that indicates their purpose. A static variable at file scope named `prev_self` is meaningless. Since they all belong together, perhaps put them all into a `struct MutexBorrowRecord` and use a `static MutexBorrowRecord g_mutex_borrowing_record`.

bunnybot (commented):
Nordfriese*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4524#issuecomment-1579957)*
*On Fri Feb 23 12:06:33 CET 2024, **Benedikt Straub (Nordfriese)** commented:*

Tested this a bit and found no more problems. Thanks for this important refactoring and the great feature :) 

Two comments inline.

src/scripting/lua_ui.cc
```diff
@@ -4084,7 +4086,9 @@ static int L_show_messagebox(lua_State* L) {
 	UI::WLMessageBox m(
```
bunnybot, line 4086: tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4524#issuecomment-1580147)*
*On Fri Feb 23 14:54:12 CET 2024, **Tóth András (tothxa)** wrote:*

---

This was based on campaign message box in `lua_game.cc`. I don't quite understand why that is done this way though...

bunnybot, line 4086: tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4524#issuecomment-1580147)*
*On Fri Feb 23 14:54:12 CET 2024, **Tóth András (tothxa)** wrote:*

---

This was based on campaign message box in `lua_game.cc`. I don't quite understand why that is done this way though...

bunnybot, line 4086: tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4524#issuecomment-1580147)*
*On Fri Feb 23 14:54:12 CET 2024, **Tóth András (tothxa)** wrote:*

---

This was based on campaign message box in `lua_game.cc`. I don't quite understand why that is done this way though...

bunnybot, line 4086: tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4524#issuecomment-1580147)*
*On Fri Feb 23 14:54:12 CET 2024, **Tóth András (tothxa)** wrote:*

---

This was based on campaign message box in `lua_game.cc`. I don't quite understand why that is done this way though...

bunnybot, line 4086: tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4524#issuecomment-1580147)*
*On Fri Feb 23 14:54:12 CET 2024, **Tóth András (tothxa)** wrote:*

---

This was based on campaign message box in `lua_game.cc`. I don't quite understand why that is done this way though...

bunnybot, line 4086: tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4524#issuecomment-1580147)*
*On Fri Feb 23 14:54:12 CET 2024, **Tóth András (tothxa)** wrote:*

---

This was based on campaign message box in `lua_game.cc`. I don't quite understand why that is done this way though...

bunnybot (commented):
Nordfriese*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4524#issuecomment-1580335)*
*On Fri Feb 23 17:15:35 CET 2024, **Benedikt Straub (Nordfriese)** **approved** this pull request:*

<@>bunnybot merge

PR #6184, opened 2023-10-13
Fix line ending issue in vcpkg update workflow

tothxa:
**Type of change**
Bugfix

**Issue(s) closed**
Failure of vcpkg update workflow run when no change is required.

**To reproduce**
See https://github.com/widelands/widelands/actions/runs/6503257512/job/17664624247

**Possible regressions**
Should fix one...

**Additional context**
I only hope this works as expected.


tothxa:
OK, doesn't work. (I actually tested on my system that `read` doesn't need the line end. Another w*ndows difference?)

PR #6185, opened 2023-10-13
Run vcpkg update step on ubuntu

matthiakl:
**Type of change**
Bugfix 

**Issue(s) closed**
Re #6184

**How it works**
Line ending issus with git on Windows, but this step does not depends on Window, so run it on Ubuntu

Test run here: https://github.com/widelands/widelands/actions/runs/6507596511


matthiakl:
Well, actually I am not sure, whether this the line ending is the real problem. There was no change in the vcpkg version, so it was correct to not push it. It only rebuild the cache, which for some reason did not exist.

tothxa:
> Well, actually I am not sure, whether this the line ending is the real problem. There was no change in the vcpkg version, so it was correct to not push it.

To me it looks like it's the problem for the failure instead of a clean exit with no commit. Without the `git` warning the push would have been skipped cleanly.

> It only rebuild the cache, which for some reason did not exist.

Indeed, that's strange. I haven't thought of that.

matthiakl:
> Maybe this is another line ending issue?

I don't see it. The cache key creation in the vcpkg update workflow did not change and yet it decided to be different from the previous one.

tothxa:
caches

I even copy-pasted the keys from the page to files and compared them, but they were identical. That's why I think there may be a line ending in one but not in the other, or something like that. But that's only a guess of course.

tothxa:
I've also tried to look for the difference between the 2 workflows' definition of the cache key, and couldn't find how they could differ.

Also the github docs explicitly say that caches can be shared between workflows.

It's really strange.

matthiakl:
This week the update was committed successfully, so I guess it was a runner issue…

tothxa:
> This week the update was committed successfully, so I guess it was a runner issue…

It succeeded because there was a change to commit. The failure happens when there's nothing to commit.

tothxa:
And the `build` workflow still doesn't find the saved cache:
https://github.com/widelands/widelands/actions/runs/6604457798/job/17938623222?pr=6197

tothxa:
Seems to be a problem with the path? From the output of post-restore from the last run (https://github.com/widelands/widelands/actions/runs/6604717119/job/17939157181?pr=6197):

> Warning: Path Validation Error: Path(s) specified in the action for caching do(es) not exist, hence no cache is being saved.

tothxa:
Sorry for messing with your branch, but it was easier this way.

tothxa:
OK, I get it now. So we don't need ubuntu after all, because that step should be skipped completely when there are no changes. So all it takes is that backslash.

matthiakl:
In the meantime, I updated my testing repo and it looks like switching to ubuntu indeed resolves not only the warning but also the cache key issue. Take a look at the runs at https://github.com/matthiakl/actions/actions

- The cache is not found: https://github.com/matthiakl/actions/actions/runs/6605503589/job/17940732697#step:4:13
- The ref file is committed from ubuntu: https://github.com/matthiakl/actions/actions/runs/6605539662/job/17940807024#step:3:15
- The cache is now found: https://github.com/matthiakl/actions/actions/runs/6605550557/job/17940824411#step:4:17

matthiakl:
> OK, I get it now. So we don't need ubuntu after all, because that step should be skipped completely when there are no changes. So all it takes is that backslash.

But this does not resolve the issue, that subsequent builds do not find the cache, or does it?

PR #6188, opened 2023-10-15
Move marble to first column of Empire warelist

MarkMcWire:
Construction materials in the first row like barbarians, so third row is identical to empire, barbarians and atlanteans

**Type of change**

All construction materials should be listed in the first column. Serves for clarity.

### DIFF
--- a/data/tribes/initialization/empire/units.lua
+++ b/data/tribes/initialization/empire/units.lua
@@ -160,6 +160,24 @@ wl.Descriptions():new_tribe {
                -- TRANSLATORS: Helptext for an Empire ware: Cloth
                purpose = pgettext("empire_ware", "Cloth is needed to build several buildings. It is also consumed in the armor smithy.")
             }
+         },
+         {
+            name = "marble",
+            default_target_quantity = 30,
+            preciousness = 3,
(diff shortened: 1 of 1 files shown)


tothxa:
I believe the grouping logic is that mined materials go to column 3. Column 1 is mostly renewable non-food, plus granite (which can also be considered much cheaper than mined materials), column 2 is food (water goes here, though it has other uses too for some tribes), column 4 is tools, and column 5 is military equipment.

MarkMcWire:
@tothxa My idea is that the third column for Barbarians, Atlanteans and Empire is designed identically, i.e. coal, ores and metals. I find it clearer and you can find the wares much more quickly at first glance.

Since barbarians have an explicit granite mine, purely by this logic the granite should also have to be placed in column 3 for them.

frankystone:
There can't be a correct, logical, order.  What do you do with Wool, which is not a construction material (at least for buildings)? Everyone has another idea of the sorting those wares. The main problem is that some wares are used for different kind of things.

In the end a player has to become familiar with the sorting. Therefor i have no personal preference of the sorting, i just have to become familiar with this.

PR #6189, opened 2023-10-15
Move quartz and diamond to first column of Atlantean warelist

MarkMcWire:
All construction materials in first row like in Barbarians and Empire

**Type of change**
All building materials should be listed in the first column. Serves for clarity.


MarkMcWire:
no consent

Issue #4744, opened 2021-03-06
registering an attribute with a building should be possible


hessenfarmer:
I was trying to register an attribute with a custom building as it should be a child of immovable. however the attribute was not registered as i got a lua error that the attribute id was not existent, while using a findobject=attrib: command.

either this is a bug or we may consider it as a feature request.

Noordfrees:
The attributes registered while loading are currently used only when creating a critter or immovable (`src/logic/map_objects/descriptions.cc:402-475 Descriptions::add_object_description`). I'd consider it a new feature to pass them to other types as well. 
But buildings are not descendants of immovables (they are siblings which are both descended from the abstract type `BaseImmovable`); what are you intending to achieve with this?

hessenfarmer:
in my attempts on atl02 i need a custom worker named "trader" to go to the barbarian tribes HQ. I had introduced an immovable named trading pole with the attrib "trdepole" for this and placed it neat the HQ. However I recently had the idea to assign the attribute to a custom HQ and let the worker go there directly. This is not urgent though. Just wante to take note of this to not forget.

Noordfrees:
This might need a slightly bigger change than just adding attributes to buildings because entering a building is a more difficult task than walking to an immovable: an immovable can be approached and left from/in any direction whereas buildings are always entered/exited from the south-east; workers should be invisible while inside a building but visible while on immovables; and the building itself also does statistics and other magic on workers inside. I'm not against implementing such a thing, we just should keep in mind that this feature might need a few additional changes downstream.

SimonHeimberg:
I am surprised to find out that this does not work for most things.

As all units need register.lua (in https://www.widelands.org/documentation/add-ons/), I expected all would support to set attributes there.
And the method `has_attribute(attr)` is [defined for MapObject](https://www.widelands.org/documentation/autogen_wl_map/#wl.map.MapObject.has_attribute), which is the parent class of BaseImmovable and Bob.

Maybe there should be a hint somewhere in the doc?

I did not have problems with this because the try to have buildings as artifacts failed before coming to this point. (see [forum post ..41](https://www.widelands.org/forum/topic/5795/?page=1#post-40441))

bunnybot:
[...]

PR #6195, opened 2023-10-18
extend documentation about attributes of units

SimonHeimberg:
**Type of change**
Doc enhancement

**Issue(s) closed**
improves #4744

**Changed behavior**
The documentation for lua api only mentions has_attribute() for ImmovableDescription *and MapObject*. But it exists for all MapObject~Descriptions~. It will return false for all types except of ImmovableDescription and critters.
The documentation about writing units tells attributes can be set, but does not mention they do not work for all units.


doc/sphinx/source/lua_tribes_units.rst.org
```diff
@@ -66,6 +66,8 @@ The matching ``register.lua`` would look like this:
 
 Every unit defined by an add-on needs to define exactly one of the special attributes ``__skip_if_exists`` or ``__replace_if_exists``. This attribute determines how the engine proceeds if a description of the same name has already been registered. Units not contained in an add-on may not use these attributes.
 
+Free attributes are only supported for immovables (and critters).
```
hessenfarmer: I would rather add a part to the sentence above like "Additionally to its normal attributes (link to the documentation of the mapObject attribute below), every unit defined by an add-on ....

doc/sphinx/source/lua_world_units.rst.org
```diff
@@ -60,6 +60,8 @@ attribute tables are empty.
 
 Every unit defined by an add-on needs to define exactly one of the special attributes ``__skip_if_exists`` or ``__replace_if_exists``. This attribute determines how the engine proceeds if a description of the same name has already been registered. Units not contained in an add-on may not use these attributes.
 
+Free attributes are only supported for immovables and critters.
```
hessenfarmer: same as for tribes units

src/scripting/lua_map.cc
```diff
@@ -2603,6 +2603,9 @@ int LuaImmovableDescription::get_size(lua_State* L) {
 
       Returns :const:`true` if the immovable has the attribute, :const:`false` otherwise.
 
+      Note: This method does exist for all MapObjects, but it only gets initialised for this class
```
hessenfarmer: It exists as it is already defined for the base class.  From my perspective this is an unneeded duplication imho and should be harmonized into the base class (i.e. merging the error reporting part into the base class definition)

hessenfarmer (changes requested):
I believe we need a refactoring here rather then a documentation change

src/scripting/lua_map.cc
SimonHeimberg: Actually this are two different object hierarchies. And I added the comment to the wrong function.
This one is `LuaImmovableDescription`, the other class is `MapObject`. They do not inherit from each other (a MapObject contains a xxDescription)

SimonHeimberg:
> I believe we need a refactoring here rather then a documentation change

What kind of a refactoring?
- Should all MapObjects read attributes (from "register.lua")?
  Attributes are passed to the constructor of two objects (Critters and Immovables) in
https://github.com/widelands/widelands/blob/aac51a3baa3dca12d6c6c8017b629bd6c7a4534b/src/logic/map_objects/descriptions.cc#L431
  , to pass it to all Subclasses of MapObjects would be quite some work. (Maybe there is a simpler way than to add it to every constructor.)
- Unify the methods named `has_attribute()`?
  They are in different class hierarchies.
- ...?

src/scripting/lua_map.cc
hessenfarmer: > Actually this are two different object hierarchies. And I added the comment to the wrong function. This one is `LuaImmovableDescription`, the other class is `MapObject`. They do not inherit from each other (a MapObject contains a xxDescription)

The base class for immovables and critters is mapobject. So immovables as well as critters inherit from mapobject.   
Both lua methods use the same c++ funtion from mapobject description.  so in c++ it is only defined in mapobject description and my proposal was to do it the same way in lua.

hessenfarmer:
> > I believe we need a refactoring here rather then a documentation change
> 
> What kind of a refactoring?
> 
>     * Should all MapObjects read attributes (from their "register.lua")?
>       Attributes are passed to the constructor of two objects (Critters and Immovables) in
>       https://github.com/widelands/widelands/blob/aac51a3baa3dca12d6c6c8017b629bd6c7a4534b/src/logic/map_objects/descriptions.cc#L431
>       
>       , to pass it to all Subclasses of MapObjects would be quite some work. (Maybe there is a simpler way than to add it to every constructor.)     
    
as you already wrote in your proposed change each mapobject has it, but it is only initialized for immovables and critters.  so no need to change anything there.  
   
> 
>     * Unify the methods named `has_attribute()`?
>       They are in different class hierarchies.
> 
>     * Should `has_attribute()` of `LuaImmovableDescription` be moved to a parent class? (Probably confusing, as only Immovable and Critter support loading the attribute values.)
> 
>     * ...?
   
I thought of determining whether we have the description already or have the object as in the end we return the value from the description.    

but we could keep it for the moment being and just shift the text changes to the correct point

doc/sphinx/source/lua_tribes_units.rst.org
```diff
@@ -64,7 +64,7 @@ The matching ``register.lua`` would look like this:
    Also, ``berry_bush_blueberry_small`` in this example has no attributes, so its
    attribute table is empty.
 
-Every unit defined by an add-on needs to define exactly one of the special attributes ``__skip_if_exists`` or ``__replace_if_exists``. This attribute determines how the engine proceeds if a description of the same name has already been registered. Units not contained in an add-on may not use these attributes.
+Additionaly to its normal attributes (:method:`MapObject.has_attribute`), every unit defined by an add-on needs to define exactly one of the special attributes ``__skip_if_exists`` or ``__replace_if_exists``. This attribute determines how the engine proceeds if a description of the same name has already been registered. Units not contained in an add-on may not use these attributes.
```
hessenfarmer: sorry i did not recognize the context in this file in the first case. As the normal attributes asre explained above we could simply skip the reference to make the ci happy
```suggestion
Additionaly to its normal attributes, every unit defined by an add-on needs to define exactly one of the special attributes ``__skip_if_exists`` or ``__replace_if_exists``. This attribute determines how the engine proceeds if a description of the same name has already been registered. Units not contained in an add-on may not use these attributes.
```

doc/sphinx/source/lua_world_units.rst.org
```diff
@@ -58,7 +58,7 @@ The matching ``register.lua`` would look like this:
 Note that ``alder_summer_pole`` and ``alder_summer_mature`` in this example have no attributes, so their
 attribute tables are empty.
 
-Every unit defined by an add-on needs to define exactly one of the special attributes ``__skip_if_exists`` or ``__replace_if_exists``. This attribute determines how the engine proceeds if a description of the same name has already been registered. Units not contained in an add-on may not use these attributes.
+Additionaly to its normal attributes (:method:`MapObject.has_attribute`), every unit defined by an add-on needs to define exactly one of the special attributes ``__skip_if_exists`` or ``__replace_if_exists``. This attribute determines how the engine proceeds if a description of the same name has already been registered. Units not contained in an add-on may not use these attributes.
```
hessenfarmer: same here
```suggestion
Additionaly to its normal attributes, every unit defined by an add-on needs to define exactly one of the special attributes ``__skip_if_exists`` or ``__replace_if_exists``. This attribute determines how the engine proceeds if a description of the same name has already been registered. Units not contained in an add-on may not use these attributes.
```

bunnybot (commented):
Nordfriese*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4537#issuecomment-1361077)*
*On Fri Dec 01 08:47:54 CET 2023, **Benedikt Straub (Nordfriese)** commented:*

"Additionaly to" → "In addition to" (2×)

In the note, please use
```
.. note::
```
and
```
:const:`false`
```

doc/sphinx/source/lua_tribes_units.rst.org
SimonHeimberg: But then we do not have mentioned that the normal attributes are only supported for Immovables and Critters. (Actually I prepared a follow up pr which initializes the attributes for all customizable map objects. Then in does not matter anymore.)
By the way, the doc check [complains about :method:](/widelands/widelands/actions/runs/7048322366/job/19184229800?pr=6195), which does not need to be resolved if we drop the reference.

SimonHeimberg:
> "Additionaly to" → "In addition to" (2×)
> 

Sounds good. Will do when the other change is clear.

> In the note, please use
> 
> ```
> .. note::
> ```
> 
> and
> 
> ```
> :const:`false`
> ```

After looking at the changed files, I found those in src/scripting/lua_map.cc and improved it.

doc/sphinx/source/lua_tribes_units.rst.org
hessenfarmer: that is correct. How about adding this information after line 14 of this file (line 178 in the other file) where attributes are first mentioned.  
the complain of the ci doc check was why I was sugesting to remove the reference.

hessenfarmer (approved):
thanks for fixing it it should be 100% clear now for everybody.

hessenfarmer:
@bunnybot update

hessenfarmer:
@bunnybot merge

PR #6198, opened 2023-10-22
Fix vcpkg cache restore

tothxa:
**Type of change**
Bugfix

**Issue(s) closed**
Re #6185: Fix restoring the vcpkg cache

**Possible regressions**
msvc builds in the workflow

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -261,7 +261,7 @@ jobs:
     - name: Restore vcpkg and its artifacts.
       uses: actions/cache@v3
       with:
-        path: ${{ env.VCPKG_ROOT }}/installed
+        path: ${{ env.VCPKG_ROOT }}\installed
         key: |
           ${{ steps.prepare.outputs.vcpkg_key }}-${{ matrix.arch }}
     - name: Installing dependencies


matthiakl (approved):
Thanks for finding this. 
I finally also understood what was going on:
The cache key is identical, but the *version* (which is not visible) was different because of the different path names ([doc](https://github.com/actions/cache#cache-version)).

tothxa:
Thank you.

Now (apart from the mingw failure just at the best moment), I think we could even remove the `git status -s` test, if that code path is unreachable without changes. I've realised though that if only the toolset version changes, but not the vcpkg commit, then everything is skipped. So I wonder whether we should include the toolset version in the cache key to allow toolset only updates too.

tothxa:
@bunnybot merge

matthiakl:
> I've realised though that if only the toolset version changes, but not the vcpkg commit, then everything is skipped. So I wonder whether we should include the toolset version in the cache key to allow toolset only updates too.

The toolset version does not change as often as the vcpkg version and if a new one is added, one or two older versions are still available, so I do not think this is necessary. But yes, to be safe we could add it to the key.

PR #6205, opened 2023-10-30
Make naval warfare a game flag, disabled by default

bunnybot:
[Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4547)*
*Created on Mon Oct 30 10:07:46 CET 2023 by [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese)*

---

**Type of change**
Feature change

**Issue(s) closed**
Re the forum thread: Naval warfare is now a game flag that has to be explicitly selected in the game configuration options. Default is disabled. There is a note that this feature is experimental and will be subject to change in the future. The checkbox is only shown for seafaring maps. Existing savegames all have naval warfare disabled for a better v1.1 → v1.2 migration experience. The flag is also exposed to the Lua API.


src/logic/game.h
```diff
@@ -405,11 +405,18 @@ class Game : public EditorGameBase {
...
 		return diplomacy_allowed_;
 	}
-	void set_diplomacy_allowed(bool d) {
-		diplomacy_allowed_ = d;
+	void set_diplomacy_allowed(bool allow) {
+		diplomacy_allowed_ = allow;
+	}
+
+	[[nodiscard]] bool naval_warfare_allowed() const {
+		return diplomacy_allowed_;
```
bunnybot: tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4547#issuecomment-1314155)*
*On Mon Oct 30 17:37:55 CET 2023, **Tóth András (tothxa)** wrote:*

---

```
return naval_warfare_allowed_;
```

bunnybot (commented):
tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4547#issuecomment-1314156)*
*On Mon Oct 30 17:37:55 CET 2023, **Tóth András (tothxa)** commented.*

frankystone:
> BTW maybe "No fog of war" and "Forbid diplomacy" should be changed in the game setup menu to "Enable fog of war" and "Enable diplomacy" (and enabled by default instead of disabled), or even just "Fog of war"/"Diplomacy"/"Naval wfarfare".

The entries in  "Show more Options" should per default set to a value for "normal" games, settings which are valid for most of the games. E.g. Fog of War  should be enabled as default because this is the "normal" way to play widelands.
For "Diplomacy" i am unsure but tend to say it should be disabled as default, because normal games are always 1vs1 ( for multiplayer, of course this is a guess). Additionally one can define teams for a game. I would say Diplomacy is an advanced option.
Same for naval ware fare. This is an advanced option and should be disabled as default.

No one should be forced to make some extra klicks to play a game.  Those extra klicks and options should only be useful for experienced players, imho.

tothxa:
> > BTW maybe "No fog of war" and "Forbid diplomacy" should be changed in the game setup menu to "Enable fog of war" and "Enable diplomacy" (and enabled by default instead of disabled), or even just "Fog of war"/"Diplomacy"/"Naval wfarfare".
> 
> The entries in "Show more Options" should per default set to a value for "normal" games, settings which are valid for most of the games.

Well, I didn't mean to change default behaviours (though they can be reconsidered too of course), only to try to use "positive" terms so that it's easier to talk about enabling/disabling them. Though ideally it should be done in a way that allows all defaults to have the same checkbox state. The above 2 use "negative" terms because currently all checkboxes are disabled by default, except for "write replay". (this isn't very good either)

But now I think that this is definitely not for this PR, which has a much higher urgency.

src/logic/game.cc
```diff
@@ -893,6 +894,7 @@ void Game::cleanup_for_load() {
 
 	pending_diplomacy_actions_.clear();
 	diplomacy_allowed_ = true;
+	naval_warfare_allowed_ = true;
```
hessenfarmer, line 897: shouldn't this be false, to recreate the default state

src/logic/game.h
```diff
@@ -495,6 +502,7 @@ class Game : public EditorGameBase {
 
 	std::list<PendingDiplomacyAction> pending_diplomacy_actions_;
 	bool diplomacy_allowed_{true};
+	bool naval_warfare_allowed_{false};
```
hessenfarmer: this needs to be initialised in true to make the BIT wise AND in line 411 of game.cc work 
this will make the checkbox work correctly in launch game.

src/logic/game.cc
hessenfarmer: forget it it needs to be true

hessenfarmer (changes requested):
found the reason for the checkbox not working.   
this needs to be changed and we can have this as one of the enablers for 1.2

bunnybot (commented):
tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4547#issuecomment-1358770)*
*On Wed Nov 29 13:12:47 CET 2023, **Tóth András (tothxa)** **approved** this pull request:*

Thank you. Tested and starting games also works now.

hessenfarmer (approved):

PR #6206, opened 2023-11-01
Revision fix windows

### LINKED ISSUE #6186: 2 day in a row: 0~cmake-REVDETECT-BROKEN-PLEASE-REPORT-THIS
## Describe the bug
This is Widelands version `0~cmake-REVDETECT-BROKEN-PLEASE-REPORT-THIS Release`

Bug reports? Suggestions? Check out the project website:
https://www.widelands.org/

Hope you enjoy this game!

> not really as unable to play :smirk: 

## To reproduce
Not really a thing for this report

## Expected behavior
Self explained

## Screenshots
image

## Version:
 - OS: Ubuntu 22.04
 - Widelands Version: See above

## Additional context

```bash
widelands  --version
This is Widelands version 0~cmake-REVDETECT-BROKEN-PLEASE-REPORT-THIS Release
[00:00:00.000 real] INFO: Set home directory: /home/$USER/.local/share/widelands
[00:00:00.000 real] INFO: Set configuration file: /home/$USER/.config/widelands/config
[00:00:00.000 real] WARNING: No locale translations found in /usr/share/games/widelands/data/locale
[00:00:00.000 real] INFO: selected language: (system language)
[00:00:00.000 real] INFO: using locale en_DK.UTF-8
=========================================================================
```

hessenfarmer:
**Type of change**
Bugfix

**Issue(s) closed**
Fixes #6186 

**New behavior**
As we require python3 we now explicitly search for it but this delivers a new Cmake variable

**Possible regressions**
linux and mac builds


matthiakl:
This will break building with CMake <= 3.12 where the old Python package is used. I just looked into the same issue this morning (https://github.com/widelands/widelands/compare/master...matthiakl:widelands:6186-revdetect), instead of replacing the PYTHON_EXECUTABLE we could set it from the new variable:
`set(PYTHON_EXECUTABLE ${Python_EXECUTABLE})`
To prevent future issues like this I am trying to add a check into the workflow to see that the version actually contains the git hash.

hessenfarmer:
I was just thinking about this as well however which of our supported OS Versions still uses Cmake below 3.12.    
I was rather thinking to bump th minimum version and cleanup other workarounds, if none of our supported OS still uses lower then 3.12. 
As we now require python3 explicitly in our python scripts will they work wit any old found PythonInterpreter or does Find PythonInterp provide the posibility to search explizitly for python3 as well

matthiakl:
Debian buster (oldoldstable) has cmake 3.13, but Ubuntu 18.04 had 3.10, but is out of normal support this year, so bumping the minimum version to 3.12 should work.

hessenfarmer:
Well I came to the same results and saw that kitware is even providing packeages to upgrade for all apt based systems so lets wait for @Noordfrees and  if he agrees lets go this way. 
Unfortunately the comments are not mirrored to codeberg at the moment

hessenfarmer:
I just checked your branch. I am wondering whether it works as in my Installation and branch the Python 3 variable is led by an underscore.

hessenfarmer:
@matthiakl 
shall I close this PR and we build the fix on your PR, as it contains the checks for a regression in the build jobs as well?

matthiakl:
I opened my branch as draft PR: https://github.com/widelands/widelands/pull/6207
We can continue there :)

PR #6208, opened 2023-11-05
Button evict builder

### LINKED ISSUE #5275: Nothing gets built if all builders are on a building site with missing supplies
The builders can refuse to build anything at all if all builders are on a building site with missing supplies for hours, even if there are other building sites with 100% supplies.

Not sure if this is a bug, but it's certainly very unexpected with no obvious solution for the player.

## How to reproduce:

(This example is quite contrived but good to demonstrate how to trigger this. For a more realistic example, see the "See also" section below.)

1. Start a singleplayer game on any map with start condition "Poor Village", "Infinite Game", as the Barbarians and close all AI slots
2. Place three expensive buildings (like the Arena) but DO NOT connect them yet
3. Set the wood log storage of the building sites to 0 to prevent logs being delivered there
4. Now place a woodcutter somewhere
5. Connect the 3 arenas to your HQ
6. Wait until the builders come out
7. Finally, connect the woodcutter

This is what will happen: Your (I guess) 2 builders will immedately run to the Arenas. Then logs get delivered to your woodcutter. Now speed up 100× and you will see the builders will do nothing and are stuck at the arena, waiting for supplies to arrive (which never come) while they all ignore the woodcutter hut with all required logs.

## Workaround (for the player)

Demolish some of the buildings with a builder on it to force them to relocate to a good building site.

## Suggestion
If all builders are busy at building sites AND they don't work (no supplies) AND there is at least one other building site with work for them (has supplies), then relocate the builder to a building site with more promising work.

## See also

https://github.com/widelands/widelands-website/issues/377

SimonHeimberg:
The PR is not complete yet. But I need some help.

- [x] How can I get the builder from the building site window? `friend` does not work as I expected.
- [x] The construction site should pause until a new builder arrived.
- [x] Not sure if the ordering of another builder is done in a good way.

**Type of change**
New feature

**Issue(s) closed**
fixes https://github.com/widelands/widelands/issues/5275

**New behavior**
A construction site shows an evict button to evict the builder.
The builder is evicted by the button.
If a builder is evicted (by button or by script), a new builder is ordered.
INCOMPLETE: The builder leaves immediately, but construction continues until the one current building material is used.

**Use case**
When I created more construction sites than there are builders, one site may have a builder but no construction material. Being able to send the builder to another site then is an advantage.

**Possible regressions**
No known, but ...


matthiakl:
> How can I get the builder from the building site window? friend does not work as I expected.

The builder should be the first worker on the construction site. Try using `workers = construction_site->get_workers()`. `workers.front()` should be the builder.

> The builder should stay on site until the one current building material is used up. Or the construction site should pause until a new builder arrived.

Instead of evicting the builder immediately, you could set a flag (e.g. in a small function `ConstructionSite::stop_work()`) and evict him in `ConstructionSite::get_building_work()` once he wants to start the next build step. Then you do not need to access the builder from `ConstructionSiteWindow`.
I suppose stopping the current task immediately will not work, the consumed ware must be reset?

SimonHeimberg:
> > How can I get the builder from the building site window? friend does not work as I expected.
> 
> The builder should be the first worker on the construction site. Try using `workers = construction_site->get_workers()`. `workers.front()` should be the builder.
> 

Getting the worker like this works. Thank you! After some tries I even found out how to check if there is a worker.

> > The builder should stay on site until the one current building material is used up. Or the construction site should pause until a new builder arrived.
> 
> Instead of evicting the builder immediately, you could set a flag (e.g. in a small function `ConstructionSite::stop_work()`) and evict him in `ConstructionSite::get_building_work()` once he wants to start the next build step. Then you do not need to access the builder from `ConstructionSiteWindow`. I suppose stopping the current task immediately will not work, the consumed ware must be reset?

Evicting a builder could be implemented with a flag in construction site. But I see two arguments against.
- Calling `builder.evict()` from a script would still evict the worker immediately but continue work.
- This would be inconsistent with every other worker. They all leave immediately and abort work.
  Examples:
  A lumber jack stops cutting the tree and walks to his hut and then to the warehouse.
  A miner leaves the mine immediately. In case he has consumed a ware but not yet produced anything,
  the consumed ware is lost. The same for a smith and ...

I suggest the builder should *leave immediately* and *abort* the work. This I could implement probably, but the
partially used ware is not requested again, so one ware is missing to complete the building. (Tried from debug console.)
Requesting the ware again would be consistent to the other workers. Choosing the right time for evicting is the players responsibility.
It could also be made consistent to the soldiers which refuse to leave when they are in a battle. The builder would *refuse to leave* while working.

SimonHeimberg:
> > > How can I get the builder from the building site window? friend does not work as I expected.
> > 
> > 
> > The builder should be the first worker on the construction site. Try using `workers = construction_site->get_workers()`. `workers.front()` should be the builder.
> 
> Getting the worker like this works. Thank you! After some tries I even found out how to check if there is a worker.
> 

Codecheck tells I should use `.empty()` instead of `.size() == 0`. Will do.

SimonHeimberg:
I see three options. Which way to go? https://github.com/widelands/widelands/pull/6208#issuecomment-1814550227

- The builder should leave immediately and abort the work.
- The builder would refuse to leave while working.
- Accept the pr as it is: The builder leaves the building, but the work is continued until one material is used up. The player will not be able to speed up building by using a worker for two sites. Walking around takes too long. But it looks strange.

tothxa:
> * The builder leaves immediately and aborts the work.

+1

> (The last used material must be requested again. Or any, can be the cheapest or ...)

This feature is most important for limited resources situations, so IMO the ware should be preserved somehow. If possible, the state of the construction site should remain as it is, i.e. when a new builder arrives, he/she should pick up the work where the old one left off. If not possible, then revert completion to what it was before the last ware was consumed and return the last ware to the construction site.

> * The builder refuses to leave while working.

If the above options are too hard to implement, this may be acceptable as long as it is marked in the status line that the builder is evicted.

hessenfarmer:
@SimonHeimberg are you willing to continue to work on this.
else @widelands/developers @cb/wl/developers  anybody willing to take this over and finalize
@w

hessenfarmer:
> [Nordfriese](https://codeberg.org/Nordfriese)_[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4551#issuecomment-2865149)_ _On Sat Feb 22 10:05:00 CET 2025, [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese) wrote:_
> 
> If I evict a builder during the last work step, then it crashes upon completion with
> 
> ```
> FATAL EXCEPTION: [.../src/base/times.h:185] Time: Subtraction result would be negative
> ```

well I think this is not related to whether evicting or not the builder, but can happen at every end of  a step due to a racecondition between the calls to draw() and get_building_work() I have seen other spots in the code where there was code to prevent that to happen, so I will try if this would solve the issue here as well.    

> 
> I don't think we need this for the dismantlesite, though it wouldn't hurt to have it either if you want it.

I will do it for consistency reasons then, to avoid questions why the button misses in a dismantle site.

hessenfarmer:
> [tothxa](https://codeberg.org/tothxa)_[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4551#issuecomment-2865618)_ _On Sat Feb 22 14:13:24 CET 2025, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) wrote:_
> 
> Sorry for nitpicking, but I don't think you should change the internal state in `draw()`. You've already intruduced a new variable (BTW don't `write()` and `read()` need updating with it too?), why not set it once in `evict_builder()`, then use either the old behaviour when `working_` and there is a worker, or the saved `remaining_time_` when there isn't?

well, the complete buildprocess is controlled by workstep_completiontime_ which is the target_time when the step shall be completed. The drawing is controlled by the completedtime value of the info_ object. As we don't know how long the builder will be away we do not know to which value to set the workstep_completiontime_ in advance. so we need to constantly push it until the builder is back. And the only task that is constantly called while the builder is away is draw().     
Furthermore to implement the setting of the internal variable in evict_worker() we would need to expose it by a public funtion tom the constructionsite window which is even more complicated and would not prevent the need to constantly pusch the internal workstep_completiontime_- 
So I don't see how to implement your suggestion. If you see a way please help me out.     

I already realized the need to save and read the new variable, but thanks for pointing to this, as I nearly overlooked it

hessenfarmer:
sorry, but the suggested solution would need a major redesign of the complete algorithm to draw the constructionsite.     
the only argument for this is to not modify an internal state in a draw function which has been the case now for quite a long time and I just extended this.    
I have come up with a working solution now. It cost me some time. And I neither have the time nor the motivation to just start over everything, just for the personal taste of a reviewer. Obviously my development and review styless do not match with the style of @cb/tothxa. So I will not change anything in this PR. If anybody from @widelands/developers or @cb/wl/Developers has a better working solution, feel free to close this and open another or push to this. Otherwise feel fre to merge.

hessenfarmer:
> [tothxa](https://codeberg.org/tothxa)_[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4551#issuecomment-2870071)_ _On Sun Feb 23 23:47:37 CET 2025, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) wrote:_
> 
> I've just tried my above suggestion, and as far as I can tell, it works. (not tried save and reload, but I don't think it should matter)
> 
> It's really just what I wrote above, no redesign, actually net code reduction instead. I'm sorry I couldn't do it in inline comments, because it touches multiple places. So here it is as a patch:
> 
> [evict_builder_suggestion.diff](https://codeberg.org/attachments/f2a2bab0-145b-4c3d-adff-edc628198d00)
> 
> This doesn't rename `working_` yet to not make it look bigger than it is.
> 
> And of course it completely builds on your work on this PR. Thank you for picking it up.
I checked the patch and it will reset the building progress and the graphics to the last finsished workstep. this means we will loose 29 seconds of the buildprocess in the worst case with the animation "jumping" back to an earlier and less visible progress state, due to line 755. 
And still you did not provide any reason why it should be done your way.

hessenfarmer:
> [tothxa](https://codeberg.org/tothxa)_[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4551#issuecomment-2872755)_ _On Mon Feb 24 22:45:30 CET 2025, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) wrote:_
> 
> Sorry, indeed, I overlooked `constructionsite.cc:755`. 

and that is exactly what causes me a lot of frustration. you are not happy with a working solution already implemented for whatever reason and make suggestions that are not thought completely to the end, while I put a lot of time into it to stumble over all of these issues previously and now have to overgo them all again.       
Personally I am just one step away of quititng due to this frustration.      
Developping widelands looses its fun currently for me and tends towards beeing either really hard but boring review work, or endles discussions over personal likings.

bunnybot (commented):
[tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4551#issuecomment-2874308)*
*On Tue Feb 25 16:05:35 CET 2025, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) **approved** this pull request:*

Thank you and I apologise for the trouble I caused to you.

src/map_io/map_buildingdata_packet.cc
```diff
@@ -302,9 +303,12 @@ void MapBuildingdataPacket::read_partially_finished_building(PartiallyFinishedBu
 			}
 
 			pfb.working_ = (fr.unsigned_8() != 0u);
-			pfb.work_steptime_ = Time(fr);
+			pfb.workstep_completiontime_ = Time(fr);
 			pfb.work_completed_ = fr.unsigned_32();
 			pfb.work_steps_ = fr.unsigned_32();
+			if (packet_version == kCurrentPacketPFBuilding) {
```
bunnybot: [Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4551#issuecomment-2874414)*
*On Tue Feb 25 17:03:00 CET 2025, [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese) wrote:*

---

Just one comment, please don't compare against the symbolic constant in legacy savegame handling, compare against the explicit version where the change was introduced instead:
```c++
			if (packet_version >= 3) { 
```
Reason, next time we bump the version this line will silently break.

bunnybot (commented):
[Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4551#issuecomment-2874415)*
*On Tue Feb 25 17:03:00 CET 2025, [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese) submitted a since-dismissed review.*

bunnybot (commented):
[Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4551#issuecomment-2874515)*
*On Tue Feb 25 17:45:56 CET 2025, [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese) **approved** this pull request:*

Thank you :)

<@>bunnybot merge

PR #6028, opened 2023-07-18
Skip CI workflows where nothing changed

Noordfrees:
**Type of change**
CI enhancement

**Issue(s) closed**
Fixes #6020 & re #5880

**New behavior**
Skip CI checks where nothing relevant has been changed

**Additional context**
Patch by @SimonHeimberg at https://github.com/widelands/widelands/discussions/5880#discussioncomment-6476511

PR #6257, opened 2023-12-01
Skip CI workflows where nothing changed

### LINKED ISSUE #6020: only trigger tests in workflow if something relevant has changed

**Describe the bug**
Tests are also run if only a file changed which has nothing to do with the code
example: https://github.com/widelands/widelands/pull/6019

**Expected behavior**
I suggest to skip tests on obvious cases by using ...

either `paths` or `paths-ignore` for the entire workflow ([workflow-syntax-... on docs.github](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore))
or conditions for some jobs ([using-conditions-... on docs.github](https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution))

**some (include) paths for ...**
compile and tests: src/, data/ (should we exclude some content?), test/, cmake/ (?), regression_test.py, (probably some more)
documentation: (is fast anyway) doc/, src/, data/scripting/, data/tribes/, data/world/, data/txts/

SimonHeimberg:
**Type of change**
CI enhancement
 
**Issue(s) closed**
Fixes #6020 & re #5880

 **New behavior**
Skip CI checks where nothing relevant has been changed
 
**Possible regressions**
- The summary job might be started late because jobs of other pr get queued first. So the total time for one workflow might not go down. (But the total still should.)
- Some changes of relevant files might not yet be listed and jobs be skipped therefore.
  If this is realised, the missing file can be added and the next commit will trigger the run.

**Additional context**
Continues pullrequest GH #6028


SimonHeimberg:
I will change the pr to master later.

.github/workflows/build.yaml
```diff
@@ -81,6 +56,14 @@ jobs:
...
   clang_tidy:
     runs-on: ubuntu-latest
+    # Set job outputs to values from changed_files step for other jobs
+    outputs:
+      run_test_suite: ${{ steps.changed_files.outputs.test_suite }}
+      create_installer: ${{ steps.changed_files.outputs.installer }}
+      inst_windows_msvc: ${{ steps.changed_files.outputs.inst_windows_msvc }}
+      inst_windows: ${{ steps.changed_files.outputs.inst_windows }}
+      inst_macos: ${{ steps.changed_files.outputs.inst_macos }}
+      appimage: ${{ steps.changed_files.outputs.appimage }}
```
tothxa: Is this really needed?

.github/workflows/build.yaml
```diff
@@ -89,39 +72,60 @@ jobs:
...
-        path: build/clang-tidy-cache
-        key: clang-tidy-cache-${{ github.sha }}
-    - name: Report
-      run: utils/check_clang_tidy_results.py clang-tidy.log
+        # groups to check for changed files (as string, yaml anchors are resolved by the tool)
+        filters: |
+          clang_tidy:
+            - 'src/**'
+            - 'utils/check_clang_tidy_results.py'
+            - 'utils/run-clang-tidy.py'
```
tothxa:
```suggestion
          clang_tidy:
            - '*compile'
            - 'utils/check_clang_tidy_results.py'
            - 'utils/run-clang-tidy.py'
            - '.github/actions/build_clang_tidy.yaml'
```

.github/workflows/build.yaml
```diff
@@ -89,39 +72,60 @@ jobs:
...
+          compile: &compile
```
tothxa: `install_dependencies` doesn't need to be separate, it should all be here.

.github/workflows/build.yaml
```diff
@@ -89,39 +72,60 @@ jobs:
...
+            - 'CMakeLists.txt'
+            - 'cmake/**'
+            - 'src/**'
```
tothxa: plus `build.yaml` :)

.github/workflows/build.yaml
```diff
@@ -89,39 +72,60 @@ jobs:
...
+          test_suite:
+            - *compile
+            - 'data/**'
```
tothxa: Not in this PR, but we'll need to divide data further. E.g. `campaigns` and most `maps` are not used in the tests. I believe sound is also disabled, so `sound` and `music` can probably be excluded too. Same for `i18n` probably.
(though we can define more tests if we can :)

Also, we'll need a "testsuite-light" task eventually, used for `data` only changes, that uses fewer builds, or even just grabs the binary from the installer of the previous master.

.github/workflows/build.yaml
```diff
@@ -89,39 +72,60 @@ jobs:
...
+          installer:
+            - *compile
+            - 'data/**'
+            - 'doc/**'
+            - 'po/**'
+          install_dependencies: &install_dependencies
+            - 'install-dependencies.sh'
+            - 'utils/download_asio.sh' # called by install-dependencies.sh
+          inst_windows_msvc:
+            - *install_dependencies
```
tothxa:
```suggestion
            - *compile
```

.github/workflows/build.yaml
```diff
@@ -89,39 +72,60 @@ jobs:
...
+            - 'utils/win32/**'
+            - '.github/scripts/vcpkg_ref'
+            - '.github/workflows/build_windows_msvc.yaml'
+          inst_windows:
+            - *install_dependencies
```
tothxa:
```suggestion
            - *compile
```

.github/workflows/build.yaml
```diff
@@ -89,39 +72,60 @@ jobs:
...
+            - 'utils/win32/**'
+            - '.github/workflows/build_windows.yaml'
+          inst_macos:
+            - 'compile.sh' # called by utils/macos/build_app.sh
+            - *install_dependencies
```
tothxa:
```suggestion
            - *compile
```

.github/workflows/build.yaml
```diff
@@ -89,39 +72,60 @@ jobs:
...
+            - 'utils/macos/*'
+            - '.github/workflows/build_macos.yaml'
+          appimage:
+            - 'xdg/org.widelands.Widelands.*'
+            - '.github/workflows/build_appimage.yaml'
+    - name: Clang-tidy steps
+      if: steps.changed_files.outputs.clang_tidy == 'true'
+      uses: ./.github/actions/build_clang_tidy.yml
```
tothxa:
```suggestion
      uses: ./.github/actions/build_clang_tidy.yaml
```

(is the `./` needed?)

SimonHeimberg: yes, it needs ./, else it interprets it as github.com/.github/workflows/...
(see https://docs.github.com/en/actions/using-workflows/reusing-workflows#calling-a-reusable-workflow)

Thanks for catching the wrong file name.

.github/workflows/build.yaml
SimonHeimberg: Yes, it needs job output because a job can not access the steps of another job.
(see https://docs.github.com/en/actions/using-jobs/defining-outputs-for-jobs)

.github/workflows/build.yaml
SimonHeimberg: clang_tidy does not need compile, as it can not validate the cmake files and cmake is not run in this job.
Of course the action needs to be listed.

.github/workflows/build.yaml
SimonHeimberg: I just found out that all installers and the test-suite use `./install-dependencies.sh`  either directly or indirectly with `.github/scripts/install_dep.sh`. So I agree. The reference might need a better name.

.github/workflows/build.yaml
```diff
@@ -89,39 +72,60 @@ jobs:
...
+          test_suite:
+            - *compile
+            - 'data/**'
+            - 'regression_test.py'
+            - 'test/**'
+            - '.github/workflows/build_testsuite.yaml'
+          installer:
+            - *compile
+            - 'data/**'
+            - 'doc/**'
```
SimonHeimberg: Do installers include doc? Which ones?

.github/workflows/build.yaml
```diff
@@ -140,303 +144,56 @@ jobs:
...
-        $env:APPVEYOR_BUILD_VERSION = "Widelands-${{ github.sha }}-${{ matrix.config }}-${{ matrix.arch }}"
-        ISCC.exe /o$env:GITHUB_WORKSPACE /fWidelands-${{ github.sha }}-msvc-${{ matrix.config }}-${{ matrix.arch }} $env:GITHUB_WORKSPACE\utils\win32\innosetup\Widelands.iss
-    - name: Uploading installer
-      uses: actions/upload-artifact@v3
-      with:
-        name: Widelands ${{ matrix.config }} ${{ matrix.arch }} Installer (MSVC)
-        path: ${{ github.workspace }}\Widelands-${{ github.sha }}-msvc-${{ matrix.config }}-${{ matrix.arch }}.exe
+    if: >
+      needs.clang_tidy.outputs.create_installer == 'true' ||
+      needs.clang_tidy.outputs.inst_windows_msvc == 'true'
```
SimonHeimberg: The `if:` for installers is differently. It checks for the general installer files or for the specific files.
I made this differently probably mainly because of it was separate before. The action `dorny/paths-filter` probably does optimisation as it tells first "2 files changed".
Maybe it should be changed to the nested file structure like the other jobs.

.github/workflows/build.yaml
SimonHeimberg: Dividing data further makes sense. But as `dorny/paths-filter` does _only maybe_ support excluding path, it _might_ be tricky to do it properly (and not forget to list new paths when they are introduced).
I add a todo comment in the code (which prevents me from optimizing the line away).
(I found advanced globbing in picomatch, the library used by the action https://github.com/micromatch/picomatch#advanced-globbing)

The "testsuite-light" sounds nice. Grabbing a binary from master is not possible because there is no binary for ubuntu (on which the tests run).

SimonHeimberg: By the way, data/i18n/locales (containing only json files) and data/txts/developers.json should move to another place, they do not need not end up on the users computer (nor trigger a build or test).

SimonHeimberg:
fixing the action, it failed on the test run https://github.com/widelands/widelands/actions/runs/7093013489/job/19305558334?pr=6257

.github/workflows/build.yaml
SimonHeimberg: I made some tries with skipping directories inside data/, it seems to work, see https://github.com/SimonHeimberg/widelands/pull/6#issuecomment-1839597867

.github/workflows/build.yaml
SimonHeimberg: Why would the installers and test run when build.yaml changes? It is unlikely that a change is related to this jobs. (The file still has many jobs inside.) In case paths for filters are changed in build.yaml, this triggers a change usually because the file was changed as well.
No job checks for build.yaml currently.

.github/workflows/build.yaml
SimonHeimberg: The action is included now (which is not shown for this comment)

.github/workflows/build.yaml
tothxa: https://github.com/widelands/widelands/pull/6028#issuecomment-1817283718

.github/workflows/build.yaml
tothxa: > The "testsuite-light" sounds nice. Grabbing a binary from master is not possible because there is no binary for ubuntu (on which the tests run).

It just needs to be cached for the 1 or 2 configs we want to run "testsuite-light" on.

> By the way, data/i18n/locales (containing only json files) and data/txts/developers.json should move to another place, they do not need not end up on the users computer (nor trigger a build or test).

I'm not quite sure about that. They're next to the files that are generated from them. There really isn't any other good place for them, and they're probably not worth a new top-level directory.

.github/workflows/build.yaml
tothxa: I actually don't quite understand why you treat clang-tidy differently and why you put it in a separate directory.

.github/workflows/build.yaml
tothxa: But still not in this PR. :) To keep it reviewable, the changes to actual workflow logic should be kept to the absolute minimum required for separating the yaml files.

.github/workflows/build_codecheck.yaml
```diff
@@ -0,0 +1,56 @@
...
+jobs:
+  codecheck:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+      with:
+        fetch-depth: 1
+    - name: Check for changed files
+      uses: dorny/paths-filter@v2
```
tothxa: This should all be done in a single dispatcher jobs/step/whatever. The actual job files shouldn't double check whether they need to run, they should assume that they only get run when they are needed and when their dependencies succeeded.

.github/workflows/build.yaml
tothxa: It does run cmake... cmake step is lines 11-19 in your clang-tidy action.yaml.

And in any way, all checks on the C++ sources need to be run with the exact same conditions as compilation jobs.

.github/workflows/build.yaml
tothxa: I still haven't checked in detail, but I believe checking changes and deciding what to run should be a separate job, on which all other jobs should depend, and use its outputs in their conditions directly to decide whether they need to call their own yaml files. All this, and nothing else should be in the main yaml file.

.github/workflows/build.yaml
tothxa: Actually is there anything that is not checked for installers? Maybe we'll have to rethink whether we really want to create installers for all PR workflow runs, because that means almost half of the compilation jobs are always run.

.github/workflows/build.yaml
SimonHeimberg: This can be changed if the checking is a separate job, what you suggest. Because it just are some steps, it is a "composite action" and not a "workflow" now.

.github/workflows/build.yaml
SimonHeimberg: Agree, cmake IS run. I missed this.

.github/workflows/build.yaml
SimonHeimberg: Separation can be done in this way.
My previous strategy was to not add an additional job to not make full runs longer. But I just found out the action "dorny/paths-filter" can run without checkout (which takes about 15 s), so the extra time is only starting and stopping one job plus triggering the dependent ones, which should be very short.

.github/workflows/build_codecheck.yaml
SimonHeimberg: can be done this way, see [my comment above](#discussion_r1417946903)

.github/workflows/build.yaml
SimonHeimberg: Yes, there are README files, utils/, debian/, test/, (xdg/ for some), ...

Rethinking makes sense definitely. But in a seperate pr.

.github/workflows/build.yaml
SimonHeimberg: > [#6028 (comment)](https://github.com/widelands/widelands/pull/6028#issuecomment-1817283718)

This is why each job detects the changes of its file (build_XXX.yaml).

.github/workflows/build.yaml
SimonHeimberg: I chose the name "compile_and_dependencies".

.github/workflows/build.yaml
SimonHeimberg: I added the cmake files

.github/workflows/build.yaml
SimonHeimberg: Implemented, there is a separate job for **checking changes** now.

The full separation of each job in a separate file is not implemented yet. I suggest to do this in a next pr as this does not have anything to do with the goal of this pull request and will make review smaller.

.github/workflows/build.yaml
SimonHeimberg: Actually the pr title says the goal is to skip unnecessary jobs, not to separate the files. This point is already reached. And today's changes made separation better.

.github/workflows/build.yaml
SimonHeimberg: Is changed, it is a workflow now.

.github/workflows/build_codecheck.yaml
SimonHeimberg: done

.github/workflows/build.yaml
tothxa: Adding a dependency doesn't necessarily mean that that dependency is changed in the given PR.

But OK, if there are no actual changes in the code, and all build instructions are in the separated sub-workflows, then indeed, running them doesn't give much value, it only wastes time. The main yaml is validated anyway, even if it just concludes that everything can be skipped.

.github/workflows/build.yaml
tothxa: Well, separation makes skipping easier. :)

.github/workflows/build.yaml
```diff
@@ -9,6 +9,77 @@ on:
...
+      with:
+        # groups to check for changed files (as string, yaml anchors are resolved by the tool)
+        filters: |
+          codecheck:
+            - 'CMakeLists.txt'
+            - '*/CMakeLists.txt'
+            - 'cmake/codecheck/**'
+            - 'cmake/WlFunctions.cmake'
+            - 'src/**'
+            - '.github/scripts/install_deps.sh'
```
tothxa:
```suggestion
            - *compile_and_dependencies
```
except it'sdefined later

.github/workflows/build.yaml
```diff
@@ -9,6 +9,77 @@ on:
...
+            - '.github/workflows/build_codecheck.yaml'
+          clang_tidy:
+            - 'CMakeLists.txt'
+            - 'cmake/**'
+            - 'src/**'
```
tothxa:
```suggestion
            - *compile_and_dependencies
```
except it's defined later

SimonHeimberg: Right, this calls .../install_deps.sh as well, so it is the same. The name `compile_and_dependencies` does not really match. Anyway.

.github/workflows/build.yaml
SimonHeimberg: and "install dependencies" is addes as well, pushed to this branch soon

SimonHeimberg:
I listed some tasks of the comments here in the original discussion https://github.com/widelands/widelands/discussions/5880#discussioncomment-7790391 Please extend it if I forgot anything.

.github/workflows/build.yaml
```diff
@@ -37,20 +37,19 @@ jobs:
...
             - '.github/workflows/build_codecheck.yaml'
           clang_tidy:
-            - 'CMakeLists.txt'
-            - 'cmake/**'
-            - 'src/**'
-            - 'utils/check_clang_tidy_results.py'
-            - 'utils/run-clang-tidy.py'
-            - '.github/workflows/build_clang_tidy.yaml'
-          test_suite:
+            # this job does not compile, but uses cmake, install-dependencies.sh and source files
```
tothxa: Well, if you think  the name is a problem, then how about just "source"?

SimonHeimberg:
After merging into master, the config on github needs to be changed: all _installers_ and `test suites` do not need to pass anymore, but the summary step needs to.
Thereafter existing pull requests might have a problem because they have no summary step. For each one this will be solved after master is merged into. Should be good enough to do before merging them. (Not all in advance.)

.github/workflows/build.yaml
SimonHeimberg: I changed the name.

.github/workflows/build.yaml
SimonHeimberg: I removed it for now. There is a note to add them to the installers really including it.

.github/workflows/build.yaml
```diff
@@ -9,6 +9,76 @@ on:
...
+          codecheck:
+            - 'CMakeLists.txt'
+            - '*/CMakeLists.txt'
+            - 'cmake/codecheck/**'
+            - 'cmake/WlFunctions.cmake'
+            - 'src/**'
+            - '.github/scripts/install_deps.sh' # calls install-dependencies.sh
+            - '.github/workflows/build_codecheck.yaml'
+          clang_tidy:
+            - &src_cmake_dependencies
```
tothxa: Please create a dummy container filter (group?), e.g. `common:` where `&src_cmake_dependencies` and `&installer` are defined, to make both the definitions and the first users more readable.

.github/workflows/build.yaml
```diff
@@ -9,6 +9,76 @@ on:
...
+      with:
+        # groups to check for changed files (as string, yaml anchors are resolved by the tool)
+        filters: |
+          codecheck:
+            - 'CMakeLists.txt'
+            - '*/CMakeLists.txt'
+            - 'cmake/codecheck/**'
+            - 'cmake/WlFunctions.cmake'
+            - 'src/**'
+            - '.github/scripts/install_deps.sh' # calls install-dependencies.sh
```
tothxa: All this should be just `*src_cmake_dependencies`.

SimonHeimberg: I kept it separate because it only needs part of cmake. But I agree, it is not a big difference.

.github/workflows/build.yaml
SimonHeimberg: There is some overhead to calculate the change for an unused filter, but it is neglectable.

.github/workflows/build.yaml
```diff
@@ -9,6 +9,76 @@ on:
...
+            - *src_cmake_dependencies
+            - 'data/**' # TODO(aDiscoverer or ...) list what needs to be checked
+            - 'regression_test.py'
+            - 'test/**'
+            - '.github/scripts/install_deps.sh' # calls install-dependencies.sh
+            - '.github/workflows/build_testsuite.yaml'
+          inst_windows_msvc:
+            - &installer
+              - *src_cmake_dependencies
+              - 'data/**'
```
SimonHeimberg: add `utils/asio_not_packaged.sh`, or probably to some installers only. No idea which ones use it.
https://github.com/widelands/widelands/blob/897ab9fdd78a888f339ea0a64b356bb8d14e236c/install-dependencies.sh#L103

tothxa: It looks to me that only OpenBSD and the Solus and Void distros need `download_asio.sh`, neither of which is used in the workflow. So it can be removed from the checks.

.github/workflows/build.yaml
tothxa: I haven't looked much into how codecheck works, but it's another C++ source analyser, so it may be affected by anything that changes how the code is built. (well, OK, `cmake` also handles linking, but I don't think we can separate those parts)

.github/workflows/build.yaml
tothxa: Actually the example of shared filters at `dorny/paths-filter` does just this: defines a dummy, even with the same name as the filter itself. If we put the 2 shared filters into the same dummy container, then we even spare one compared to that.

.github/workflows/build.yaml
SimonHeimberg: codechecker is a python script.
The codechecker code only runs `cd build/; cmake .` (or `cmake build/`). For compiling, linking, ... it needs another command.
But let us use `*src_cmake_dependencies` for now. We might discuss this in an other pr (it is prepared).

.github/workflows/build.yaml
SimonHeimberg: It already was in `src_cmake_dependencies`, I forgot about.
https://github.com/widelands/widelands/blob/c6522ff016211307224c366ced1b8916399bfb77/.github/workflows/build.yaml#L38
So it could be removed.

.github/workflows/build.yaml
SimonHeimberg: This is pushed now.

.github/workflows/build.yaml
SimonHeimberg: I removed it, as I had to resolve a merge conflict anyway.

SimonHeimberg:
Resolving merge conflicts is not very complicated:
- look which commit changed .github/workflows/build.yaml on master: `git log origin/master -- .github/workflows/build.yaml`
- look in which job (in the workflow) the change is done
- apply the change: `git show $COMMIT_HASH -- .github/workflows/build.yaml | patch .github/workflows/build_JOB.yaml -`
- add and commit this change (referencing the commit hash)
- merge with master, resolving the conflict in .github/workflows/build.yaml (using our version)
- push this

But it would be nice to not have to do it many times more.

tothxa:
> Resolving merge conflicts is not very complicated:

Sorry about that. I very much hope we are good for a while now...

I think this is now OK, but I'd like to verify the changes once again and see a merge with translation updates only. (but now I'm inclined to approve even without the latter)

SimonHeimberg:
> > Resolving merge conflicts is not very complicated:
> 
> Sorry about that. I very much hope we are good for a while now...
> 
No worries.

> I think this is now OK, but I'd like to verify the changes once again and see a merge with translation updates only. (but now I'm inclined to approve even without the latter)

There will not be any difference in the test run. It is always about all the files changed compare to the merge base (master here), so no skipping will happen.
In case you want to see a skipped tests with current branch, I created a pr in my repo containing only the newest commit: https://github.com/SimonHeimberg/widelands/pull/8

tothxa:
> There will not be any difference in the test run. It is always about all the files changed compare to the merge base (master here), so no skipping will happen.

Oh, that decreases the efficiency of the whole thing a lot... I see there are feature requests and not entirely working PRs to this action to implement checking changes between PR commits too...

> In case you want to see a skipped tests with the current branch, I created a pr in my repo containing only the newest commit: [SimonHeimberg#8](https://github.com/SimonHeimberg/widelands/pull/8)

That run actually shows that `build.yaml` can still be an important dependency... ;)

bunnybot (commented):
tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4644#issuecomment-1434722)*
*On Thu Jan 11 00:49:29 CET 2024, **Tóth András (tothxa)** **approved** this pull request:*

I think this already helps some cases and gives a reasonable framework for further improvements. All new changes should go to new PRs.

<@>Nordfriese Could you please have a look at this too? Much of the discussion is not mirrored yet, so please have a look at github too. (no need to log in for that :)

bunnybot (commented):
Nordfriese*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4644#issuecomment-1435430)*
*On Thu Jan 11 11:04:54 CET 2024, **Benedikt Straub (Nordfriese)** **approved** this pull request:*

Looks good to me as far as I can tell, though I lost track of the details a bit ;P 
Feel free to tell bunnybot to merge

SimonHeimberg:
> > There will not be any difference in the test run. It is always about all the files changed compare to the merge base (master here), so no skipping will happen.
> 
> Oh, that decreases the efficiency of the whole thing a lot... I see there are feature requests and not entirely working PRs to this action to implement checking changes between PR commits too...
> 

On a __pr__ (its creation or its change), only the pr is checked, but always the whole one.
On commit on a __branch__ (like on master), the new changes are checked.
This is close to what I expect. (But I agree it is not handy for the pr introducing it.)

> > In case you want to see a skipped tests with the current branch, I created a pr in my repo containing only the newest commit: [SimonHeimberg#8](https://github.com/SimonHeimberg/widelands/pull/8)
> 
> That run actually shows that `build.yaml` can still be an important dependency... ;)

With pr XXX (branch validate-github-workflow) there is a tool to check if the filters still match the workflow. (Will get included in a check once all the workflows are split up.) So such changes of paths will be done in the pr.

:arrow_right: Do not forget to [update the required checks](https://github.com/widelands/widelands/pull/6257#issuecomment-1853569509).

bunnybot (commented):
matthiakl*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4644#issuecomment-1435883)*
*On Thu Jan 11 18:26:47 CET 2024, ** (matthiakl)** **approved** this pull request:*

Fine for me as well.
For the upcoming PRs we should check that the expected runs were triggered, to verify the filters.
I like the clean structure of the CI runs now 👍

> After merging into master, the config on github needs to be changed: all installers and test suites do not need to pass anymore, but the summary step needs to.

Is this something for the bunnybot configuration now?

tothxa:
Oh, great, codeberg is down... Does the bot still work? Anyway, let's hope for the best, and get this out of the way...

@bunnybot merge

tothxa:
@bunnybot update

Noordfrees:
Codeberg is still partially down after a DDos attack, so the bot is unable to act until it can access Codeberg's API again. This also includes one-sided update orders. There's no timeline how long until it works normally again.
See https://status.codeberg.eu/status/codeberg


P.S. this comment was written via GitHub's API :)

tothxa:
Hmm, looks like things should be kind of working now...

@bunnybot merge

(wasn't he supposed to do it when the workflow succeeded?)

bunnybot (commented):
tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4644#issuecomment-1437096)*
*On Fri Jan 12 23:54:57 CET 2024, **Tóth András (tothxa)** **approved** this pull request.*

SimonHeimberg:
> > Is this something for the bunnybot configuration now?
> 
> There are no required checks on GitHub because we don't use GitHub for merging any more. The bot requires all checks to succeed or be skipped, there is no explicit list of required vs non-required checks.
> 

Good to know. If I am not wrong, this means the summary step is not needed. It was just needed for the required steps which were set on github.

PR #6261, opened 2023-12-04
Fix building on FreeBSD / OpenBSD

brad0:
OS's that use compiler-rt do not use libatomic like a traditional
libgcc based runtime environment does.

### DIFF
--- a/cmake/WlFunctions.cmake
+++ b/cmake/WlFunctions.cmake
@@ -112,7 +112,8 @@ macro(_common_compile_tasks)
       endif()
   endif()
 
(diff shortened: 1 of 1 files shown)


bunnybot (commented):
Nordfriese*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4600#issuecomment-1373057)*
*On Sun Dec 10 15:19:56 CET 2023, **Benedikt Straub (Nordfriese)** **approved** this pull request:*

Thanks for the patch :)

<@>bunnybot merge force

PR #6262, opened 2023-12-05
Documentation: place_road() needs a flag, not a field

frankystone:
Fixes a mistake in the documentation for `place_road()` which needs a flag as start, not a field. See https://www.widelands.org/documentation/autogen_wl_bases/#wl.bases.PlayerBase.place_road
Additionally:
* added an example because the road directions in `place_road()` are comma separated strings whereas `connected_road()` uses all directions in one string.
* added a note that `connected_road()` does not return the road, in contrary to `place_road()`

**Issue(s) closed**
The wrong documentation of those two functions took me several hours of "Why the hell is this not working??" :-D

### DIFF
--- a/data/scripting/infrastructure.lua
+++ b/data/scripting/infrastructure.lua
@@ -46,6 +46,9 @@
 --    :arg create_carriers: If this is :const:`true` carriers are created for
 --       the roads. Otherwise no carriers will be created.
 --    :type create_carriers: :class:`boolean`
+--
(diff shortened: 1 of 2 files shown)


hessenfarmer (approved):
LGTM

hessenfarmer:
@bunnybot merge

hessenfarmer:
@bunnybot merge force

PR #6264, opened 2023-12-06
win condition artifacts ignores hidden players

SimonHeimberg:
**Type of change**
New feature

**Issue(s) closed**
Fixes -

**New behavior**
For addons giving a special job to a player and hiding it therefore, it does not make sense to let this player get points or even win.

**How it works**
Artifacts owned by a hidden player are not counted.

**Possible regressions**
I did not find any existing use of a hidden player which could be combined with win condition "artifacts".

**Additional context**
My addon [artifacts-defended](https://github.com/widelands/wl_addons_server/tree/master/addons/artifacts-defended.wad) can not be played with win condition "artifacts" without this change.

### DIFF
--- a/data/scripting/win_conditions/artifacts.lua
+++ b/data/scripting/win_conditions/artifacts.lua
@@ -128,7 +128,7 @@ local r = {
          check_player_defeated(plrs, lost_game.title, lost_game.body, wc_descname, wc_version)
          local all_artifacts_found = true
          for idx, f in ipairs(artifact_fields) do
-            if f.owner then
+            if f.owner and not f.owner.hidden_from_general_statistics then -- owner is normal player
                if artifacts_owner[f] then
                   if f.owner.team ~= artifacts_owner[f].team or (f.owner.team == 0 and f.owner.number ~= artifacts_owner[f].number) then
                      -- a new team has this artifact
@@ -142,6 +142,7 @@ local r = {
                end
             else
                all_artifacts_found = false
+               artifacts_owner[f] = nil -- delete in case owned by hidden player
             end
          end
       until all_artifacts_found


data/scripting/win_conditions/artifacts.lua
```diff
@@ -128,7 +128,7 @@ local r = {
          check_player_defeated(plrs, lost_game.title, lost_game.body, wc_descname, wc_version)
          local all_artifacts_found = true
          for idx, f in ipairs(artifact_fields) do
-            if f.owner then
+            if f.owner and not f.owner.hidden_from_general_statistics then -- owner is normal player
```
tothxa, line 131: I think `hidden_from_general_statistics` has a different purpose. (mostly for scenarios to increase fog of war)

OTOH as this doesn't require any code changes, you can package your own modified version of the artifacts win condition as an add-on, and your `artifacts-defended` add-on can depend on it.

tothxa: I've looked at the `Player` class in the scripting reference, and I think you should use an objective to uniquely identify the defender. But any change to the win condition for the sake of an add-on should still be in an add-on too.

SimonHeimberg: Objectives are shared between players (see [doc]), so they can just be used for the interactive one.

The attribute `hidden_from_general_statistics` has the purpose the hide the player in the general statistics window. In a scenario, you do not know in advance that there is another player nor anything about his strength.
I use this attribute in the addon because the players shall not know how many artifacts are currently defended by the "defending player", they shall not know if other players have found an attribute but not yet conquered it. They shall not see the military strange of the defender neither (and be able to guess how many soldiers are still in one artifact).

My assumption was that a hidden player should not be able to win a game. This would be a unexpected surprise. I assumed if a player is hidden from statistic, he is not (yet) supposed to win.

tothxa: > Objectives are shared between players (see [doc]), so they can just be used for the interactive one.

I've tested it, and they're indeed shared, but that wasn't at all clear to me from the docs. `add_objective()` is a method of the `Player` object, so one would think it's associated with that player, and even https://www.widelands.org/documentation/autogen_wl_game/#wl.game.Player.add_objective suggests that different players can have different objectives. So the latter should clearly be changed.

> I assumed if a player is hidden from statistic, he is not (yet) supposed to win.

Why not?

SimonHeimberg: > > Objectives are shared between players (see [doc]), so they can just be used for the interactive one.
> 
> I've tested it, and they're indeed shared, but that wasn't at all clear to me from the docs. `add_objective()` is a method of the `Player` object, so one would think it's associated with that player, and even https://www.widelands.org/documentation/autogen_wl_game/#wl.game.Player.add_objective suggests that different players can have different objectives. So the latter should clearly be changed.
> 
👍 to your suggestions

> > I assumed if a player is hidden from statistic, he is not (yet) supposed to win.
> 
> Why not?

I can not find a case to hide a single player.
But I wrote an extension just to show you are right. (Not yet published.) Hiding statistics of all players (except of yours) increases "fog". (Additional idea: Having a scout and having seen the main quarter could show statistics of this player.)

If I have to copy the win condition, I will try to make it compatible to the extension "hidden artifacts" which I will publish once I manage not only the logics but also the statistics to make it an interesting game.

hessenfarmer: ok, @SimonHeimberg I agree with @tothxa that you should provide the appropriate win condition for your addon with the defended artifacts in the adon itself. maybe you might name it "Defended Artifacts", as I believe there might be a good chance for this to enter the game after some public testing via the addon.       
   
If you agree to that please close this PR.
BTW:      
In a scenario a hidden player might be supposed to win, e.g. Imagine a undeclared race for an objective or an objective to achieve before others which you do not know yet.     
It is the duty of the scenario code to unhide the player as soon as he is seen.

SimonHeimberg: I closed the PR.

SimonHeimberg: I published mentioned addons:
- new version of artifacts-defended.wad, depending on next addon
- artifacts-extendable.wad
- no-stats.wad
- no-stats-till-spied.wad

PR #6273, opened 2023-12-15
Make spinbox current value and change-by value translatable separately

### LINKED ISSUE #6267: Spinbox unit_text() translation conflict
tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/issues/4606)*
*Created on Sun Dec 10 17:45:50 CET 2023 by **Tóth András (tothxa)***

---

**Describe the bug**
Since 1034766 the spinbox button tooltips include the amount of change with units when the spinbox is using units. However, to do this, it uses the `unit_text()` member function, that is also used to display the current value of the spinbox. So the translatable unit name returned by `unit_text()` must work both in nominative case and in whatever case or form is required to translate "{in,de}crease by *n* [units]", or some rather convoluted translation is required for the "{in,de}crease the value by %s" format strings.

**Expected behavior**
"*n* [unit(s)]" and "by *n* [unit(s)]" should be translatable separately, so either `unit_text()` should be changed to have 2 modes, or a separate new function should be added to return the latter form. (Maybe some languages even require different forms after increase than after decrease?)

**Version:**
 - OS: n/a
 - Widelands Version: any since 1034766 (one of the first commits in the v1.2 cycle)
 - Enabled Add-Ons: n/a

**Additional context**
I've been planning to address this for a while, but I have too many open PRs waiting for review, so now with the freeze coming, I'd like to know whether we can fix this for v1.2.

bunnybot:
[tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4612)*
*Created on Fri Dec 15 20:15:14 CET 2023 by [**Tóth András (tothxa)**](https://codeberg.org/tothxa)*

---

**Type of change**
i18n fix

**Issue(s) closed**
Fixes #6267

**Possible regressions**
none?

**Additional context**
I chose to not use "by %d [unit(s)]" for English, so it uses the same strings only in different contexts. I hope this makes things easier for other languages that don't need different forms, so they can just copy the existing translations to the new contexts.

Should the full text of "{In,De}crease value by %d [unit(s)]" be translatable instead? Do we know of any language that needs different forms for increase and decrease?


src/ui_basic/spinbox.cc
```diff
@@ -469,46 +473,111 @@ void SpinBox::add_replacement(int32_t value, const std::string& text) {
...
 
-const std::string SpinBox::unit_text(int32_t value) const {
+const std::string SpinBox::unit_text(int32_t value, const bool change) const {
 	switch (sbi_->unit) {
 	case (Units::kMinutes): {
 		if (value < 60) {
-			/** TRANSLATORS: A spinbox unit */
-			return format(ngettext("%d minute", "%d minutes", value), value);
+			// auto-formatting completely messes up the indentation of these comments
+			// clang-format off
```
bunnybot: Nordfriese*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4612#issuecomment-1417439)*
*On Thu Jan 04 17:55:52 CET 2024, **Benedikt Straub (Nordfriese)** wrote:*

---

Please don't do this. Shouldn't this work:
```c++
return change ? 
	format(
		/** TRANSLATORS: ... */
		npgettext(...)
	)
	:
	format(
		/** TRANSLATORS: ... */
		npgettext(...)
	)
;
```

bunnybot (commented):
Nordfriese*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4612#issuecomment-1417441)*
*On Thu Jan 04 17:55:52 CET 2024, **Benedikt Straub (Nordfriese)** commented:*

Looks safe, one comment inline, not tested

bunnybot (commented):
Nordfriese*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4612#issuecomment-1418173)*
*On Fri Jan 05 10:21:07 CET 2024, **Benedikt Straub (Nordfriese)** **approved** this pull request:*

LGTM :) 

<@>bunnybot merge

PR #6279, opened 2023-12-22
Fix building with external minizip

### LINKED ISSUE #6266: Cannot build with external minizip
**Describe the bug**
Building with external minizip does not work. CMake does not pass along the library path.

**To reproduce**
On a non-Linux OS..
CMake...
make / ninja..
..builds..
..gets to linking stage..
..fails..

**Expected behavior**
CMake passes along the library path and links properly instead of passing a library name without the path.

**Version:**
 - OS: OpenBSD. Applicable to anything non-Linux.
 - Widelands Version: v1.1, but also applies to ToT git.
 - Enabled Add-Ons: whatever the default build is without any additional parameters.

```
ld: error: unable to find library -lminizip
```

Looking at cmake/WlFunctions.cmake I see the header path being added but not the library path.

brad0:
The CMake bits were passing along the header path but not the library path. Building on a non-Linux system or one where the path is other than /usr/local would fail.

**Type of change**
Bugfix

**Issue(s) closed**
Fixes #6266

**To reproduce**
On a non-Linux OS..
CMake...
make / ninja..
..builds..
..gets to linking stage..
..fails..

**New behavior**
Passes path to minizip library when linking.


tothxa (approved):
Thank you. Looks OK, and it works on my Debian system too.

Though I don't know cmake and pkg-config...

tothxa:
BTW, maybe we should add building on OpenBSD to the workflow too? If possible, also running the testsuite. Does github offer runners for it?

cmake/WlFunctions.cmake
```diff
@@ -104,7 +104,7 @@ macro(_common_compile_tasks)
 
   if(ARG_USES_MINIZIP)
       if(MINIZIP_STATIC_LIBRARIES)
-          target_link_libraries(${NAME} minizip)
+          target_link_libraries(${NAME} ${pkgcfg_lib_MINIZIP_minizip})
           message(STATUS "Link ${NAME} with minizip")
```
matthiakl: I would prefer to use `MINIZIP_LINK_LIBRARIES` as it is a documented variable, I could not find anything about `pkgcfg_lib_*` in cmake documentation (https://cmake.org/cmake/help/latest/module/FindPkgConfig.html#command:pkg_check_modules). 
```suggestion
          target_link_libraries(${NAME} ${MINIZIP_LINK_LIBRARIES})
          message(STATUS "Link ${NAME} with ${MINIZIP_LINK_LIBRARIES}")
```
Does this still work for you?

matthiakl (commented):
> BTW, maybe we should add building on OpenBSD to the workflow too? If possible, also running the testsuite. Does github offer runners for it?

No. There is an unofficial cross-platform environment (https://github.com/cross-platform-actions/action), but I don't know whether we really want the trouble.

cmake/WlFunctions.cmake
brad0: > I would prefer to use `MINIZIP_LINK_LIBRARIES` as it is a documented variable, I could not find anything about `pkgcfg_lib_*` in cmake documentation (https://cmake.org/cmake/help/latest/module/FindPkgConfig.html#command:pkg_check_modules).
> 
> Does this still work for you?

I would have used that if I had seen it in CMakeCache.txt. The contents of the variable are closer to what I was looking for.

```
-- Link io_filesystem with /usr/local/lib/libminizip.so.2.0;/usr/lib/libz.so.7.0
```

brad0:
Your bot is quite strange when it pushes commits that break the submitted patch.

matthiakl:
Maybe it is his way to tell you that force-pushing is not always appropriate…

brad0:
> BTW, maybe we should add building on OpenBSD to the workflow too? If possible, also running the testsuite. Does github offer runners for it?

I wouldn't worry about it. Overall things are good. 

The only other thing I wanted to look into was how to fix the man page path due to the only other local patch we have.

klaus-halfmann:
> Your bot is quite strange when it pushes commits that break the submitted patch.

@brad0 I hope you are aware that our repo is mirrored to https://codeberg.org/wl/widelands
 I was hit my similar Issues when I tried to add multiple tags in quick sucession.

See: https://www.widelands.org/wiki/GitPrimer/ 

We might remove the part about bazaar, I already forgor what this was 🤔

tothxa:
@matthiakl I think this is good now. Can we merge it?

@bunnybot update

tothxa (approved):

matthiakl (approved):
Yes, it is fine now.

brad0:
> Maybe it is his way to tell you that force-pushing is not always appropriate…

Thanks. I'll keep that in mind. I'm used to other projects where they would either print a message of some kind or just format but based on the latest commits.

brad0:
> @brad0 I hope you are aware that our repo is mirrored to https://codeberg.org/wl/widelands I was hit my similar Issues when I tried to add multiple tags in quick sucession.
> 
> See: https://www.widelands.org/wiki/GitPrimer/
> 
> We might remove the part about bazaar, I already forgor what this was 🤔

Oh. I was not aware of that. Thanks.

Can't say as I have ever run into anyone using Bazaar other than Canonical even years ago.

hessenfarmer:
@bunnybot update

hessenfarmer:
@bunnybot merge

brad0:
Was this to be commited?

tothxa:
> Was this to be commited? That is to say I thought merge would mean commited.

Yes it means that, but our bot is busy with re-mirroring everything because of a fix, which is extremely slow because of rate limiting, so it didn't get around to handling PRs while doing that.

brad0:
> Yes it means that, but our bot is busy with re-mirroring everything because of a fix, which is extremely slow because of rate limiting, so it didn't get around to handling PRs while doing that.

Good to know. Thanks.

matthiakl:
@bunnybot update

matthiakl:
@bunnybot merge

Issue #6275, opened 2023-12-20
msys linking failure: libSDL2-image IMG-avif missing?


tothxa:
**Describe the bug**
mingw build failure in the workflow.

**To reproduce**
See https://github.com/widelands/widelands/actions/runs/7279661441/job/19836541027?pr=6150

**Build log**
```
[2352/2355] Linking CXX executable src\ai\test\test_ai.exe
FAILED: src/ai/test/test_ai.exe 
C:\Windows\system32\cmd.exe /C "cd . && C:\msys64\mingw32\bin\g++.exe -O3 -DNDEBUG -Wl,--large-address-aware @CMakeFiles\test_ai.rsp -o src\ai\test\test_ai.exe -Wl,--out-implib,src\ai\test\libtest_ai.dll.a -Wl,--major-image-version,0,--minor-image-version,0  && cd ."
C:/msys64/mingw32/bin/../lib/gcc/i686-w64-mingw32/13.2.0/../../../../i686-w64-mingw32/bin/ld.exe: C:/msys64/mingw32/lib/libSDL2_image.a(libSDL2_image_la-IMG_avif.o):(.text+0xdd): undefined reference to `_imp__avifDecoderSetIO'
C:/msys64/mingw32/bin/../lib/gcc/i686-w64-mingw32/13.2.0/../../../../i686-w64-mingw32/bin/ld.exe: C:/msys64/mingw32/lib/libSDL2_image.a(libSDL2_image_la-IMG_avif.o):(.text+0xe5): undefined reference to `_imp__avifDecoderDestroy'
C:/msys64/mingw32/bin/../lib/gcc/i686-w64-mingw32/13.2.0/../../../../i686-w64-mingw32/bin/ld.exe: C:/msys64/mingw32/lib/libSDL2_image.a(libSDL2_image_la-IMG_avif.o):(.text+0xed): undefined reference to `_imp__avifDecoderCreate'
C:/msys64/mingw32/bin/../lib/gcc/i686-w64-mingw32/13.2.0/../../../../i686-w64-mingw32/bin/ld.exe: C:/msys64/mingw32/lib/libSDL2_image.a(libSDL2_image_la-IMG_avif.o):(.text+0xf9): undefined reference to `_imp__avifPeekCompatibleFileType'
[...]

PR #6281, opened 2023-12-25
New Msys dependencies

hessenfarmer:
Finally the underlying issue of #6275 has been fixed in the MSYS repo. 
so we are finally able to build again without downgrading


.github/workflows/build.yaml
```diff
@@ -330,7 +330,7 @@ jobs:
         pacman.exe --noconfirm -Syyuu --ignore ca-certificates
         pacman.exe --noconfirm -Syyuu --ignore ca-certificates
         bash.exe -c "./install-dependencies.sh msys${{ matrix.bits }} --noconfirm"
-        pacman.exe --noconfirm -U https://repo.msys2.org/mingw/mingw${{ matrix.bits }}/mingw-w64-${{ matrix.archname }}-SDL2_image-2.8.0-1-any.pkg.tar.zst
+        # pacman.exe --noconfirm -U https://repo.msys2.org/mingw/mingw${{ matrix.bits }}/mingw-w64-${{ matrix.archname }}-SDL2_image-2.8.0-1-any.pkg.tar.zst
```
tothxa, line 334:
```suggestion
        # Downgrade incompatible packages to working version
        # pacman.exe --noconfirm -U https://repo.msys2.org/mingw/mingw${{ matrix.bits }}/mingw-w64-${{ matrix.archname }}-SDL2_image-2.8.0-1-any.pkg.tar.zst
```

hessenfarmer:
works only for 64 bit. so lets wait if this will get fixed in the MSYS packages. (problem is already reported there

hessenfarmer:
ok it seems we might need to deprecate 32 bit Windows builds at some time in the future (Maybe after 1.2) 
see https://github.com/msys2/MINGW-packages/issues/19510

tothxa (approved):
Thank you for all your work to sort this out.

It built successfully, and the changes are not unreasonable. I still find it unfortunate though that we now have such a long list of dependencies forced on us that we don't actually use ourselves.

matthiakl (approved):
> ok it seems we might need to deprecate 32 bit Windows builds at some time in the future (Maybe after 1.2)

+1

> I still find it unfortunate though that we now have such a long list of dependencies forced on us that we don't actually use ourselves.

At least the current setup allows for relatively easy building on Windows. If we compile the packages ourselves (like we do for MSVC) there is additional effort required to get a development environment running. Dynamic linking may pull less dependencies in, but causes other issues which are not alwalys found during build, but only during runtime.

@bunnybot update

matthiakl:
@bunnybot update

matthiakl:
@bunnybot merge

PR #6287, opened 2024-01-01
Install fixed version of 32bit glbinding

### LINKED ISSUE #6285: Workflow mingw build doesn't find `ninja` and C/C++ compiler
**Describe the bug**
Workflow builds are failing in mingw job with:
```
CMake Error: CMake was unable to find a build program corresponding to "Ninja".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.
CMake Error: CMAKE_C_COMPILER not set, after EnableLanguage
CMake Error: CMAKE_CXX_COMPILER not set, after EnableLanguage
-- Configuring incomplete, errors occurred!
```

**To reproduce**
See:
 - https://github.com/widelands/widelands/actions/runs/7372058568/job/20060435425
 - https://github.com/widelands/widelands/actions/runs/7371563696/job/20059372674?pr=6257
 - https://github.com/widelands/widelands/actions/runs/7371817137/job/20059878798?pr=6282

**Additional context**
First happened in https://github.com/widelands/widelands/actions/runs/7348943728/job/20010113598 but went away on re-running the job. But now it looks like all workflows are failing on this?

matthiakl:
**Issue(s) closed**
Fixes #6285 

**New behavior**
Install the last known version of glbindings for 32bit

**Additional context**
This step will become necessary for other 32bit packages in the future

### DIFF
--- a/install-dependencies.sh
+++ b/install-dependencies.sh
@@ -126,7 +126,7 @@ install_msys_pkgs() {
    pacman -S $@ mingw-w64-$ARCH-toolchain git mingw-w64-$ARCH-cmake \
     mingw-w64-$ARCH-ninja mingw-w64-$ARCH-asio mingw-w64-$ARCH-SDL2_ttf \
     mingw-w64-$ARCH-SDL2_mixer mingw-w64-$ARCH-SDL2_image \
-    mingw-w64-$ARCH-glbinding mingw-w64-$ARCH-glew mingw-w64-$ARCH-icu
+    mingw-w64-$ARCH-glew mingw-w64-$ARCH-icu
 }
 
 # Install the dependencies
@@ -178,10 +178,12 @@ elif [ "$DISTRO" = "openbsd" ]; then
 elif [ "$DISTRO" = "msys32" ]; then
    echo "Installing dependencies for 32-bit Windows..."
    install_msys_pkgs i686 $@
+   # No more updates for 32bit version...
+   pacman.exe --noconfirm -U https://repo.msys2.org/mingw/i686/mingw-w64-i686-glbinding-3.1.0-3-any.pkg.tar.zst
 
 elif [ "$DISTRO" = "msys64" ]; then
    echo "Installing dependencies for 64-bit Windows..."
-   install_msys_pkgs x86_64 $@
+   install_msys_pkgs x86_64 $@ mingw-w64-$ARCH-glbinding
 
 elif [ "$DISTRO" = "homebrew" ]; then
    echo "Installing dependencies for Mac Homebrew..."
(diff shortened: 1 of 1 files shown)


tothxa:
Also doing this in `install-dependencies.sh` will not play nice when #6257 is merged.

matthiakl:
> Can't we use a whole fixed build environment?

You mean to use `pacman -U  https://repo.msys2.org/mingw/i686/mingw-w64-i686-*` for all packages with the currently available version?

> Also doing this in install-dependencies.sh will not play nice when https://github.com/widelands/widelands/pull/6257 is merged.

Trying to install the glbindings packages within the script is causing the error, because it is not longer available in the current repo. So there must be a change in the script.

tothxa:
> > Can't we use a whole fixed build environment?
> 
> You mean to use `pacman -U https://repo.msys2.org/mingw/i686/mingw-w64-i686-*` for all packages with the currently available version?

I don't know pacman, nor msys, nor mingw... Does that mean specifying the version of each package one by one? That would be... suboptimal. Doesn't msys have actual releases or something? Or a specific commit of the package list that we can pin, and get versions from there?

> > Also doing this in install-dependencies.sh will not play nice when #6257 is merged.
> 
> Trying to install the glbindings packages within the script is causing the error, because it is not longer available in the current repo. So there must be a change in the script.

Actually this issue got me thinking about that script, whether we should move the actual package lists for each OS/distro into separate config files, at least for the ones we use in the workflows. But as long as they don't change often it's not a major problem. The problem here is the individual package version pinning, that is likely to need more changes soon.

matthiakl:
> Does that mean specifying the version of each package one by one?

Yes.

> Doesn't msys have actual releases or something?

No, it works like other packaging systems on Linux distributions, there is a package index which lists all current versions, but you cannot load an older one.

> The problem here is the individual package version pinning, that is likely to need more changes soon.

We can extract the pinned versions for 32bit into a separate file, all other dependencies rarely change.

tothxa:
> No, it works like other packaging systems on Linux distributions, there is a package index which lists all current versions, but you cannot load an older one.

Well, that sounds more like "appstores" as far as I'm concerned. Most GNU/linux distros have stable releases and archives of them.

OK, I've had a look at https://repo.msys2.org/mingw/i686/. So they do keep some older versions around, but it's not clear how long. Do you think they'll keep that repo itself when they stop supporting 32bit?

So I can't add much to the solution to this problem, so I can only summarise my whishlist, that you may ignore if you like:

 - The solution should ensure that 32bit mingw builds work in the workflow as long as possible with as little later maintenance as possible, while we can expect that upstream will keep dropping our dependencies.
 - It looks to me that we have to prepare for dropping 32bit mingw, so the "as long as possible" above means "preferably at least until WL v1.2" (whenever that may happen). So using ugly hacks may be OK if they're needed to eliminate maintenance burden.

hessenfarmer (approved):
seems to work and provides a template for a workaround

hessenfarmer:
@bunnybot merge

PR #6290, opened 2024-01-02
mingw dependency system

matthiakl:
**Type of change**
New feature

**New behavior**
As requested by @tothxa in https://github.com/widelands/widelands/pull/6287#issuecomment-1873497978 this provides a system to react semi-automatically on ceased 32bit packages.

**How it works**
The msys/mingw package dependencies are read from a file. If the install script does not find a package in the default repo (as it happened for glbinding) it tries to locate an older version on the mirror and installs that one together with its manually listed dependencies.
The file parser also checks for version suffixes (SDL2_image=2.8.0-1), to allow easy rollback in case of breaking through an update (as it happend for SDL_image), see https://github.com/matthiakl/widelands/actions/runs/7386349276/job/20092832790#step:3:446

**Possible regressions**
Other CI breakings


bunnybot (commented):
tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4653#issuecomment-1418395)*
*On Fri Jan 05 13:27:27 CET 2024, **Tóth András (tothxa)** **approved** this pull request:*

Thank you. Looks good and worked well in the workflow. I also tested the script locally by just echoing the `pacman` commands, and also using `wget` to download the packages instead of `pacman -U`.

PR #6309, opened 2024-01-13
Fix "changes" workflow step in master

tothxa:
**Type of change**
Bugfix

**Issue(s) closed**
"changes" step failing in workflow for master, and said failure being ignored by summary

**To reproduce**
See https://github.com/widelands/widelands/actions/runs/7508842696/job/20445046588

**Possible regressions**
More workflow breakage

**Additional context**
From https://github.com/dorny/paths-filter/blob/master/README.md

> Master, Release, or other long-lived branches:
> Uses git commands to detect changes - repository must be already [checked out](https://github.com/actions/checkout)

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -6,8 +6,7 @@ on:
   pull_request:
     types: [ opened, reopened, synchronize ]
   push:
-    # NOCOM
-    branches: [ master, fix-workflow-changes-step ]
+    branches: [ master ]
 
 jobs:
   changes:
@@ -25,8 +24,7 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v3
-      # NOCOM
-      if: github.ref == 'refs/heads/master' || github.ref == 'refs/heads/fix-workflow-changes-step'
+      if: github.ref == 'refs/heads/master'
       with:
         fetch-depth: 1
     - name: Check for changed files
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -24,10 +24,8 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v3
-      # TODO(tothxa): needed for testing in PR, uncomment before merge
-      # if: github.ref == 'refs/heads/master'
+      if: github.ref == 'refs/heads/master'
       with:
-        # TODO(tothxa): this is probably wrong
         fetch-depth: 1
     - name: Check for changed files
       uses: dorny/paths-filter@v2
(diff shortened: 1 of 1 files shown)


tothxa:
Maybe `fetch-depth: 1` is OK, as that's the default and this is the example for master from above readme:

```
jobs:
  build:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v3
    - uses: dorny/paths-filter@v2
      id: filter
      with:
        # Use context to get the branch where commits were pushed.
        # If there is only one long-lived branch (e.g. master),
        # you can specify it directly.
        # If it's not configured, the repository default branch is used.
        base: ${{ github.ref }}
        filters: ... # Configure your filters
```

matthiakl:
> Maybe fetch-depth: 1

No, the tool needs at least 2 commits to get the diff between them, otherwise all files would have changed.

> But we probably need base:?

Against master should be correct in this case.

tothxa:
> > Maybe fetch-depth: 1
> 
> No, the tool needs at least 2 commits to get the diff between them, otherwise all files would have changed.

Looks like the doc is correct, and default branch is treated differently, it looks up previous commit in main repo if needed. See https://github.com/tothxa/widelands/actions/runs/7511688331/job/20451551855

There is a huge fundamental problem though: What if there have been more than 1 commits since the last completed workflow run? Also, if last one was a failure, unchanged failed steps should be marked failed. But that needs more possible states for the output (at least needs-running/skip-as-failed/skip-as-success).

I guess we can try to solve at least the first one (commit to compare against) by caching results and using `base:` with the cached commit hash. Maybe the 3-state thing can also be solved by caching and adding a little bit of our own logic?

.github/workflows/build.yaml
```diff
@@ -176,7 +181,8 @@ jobs:
   # (see https://github.com/actions/runner/issues/952 and https://github.com/orgs/community/discussions/9141)
   summary:
     name: CI Results Summary
-    needs: [testsuite, windows, macos, appimage]  # We do not require MSVC due to frequent unrelated runner failures
+    # "changes" has to be included in needs here too, because all others are skipped if it fails
+    needs: [changes, testsuite, windows, windows-msvc, macos, appimage]
```
matthiakl, line 185: So, do you consider msvc stable now? At least recently it was more stable than mingw...

matthiakl (approved):
>  it looks up previous commit in main repo if needed.

Okay

>  There is a huge fundamental problem though

Such ideas always come a bit late ;). 

> so we can probably live with it for a while. (ie. not necessary for v1.2)

So let's hope that the master build stay green.

.github/workflows/build.yaml
tothxa: Yes, we have now mitigations for instabilities in both. And also we hadn't excluded it before either. I just forgot to request removing that exception from the previous PR in time.

tothxa:
Thanks for the review. :)

@bunnybot merge

PR #6314, opened 2024-01-15
Testsuite skip irrelevant testruns

SimonHeimberg:
**Type of change**
Enhancement of testsuite

**Issue(s) closed**
Enhances #5880

**New behavior**
The job testsuite is skipped for changes in irrelevant directories.

**How it works**
For the list of dorny/paths-filter, it uses general excludes for some subdirectories. Some files inside are explicitly included after.

**Possible regressions**
I do not think I excluded too much, but I might have.

**Additional context**


.github/workflows/build.yaml
```diff
@@ -57,7 +57,15 @@ jobs:
             - '.github/workflows/build_clang_tidy.yaml'
           test_suite:
             - *src_cmake_dependencies
-            - 'data/**' # TODO(aDiscoverer) list what needs to be checked
+            # ignore some subdirectories
+            - 'data/!(campaigns|maps|music|sound|tribes|world)/**'
+            # but track parts of them
+            - 'data/tribes/**.lua' # track lua files
```
bunnybot: tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4677#issuecomment-1505241)*
*On Mon Jan 15 13:53:50 CET 2024, **Tóth András (tothxa)** wrote:*

---

This is not right. We must test new images too.

bunnybot (commented):
tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4677#issuecomment-1505242)*
*On Mon Jan 15 13:53:49 CET 2024, **Tóth András (tothxa)** commented.*

.github/workflows/build.yaml
hessenfarmer: @cb/tothxa what do you mean with testing images. which of our testsuite testcases does check images?

for me this selection of relevant files for our testsuite does make sense

hessenfarmer:
> [tothxa](https://codeberg.org/tothxa)_[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4677#issuecomment-2511762)_ _On Thu Dec 12 14:06:09 CET 2024, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) wrote:_
> 
> OK, so replies in this case are not mirrored... Here's what I wrote:
> 
> > what do you mean with testing images. which of our testsuite testcases does check images?
> 
> All of them? I mean, removing an image file of any item will cause the game to fail loading. Also the lua definitions may include references to new graphics, which are then not actually committed to git, with the same effect. _edit: Sorry, that's wrong. Actually the lua change itself will be checked and will fail, but if we implement changed checks within PRs (which we should), then the fix adding the images will not be checked._
> 
> > for me this selection of relevant files for our testsuite does make sense
> 
> I think at least all of `data/tribes` and `data/world` must be checked. Sounds and music are different, since the testsuite runs with sound disabled, so problems there are probably missed completely. :( Same for translations.

well we do have only *.lua files and *.png files (images) in data/tribes/** and data/world/**.
the lua files are checked if changed according to the filter in lines 63 and 64.
to my understanding any other file in these 2 folders (which can normally be only .png) that is added or deleted from the repo  is checked by lines 65 and 66. we can't test anything about changed graphics so this is correctly omitted.

hessenfarmer:
As I agree on the safe side argument and the Ci seems to performant enough at the moment I'll close this.

SimonHeimberg:
reopened the pr now because addressing the reviews on my 3 pullrequests filled the build queue. But none of them have changes that the regression tests can check. (A waste of time and energy. Time would better be used for the other commits on master.)

I hope I have addressed all the review comments: data/world/ and data/tribe/ are never ignored. (For this seldom case, the filter was too complicated, agree. And not on the safe side.)

.github/workflows/build.yaml
bunnybot: [tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4677#issuecomment-2511712)*
*Created on Thu Dec 12 13:38:23 CET 2024 by [**Tóth András (tothxa)**](https://codeberg.org/tothxa)*

---

> what do you mean with testing images. which of our testsuite testcases does check images?

All of them? I mean, removing an image file of any item will cause the game to fail loading. Also the lua definitions may include references to new graphics, which are then not actually committed to git, with the same effect. *edit: Sorry, that's wrong. Actually the lua change itself will be checked and will fail, but if we implement changed checks within PRs (which we should), then the fix adding the images will not be checked.*

> for me this selection of relevant files for our testsuite does make sense

I think at least all of `data/tribes` and `data/world` must be checked. Sounds and music are different, since the testsuite runs with sound disabled, so problems there are probably missed completely. :( Same for translations.

SimonHeimberg:
> [tothxa](https://codeberg.org/tothxa)_[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4677#issuecomment-5872264)_ _On Fri Jul 18 01:38:03 CEST 2025, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) wrote:_
> 
> I think `data/i18n/translations` should also be ignored for the testsuite, because it runs with the default locale only.

Agree, implemented.  (When the pr was created, translations where in another path and excluded therefore.)

> Ideally we shouldn't rebuild w*ndows and mac binaries either for translation updates, but `datadirversion` is a problem... (though we could replace the translations and repackage with the old datadirversion for the sake of the executable, but that could lead to problems for identifying problematic ones)

I propose to optimise triggering installers in a separate pr.
When would the installers be built? Only when the binary will change? Then we could remove triggering on `data/**` altogether. (By the way, we also trigger on cmake/*codecheck*.)
I would not worry too much about (data/)datadirversion . A developer will learn quickly how to adapt it from output of `widelands --version`.

Edit: I see, skipping "data/i18n/translations" for installers would help because translations are updated automatically frequently. And most of the few people using a dev installer will not care about translation changes (getting delivered with the installer).  What problem would be with datadirversion when no installer is built when only translation changes? Or maybe problems with the job dev_release?

SimonHeimberg:
> > [tothxa](https://codeberg.org/tothxa)_[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4677#issuecomment-5872264)_ _On Fri Jul 18 01:38:03 CEST 2025, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) wrote:_
> > Ideally we shouldn't rebuild w*ndows and mac binaries either for translation updates, but `datadirversion` is a problem... (though we could replace the translations and repackage with the old datadirversion for the sake of the executable, but that could lead to problems for identifying problematic ones)
> 
> I propose to optimise triggering installers in a separate pr. When would the installers be built? Only when the binary will change? ...

let us continue discussion about skipping to build installers in my proposed pr #6852

hessenfarmer (approved):
LGTM.
@cb/tothxa?

SimonHeimberg:
> [tothxa](https://codeberg.org/tothxa)_[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4677#issuecomment-6643081)_ _On Thu Aug 21 16:12:22 CEST 2025, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) wrote:_
> 
> I think `i18n/translation_stats.conf` is usually updated together with the translations, so it should be ignored too to really ignore translation updates. Syntax errors in it probably don't affect the testsuite either, as it should only be used by the options menu, which is not used in the test... is that correct?

Right, data/i18n/translation_stats.conf will change on almost all translation updates. Does not make much sense to exclude translations but not this file. In case we want a syntax check for this file, it should be cheaper than running all tests.
There are also data/i18n/locales.lua data/txts/translators_data.lua data/txts/developers.lua data/i18n/locales/* , which will change much less frequent. I doubt we have to care about those. Agree?

bunnybot (commented):
[tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4677#issuecomment-6669583)*
*On Sat Aug 23 00:09:29 CEST 2025, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) **approved** this pull request:*

Thank you. Let's hope the matching will work as expected — I don't feel like creating a test branch on my gh fork to test it right now...

hessenfarmer (approved):
lgtm 
@bunnybot merge

hessenfarmer:
well it looks like it still runs the test suite on translation updates.

SimonHeimberg:
> well it looks like it still runs the test suite on translation updates.

True, ...
The output tells data/i18n/translation_stats.conf matches: https://github.com/widelands/widelands/actions/runs/17198145620/job/48783596331#step:3:117
I do not yet see why it matches. 

Testing (each line) on https://codesandbox.io/p/sandbox/picomatch-playground-ebjlxm
...

SimonHeimberg:
It matches because `'data/i18n/!(translations)/**'` also matches on files inside `i18n/` (despite the / before `**`).
Fix in progress.

SimonHeimberg:
fix in #6878

SimonHeimberg:
Finally, translation changes do not trigger test runs. This is true at least for 275fa27d5152344c69cb58aff6d61b4ebb164507.

PR #6315, opened 2024-01-15
improve build_codecheck

SimonHeimberg:
**Type of change**
Enhances testsuite

**Issue(s) closed**
Fixes -

**New behavior**
Do not use cmake. All files are checked anyway. Using cmake is no advantage.

CodeCheck.py produces output for failures only, resulting in nicer output.

This spares to install dependencies, making the run about 20sec faster.

**Possible regressions**
None I am aware of.
If `make codecheck` needs to be checked, it can be tested separately (with constructed changes).


hessenfarmer:
As @SimonHeimberg seems to have left the project (did not respond to any comment adressing him by name) and I agree with @tothxa that using unofficial extensions to github actions has always the risk of making our CI environment less stable due to addtional dependencies I propose to close this.
@widelands/developers @cb/Nordfriese  any objections?

SimonHeimberg:
I have removed the problem matcher.
I already open this pr now because I will probably be unreachable after release. But it is not to be consider it before.

PR #6317, opened 2024-01-15
cleanup check_vcpkg.yaml, remove option CMAKE_BUILD_TYPE, ...

SimonHeimberg:
**Type of change**
Enhancement of a github workflow

**Issue(s) closed**
Fixes -

**New behavior**
remove option CMAKE_BUILD_TYPE from check_vcpkg.yaml, it got an empty value

There is no "config" in matrix.

The fixed line was introduced in 1be5a4e5.

**Additional context**
Maybe a better approach is to configure a matrix with just one value for CMAKE_BUILD_TYPE.
And add a comment above line 76 (which uses it), telling it is exactly the same as in ./build_windows*.yaml


matthiakl:
> Maybe a better approach is to configure a matrix with just one value for CMAKE_BUILD_TYPE.
> And add a comment above line 76 (which uses it), telling it is exactly the same as in ./build_windows*.yaml

+1 for having a reference. Does it make sense to move this step into a shared file called from both workflows?

hessenfarmer:
@matthiakl @cb/tothxa
I would vote for just having this PR as it seems to be just a fix from a copy paste leftover.

hessenfarmer (approved):

hessenfarmer:
@bunnybot merge

PR #6329, opened 2024-01-28
Add-Ons Website Maps Integration

bunnybot:
[Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4692)*
*Created on Sun Jan 28 19:37:38 CET 2024 by [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese)*

---

**Type of change**
New feature (for v1.3)

Counterpart to https://codeberg.org/wl/wl_addons_server/pulls/84

**How it works**
The add-ons server now provides info about website maps. The client displays maps usefully with some map-specific filtering options and allows one-click installing and uninstalling.

For testing, the server-side counterpart is currently served on Alpha:
```
--addons_server_port=7377
```

**Not yet implemented** are:
- Translations for website maps
- Commenting, voting, and download counter. You can view comments and votes like for add-ons (except for relationships between comment threads), but creating and editing them is not yet possible due to server-side database complications.

**Possible regressions**
All add-ons networking

**Screenshots**
grafik


hessenfarmer:
I tested this now. 
I appreciate that maps are stored in the downloaded maps folder, as maps downloaded before from the website were detected and shown as installed.      
However I support Kaputtniks view that we should give a hint that th new tab is about website Maps. My suggestion would be to change the name of the Tab to "Website Maps" and provide a tooltip to state where the maps are stored. Last but not least we should include some documentation for our Addons Documentation.     
Everything worked ok for me except the filtering by name or other criteria, as the criteria is not applied after setting directly only after a refresh of the view. (e.g by leaving and revisiting the tab). in the normal addons tab this works fine.

hessenfarmer:
> [Nordfriese](https://codeberg.org/Nordfriese)_[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4692#issuecomment-2467117)_ _On Sat Nov 23 11:18:58 CET 2024, [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese) wrote:_
> 
> > However I support Kaputtniks view that we should give a hint that th new tab is about website Maps. My suggestion would be to change the name of the Tab to "Website Maps" and provide a tooltip to state where the maps are stored. Last but not least we should include some documentation for our Addons Documentation.
> 
> Done

Happy with that. Thanks.

> 
> > Everything worked ok for me except the filtering by name or other criteria, as the criteria is not applied after setting directly only after a refresh of the view. (e.g by leaving and revisiting the tab). in the normal addons tab this works fine.
> 
> I cannot reproduce this, all filters are applied immediately for me. Can you give exact examples to reproduce?

I activated the alpha server and got shown the test maps. I selected sort by name and the list got not resorted. I then tried to select the sorting by name in reverse order and the list got not resorted as well. Only after switching to the 2nd tab and back the list got resorted. selecting a filter from the right like number of players did have the same effect. But just selecting a filter from the dropdownlist did not trigger the resorting.

hessenfarmer (approved):
Code looks reasonable, although I m not very familiar with the UI parts.
Tested and working.

Issue #1914, opened 2019-09-09
Auto-generate Doxygen-generated documentation


bunnybot:
We have the capability to generate source code documentation using Doxygen. While everybody is free to generate it his or herself, it would be really nice to have this available automatically generated via the website from trunk, similarly to the Lua documentation.

------------------------------------
Imported from Launchpad using lp2gh.

 * date created: 2018-02-03T13:54:23Z
 * owner: franku
 * the launchpad url was https://bugs.launchpad.net/bugs/893275

bunnybot:
(by shevonar)
I also think that having the documentation online is easier than everyone building it for himself. I will see if I can figure out a good way to auto-generate it on a server and then ask Sirver to do so on the Widelands server.

bunnybot:
(by franku)
The lua documentation has a big shortcoming: There is currently no possibility for searching and the related django-app isn't maintained anymore, a search wouldn't be implemented.

Since both, doxygen and sphinx can make full featured html files, why do we not serve just the produced html?

Or/and provide the documentations as zip files for download? Build them daily (or once a month) on the server and provide a download link. Some starter pages, which will likely not change that often, can be included in the wiki.

BTW: Building the doxygen generated html results in lot of warnings and undocumented parameters  e.g. 

src/economy/input_queue.h:175: warning: argument 'fr' of command @param is not found in the argument list of Widelands::InputQueue::read(FileRead &f, Game &g, MapObjectLoader &mol)

src/economy/economy.h:205: warning: unable to resolve reference to `Cmd_Call_Economy_Balance' for \ref command

I looked into one and it seems those warnings are just there because a change of a variable name was not reflected for the documentation part. Or are forgotten to remove during restructuring? Anyway the produced documentation is working. No idea if it is also valid :-D

bunnybot:
(by gunchleoc)
Isn't the Sphinx documentation already available as HTML? It's on the website after all, so I don't understand what you mean here.

I have never generated the Doxygen documentation, because I just use my IDE to read the comments and header files. Fixing the errors & warnings is desirable though.

bunnybot:
(by franku)
The sphinx documentation is produced as json files. django_sphinx reads them out and passes them through the django template system.

The documentation on the website lacks searching: https://wl.widelands.org/docs/wl/search/ 
[...]

PR #6341, opened 2024-02-04
Lua styles documentation

frankystone:
**Type of change**
New feature: Added lua_styles.cc to documentation and added links to appropriate sections in themes. Added also an example for `as_p_with_attr()`.

**Issue(s) closed**
None.

The documentation missed this file.

**New behavior**
Complete the documentation.

**Possible regressions**
None.

**Screenshots**

styles_in_core

**Additional context**
Don't know if the "core functionality" is the right place for this because the provided functions are exclusively richtext related.


src/scripting/lua_styles.cc
```diff
@@ -26,6 +26,9 @@ namespace LuaStyles {
 :mod:`styles`
 =============
 
+These functions provides functionality to format :ref:`richtext <richtext.lua>` with the styles of
```
tothxa:
```suggestion
These functions provide functionality to format :ref:`richtext <richtext.lua>` with the styles of
```

src/scripting/lua_styles.cc
```diff
@@ -86,16 +89,30 @@ static int L_as_paragraph(lua_State* L) {
...
+   Example:
+
+   .. code-block:: lua
+
+      include "scripting/richtext.lua"
+
+      rt(
+         styles.as_paragraph("wui_text", _("This paragraph is left aligned"))
+         .. styles.as_p_with_attr("wui_text", "align=center", _("This paragraph is centered"))
+        )
```
tothxa:
```suggestion
      formatted_text = rt(
         styles.as_paragraph("wui_text", _("This paragraph is left aligned"))
         .. styles.as_p_with_attr("wui_text", "align=center", _("This paragraph is centered"))
        )
```

src/scripting/lua_styles.cc
```diff
@@ -163,13 +183,13 @@ static int L_close_p(lua_State* L) {
 /* RST
 .. function:: as_font_from_p(p_style, text)
 
-   Format the given text with the font style of the given paragraph style, without the paragraph
-   tags.
+   Format the given text with the font of the given :ref:`paragraph style <theme_paragraphs>`
```
tothxa:
```suggestion
   Format the given text with the font style of the given :ref:`paragraph style <theme_paragraphs>`
```

tothxa (commented):
Thank you very much for this PR.

I reviewed it as far as I could, but I'm not really familiar with RST and Sphinx. Also I can't test it, because I don't have Sphinx installed.

> Don't know if the "core functionality" is the right place for this because the provided functions are exclusively richtext related.

I don't think `path` and "Global functions" are any more "core functionality" than this.

frankystone:
Will apply your suggestions the next days :-) And upload a zip containing the whole documentation (without website style) so you can check if everything is fine.

> I don't think `path` and "Global functions" are any more "core functionality" than this.

The main difference is that `path` and "Global functions" can be used in every file, without any other functions, whereas this can only be used in combination with richtext, with `rt()`. Can this be used also with `<rt>`?
But if every other @developers is fine to have this under "Core functionality" this is it then, although i would vote against having it there. Have to think where this is better linked, though.

tothxa:
> The main difference is that `path` and "Global functions" can be used in every file, without any other functions, whereas this can only be used in combination with richtext, with `rt()`. Can this be used also with `<rt>`?

Of course. All these do is return strings with the appropriate richtext tags:

This script:

```
print(styles.get_size("editor_tooltip_icon_gap"))
print(styles.as_paragraph("wui_heading_2", "Some Title"))
print(styles.color("campaign_emp_saledus"))
```

Prints this (on stdout):

```
2
<p><vspace gap=12><font face=sans size=14 color=d1d1d1 bold=1>Some Title</font><vspace gap=12></p>
8f9131
```

> But if every other @Developers is fine to have this under "Core functionality" this is it then, although i would vote against having it there. Have to think where this is better linked, though.

Maybe the section could be renamed to make it clear that these are just the top level lua modules and functions that Widelands defines. But I don't have any suggestion for a different title right now.

frankystone:
From my point of view (a users view) this should be part of [The Widelands richtext system](https://www.widelands.org/documentation/wlrichtext/), like the link to "Scenario Markup Functions". Just because this __is a part__ of the widelands richtext sytem. And probably "The widelands richtext system" should be part of "Core functionality"? Or, even better, make "The Widelands richtext system" a top level entry in [Widelands scripting reference](https://www.widelands.org/documentation/reference/)

> Maybe the section could be renamed to make it clear that these are just the top level lua modules and functions that Widelands defines.

Explain "top level lua modules". A module which depends on an other function can't be a top level module, from my point of view. 

Let's wait what other developers say.

tothxa:
> From my point of view (a users view) this should be part of [The Widelands richtext system](https://www.widelands.org/documentation/wlrichtext/), like the link to "Scenario Markup Functions". Just because this **is a part** of the widelands richtext sytem.

It should certainly be linked from there, along with themes. There are also 2 functions from `styles` mentioned at the end of that page after the example that should be turned into links. Actually all of richtext system, the documentation of `richtext.lua`, themes and the styles module should link to each other.

And `wl.ui` is probably the main user of all of them. BTW `styles.color()` and `styles.get_size()` can be used by UI without ever touching richtext.

> Or, even better, make "The Widelands richtext system" a top level entry in [Widelands scripting reference](https://www.widelands.org/documentation/reference/)

Yes, that's a good idea. Though richtext itself is not lua, it is an important part of the Widelands API that is accessible from lua.

> > Maybe the section could be renamed to make it clear that these are just the top level lua modules and functions that Widelands defines.
> 
> Explain "top level lua modules". A module which depends on an other function can't be a top level module, from my point of view.

`styles` doesn't depend on anything from lua's perspective. It refers to themes (but that's inside Widelands, themes are just strings for lua) and emits richtext, but technically it doesn't depend on it.

> Let's wait what other developers say.

OK.

frankystone:
> On Mon Feb 05 11:06:41 CET 2024, **Benedikt Straub (Nordfriese)** wrote:_
> 
> In my view, Core Functionality refers to all those Lua functions that we implement in C++, and the other categories are for utility functions that are coded in Lua using other functions.

So all entries in https://github.com/widelands/widelands/blob/5e7fbd522f19b436bf65c03905643b2160092c60/doc/sphinx/extract_rst.py#L17-L31

are Core functionality and should be added to the section "Core functionality"?

I think the problem i have is, that i think this documentation is mainly for users who want to do some lua scripting (without knowing anything about c++), e.g. make an add-on or create a scenario, and the documentation should be a somehow a guide for users who wants to do that. And that is to bundle things together with the meaning of "i want to use richtext, where do i find all richtext related functions?", not to bundle things with the meaning "this is c++ so this is core functionality". This wouldn't help add-on or scenario authors, because all available information would be spread over the documentation.  So this documentation is mainly not a developers documentation.
BTW: For the developer documentation (c++) it was planned to add an additional link on the website, see https://github.com/widelands/widelands/issues/1914#issuecomment-529464037 ff 

Anyway, i'll leave the styles in "Core Functionality" and try to do some more linking.

tothxa:
> ```
> print(styles.get_size("editor_tooltip_icon_gap"))
> print(styles.color("campaign_emp_saledus"))
> ```

That wasn't good API design. I didn't want to use `styles.size()`, but I could change colour to `styles.get_color()` after this PR is merged (to avoid conflicts). We have to do it in v1.2 while we can still change it freely. Should I do it?

frankystone:
Here is the current state. Just unpack and open an html file in your browser.

[html.zip](https://github.com/widelands/widelands/files/14178755/html.zip)

tothxa:
> Here is the current state. Just unpack and open an html file in your browser.

Thank you. Checked it and looks OK. (except for the verisonadded fixes that will be done in the other PR)

tothxa (approved):

src/scripting/lua_styles.cc
```diff
@@ -86,16 +89,31 @@ static int L_as_paragraph(lua_State* L) {
...
+   Example:
+
+   .. code-block:: lua
+
+      include "scripting/richtext.lua"
+
+      formatted_text = rt(
+                          styles.as_paragraph("wui_text", _("This paragraph is left aligned"))
+                          .. styles.as_p_with_attr("wui_text", "align=center", _("This paragraph is
+                          centered"))
```
bunnybot: Nordfriese*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4704#issuecomment-1553805)*
*On Wed Feb 07 08:40:40 CET 2024, **Benedikt Straub (Nordfriese)** wrote:*

---

This needs better linebreaking

bunnybot (commented):
Nordfriese*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4704#issuecomment-1553806)*
*On Wed Feb 07 08:40:40 CET 2024, **Benedikt Straub (Nordfriese)** commented.*

src/scripting/lua_styles.cc
frankystone: Done

bunnybot (commented):
Nordfriese*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4704#issuecomment-1555220)*
*On Thu Feb 08 10:23:33 CET 2024, **Benedikt Straub (Nordfriese)** **approved** this pull request:*

Thank you.

<@>bunnybot merge force

PR #6345, opened 2024-02-06
Economic Districts

bunnybot:
[Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4708)*
*Created on Tue Feb 06 19:59:44 CET 2024 by [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese)*

---

**Type of change**
New feature for v1.3

**Issue(s) closed**
This addresses several points that arise frequently in the context of very large games – namely the problem of wares (and also workers) being transported across long distances instead of preferring local supplies.

**How it works**
These points all happen in the background. The UI is unchanged.

Each economy automatically divides itself into **districts**. A district is centered around a warehouse (multiple warehouses in close proximity are clustered for this purpose), and each flag belongs to the closest warehouse. Thus, a district is a circle around a warehouse or cluster of warehouses.

Districts try to be self-sufficient and minimize imports.
Supply/request matching always prefers to use supplies from the same district. Imports are only accepted if no local supply is available.

New: All active long-distance imports are frequently monitored, and if a ware becomes available in the local district, the import is cancelled and replaced with the local ware. This type of supply exchanging was previously always rejected as too CPU-intensive, but this approach is performant and catches all the cases where it matters most.

Additionally, economy targets are distributed across districts. Global targets (global here meaning within one economy) continue to work as before. In addition, if a district is short of a specific ware, then the productionsites within this district keep producing even if the global stock is above target. Each district's local target is simply the global target divided by number of districts, rounding up.

**Possible regressions**
Economic request/supply matching, economy targets.

**Screenshots**
grafik grafik

**Additional context**
Nothing here needs to be saveloaded. This is all recalculated lazily every few seconds.
Benchmark for a nearly completely conquered Accurate Europe 1.0 map, Release build:
```
[00:02:38.430 real] INFO: NOCOM Recalcing districts for 5376 flags and 34 warehouses took 7 ms
[00:02:38.434 real] INFO: NOCOM Checking imports for 5376 flags and 34 warehouses took 3 ms
```

Flags at the border between two districts tend to jitter between the two possible districts due to pseudo-randomness in the routing algorithm. I would not consider this a problem.

Could do with more testing, both with big economies and with small ones. I'd therefore like to have this early in the release cycle for v1.3.


src/economy/flag.h
```diff
@@ -204,6 +216,10 @@ struct Flag : public PlayerImmovable, public RoutingNode {
 	Coords position_;
 	Time animstart_{0};
 
+	Warehouse* district_center_[2] = {
+	   nullptr,
+	   nullptr};  ///< Warehouse at the center of our district, indexed by WareWorker (may be null).
```
bunnybot, line 221: tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4708#issuecomment-2012204)*
*On Tue Jun 11 15:45:41 CEST 2024, **Tóth András (tothxa)** wrote:*

---

Is indexing this directly by an unscoped enum safe?

src/economy/flag.cc
```diff
@@ -1015,6 +1018,19 @@ void Flag::log_general_info(const Widelands::EditorGameBase& egbase) const {
 
 	Widelands::PlayerImmovable::log_general_info(egbase);
 
+	if (district_center_[0] == nullptr) {
```
bunnybot: tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4708#issuecomment-2012208)*
*On Tue Jun 11 15:51:36 CEST 2024, **Tóth András (tothxa)** wrote:*

---

Maybe `wwWARE` (and `wwWORKER` below) should be used for consistency?

src/economy/economy.cc
```diff
@@ -504,12 +513,28 @@ Worker& Economy::soldier_prototype(const WorkerDescr* d) {
...
+	if (target_global > 0 && !warehouses_.empty()) {
+		// We have a target quantity set.
+		Quantity target_district;
+		if (flag != nullptr && nr_districts_ > 0) {
+			target_district = target_global / nr_districts_;
+			if (target_district * nr_districts_ < target_global) {
+				++target_district;  // Rounding up is important for wares with small targets
+			}
+		} else {
+			target_district = 0;
```
bunnybot: tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4708#issuecomment-2012254)*
*On Tue Jun 11 16:26:04 CEST 2024, **Tóth András (tothxa)** wrote:*

---

Simply initialise as 0 for readability instead of this `else` clause?

src/economy/economy.cc
```diff
@@ -1147,6 +1223,182 @@
...
+			}
+
+			upcast(PlayerImmovable, imm, location);
+			if (imm == nullptr) {
+				continue;
+			}
+
+			Flag& supply_flag = imm->base_flag();
+			Warehouse* supply_district = supply_flag.get_district_center(type());
+			if (supply_district == nullptr || supply_district == target_district) {
```
bunnybot: tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4708#issuecomment-2012360)*
*On Tue Jun 11 17:24:19 CEST 2024, **Tóth András (tothxa)** wrote:*

---

Is there a valid case when `supply_district` can be `nullptr`? Doesn't that mean there's some inconsistency somewhere?

src/economy/economy.cc
```diff
@@ -1147,6 +1223,182 @@
...
+		return;
+	}
+
+	// First pass: Assign each flag to the closest warehouse.
+	RouteAStar<AStarZeroEstimator> astar(*router_, type_, AStarZeroEstimator());
+	for (Warehouse* wh : warehouses_) {
+		wh->base_flag().set_district_center(type(), wh);
+		astar.push(wh->base_flag());
+	}
+	while (RoutingNode* current = astar.step()) {
```
bunnybot: tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4708#issuecomment-2012371)*
*On Tue Jun 11 17:34:44 CEST 2024, **Tóth András (tothxa)** wrote:*

---

Isn't this an implicit cast to bool?

src/economy/economy.cc
```diff
@@ -1147,6 +1223,182 @@
...
+		current->base_flag().set_district_center(
+		   type(), astar.route_start(*current).base_flag().get_district_center(type()));
+	}
+
+	// Second pass: Identify minimal monodirectional distances between warehouses.
+	constexpr uint32_t kClusterThreshold = 12 * 1800;  // 12 nodes on flat terrain.
+	std::map<std::pair<Warehouse*, Warehouse*>, uint32_t> warehouse_distances;
+
+	for (Flag* flag1 : flags_) {
```
bunnybot: tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4708#issuecomment-2012409)*
*On Tue Jun 11 18:02:55 CEST 2024, **Tóth András (tothxa)** wrote:*

---

Why don't you only iterate over `warehouses_`?

src/economy/economy.cc
```diff
@@ -1147,6 +1223,182 @@
...
+		Cluster current_cluster = {current};
+		for (auto& pair : warehouse_distances) {
+			if (pair.first.first == current) {
+				current_cluster.push_back(pair.first.second);
+			} else if (pair.first.second == current) {
+				current_cluster.push_back(pair.first.first);
+			}
+		}
+
+		// Now find all existing clusters to merge this one with
```
bunnybot: tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4708#issuecomment-2012441)*
*On Tue Jun 11 18:21:07 CEST 2024, **Tóth András (tothxa)** wrote:*

---

Do I get it right that if there's a long chain of warehouses where neighbouring ones are within the threshold, then all of them will get clustered even if the distance between the opposite ends is many times the threshold? Do we really need clusters at all?

bunnybot (commented):
tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4708#issuecomment-2012447)*
*On Tue Jun 11 15:45:41 CEST 2024, **Tóth András (tothxa)** commented:*

I tested it in a game too and it seems to work nicely. Thanks a lot for solving this! :)

Some comments inline.

src/economy/economy.cc
```diff
@@ -1147,6 +1221,190 @@ void Economy::handle_active_supplies(Game& game) {
...
+		for (size_t i = 0; i < request->get_num_transfers(); ++i) {
+			MapObject* location;
+			if (request->get_transfers()[i]->get_ware() != nullptr) {
+				location = request->get_transfers()[i]->get_ware()->get_location(game);
+			} else {
+				location = request->get_transfers()[i]->get_worker()->get_location(game);
+			}
+
+			upcast(PlayerImmovable, imm, location);
+			if (imm == nullptr) {
```
bunnybot, line 1330: Nordfriese*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4708#issuecomment-2012579)*
*On Tue Jun 11 19:19:35 CEST 2024, **Benedikt Straub (Nordfriese)** wrote:*

---

Since districts are recalculated lazily, they can be `nullptr` for short periods during and after splits and merges and warehouse deletions. In this case the import will simply be caught on the next cycle, after the districts have been updated again.

src/economy/economy.cc
```diff
@@ -1147,6 +1221,190 @@ void Economy::handle_active_supplies(Game& game) {
...
+	using Cluster = std::vector<Warehouse*>;
+	std::vector<Cluster> all_clusters;
+	using ClusterIterator = std::vector<Cluster>::iterator;
+
+	for (Warehouse* current : warehouses_) {
+		// Find all warehouses this one should be clustered with
+		Cluster current_cluster = {current};
+		for (auto& pair : warehouse_distances) {
+			if (pair.first.first == current) {
+				current_cluster.push_back(pair.first.second);
```
bunnybot: Nordfriese*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4708#issuecomment-2012570)*
*On Tue Jun 11 19:17:28 CEST 2024, **Benedikt Straub (Nordfriese)** wrote:*

---

The case I was thinking of was the strategy that some players use of building two or three warehouses in close proximity for training centres, one to store food, one to store weapons, etc. If each of these warehouses defined its own district then this would mess up all productionsites around this centre. This new feature should not break existing valid strategies and clustering was the solution I came up with.

A long chain would be clustered too, but IMHO a really long warehouse chain is an indication that the player is doing something highly unusual and unexpected and we can't really guess what he might wish to accomplish with this, except that such an economy seems to be tightly chain-linked and therefore may be meant to be treated as one unit.

bunnybot: tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4708#issuecomment-2012842)*
*On Tue Jun 11 20:34:08 CEST 2024, **Tóth András (tothxa)** wrote:*

---

> The case I was thinking of was the strategy that some players use of building two or three warehouses in close proximity for training centres, one to store food, one to store weapons, etc. If each of these warehouses defined its own district then this would mess up all productionsites around this centre.

Have you experienced that in testing before adding clusters?

bunnybot: tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4708#issuecomment-2013191)*
*On Wed Jun 12 01:12:10 CEST 2024, **Tóth András (tothxa)** wrote:*

---

> A long chain would be clustered too, but IMHO a really long warehouse chain is an indication that the player is doing something highly unusual and unexpected and we can't really guess what he might wish to accomplish with this, except that such an economy seems to be tightly chain-linked and therefore may be meant to be treated as one unit.

Or maybe it's just a player who took the advice to build warehouses a bit too overzealously. :) They shouldn't be punished for it. But I really don't know what we could do that's simple enough.

I also have a doubt about how we can decide what's an appropriate cluster threshold.

bunnybot: Nordfriese*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4708#issuecomment-2013677)*
*On Wed Jun 12 11:38:08 CEST 2024, **Benedikt Straub (Nordfriese)** wrote:*

---

Well, in the worst case aggressive clustering simply means that the entire economy (or a large chunk thereof) becomes one big district, which will simply behave exactly as in current master. So the only downside is that such an economy won't be able to fully utilize the benefits of this addition, but there's no regression or penalty compared to master.
On the other hand, splitting a zone that should logically form a single unit will lead to two or three close-by districts that are constantly importing wares from each other. While not super-harmful, this results in an overhead of production for goods that are needed only in smaller quantity (and thus waste of precious input materials), as well as an overhead for the many short-distance imports. That's why I think it is preferable to err on the side of clustering some more when in doubt.

bunnybot: tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4708#issuecomment-2014775)*
*On Thu Jun 13 01:06:49 CEST 2024, **Tóth András (tothxa)** wrote:*

---

> So the only downside is that such an economy won't be able to fully utilize the benefits of this addition, but there's no regression or penalty compared to master.

I don't think master is the most appropriate base for comparison here. As districts are not visible to the players, they won't be able to understand why the otherwise smart new supply matching algorithm breaks down in such situations.

bunnybot (commented):
[tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4708#issuecomment-2151766)*
*On Sun Aug 04 19:23:08 CEST 2024, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) **approved** this pull request:*

I've played several games with my branch (should be equivalent with the current state) recently. My main observation WRT districts was that there are often quite big production overshoots, far above the allowance for the local warehouse.

I've been also thinking about `check_imports()`. First of all, it ignores wares on ships (big problem for new colonies with fresh production sites) and with carriers. And in general, it may be better to schedule a check for replacement for individual long-range transports, instead of iterating over all transports regularly.

Both of these can be solved in follow-up PRs though. I'd certainly try the `check_imports()` replacement myself.

I'm also a bit afraid whether my changes to `find_best_supply()` are sync-safe across different systems. Should we test that? (I tested that replays work locally.)

But in general, I think this is probably OK now.

Issue #6335, opened 2024-01-30
macOS builds fail in CI workflow in dependency installation step


bunnybot:
<img align='left' src='https://codeberg.org/avatars/593dc43e2444c25db9481960e5b693ab' alt='tothxa' height='48' />*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/issues/4698)*
*Created on Wed Jan 31 00:15:18 CET 2024 by **Tóth András (tothxa)***

---

**Describe the bug**
It looks like we're starting to get random breakage in macOS builds in the dependency installation step. We should probably consider caching those similar to vcpkg.

**Expected behavior**
Stable workflow runs

tothxa:
Although the end of the logs shows a big scary deprecation warning for macos11, that's actually an old thing that didn't cause problems so far.

```
==> Installing graphviz dependency: libvmaf
==> meson build
==> ninja -vC build
Last 15 lines from /Users/runner/Library/Logs/Homebrew/libvmaf/02.ninja:
src/vmaf_v0.6.1.json.c:1595:1: error: unknown type name 'size_t'
size_t src_vmaf_v0_6_1_json_len = 19101;
^
1 error generated.
[42/165] clang -Isrc/liblibvmaf_feature.a.p -Isrc -I../src -Iinclude -I../include -I../src/feature -I../src/feature/common -I../src/cuda -fdiagnostics-color=always -Wall -Winvalid-pch -Wextra -std=c11 -O3 -D_DARWIN_C_SOURCE -MD -MQ src/liblibvmaf_feature.a.p/feature_iqa_convolve.c.o -MF src/liblibvmaf_feature.a.p/feature_iqa_convolve.c.o.d -o src/liblibvmaf_feature.a.p/feature_iqa_convolve.c.o -c ../src/feature/iqa/convolve.c
[...]

PR #6352, opened 2024-02-09
Workflow tweaks

tothxa:
**Type of change**
Maintenance / New feature / Cleanup

**Issue(s) closed**
 - node.js 16 deprecation warnings
 - parallel workflow runs holding up older ones because summary job doesn't get a runner

**New behavior**
 - The documentation job uplads the generated files as an artifact. HTML is also generated for user testing. Graphs are also created.
 - The `summary`  job was deleted as not necessary and causing delays

**Possible regressions**
Workflow problems, particularly due to new behaviour of new versions of actions

**Additional context**
This **does NOT** fix the mac build problem. (#6335)

The node.js update is forced on us until "spring 2024", so we probably have to do it before v1.2. If the other changes are found too disruptive, I'll undo them.

I tried to enable `--nitpicky` for `sphinx-build`, but it generated warnings I don't know how to solve, and not belonging to this PR anyway.


bunnybot (commented):
Nordfriese*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4715#issuecomment-1556664)*
*On Fri Feb 09 12:02:21 CET 2024, **Benedikt Straub (Nordfriese)** **approved** this pull request:*

LGTM :)

.github/workflows/pre-build_checks.yaml
```diff
@@ -45,26 +45,39 @@ jobs:
...
         mkdir source/_static
         ./extract_rst.py
+        # Test building for the homepage
         sphinx-build -W -b json -d build/doctrees source build/json
+        # This is suitable for user testing
+        ./make.py html -graphs
+    - name: Upload artifact
+      uses: actions/upload-artifact@v4
+      with:
+        name: Widelands_Developer_Documentation_${{ github.sha }}
```
matthiakl: You may want to rename this to `Widelands-${{ github.sha }}-Developer-Documentation`.
Then it should be picked up by the automated development releases (Pattern `Widelands-<hash>-<os matrix>`). Otherwise the artifact is only available for logged in users.

tothxa: Thanks for the tip! :)

matthiakl (approved):
I am not sure whether the nested zip is necessary, but this is definitely the safest way as it looks the same for the other artifacts.

> I wonder whether I should exclude the doctrees subdirectory.

Just keep it, it is well compressed anyway.

tothxa:
Thank you.

@bunnybot merge force

PR #6362, opened 2024-02-15
Update performance helptext for farm-like buildings

bunnybot:
[palino](https://codeberg.org/palino)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4725)*
*Created on Thu Feb 15 14:47:55 CET 2024 by [**Pavol Gono (palino)**](https://codeberg.org/palino)*

---

**Type of change**
Fix and add missing performance help texts in lua scripts. Subtask of https://github.com/widelands/widelands/issues/5268
First ideas here: https://github.com/widelands/widelands/pull/6068

Now I want to focus on farm-like buildings.

Initial proposal is for atlantean farm as an example (the same timings are for barbarian and empire farms). Not sure if 5 nearest fields is the correct number here.

First describing average calculation for single farm without nearby obstacles on flat terrain. Additionally describing the ideal (fastest) case and providing info how many fields need to remain free.

### Alternative 1: describing ideal and worst cases, like for small outdoor buildings
- I don't like it because worst case is highly improbable to achieve
- average case can be easily calculated exactly (except maybe Frisian Fruit Collector)

#### Advantages:
- removes need of silly micromanagement (building artificial roads on the "bad" spots of farm work area) to have high-performance farms
- it is then enough to describe ideal cases in performance helptexts
#### Disadvantages:
- considerably changes logic of other workers too (like fishers and foresters)
- (I am not able do this change alone)

### Question
What I am missing is understanding how often the field states change.
tiny -> small -> medium -> ripe
After few measurements, these 3 transitions together took between 75 sec - 189 sec
What is the possible time range here?
The files data/tribes/immovables/cornfield/*/init.lua contain certain durations, but I am not sure if they are used at all.


data/tribes/initialization/atlanteans/units.lua
```diff
@@ -1808,8 +1808,12 @@ wl.Descriptions():new_tribe {
             lore = pgettext("atlanteans_building", "I use my feet, I use my arm, I’m caring for this farm!"),
             -- TRANSLATORS: Lore author helptext for an Atlantean production site: Farm
             lore_author = pgettext("atlanteans_building", "Atlantean harvesting song"),
-            -- TRANSLATORS: Performance helptext for an Atlantean production site: Farm
-            performance = pgettext("atlanteans_building", "The farmer needs %1% on average to sow and harvest a sheaf of corn."):bformat(format_minutes_seconds(1, 20))
+            performance = {
+               -- TRANSLATORS: Performance helptext for an Atlantean production site: Farm, part 1
+               pgettext("atlanteans_building", "The farmer needs %1% on average to sow and harvest a sheaf of corn."):bformat(format_minutes_seconds(1, 4)),
+               -- TRANSLATORS: Performance helptext for an Atlantean production site: Farm, part 2
+               pgettext("atlanteans_building", "It is possible to decrease time to %1$s on average, if you leave %2$d fields near the flag free and other fields of work area occupied, e.g. by roads and buildings."):bformat(format_seconds(56), 5)
```
bunnybot: Nordfriese*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4725#issuecomment-1567513)*
*On Thu Feb 15 18:55:14 CET 2024, **Benedikt Straub (Nordfriese)** wrote:*

---

`%2$d fields` needs an `ngettext` call

bunnybot (commented):
Nordfriese*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4725#issuecomment-1567514)*
*On Thu Feb 15 18:55:14 CET 2024, **Benedikt Straub (Nordfriese)** commented.*

hessenfarmer:
Thanks for finding this out.      
However I beleive it just does affect the number of fields you need, not the production time.    
If you have enough space the farmer will (after a some time) always find a spot to harvest from and a spot for new seeds. By this the production time after the initial phase will be only dependant on walking time.       
Having some randomness here is not to bad imho as otherwise we might have synchronized growings which would look unnatural.      
so for me we should leave the formula in and take only the walking Production Programs from the farmer into account for the performance texts

data/tribes/initialization/atlanteans/units.lua
```diff
@@ -1815,8 +1815,12 @@ wl.Descriptions():new_tribe {
             lore = pgettext("atlanteans_building", "I use my feet, I use my arm, I’m caring for this farm!"),
             -- TRANSLATORS: Lore author helptext for an Atlantean production site: Farm
             lore_author = pgettext("atlanteans_building", "Atlantean harvesting song"),
-            -- TRANSLATORS: Performance helptext for an Atlantean production site: Farm
-            performance = pgettext("atlanteans_building", "The farmer needs %1% on average to sow and harvest a sheaf of corn."):bformat(format_minutes_seconds(1, 20))
+            performance = {
+               -- TRANSLATORS: Performance helptext for an Atlantean production site: Farm, part 1
+               pgettext("atlanteans_building", "The farmer sows and harvests one sheaf of corn between %1$s and %2$s, depending on how far he has to walk."):bformat(format_seconds(57), format_minutes_seconds(1, 12)),
+               -- TRANSLATORS: Performance helptext for an Atlantean production site: Farm, part 2
+               pgettext("atlanteans_building", "The farm needs at least %s to achieve 100%% productivity."):bformat(ngettext("%d growing field", "%d growing fields", 6):bformat(6))
```
bunnybot: [tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4725#issuecomment-2552067)*
*On Mon Dec 30 14:52:14 CET 2024, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) wrote:*

---

Can't we just include the number of fields in the main text without `bformat()` and `ngettext()`? What do we gain by using them here? (it's different from time units with the helper functions)

bunnybot (commented):
[tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4725#issuecomment-2552068)*
*On Mon Dec 30 14:52:14 CET 2024, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) commented.*

data/tribes/initialization/atlanteans/units.lua
bunnybot: [palino](https://codeberg.org/palino)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4725#issuecomment-2552771)*
*On Mon Dec 30 21:59:50 CET 2024, [**Pavol Gono (palino)**](https://codeberg.org/palino) wrote:*

---

I based my change on Nordfriese's request:
https://codeberg.org/wl/widelands/pulls/4725#issuecomment-1567513
I find it more flexible - in case we change number of recommended fields, the translation don't need any change, just number update.
E.g. in Slovak and Czech language, there are different forms both for adjective and noun, in case count is 4 or count is 5. So in case we use just simple %d and count changes from 5 to 4, content needs to be retranslated.

data/tribes/buildings/productionsites/empire/farm/init.lua
```diff
@@ -49,6 +49,15 @@ wl.Descriptions():new_productionsite_type {
          -- TRANSLATORS: Completed/Skipped/Did not start working because ...
          descname = _("working"),
          actions = {
+            -- longest possible transition from tiny to ripe field:
+            --   2 * (30 + 45 + 50) = 250 sec
+            --   see data/tribes/immovables/wheatfield/ and src/logic/map_objects/immovable_program.cc - ImmovableProgram::ActAnimate::execute()
+            -- how many fields are needed for reliable 100% productivity: 6
+            --   with 5 fields, this inequation is not fulfilled:
+            --   (250 + 55.84 - (6 + 2.2 * 1.8 + 10)) / 55.84 < 5
```
bunnybot: [tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4725#issuecomment-2572396)*
*On Wed Jan 08 03:47:05 CET 2025, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) wrote:*

---

Maybe you should turn these around, and write the calculation of the required number of fields as:

1 + ceil((t_max_grow - (t_post_seed + t_walkback_min + t_sleep_after_seed)) / t_total_with_min_walk)

However, calculating with the extremely unlikely maximum theoretical growing time is overkill IMO. The total growing time is the sum of 6 random values with uniform distribution. This is basically a [Bates distribution](https://en.wikipedia.org/wiki/Bates_distribution) (though with slightly varying intervals), so the standard deviation should be ~1/sqrt(12*6) times the nominal value, so using 3σ to get well over 99% probability would give us ~1.36 times the nominal growing time as practical maximum.

(I repeated the test, and I was able to get 100% productivity with only the 4 nearest or the 3 farthest fields. I don't understand why I needed more the last time. The new values are consistent with the above calculation.)

bunnybot (commented):
[tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4725#issuecomment-2572397)*
*On Wed Jan 08 03:47:05 CET 2025, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) commented.*

data/tribes/buildings/productionsites/empire/farm/init.lua
bunnybot: [palino](https://codeberg.org/palino)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4725#issuecomment-2573013)*
*On Wed Jan 08 12:36:45 CET 2025, [**Pavol Gono (palino)**](https://codeberg.org/palino) wrote:*

---

Do you mean to formulate it following?
```
            -- how many fields are needed for reliable 100% productivity:
            --   1 + ceil((250 - (6 + 2.333 * 1.8 + 10)) / 56.8) = 6
            -- min. time total (free 6 nearest fields):  30.4  + 26.4  = 56.8  sec
            -- max. time total (free 6 furthest fields): 38.2  + 34.2  = 72.4  sec
```
The `t_walkback_min` and `t_total_with_min_walk` are also variable, depending on the result, so it not an easy universal formula. With `N-1` field count I wanted to have easy verification, that it is too low count for 100% productivity.

I agree the maximum theoretical growing time is highly improbable. We could give to helptext number of fields for 100% and for 99% "average" productivity, but do you think is it practical to have both?

What concerns me more is my testing with Barbarian reed farms. The ceiling here is minimal among farms:
1 + (180 - (6.5 + 2.5 * 1.8 + 8)) / 56 = 3.875
With 4 free fields I can sometimes reproduce drop of 100% productivity, I can't explain why. We may miss something in calculations, or the ingame production ratio calculation is faulty.

bunnybot: [tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4725#issuecomment-2573212)*
*On Wed Jan 08 14:40:56 CET 2025, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) wrote:*

---

> Do you mean to formulate it following?
> ```
>             -- how many fields are needed for reliable 100% productivity:
>             --   1 + ceil((250 - (6 + 2.333 * 1.8 + 10)) / 56.8) = 6
> ```

Yes.

> The `t_walkback_min` and `t_total_with_min_walk` are also variable, depending on the result, so it not an easy universal formula. With `N-1` field count I wanted to have easy verification, that it is too low count for 100% productivity.

It's safe to always calculate the required number of nearest fields with 2 steps, and that's the only one that you put in `units.lua` anyway.

If you also want the number for the farthest fields, you can calculate with (max_steps - 1), and if that gives more than what's available at at least that distance, then verify with (max_steps - 2) (without averaging).

> I agree the maximum theoretical growing time is highly improbable. We could give to helptext number of fields for 100% and for 99% "average" productivity, but do you think is it practical to have both?

No. What I mean is that if the probability of one field to miss its deadline is less than 0.5%, then, once the productivity reaches 100%, it will stay so for 99+% of the time, ie. practically always. If the number of fields and the probability of another field getting ready early  is large enough (like the Empire farm with the 4 nearest fields), then it easily gets to 99.99+%.

(within ±3σ with normal distribution is 99.7%, so 0 to mean+3σ is 99.85%, but I believe ours is slightly worse than normal distribution, except the standard deviation may already account for that, I really don't know...)

bunnybot: [palino](https://codeberg.org/palino)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4725#issuecomment-2576739)*
*On Fri Jan 10 06:49:33 CET 2025, [**Pavol Gono (palino)**](https://codeberg.org/palino) wrote:*

---

What about using the Cumulative distribution function, so that we don't need approximations? SciPy recently added Irwin-Hall distribution https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.irwinhall.html
Actually, an inverse function to CDF would be the best, but with manual bisection it is easy to find what we need.
For example if we target 99%, we could use following numbers for 4 farms and 6 farms:
```
>>> from scipy.stats import irwinhall
>>> irwinhall(8, scale=1/8).cdf(0.2663)
0.009994741336539768
>>> irwinhall(12, scale=1/12).cdf(0.3081)
0.009983824644571366
```
(the doubled N is needed, because `ImmovableProgram::ActAnimate::execute()` is using sum of 2 random numbers)

I would like to keep existing calculations in lua scripts (for 100% productivity). In `units.lua` will remain only results of 99% (or whatever else) calculations.

And I am not sure - which percentage should I target: 99% or 99.5% ?

gpalino: I am correcting myself, there exists inverse function of `cdf()` - `ppf()`:
```
>>> from scipy.stats import irwinhall
>>> irwinhall(8, scale=1/8).ppf(0.01)
0.266318370224793
>>> irwinhall(12, scale=1/12).ppf(0.01)
0.30814770483468445
```

bunnybot: [tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4725#issuecomment-2577350)*
*On Fri Jan 10 12:02:22 CET 2025, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) wrote:*

---

The actual percentage doesn't really matter, since we have to round (preferably up) to an integer in the end. Because of that, I'd say it's probably better to say in the encyclopedia that a farm needs 3 fields if that's enough for 95% probability, than to make it 99.99% with 4 fields. But let's hope we won't need such trade-offs. The ones I checked so far all needed just a small upward rounding with 1.36 times the nominal growth time fortunately. If we find any that's only slightly above an integer, then we can evaluate those few one by one in more detail to decide whether to round them up or down.

(BTW ±3 times observed σ is a well established practice for confidence interval of measurements, i.e. with unknown distribution)

bunnybot: [palino](https://codeberg.org/palino)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4725#issuecomment-2578483)*
*On Fri Jan 10 23:39:36 CET 2025, [**Pavol Gono (palino)**](https://codeberg.org/palino) wrote:*

---

I checked the farm for empire, atlanteans and barbarians with SciPy:
4 fields:
```
1 + (250 - (6 + 2 * 1.8 + 10)) / 54.4 = 5.2353
irwinhall(8, scale=1/8).sf(1 - 4/5.2353) = 0.99608
```
-> 99.6% average productivity

Barbarian reed yard:
3 fields:
```
1 + (180 - (6.5 + 2.3333 * 1.8 + 8)) / 54.8 = 3.9434
irwinhall(6, scale=1/6).sf(1 - 3/3.9434) = 0.98791
```
-> 98.8% average productivity

`sf` is defined as `1 - cdf`
With +1 fields, the probability is 100% or very close to 100%.

I can recalculate each farm. I agree with providing the percentage in helptext. Are the following formulations OK for you?
```
The farm needs at least 4 growing fields to achieve 99.6% productivity.
The reed yard needs at least 3 growing fields to achieve 98.8% productivity.
```

bunnybot: [tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4725#issuecomment-2578717)*
*On Sat Jan 11 03:41:47 CET 2025, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) wrote:*

---

Yes, it's OK to me if <@>hessenfarmer doesn't object. Or maybe most players would find something like this more friendly:
`... to work at full productivity most of the time.`

data/tribes/buildings/productionsites/empire/vineyard/init.lua
```diff
@@ -48,6 +48,18 @@ wl.Descriptions():new_productionsite_type {
          -- TRANSLATORS: Completed/Skipped/Did not start working because ...
          descname = _("working"),
          actions = {
+            -- the longest possible transition from tiny to ripe field:
+            --   2 * (22 + 28 + 40) = 180 sec
+            --   see data/tribes/immovables/grapevine/ and src/logic/map_objects/immovable_program.cc - ImmovableProgram::ActAnimate::execute()
+            -- how many fields are needed for 100% productivity in worst case (nearest fields, longest field transition):
+            --   1 + (180 - (5 + 2.5 * 1.8 + 5)) / 48 = 4.448
+            -- calculation of productivity with 4 farms:
```
bunnybot: [tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4725#issuecomment-2586907)*
*On Wed Jan 15 02:27:13 CET 2025, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) wrote:*

---

farms → fields (in many places)

data/tribes/buildings/productionsites/empire/vineyard/init.lua
```diff
@@ -48,6 +48,18 @@
...
+            --   irwinhall(8, scale=1/8).cdf(4/4.448) = 0.999996
```
bunnybot: [tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4725#issuecomment-2586909)*
*On Wed Jan 15 02:29:11 CET 2025, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) wrote:*

---

`irwinhall(6, scale=1/6)`

bunnybot (commented):
[tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4725#issuecomment-2586923)*
*On Wed Jan 15 02:27:12 CET 2025, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) commented.*

data/tribes/buildings/productionsites/empire/vineyard/init.lua
bunnybot: [palino](https://codeberg.org/palino)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4725#issuecomment-2592680)*
*On Fri Jan 17 17:09:14 CET 2025, [**Pavol Gono (palino)**](https://codeberg.org/palino) wrote:*

---

Thanks for finding this

data/tribes/buildings/productionsites/empire/vineyard/init.lua
bunnybot: [palino](https://codeberg.org/palino)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4725#issuecomment-2592683)*
*On Fri Jan 17 17:10:17 CET 2025, [**Pavol Gono (palino)**](https://codeberg.org/palino) wrote:*

---

I am not sure what you are pointing to - do you prefer having 3-field calculation in helptext? 91.1% gives about 7 misses per hour, which is too much in my opinion.

data/tribes/buildings/productionsites/empire/vineyard/init.lua
```diff
@@ -48,6 +48,18 @@ wl.Descriptions():new_productionsite_type {
          -- TRANSLATORS: Completed/Skipped/Did not start working because ...
          descname = _("working"),
          actions = {
+            -- the longest possible transition from tiny to ripe field:
+            --   2 * (22 + 28 + 40) = 180 sec
+            --   see data/tribes/immovables/grapevine/ and src/logic/map_objects/immovable_program.cc - ImmovableProgram::ActAnimate::execute()
+            -- how many fields are needed for 100% productivity in worst case (nearest fields, longest field transition):
+            --   1 + (180 - (5 + 2.5 * 1.8 + 5)) / 48 = 4.448
+            -- calculation of productivity with 4 fields:
+            --   irwinhall(8, scale=1/8).cdf(4/4.448) = 0.999996
```
bunnybot: [tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4725#issuecomment-2592765)*
*On Fri Jan 17 18:07:39 CET 2025, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) wrote:*

---

No. I mean that we calculate the distribution of the sum of 6 variables with uniform distribution for the growth of one field (2 times the durations of the 3 growth steps). Isn't that what `irwinhall(6)` means? The number of fields is in the target value (the 4 in 4/4.448).

bunnybot: [palino](https://codeberg.org/palino)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4725#issuecomment-2593968)*
*On Sat Jan 18 08:38:28 CET 2025, [**Pavol Gono (palino)**](https://codeberg.org/palino) wrote:*

---

Yes, `irwinhall(6)` is for sum of 6 random variables (2 times 3 growth steps). Question is, how many growth steps should we consider here.
I started calculations using the numerator (4) of the 4/4.448 - actual number of free fields.
But when thinking about it now, the denominator (ceil(4.448)) seems better for me - number of fields needed for 100% productivity. Actual number of free fields should not change the shape of cumulative distribution function, just position on X axis (with the exception of case when denominator overflows/underflows the whole number, because of walking distance change).
So I am inclined to following change for vineyard:
```
-- calculation of productivity with 4 fields:
--   irwinhall(10, scale=1/10).cdf(4/4.448) = 0.9999997
-- calculation of productivity with 3 fields:
--   irwinhall(10, scale=1/10).cdf(3/4.543) = 0.961
```

You proposed the opposite "direction" - `irwinhall(6)`. Why do you think this is better?

bunnybot: [tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4725#issuecomment-2594202)*
*On Sat Jan 18 12:06:05 CET 2025, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) wrote:*

---

On one side we have the random growth time of one field: `irwinhall(6, scale=1/6) * 180`
On the other we have the fixed[^a] time of the farmer tending to the other fields, plus the parallel time spent on the current field: `t_other = (n - 1) * 48 + (5 + 2.5 * 1.8 + 5)`
So the probability of the field getting ripe on time is `irwinhall(6, scale=1/6).cdf(t_other/180)`

[^a]: if we're looking for the smallest possible number, then he/she has no choice, there's always only one fully grown field, that gets harvested and planted again, so all other fields get their turns exactly once before the farmer can return to harvest the field under consideration

bunnybot: [palino](https://codeberg.org/palino)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4725#issuecomment-2819468)*
*On Fri Feb 14 10:54:07 CET 2025, [**Pavol Gono (palino)**](https://codeberg.org/palino) wrote:*

---

I totally forgot about 3 field transitions, so indeed, `irwinhall()` should have parameter 6, when considering one field.
I used the ratio of field numbers (available / total needed for 100% productivity) so far, you used the ratio of times. I like your proposal, it is possible then to use actual time of cycle.
What I am not convinced of is the parameter of Irvin-Hall distribution - whether it should be 6 or 6 multiplied by N fields (for vineyard 6 or 6*5 or eventually 6*4).
I would like to run a game simulation to prove which distribution function is correct - I'll describe it in the main thread.

bunnybot (commented):
[tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4725#issuecomment-4613924)*
*On Sat May 17 01:17:13 CEST 2025, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) **approved** this pull request:*

Thank you, that was a lot of work.

I only approve reluctantly, honouring the long time this took, but I'd [still](https://codeberg.org/wl/widelands/pulls/4725#issuecomment-1567484) prefer the min and max times to be given for the single closest and farthest fields.

bunnybot (commented):
[tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4725#issuecomment-4889865)*
*On Wed May 28 00:37:45 CEST 2025, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) **approved** this pull request:*

Thank you very much.

Issue #6296, opened 2024-01-06
vcpkg dependency caching still needs improvement


bunnybot:
[<img align='left' src='https://codeberg.org/avatars/593dc43e2444c25db9481960e5b693ab' alt='tothxa' height='48' />](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/issues/4659)*
*Created on Sat Jan 06 17:10:39 CET 2024 by [**Tóth András (tothxa)**](https://codeberg.org/tothxa)*

---

**Is your feature request related to a problem? Please describe.**
When `install-dependencies.sh` changes, it affects the vcpkg cache key. In theory, the first consequent run should store the new cache, but that doesn't work for some reason (e.g. https://github.com/widelands/widelands/actions/runs/7431196023/job/20222167930 (Post Restore vcpkg and artifacts)):

```
Post job cleanup.
Warning: Path Validation Error: Path(s) specified in the action for caching do(es) not exist, hence no cache is being saved.
```

Furthermore, in this case the next weekly run reports failure after otherwise correctly storing the new cache, due to the line ending issue (https://github.com/widelands/widelands/actions/runs/7431372589/job/20222607271):

```
warning: in the working copy of '.github/scripts/vcpkg_ref', LF will be replaced by CRLF the next time Git touches it
```

**Describe the solution you'd like**
 - normal builds of master should be able to store a new cache when needed
 - weekly runs shouldn't report failure when they built a new cache, but didn't need to change `vcpkg_ref`

**Implementation steps**
I have no idea for the first problem. For the second maybe we could use `git diff --stat` instead of `git status`? Or better yet, compare the content of `.github/scripts/vcpkg_ref` with `"${{ needs.vcpkg-packages.outputs.vcpkg_ref }}"` directly.

bunnybot:
<img align='left' src='https://codeberg.org/avatars/9c1173656bdf1f05abb2e8485b1d8c835be2116bbbb436b3bd2a3b4bb636c8f6' alt='matthiakl' height='48' />*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/issues/4659#issuecomment-1435920)*
*On Thu Jan 11 18:51:59 CET 2024, ** (matthiakl)** wrote:*

---

> due to the line ending issue

Can probably fixed by adding `git config core.autocrlf false` at the beginning of the job, to tell git to not mess with line endings.

> but that doesn't work for some reason

[...]

PR #6389, opened 2024-03-01
Add gettext tools to vcpkg dependencies

### LINKED ISSUE #6383: msvc builds are broken in the workflow: missing gettext
**Describe the bug**
msvc builds fail in the workflow, cmake fails because it doesn't find gettext, even though it seems to be installed in the dependency installation step. Both 32 and 64 bit builds are affected.

vcpkg dependencies are restored from 3 weeks old cache, last 2 attempts to update also failed in link step of test Widelands build due to missing wavpack symbols.

**To reproduce**
See:
https://github.com/widelands/widelands/actions/runs/8082106344/job/22087237388
https://github.com/widelands/widelands/actions/runs/8078369949/job/22077932922

vcpkg update failures:
https://github.com/widelands/widelands/actions/runs/8013830338/job/21891473555
https://github.com/widelands/widelands/actions/runs/7925133118/job/21637803152

matthiakl:
**Type of change**
Bugfix 

**Issue(s) closed**
Fixes #6383 
Re #6296

**New behavior**
Explicitly install gettext tools, which have been used from the git distribution before.

**Additional context**
This will build a new cache but remains at the old vcpkg version where wavpack is not needed.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -24,7 +24,7 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v4
-      if: github.ref == 'refs/heads/master'
+      # if: github.ref == 'refs/heads/master'
       with:
         fetch-depth: 1
     - name: Check for changed files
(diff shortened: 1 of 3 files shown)


tothxa (approved):
It looks like it works. Thank you. :)

@bunnybot merge

Issue #3618, opened 2019-12-20
Convert Blender Files


stonerl:
With some of our blender files seem to incompatible with Blender versions newer than 2.79. They crash Blender on opening. We need go through the files and see which of them crash on the latest version.

Those files need to be saved (overwritten) again with blender 2.79. Maybe we can automate this and have all files rewritten to future proof them.

See also [this](https://www.widelands.org/forum/topic/4660/) forum topic.

gunchleoc:
Some of the files will also need manual editing, e.g. the Barbarian roof grass textures.

Automation is a good idea anyway though.

tothxa:
My experience with the hi-res re-exports was that the media repo on launchpad already contains many models that were (auto?)"converted" to newer blender version(s) without verifying the render results. It looks like blender development doesn't really care about breaking compatibility, and there were already several subsystems that got completely rewritten in an incompatible way between 2.4x (when most models of the first 3 tribes were developed) and 2.7x (the latest major release in which some models actually render properly). Blender 2.8x completely removed the old renderer for which all Widelands models were developed, so now all lighting and shadow work needs to be redone. All these changes in blender require a lot of manual work just to bring the models back to where they were before. I don't think it is worth it. blender's solution to backwards compatibility seems to be that they have an archive of all old release binaries.

So instead of converting, the problematic models should be actually reverted to the last blender version that can render them properly.

PR #6392, opened 2024-03-02
Optimize images with PNGQuant

bunnybot:
[Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4754)*
*Created on Sat Mar 02 20:48:34 CET 2024 by [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese)*

---

**Type of change**
Pre-release image optimizations.

**Issue(s) closed**
Reduces data dir size by 13.2 MB.

This is net 41 MB up from Release 1.1 due to the new high-resolution exports.

**Possible regressions**
broken images


bunnybot (commented):
tothxa*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4754#issuecomment-1638219)*
*On Mon Mar 04 20:43:10 CET 2024, **Tóth András (tothxa)** **approved** this pull request:*

Thank you. This is now much safer. I tested Amazons and the opening messages of atl02 and fri04, and the images are fine on my system.

> Here's with only these images converted:
> ```
> $ utils/run_pngquant.sh data/campaigns/
> Running pngquant...
> 40 images converted, 322 images skipped, 1207817 bytes saved.
> $ utils/run_pngquant.sh data/tribes/ships/amazons/
> Running pngquant...
> 64 images converted, 1 images skipped, 11206704 bytes saved.
> $ utils/run_pngquant.sh data/tribes/initialization/amazons/
> Running pngquant...
> 14 images converted, 39 images skipped, 198444 bytes saved.
> ```

So it looks like Amazon ship had been most of the gain all this time.

> I don't really care that much about which tools we use in which order

Well, the point of `advpng` is that it needs to be last. :)

> the point as I see it is to reduce filesize by altering the image in a way that is not perceptible to humans, and the pixel values do not need to stay perfectly constant.

OK, but using multiple runs of lossy transformations makes me cringe. What isn't supposed to be noticeable can build up to become quite ugly, as the baseline of each new run is the result of the previous one, not the original (which is not known at that point anymore).

> And while several MB are a good saving, a few hundred bytes do not make a significant difference IMHO.

Agreed. Especially for actual disk usage, considering block/cluster sizes.

All in all, if we ever get around to dealing with #3618, it would be nice to create makefiles/scripts to rebuild all images automatically, and that could use `pngquant`+`advpng` as post-processing.

PR #6447, opened 2024-05-03
update macOS versions in the CI

hessenfarmer:
Update of our macOS runners due to dropped support for MacOS 12


tothxa:
Maybe [using APFS for the DMG](https://github.com/actions/runner-images/issues/7522#issuecomment-2299918092) (in `utils/macos/build_app.sh`)  could help with the build failures?

hessenfarmer:
I updated the runners.
As macOS 12 is out of support officially since september and deprecated by github we should switch to macOS13 for x86. 
I added the macOS 15 beta image to possibly identify any issues early. However for our releases we should stay at the lowest images available as afaik newer macOS versions always could run builds from older versions. 

so the only official version of macOS we do not support due to runner limitations is macOS 13 on arm64.

.github/workflows/build_macos.yaml
```diff
@@ -9,14 +9,19 @@ jobs:
...
-    name: MacOS ${{ matrix.config }} ${{ matrix.compiler }} ${{ matrix.arch }}
+          - os: 13
+            osname: macOS13 x86
+          # - os: 14-large
+            # osname: macOS14 x86
+          - os: 14
+            osname: macOS14 arm64
+          - os: 15
+            osname: macOS15 arm64
+    name: MacOS ${{ matrix.config }} ${{ matrix.compiler }} ${{ matrix.osname}}
```
tothxa:
```suggestion
    name: MacOS ${{ matrix.os }} ${{ matrix.config }} ${{ matrix.compiler }} ${{ matrix.arch }}
```

(or without arch if we remove the include block and get arch inside the runner)

.github/workflows/build_macos.yaml
```diff
@@ -9,14 +9,19 @@ jobs:
       matrix:
         config: [debug, release]
         compiler: [clang]
-        arch: [x86, arm64]
+        os: [13, 14, 15]
+        # arch: [x86, arm64]]
```
tothxa: You could get arch from `uname -m`. I've tested it [here](https://github.com/tothxa/widelands/pull/22/files). OK, it won't be available for the matrix job name, where it's not very important IMO, but you can remove the `include` block, which is a big clean-up IMO.

.github/workflows/build_macos.yaml
```diff
@@ -9,14 +9,19 @@ jobs:
...
         # Check https://github.com/actions/runner-images#available-images
         include:
-          - arch: x86
-            os: 12
-          - arch: arm64
-            os: 14
-    name: MacOS ${{ matrix.config }} ${{ matrix.compiler }} ${{ matrix.arch }}
+          - os: 13
+            osname: macOS13 x86
```
tothxa: You use this for both the artifact name, which deliberately uses spaces, and for the DMG filename, which avoids spaces - but fails doing so this way. Oh, and in the matrix job name, where it duplicates "MacOS".

So I think you should just stick to the old formats, and add any extra you'd like just there.

.github/workflows/build_macos.yaml
```diff
@@ -35,15 +40,15 @@ jobs:
...
         mkdir build_wl
         cd build_wl/
-        ../utils/macos/build_app.sh "--${{ matrix.config }}" "--${{ matrix.compiler }}"
+        ../utils/macos/build_app.sh "--${{ matrix.config }}" "--${{ matrix.compiler }}" "--no-website"
         grep -v REVDETECT-BROKEN ../build/VERSION
         DMGPATH="$(pwd)"
         DMGFILE="$(ls *.dmg)"
         echo "DMG file is: $DMGPATH/$DMGFILE"
-        mv "$DMGFILE" "${{ github.workspace }}/Widelands-${{ github.sha }}-${{ matrix.config }}-${{ matrix.compiler }}-${{ matrix.arch }}.dmg"
+        mv "$DMGFILE" "${{ github.workspace }}/Widelands-${{ github.sha }}-${{ matrix.config }}-${{ matrix.compiler }}-${{ matrix.osname }}.dmg"
```
tothxa:
```suggestion
        mv "$DMGFILE" "${{ github.workspace }}/Widelands-${{ github.sha }}-${{ matrix.config }}-${{ matrix.compiler }}-macos${{ matrix.os }}-${{ matrix.arch }}.dmg"
```

(or:

```suggestion
        ARCH="$(uname -m)"
        DMGNEWNAME="${{ github.workspace }}/Widelands-${{ github.sha }}-${{ matrix.config }}-${{ matrix.compiler }}-macos${{ matrix.os }}-${ARCH}.dmg"
        mv "$DMGFILE" "$DMGNEWNAME"
        echo "arch=$ARCH" >>"$GITHUB_ENV"
        echo "dmg_name=$DMGNEWNAME" >>"$GITHUB_ENV"
```
and use
```
path: ${{ env.dmg_name }}
```
in the upload step to remove the duplication)

.github/workflows/build_macos.yaml
```diff
@@ -35,15 +40,15 @@ jobs:
...
 
     - name: Uploading DMG
       uses: actions/upload-artifact@v4
       with:
-        name: Widelands ${{ matrix.config }} ${{ matrix.compiler }} MacOS ${{ matrix.os }} ${{ matrix.arch }} AppImage
-        path: ${{ github.workspace }}/Widelands-${{ github.sha }}-${{ matrix.config }}-${{ matrix.compiler }}-${{ matrix.arch }}.dmg
+        name: Widelands ${{ matrix.config }} ${{ matrix.compiler }} ${{ matrix.osname }} AppImage
```
tothxa, line 53:
```suggestion
        name: Widelands ${{ matrix.config }} ${{ matrix.compiler }} MacOS ${{ matrix.os }} ${{ env.arch }} AppImage
```
if you go with the above.

.github/workflows/build_macos.yaml
```diff
@@ -35,15 +40,15 @@ jobs:
...
+        path: ${{ github.workspace }}/Widelands-${{ github.sha }}-${{ matrix.config }}-${{ matrix.compiler }}-${{ matrix.osname }}.dmg
```
tothxa: (see above)

.github/workflows/build_macos.yaml
hessenfarmer: well I chose this due to the attempt to use more runners.    
I did not realize that not all runners are provided for FOSS. However the different runners for one macOS version are distinguished by a suffix like -large or -xlarge this would require the os variable to be for example 13 (or 13-large) which is  a x86 runner or 13-xlarge which is a arm runner.

as this would be of no benefit in the Artifact name I chose to go with the osname to combine Arch and OS Version to be used for artifacts and Jobnames 

However as for FOSS the runners with the suffix are not currently available we could just skip this PR and use your branch instead

tothxa: OK, I guess it can stay this way in case we ever can use the suffixed runners.

tothxa (approved):
Thank you. As far as I'm concerned, the names are OK this way. @CB/Nordfriese ?

tothxa:
BTW if we get hdiutil failures again on macos 13, then I'll open a new PR with my changes to `build_app.sh` with all suggested workarounds.

hessenfarmer:
> BTW if we get hdiutil failures again on macos 13, then I'll open a new PR with my changes to `build_app.sh` with all suggested workarounds.

you can start doing this as it happened although I merged in master to avoid this. 
see 
https://github.com/widelands/widelands/actions/runs/11919138070/job/33223251508

hessenfarmer:
thanks for the review

bunnybot (commented):
[Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4809#issuecomment-2463573)*
*On Thu Nov 21 11:24:53 CET 2024, [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese) **approved** this pull request:*

<@>bunnybot update merge

PR #6411, opened 2024-03-14
Replace gettext with tinygettext

bunnybot:
[Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4772)*
*Created on Thu Mar 14 13:41:47 CET 2024 by [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese)*

---

**Type of change**
Refactoring

**Issue(s) closed**
Fixes #4978
Fixes #6450

**How it works**
Embed tinygettext to replace gettext.

Language changes are now applied immediately. We do not mess with any environment variables, except to read `LANG` and/or `LANGUAGE` once at startup to find out the platform's system language.

This removes the dependencies on gettext and libintl. We still need xgettext to build the catalogues in the automated workflow and that's it.

Tinygettext reads the PO files directly, so I have moved them all into the data/i18n directory. I hope I updated all occurrences and that builds will still work.

Tinygettext can not use MO files. Therefore, add-on translations are broken until I make a branch for the server to serve us the raw PO files instead of precompiled MO files. (Which I will do after this branch has been merged.)

Since translations no longer need to be compiled, I removed the `-DOPTION_BUILD_TRANSLATIONS` option.

Since the diff is a bit cluttered by the many renamings and the new module, I recommend reviewing this commit by commit.

**Possible regressions**
Strings not being translated correctly
The translation maintenance scripts

PR #6453, opened 2024-05-17
Update add-on i18n for tinygettext

bunnybot:
[Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4815)*
*Created on Fri May 17 13:15:06 CEST 2024 by [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese)*

---

**Type of change**
Follow-up to #6411
Counterpart to wl/wl_addons_server#88 (currently served on alpha)

**New behavior**
We now get add-on translations as PO files from the server. We can now have a cleaner directory structure, and since we control our own cache now, we no longer need the hack of adding the i18n version to the filename.

**Possible regressions**
Add-on translations, and add-on translation upgrades


src/ui_fsmenu/addons/manager.cc
```diff
@@ -1334,36 +1334,23 @@ bool AddOnsCtrl::is_remote(const std::string& name) const {
...
 
 	assert(!g_fs->is_directory(new_locale_path));
-	// Delete outdated translations if present.
-	for (const std::string& mo : g_fs->list_directory(new_locale_dir)) {
-		if (strncmp(FileSystem::fs_filename(mo.c_str()), addon_name.c_str(), addon_name.size()) ==
-		    0) {
-			g_fs->fs_unlink(mo);
-		}
+	if (g_fs->file_exists(new_locale_path)) {
+		g_fs->fs_unlink(new_locale_path);
```
hessenfarmer, line 1353: we had problems in the past and still have imhgo with unlink on windows.  Is there an addon with which I could test this from the server.

hessenfarmer (commented):
LGTM however needs testing see inline comment

hessenfarmer:
> Nordfriese_[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4815#issuecomment-1835408)_ _On Mon May 20 13:16:34 CEST 2024, **Benedikt Straub (Nordfriese)** wrote:_
> 
> You can use the Alpha server `--addon_server_port=7377` Not all add-ons are functional in the Alpha database, but at least Foreign Planet and Frisians Economy Ultra both have working translations there and should be 100% translated in German.

ok. so to properly test this I need to use the alpha, and download the frisians Ultra, afterwads I need to change the version manually and redownload it to have the files in question existing. correct?

hessenfarmer:
ok der File Handle und unlink funktioniert auch unter Windows.

hessenfarmer (approved):

hessenfarmer:
vielleicht nur ne Randnotiz aber ich hatte jede Menge could not translate: (0 oder 1) messages im log.    
Wir sollten diesen Part und den Server Part aber wohl zusammen comitten, damit es kompatibel bleibt.    
Wie sollen wir denn mit den Mo Leiochen umgehen, die in den Addon Verzeichnissen noch existieren?

hessenfarmer:
@bunnybot update

hessenfarmer:
> IMHO we should just keep the MO files around for now, so if users switch between v1.2 and master they don't get their translations autodeleted every time. It's just a few kB so it doesn't really matter much, but after v1.3 we could maybe add a simple cleanup scanner.

I had exactly the same in mind, but just wanted to be sure that we do not forget planning for it. 
Thanks

PR #6483, opened 2024-07-30
Dropdown Listselect: horizontal centering of pics

cake-pie:
**Type of change**
Polish

**Issue(s) closed**
Fixes pictures displayed off-center when different entries in a pictorial dropdown list have differently sized images.

**How it works**
When drawing a list entry, compensate for difference between overall `max_pic_width_` of the whole list vs width of the picture in the entry.

**Screenshots**

Before:
dropdown-listselect-pic-before

After:
dropdown-listselect-pic-centered

Full-blown implementation with left/center/right horizontal alignment options. Not clear this would really be required. Also more complicated due to need to consider localization and text direction.

**Possible regressions**
Maybe there's a dropdown or other kind of listselect somewhere that requires its pictures to be flush left (or flush right for RTL text direction)

### DIFF
--- a/src/ui_basic/listselect.cc
+++ b/src/ui_basic/listselect.cc
@@ -474,11 +474,13 @@ void BaseListselect::draw(RenderTarget& dst) {
 
 		// Now draw pictures
 		if (er.pic != nullptr) {
-			dst.blit(Vector2i(UI::g_fh->fontset()->is_rtl() ?
-                              get_eff_w() - er.pic->width() - (max_pic_width_ - er.pic->width()) / 2 - 1 - kIndentStrength * er.indent :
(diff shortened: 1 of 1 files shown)


hessenfarmer (approved):
code looks reasonable to me
@tothxa

bunnybot (commented):
Nordfriese*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4845#issuecomment-2140264)*
*On Thu Aug 01 09:42:05 CEST 2024, **Benedikt Straub (Nordfriese)** **approved** this pull request:*

LGTM, thanks for fixing this :)

<@>bunnybot merge force

PR #6498, opened 2024-08-10
Keep vcpkg_ref in a repository variable instead of a file

tothxa:
**Type of change**
New feature / Improvement

**Issue(s) closed**
Re https://codeberg.org/wl/widelands/pulls/4858#issuecomment-2164355

**New behavior**
The latest tested vcpkg version is stored in a repository value instead of in a file within the repository. This gets rid of the regular update commits.

It is also stored in a cache for the sake of PRs from forked github repos.

**Possible regressions**
msvc builds and dependency caching

**Additional context**
See test runs in https://github.com/tothxa/widelands/actions


tothxa:
OK, I'm not allowed to test using the secret here, but it works in my repo, just like everything else.

@cb/Nordfriese Could you please set the `VCPKG_REF` variable to `f5398d978` to let the CI succeed?

Then this should be ready... Hopefully... But we'll only know for sure after the first new PR run and the first vcpkg update... :(

bunnybot (commented):
[Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4860#issuecomment-2168713)*
*On Mon Aug 12 11:46:06 CEST 2024, [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese) **approved** this pull request.*

tothxa:
> There are currently no more pending builds, so we can delete the vcpkg caches and then run check_vcpkg, so it could create a fresh build and then the vcpkg_ref cache.

OK, I just did it. At least that'll properly test the whole thing at once.

tothxa:
> > There are currently no more pending builds, so we can delete the vcpkg caches and then run check_vcpkg, so it could create a fresh build and then the vcpkg_ref cache.
> 
> OK, I just did it. At least that'll properly test the whole thing at once.

Worked nicely. I've just triggered re-runs of your old PRs, and they all got past the dependencies installation phase without problems.

PR #6499, opened 2024-08-11
Make Show/Hide buildings option in game also affect immovables

### LINKED ISSUE #6481: Show/Hide immovables option in game

(Some) trees, rocks and other immovables can obscure census/status text behind them.
rockblock seetrees

Buildings can be hidden, but there does not exist a similar option for immovables.
bldg-hide-transparent

To be fair, most of the time the text isn't completely covered over, but the player shouldn't have to "guess the text", which can be particularly annoying in the case of obscured first digit of percentage (e.g. Ranger's Hut in screenshot above). Also, can't discount the possibility of user-created content with some kind of custom oversized immovable.

A "Hide Immovables" option. This option already exists in editor, albeit the objects disappear completely rather than turning partially transparent.

Force all "text" such as census/status to be on a different "layer" than all buildings/immovables so they never get obstructed. This option is likely not aesthetically ideal.

cake-pie:
**Type of change**
New feature

**Issue(s) closed**
Fixes #6481

**New behavior**
The in-game show/hide menu option for buildings now also affects immovables like rocks and trees

**How it works**
Fundamentally same mechanism as show/hide of buildings.
Opacity level now shared for buildings and immovables.

Separate entry for immovables; this was rejected.

**Screenshots**
barbarian campaign mission 1 start; both buildings and immovables are hidden

**Additional context**
Coincidentally fixes build progress percent (int) being passed to opacity (float):
https://github.com/widelands/widelands/blob/ea759febc15901c5dbc6888f8aa173ae338a3c9b/src/logic/map_objects/immovable.cc#L485-L486


cake-pie:
> Flags and border markers

I don't think those are as big a problem in terms of visual obstruction; have not actually been bothered by those, and don't think there really needs to be a way to hide them

I would in fact strongly discourage adding flags and borders under the same show/hide toggle option as buildings and trees/rocks type of immovables. Couple of use cases:

First, if you look at the screenshot above, I think it shows potential to be quite useful for inspecting the road network without the clutter of / blocking by building and tree sprites. You lose that utility if flags cannot be show/hide separately from the buildings+immovables option implemented here.

Similar to that would be borders in a dense forest area with lots of trees (e.g. barbarian campaign, mission 2). Particularly for barbarian, the crossed spears can be rather difficult to see in that situation. You can't use buildhelp as a rough proxy (construction not possible due to trees) so prior to this PR I think the best alternative would be to find your military buildings and toggle their workarea -- quite troublesome.

bunnybot (commented):
[Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4861#issuecomment-2171313)*
*On Tue Aug 13 19:47:32 CEST 2024, [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese) **approved** this pull request:*

Ok, these are valid points. Let's have it then :)

hessenfarmer:
I fear this lead to the immovables permanently hidden in the editor

cake-pie:
> I fear this lead to the immovables permanently hidden in the editor

investigating.

PR #6505, opened 2024-08-14
Use silk rather than cloth for fishing nets and bows

### LINKED ISSUE #3145: Use silk rather than spider cloth for fishing nets and bows
It doesn't make sense to create bowstrings or nets out of cloth. So, unless it's a balancing problem, we should use silk rather than spider cloth for creating Atlantean fishing nets and bows.

 * date created: 2016-09-21T16:20:05Z
 * owner: gunchleoc
 * the launchpad url was https://bugs.launchpad.net/bugs/1626176

cake-pie:
**Type of change**
Bugfix? Commonsense fix?

**Issue(s) closed**
Fixes #3145

**New behavior**
Atlantean toolsmithy uses silk instead of cloth for nets and bows.

**Possible regressions**
Potential game balance issues, which maybe could be tackled by adjusting inputs or production time.

**Additional context**
I considered making the production time 44s longer to compensate for skipping the weaving mill's silk->cloth step, but it doesn't really make sense, the whole point is that it shouldn't need whole sheets of fabric from a commonsense perspective.

A different approach might be to introduce a new ware, some kind of fiber / rope, as the intermediate product i.e. thin spider silk has to be spun into thicker bundles (at the mill, or a new production building) before it can be used at the toolsmithy. It could also have potential to be used as a construction material for building sites or ships. But that is a bigger job requiring graphics for the new ware (and possibly associated buildings/workers) and has even wider implications for game balance. Should probably open a separate issue to discuss it more first, if there is interest in going this direction.

For the time being I think it's better to make a small change first and see what the feedback is like after it gets playtested a bit.

----

**Status**

- [x] Change toolsmithy input queues and production programs
- [x] Adjust affected helptexts
- [x] Adjust economy profiles (default, stockpile)
- [x] Adjust affected starting conditions
- [x] Screen campaign / tutorial texts: none impacted
- [x] Savegame compatibility: out of scope, universal approach to be taken in a separate PR.
- [x] Balancing adjustments to production time / starting wares / etc. : up for discussion


hessenfarmer (changes requested):
as silk is now much more important we should add 5 to the default quantity of silk and reduce the quantity for cloth by 5.
furthermore we should exchange the values for cloth an d silk in the stockpile economy profile as well.
Edit: 
the campaign texts of the atlantean campaigns might need a change as well.

hessenfarmer:
I agree with the savegame compatibility argument and the increased difficulty for deadlock recovery argument of @Nordfriese. 
However it is just one building (the weaving mill) that is not needed for the production of hunting bows and fishnets in this PR, where hunting bows are not needed that much due to the Atlanteans not being able to replenish game. 
The need for the weaving mill is present there anyhow to avoid the deadlock and could be increased by lowering the cloth in the starting wares. But this would increase the risk for the deadlock again, while the risk to run into it is decreased with this PR. so for me having a lower risk to run into the deadlock and an increased difficulty to recover from it might me an acceptable change.  


so for logical reasons as outlined in the issue by our former chieftain I would vote +1 as long as we can find a solution for the compatibility issue.

cake-pie:
Game balance discussion first:

---

from Nordfriese:

> ...  most important economical deadlock ...
> Replacing any use of cloth with raw silk would thus make the early game far easier for Atlanteans

I don't think this is true at all, considering that cloth is made from silk at 1:1 input to output parity, and before this PR there are no other uses for silk anyway. We'd want to adjust the initial amount of these in warehouse at game start, reduce cloth and increase silk by (roughly) the same amount. So, before you are able to get the whole production chain up and running, the net potential maximum quantity of cloth -- supposing you do build a weaving mill and turn all the silk into cloth -- still remains the same.

I think once the amount of starting wares have been adjusted appropriately, the changes in this PR can actually make things slightly harder for the Atlantean player:

1. Assume unchanged demand for nets/bows. Less starting cloth -> higher likelihood of early game cloth shortage for construction, meanwhile more starting silk -> increased probability of silk surplus. This can put pressure on the player to build the weaving mill sooner, perhaps even before they have spider farm to provide steady supply of required silk input.

2. Lower cap on potential maximum fish/meat production before establishing silk production. Given current starting silk (A) and cloth (B) amount and assume fixed construction demand (C) we can put (A+B-C) into making bows/nets. If we change some (D) cloth into silk in starting conditions with this PR, max number of bows/nets that can be made is reduced (A+D-C).

3. Whereas before this PR you can blindly turn all silk into cloth and not worry about keeping any in stock, now there is a reason to keep some in reserve for toolmaking. The player now has to care about managing how much silk is being distributed to toolsmithy vs weaving mill.


> ...even if the time were increased.

Although the weaving mill is no longer required for making nets/bows, it remains an essential enabler for the resulting increased food output to deliver any actual impact. All fish/meat goes through the smokery, which needs cloth to construct. Mines and training sites are the consumers of smoked fish/meat, they too need cloth to build. There is no use adding more fishers/hunters if number of smokeries and food consuming buildings remain capped due to having no cloth production. With this in mind, I'd argue that the greater ease of making nets/bows can in fact be offset by increasing their production time.

---

from hessenfarmer:

> add 5 to the default quantity of silk and reduce the quantity for cloth by 5.

Sorry, which "default" is this? Do you mean initial wares at game start?

Excluding hard mode and shipboard starts, it is currently 9 silk and 5 cloth in the starting HQ/warehouse. I do think cloth should be reduced, but going to zero is probably overkill.


> exchange the values for cloth and silk in the stockpile economy profile.

makes sense.


> just one building (the weaving mill) that is not needed for the production of hunting bows and fishnets in this PR

And as I outline above, it is useless to make more bows and nets without also having weaving mill -> cloth -> smokery, mines & trainingsites.


> where hunting bows are not needed that much due to the Atlanteans not being able to replenish game.

True, bows have diminishing usefulness as game goes on longer. But at 1×log + 1×(silk/cloth) it is cheaper than 2×(silk/cloth) fishing nets, thus offering an interesting tradeoff in the early game: if you have enough animals around, for the same amount of (silk/cloth) resource you can boost food production more with hunters than fishers -- albeit only temporarily, until game is depleted. So we should not be too dismissive about the bows.


> The need for the weaving mill is present there anyhow to avoid the deadlock and could be increased by lowering the cloth in the starting wares. But this would increase the risk for the deadlock again, while the risk to run into it is decreased with this PR. so for me having a lower risk to run into the deadlock and an increased difficulty to recover from it might me an acceptable change.

Correct, it should be possible tune the starting wares to find some desired levels of deadlock risk and difficulty of recovery.

cake-pie:
> the campaign texts of the atlantean campaigns might need a change as well.

Will inspect later, higher priority to resolving savegame compatibility first.


> loading itself is not a problem, existing buildings are loaded with the old inputqueues, while newly built ones will have the new. However even the old buildings will try to use the new programs, so they won't be able to produce the wares that need the new input. 

Thanks for investigating!


> I tried changing the programs to accept either silk or cloth ...
> ... adding a building and input specific exemption in the production program check for v1.3?

I don't think it is advisable to let the old buildings continue function with old inputs and old production programs. Depending on number of old vs new toolsmithy they have, players will be at a comparative (dis)advantage. Human players should react by swapping out old buildings and start using new ones instead. I don't think AI knows how to respond to the change. It might also get hard to keep track of old vs new buildings, as you have to click each one to check what kind of inputqueue they have.

The old buildings will continue to exist in subsequent saves of the game; they could potentially remain an issue past v1.3.

My preference is to detect existing buildings with the old cloth inputqueues and change them to equivalent number of silk. The cost is losing 44s of weaving mill production time for each (max 4) ware already present in the building. It ensures consistent behavior across buildings, and no old zombie buildings can be in v1.3 savefiles.

This can be done as part of the savegame loading (automatic, more convenient), or with a separate savegame migration tool (more user control). A possible additional safety measure is to refuse to overwrite any existing < v1.3 when saving games, so players can revert to an earlier state if they want -- e.g. player might decide to empty out the cloth inputqueue first before migrating the savefile.

hessenfarmer:
> from hessenfarmer:
> 
> > add 5 to the default quantity of silk and reduce the quantity for cloth by 5.
> 
> Sorry, which "default" is this? Do you mean initial wares at game start?

no, I meant the defined default quantity in the tribes units.lua file where this is defined for all wares (in fact this defines the default eco profile.  

> 
> Excluding hard mode and shipboard starts, it is currently 9 silk and 5 cloth in the starting HQ/warehouse. I do think cloth should be reduced, but going to zero is probably overkill.

I don't think we need to change starting wares, as we still need the same amount of cloth for construction, otherwise we would significantly increase the risk for the deadlock, togehter with more difficulty to resolve it.

cake-pie:
> I meant the defined default quantity in the tribes units.lua file

Found it, thanks for clarifying!

> I don't think we need to change starting wares, as we still need the same amount of cloth for construction,

Okay, will hold off on that for now. I suppose it is most common to put the starting cloth primarily into construction and very rare to make a fishnet or bow and boost food production super early in the game.

> The official policy is that we don't support continuing old games through more than 2 subsequent stable releases.

Good to know, thanks.

> universal, version independent solution is possible for new inputs, changing `MapBuildingDataPacket::read_productionsite()` ...

Thanks for looking into this; I concur based on my own investigation in parallel. Am testing with preliminary implementation [here](https://github.com/cake-pie/widelands/tree/inputqueue-compat). Will dig further re: any interactions with pending requests due to decrease in queue capacity.

cake-pie:
Insofar as I can tell, ware requests are not written to / read from the save file (although worker requests are) so I think it should not cause any problems.

It is probably not safe to outright remove an input queue, but we can gradually reduce its capacity while ejecting existing wares. Per this snippet:

https://github.com/widelands/widelands/blob/71c29e2eb78874b1eb8ebaab99d654df53307193/src/logic/map_objects/tribes/productionsite.cc#L1056-L1067

As long as the building's worker is not busy with other higher priority task, they will transport "excess" wares from the building to its flag. As for incoming wares that are no longer wanted, the carrier ("input-worker") is responsible for taking it away (delivery declined).

<details>
<summary>Click for steps to replicate my testing</summary>

- `git remote add` my fork
- checkout  [`inputqueue-compat`](https://github.com/cake-pie/widelands/tree/inputqueue-compat) branch at cake-pie@666288c; compile
- `merge --squash` this `issue-gh3145` branch on top, but do not commit it
- further edit `data/tribes/buildings/productionsites/atlanteans/toolsmithy/init.lua`,
reduce size of logs input queue to 2
- obtain save file: [inputqueue-test.zip](https://github.com/user-attachments/files/16682778/inputqueue-test.zip)
  - (optional) load this in v1.2, pause immediately to inspect: see that it has 4/6 logs, 3/4 cloth, no silk queue. 2 logs and 1 cloth en route, with imminent arrival of one of the logs -- the carrier is already bringing it from the flag into the toolsmithy.
- run and load the save
  - log messages indicate input queue changes: logs decreased (6->2), cloth gone (4->0), silk added
  - carrier bringing log into toolsmithy enters and comes back out immediately, still carrying the log as it is no longer wanted.
  - the enroute wares (2 logs, 1 cloth) get returned to the warehouse
  - open building window of toolsmithy:
    - 4/4 logs, with desired fill at 2; can be adjusted <= 2 but not more
    - 3/3 cloth, desired fill locked at 0
    - silk queue added; wares should start arriving
  - the toolsmithy is stopped, wait for it to finish its current job (tongs) and taking it out the the flag
    - the worker will then start taking out excess logs and cloth
  - cloth queue no longer appears in building window after last cloth removed -- technically, it still exists with capacity 0, just doesn't get displayed
- save the game under a different filename
- load this new save game to see the log message about cloth gone again, but this time the input queue is truly removed
- exit from the application, `git reset --hard` to discard all the data file changes
- run and load the new save game, this time it should increase logs 2->6 and cloth 0->4 while silk is detected as gone
</details>

This should take care of savefile compatibility; will check campaign stuff next.

cake-pie:
> the campaign texts of the atlantean campaigns might need a change as well.

Reviewed atleantean campaign texts (as well as tutorial 3) returned by a search of these terms, did not see any issues:

```
grep -r "[Tt]oolsmith" data/campaigns/atl*
grep -r "[Ss]pidercloth" data/campaigns/atl*
grep -r "[^r][Cc]loth" data/campaigns/atl*
grep -r "[Ss]ilk" data/campaigns/atl*
grep -r "[Bb]ow" data/campaigns/atl*
grep -r "[Nn]et" data/campaigns/atl*
```

I did stumble across this (unrelated):
https://github.com/widelands/widelands/blob/bf9af9cbc3d16214d01e8e34ec5fc65778e7dee3/data/campaigns/atl01.wmf/scripting/mission_thread.lua#L151

For a food production objective, this should probably be a mill instead of sawmill, right?

---

> Just delete (in a safe way) / don't add the wares and queues immediately right on loading.

If the queue happens to be empty, it does get deleted in a safe way.

The issue lies with non-empty queues; what to do with wares currently sitting in them? Present implementation takes advantage of existing mechanisms to have the worker eject the excess items onto the building's flag, and from there the usual behaviors apply as they are moved to a warehouse or reallocated elsewhere directly. This is logical and predictable.

Possible alternatives:
- wares disappear into thin air -- no, this is player-unfriendly 
- teleport wares to a warehouse instantly -- `find_closest_warehouse()` of the building's wares economy would be the logical choice, if it exists, but we cannot guarantee that the building is connected to a warehouse. Using any other arbitrary warehouse is too unpredictable; we really don't know what is appropriate for the player's current logistics network situation; it could end up sending the wares somewhere really silly / inconvenient.
- teleport wares to the building's flag directly -- not safe. flags have limited capacity, no guarantee it will be able to accommodate instantaneous dumping of all excess wares


> See `economy/inputqueue.c` / `InputQueue::read()`. Called as `wq->read()` in `MapBuildingDataPacket::read_productionsite()`.

Looked into it, not an issue. `InputQueue::update()` will re-assess how many wares the building is missing (or has excess) and then adjust/delete the request accordingly; ongoing transfers are cancelled as appropriate.

cake-pie:
> > * wares disappear into thin air
>
> This is what I meant. [snip] a few lost wares don't count

Apparently, spidercloth is a "rare and precious ware" but it's also okay to have it just vanish from a save game. :shrug:

[This](https://github.com/widelands/widelands/compare/master...cake-pie:widelands:inputqueue-compat) should handle any current and future change to inputqueues of a productionsite (add, remove, or grow/shrink in size).
FWIW, it's not much more complexity -- just 12 LoC -- to not unfairly expropriate wares from the player.

In any case, savegame compatibility can be treated as out of scope for this PR. It doesn't matter whether my approach is adopted or something different is implemented -- either way, it  no longer poses a technical hurdle.

All that remains is the (somewhat subjective) question of balance adjustments to compensate, now that it is easier for atlantean players to start boosting their food production -- though corresponding expansion of processing capacity (smokery) and consumption remain dependent on spidercloth supply. This might include tweaks to production time, starting wares, or perhaps something else.

hessenfarmer:
Well, 
I believe we should only have this if the solution to handle such an exchange properly for all cases is at least agreed and reviewed in this or an own PR.      
I really would like to have a proper solution as it would allow for tribe defin itin adjustments in the future as well, while maintaining savegame compatibility. 
However it seems the solution is based on the name of the PS. this brings 2 concerns to my mind that I am not sure whether they are properly handled.
First is the handling of savegames with addons enabled. how is this solution dealing with savegames saved without an addon that changes the PS while loading with the addon enabled. and vice versa. Same is the handling of scenario specific extensions.    
second is how is the given fix interacting with the legacy lookuptable mechanism for saveloading?

cake-pie:
> savegames saved without an addon that changes the PS while loading with the addon enabled. and vice versa. Same is the handling of scenario specific extensions.

No scope creep, please.

Savegame compatibility for this PR only cares about coping with impacts due to changes of this PR. Its job is to make sure that a game that is saved before this PR can still be loaded and played after this PR, **provided that nothing else has changed**. Likewise for #6150, which also changes production buildings inputqueues -- so we will most likely have a separate PR providing a common solution for inputqueue changes. Yes, having this implemented should help facilitate savegame compatibility for other similar changes to production sites inputqueues in future. But addons, etc remain out of scope regardless.

Currently it looks like `WidelandsMapLoader::preload_map()` drops unwanted addons, so you cannot load a game with a new addon enabled if it was not already in the saved game. Conversely, if the save game requires an addon that is no longer available, it cannot be loaded. Even the same addon, but with different version number, will trigger a warning that compatibility is not assured. This is the correct stance, as addons can be made by third-parties, beyond the control of the project team. We cannot and should not vouch that a modded game will work as desired -- especially if the user tries to do things that are not intended by developers, like loading without necessary addons.

hessenfarmer:
i didn't wan't to crep the scope.
I just wanted to make sure we have a working savegame compatibility for this, which does not cause any other problems elsewhere.

bunnybot (commented):
[tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4866#issuecomment-2278537)*
*On Tue Sep 03 20:47:56 CEST 2024, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) **approved** this pull request:*

As far as I'm concerned, the actual changes of this PR are fine now...

hessenfarmer (approved):
changes are fine for me I just wanted to point out that I would recommend having the compatibility PR firt and then this afterwards. Otherwise we would loose savegame compat if the second PR is delaye for wahtever reason.

cake-pie:
> fake WIP PR that merges this, the compatibility PR, and ... work around the chicken-and-egg problem

SGTM.

---

No other opinions on other adjustments (production time, etc) for game balance?

bunnybot (commented):
[hessenfarmer](https://codeberg.org/hessenfarmer)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4866#issuecomment-2493097)*
*On Tue Dec 03 23:15:02 CET 2024, [**Stephan Lutz (hessenfarmer)**](https://codeberg.org/hessenfarmer) **approved** this pull request:*

after the enabling savegame change has now been merged, this is ready to go imho.

PR #6510, opened 2024-08-16
vcpkg workflow fix for v1.2.1

tothxa:
**Type of change**
Bugfix

**Issue(s) closed**
This should fix msvc builds in the release-1.2 branch

**Possible regressions**
dependency caching may not work in all cases

**Additional context**
`check_vcpkg.yaml` is not actually needed, it's only a sync-up

### DIFF
--- a/.github/workflows/build_windows_msvc.yaml
+++ b/.github/workflows/build_windows_msvc.yaml
@@ -44,7 +44,6 @@ jobs:
         [ -n "$REF" ] && git fetch
         git checkout $REF || git pull origin $REF
         echo "vcpkg_key=${{ hashFiles( './install-dependencies.sh' ) }}-${REF}" >> $GITHUB_OUTPUT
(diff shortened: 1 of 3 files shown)


tothxa:
OK, there's a probable unavoidable regression: I'm afraid caching of dependencies won't work properly. :( The cache key contains the hash of `install-dependencies.sh`, which is different due to dropping `gettext` in master. But IIRC only caches from master or the same PR can be restored in a workflow run.

bunnybot (commented):
[Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4871#issuecomment-2280044)*
*On Wed Sep 04 19:59:01 CEST 2024, [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese) **approved** this pull request.*

PR #6513, opened 2024-08-17
BuildingStatisticsMenu: fix mismatched logic of update_building_list() vs init()

cake-pie:
**Type of change**
Bugfix

**Issue(s) closed**

The building statistics window has entries for all the different buildings that the player can build (own tribe, subject to any building restrictions in scenario play) plus any enemy building types that the player has managed to capture. In its code, `update_building_list()` is intended to detect situations when the set of possible buildings has changed -- either due to scenario script (`allow_buildings`, `forbid_buildings`) or loss/capture of other tribes' building type -- and then trigger `reset()` in order to refresh the entries shown in the window by re-initialization (`init()`).

Part of the logic does not match the intended behavior and deviates from `init()`. Under certain conditions, this mismatch can lead to `reset()` being called every `think()` cycle.

**To reproduce**

1. Add logging outputs for debugging

```diff
void BuildingStatisticsMenu::think() {
	// Update statistics
	const Time& gametime = iplayer().game().get_gametime();

	if (!is_minimal() && (was_minimized_ || (gametime - lastupdate_) > kUpdateTimeInGametimeMs)) { // <-- GH#6512, CB#4873
+		log_info("thinking...");
		update_building_list();
		update();
		lastupdate_ = gametime;
	}
	// Make sure we don't have a delay with displaying labels when we restore the window.
	was_minimized_ = is_minimal();
}
```
```diff
void BuildingStatisticsMenu::update_building_list() {
	// ... snipped ...
	for (Widelands::DescriptionIndex index = 0; index < nr_building_types_; ++index) {
		// ... snipped ...
		if (should_have_this_building != has_this_building) {
+			log_info("update_building_list will now call reset()");
+			log_info("triggered by: %i:%s -- %s%s%s", index,
+				 iplayer().egbase().descriptions().get_building_descr(index)->name().c_str(),
+				 own_building_is_valid(player, index, map_allows_seafaring, map_allows_waterways) ?"Y":"N",
+				 foreign_tribe_building_is_valid(player, index) ?"Y":"N",
+				 has_this_building?"Y":"N");
			reset();
			return;
		}
	}
}
```
```diff
void BuildingStatisticsMenu::reset() {
+	log_info("reset called");
	update();  // In case a building got removed, make sure to deselect it first

```

2. compile, execute. start the second tutorial (warfare), click past the initial tutorial messages, then `b` to open building stats window

3. observe `update_building_list()` calls `reset()` due to `empire_sentry`. this is an enemy tribe building that the player cannot build, and has not captured any yet at the start of the tutorial scenario
```
[00:00:17.149 real] INFO: thinking...
[00:00:17.149 real] INFO: update_building_list will now call reset()
[00:00:17.149 real] INFO: triggered by: 94:empire_sentry -- YNN
[00:00:17.149 real] INFO: reset called
[00:00:18.157 real] INFO: thinking...
[00:00:18.158 real] INFO: update_building_list will now call reset()
[00:00:18.158 real] INFO: triggered by: 94:empire_sentry -- YNN
[00:00:18.158 real] INFO: reset called
[00:00:19.153 real] INFO: thinking...
[00:00:19.153 real] INFO: update_building_list will now call reset()
[00:00:19.153 real] INFO: triggered by: 94:empire_sentry -- YNN
[00:00:19.153 real] INFO: reset called
[00:00:20.181 real] INFO: thinking...
[00:00:20.181 real] INFO: update_building_list will now call reset()
[00:00:20.181 real] INFO: triggered by: 94:empire_sentry -- YNN
[00:00:20.182 real] INFO: reset called
[00:00:21.207 real] INFO: thinking...
[00:00:21.207 real] INFO: update_building_list will now call reset()
[00:00:21.207 real] INFO: triggered by: 94:empire_sentry -- YNN
[00:00:21.207 real] INFO: reset called
```

4. if we additionally comment out the return statement in `update_building_list()`, it can be seen that every empire military building is able to trigger the spurious reset call.

**Diagnosis**

Intended behavior per header file is:
```cpp
/// Whether a building that is used by the player's tribe should be added
bool own_building_is_valid(const Widelands::Player& player,
                           Widelands::DescriptionIndex index,
                           bool map_allows_seafaring,
                           bool map_allows_waterways) const;
/// Whether a building that isn't used by the player's tribe should be added
bool foreign_tribe_building_is_valid(const Widelands::Player& player,
                                     Widelands::DescriptionIndex index) const;
```

`init()` adheres to this and feeds only buildings of the player's tribe to `own_building_is_valid`; then feeds all buildings to `foreign_tribe_building_is_valid` which is responsible for filtering out anything other than building types of other tribes that the player actually currently owns (e.g. by capture).

`update_building_list()` feeds all buildings to both methods. Furthermore, within `own_building_is_valid`:

https://github.com/widelands/widelands/blob/f915c8c5aab0b44dcc08eeeb5b083d6aff39df76/src/wui/building_statistics_menu.cc#L395-L398

This does not adhere to the advertised contract:
- it will accept (return true) building types not of the player's tribe, but currently owned by the player
- `player.is_building_type_allowed()` returns true for all military buildings of other tribes (because it is legal to capture and own them) -- even if the player doesn't currently own any of them

Consequently, under certain conditions, `should_have_this_building` in `update_building_list()` has an incorrect value.

**Solution**
rename `own_building_is_valid` to `own_tribe_building_is_valid` for better clarity
explicitly screen out other tribes' buildings from `own_tribe_building_is_valid`
remove check for owned buildings in `own_tribe_building_is_valid`; that's the job of `foreign_tribe_building_is_valid`

**New behavior**
No longer misbehaving.
Confirmed that reset called to add enemy building types the window upon actually capturing them: play tutorial 2 to conclusion with the building stats window open, and including same debug logging as before to monitor reset calls upon capture of enemy military sites


hessenfarmer (approved):
LGTM.

PR #6514, opened 2024-08-17
Flag: fix nr_of_roads()

cake-pie:
**Type of change**
Bugfix

**Issue(s) closed**
`nr_of_roads()` is identical to `nr_of_roadbases()`

https://github.com/widelands/widelands/blob/f915c8c5aab0b44dcc08eeeb5b083d6aff39df76/src/economy/flag.cc#L357-L367

https://github.com/widelands/widelands/blob/f915c8c5aab0b44dcc08eeeb5b083d6aff39df76/src/economy/flag.cc#L369-L379

**New behavior**
`nr_of_roads()` will now return number of roads (excludes waterways)
`nr_of_roadbases()` continues to return the number of roadbases (roads + waterways)

### DIFF
--- a/src/ai/defaultai.cc
+++ b/src/ai/defaultai.cc
@@ -3705,10 +3705,10 @@ bool DefaultAI::improve_roads(const Time& gametime) {
 
 	// when deciding if we attempt to build a road from here we use probability
 	uint16_t probability_score = 0;
-	if (flag.nr_of_roads() == 1) {
+	if (flag.nr_of_roadbases() == 1) {
 		probability_score += 20;
 	}
-	if (is_warehouse && flag.nr_of_roads() <= 3) {
+	if (is_warehouse && flag.nr_of_roadbases() <= 3) {
 		probability_score += 20;
 	}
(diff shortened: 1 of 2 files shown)


hessenfarmer:
good catch, however I am not sure, whether this was intentional, to keep some compatibility. Have you checked all of the usages of nr_of_roads for whether this change makes any difference

cake-pie:
> checked all of the usages of nr_of_roads

Three mentions in `defaultai.cc` which predate introduction of ferries and waterways. Looks fine to me but would probably be best if someone with more familiarity with the AI code could check.

hessenfarmer:
I have had a look into this. the conclusion is:   
Although the AI does not know how to handle waterways (yet) all three instances should take them into account if this will be implemented, so please change these instances to nr_of_roadbases() in this PR

cake-pie:
My reasoning was to leave it as is until AI gets waterway/ferry support, but I'll take your word for it.

hessenfarmer (approved):
LGTM.

hessenfarmer:
@bunnybot update merge

PR #6531, opened 2024-09-08
WIP: Production site input queues save compatibility

cake-pie:
**Type of change**
New feature

**Issue(s) closed**
Universal savegame compatibility for whenever a PR alters a productionsite by adding, removing, or changing the queue capacity of its inputqueues. Will be used by #6150, #6505.

**How it works**
In `read_productionsite` when loading a production site from a save file, check the ingested inputqueues against the `input_wares()` defined in `ProductionSiteDescr` for that building.

**New behavior**
- [x] New ware added: adds new, empty input queue for that ware to the building
- [x] Queue size increased: queue capacity expanded, existing ware count preserved. Existing max fill setting left unchanged if originally lower than old capacity (i.e. was capped by player), or increased to new capacity if was originally equal to old capacity (i.e. was unconstrained)

- [ ] Queue size decreased / ware removed: see below


cake-pie:
After thinking more on this issue, there are a couple more possible approaches (options A, B) that may be worth considering here, in comparison with the options (C, D) already previously discussed in #6505.

### Option A: no further action
Nothing else implemented beyond 50d2a330. Does nothing in the cases of queue size decreased and ware removed.

In the former case, the queue for that ware is bigger than intended in the current version, but this is harmless. For the latter, it can result in wares sitting and gathering dust in the building which no longer uses them. This is also mostly harmless, the main downside being that it can tie up the player's resources unproductively -- could be alleviated by a message to the player to inform them that the ware is no longer needed, so they can set max fill = 0 if desired.

If minimalism is the goal, then this is close to the bare minimum needed to allow finishing up old saves in the new version.


### Option B: update buildings to the extent possible without sacrificing wares
(adds: 30b62268)

This improves on option A by updating the building -- deleting the queue or reducing its size -- where possible. If this is prevented by the presence of excess wares, sets max fill so that they will get ejected from the building when simulation resumes. Like option A, an oversized or superfluous queue could still exist, but this is harmless. As long as the player doesn't interfere by manually fiddling with max fill settings (can be prevented by informing them of the building changes), a subsequent save-and-load cycle after all excess wares have been hauled out of the building will bring everything in line with the current version.

Like option C, imposes no penalty on the player, but does so using only pre-existing features of `InputQueue`. Similar complexity to option D.


### Option C: aggressively update building while ejecting excess wares
(adds: 3f4a190e)

Already voted down for being too complicated and is only listed for comparison purposes.  
Like option B, but adds machinery to `InputQueue` to prevent player from increasing max fill, and aggressively adjusts queue capacity downwards as excess wares get carried out of the building. This brings the building up to date with the current version as quickly as possible without sacrificing wares.


### Option D: outright ware deletion
(adds: 90a9bb0e)

Favored over option C in prior discussion.  
Deletes excess wares to immediately bring building into full compliance with current version behavior.

cake-pie:
> implement the message to the players in a follow-up PR

Or you can tack it onto this PR along with your desired option by pushing to the `mirror/cake-pie/widelands/psiq-compat` branch on codeberg, if you have maintainer/write access to that; I think it should mirror correctly.

PR #6525, opened 2024-08-28
Expanded dropdown and listselect types; facilitates checkmarks in show/hide menu (GH-6407)

cake-pie:
**Type of change**
New feature

**Issue(s) closed**
Fixes #6407 (by alternative means)

**New behavior**
- Expanded dropdown and listselect types. (See detailed feature breakdown below.)
- The "show/hide" menu in the toolbar now displays a checkmark next to each entry that is enabled (shown).
  - Very easy to tell what is shown/hidden from just a quick glance at checkmarks and icons<br>(compared with having to parse the menu text with the previous behavior)

[demo screencast](https://github.com/user-attachments/assets/0ffd8939-944d-48b1-8fd4-c14d8200d998)

**How it works**
(See implementation details below.)

**Possible regressions**
Representative existing usages of `Dropdown` and `Listselect` have been tested for possible breakage; but it is possible that some unforeseen regression might have been missed in the process. Examples tested:
- Listselect: encyclopedia window, left pane
- `kTextual` dropdown: mapselect window
- `kTextualNarrow` dropdown: singleplayer start conditions
- `kPictorial` dropdown: singleplayer tribe / type
- `kPictorialMenu` dropdown: editor/game toolbar
- `kTextualMenu` dropdown: main menu buttons

----

#### Allow both pic and checkmark in entries

It is now possible for list entries of `ListselectLayout::kShowCheck` to have an associated picture.
This was previously not possible because the `pic` in `EntryRecord` was being used to display the green checkmark.

#### `ListselectLayout` expanded from 3 to 6 types

Three types of standalone listselects (not belonging to a dropdown):
- (existing) `ListselectLayout::kPlain`: selected entry is highlighted
- (existing) `ListselectLayout::kShowCheck`: additionally displays green checkmark next to selected entry
- (new) `ListselectLayout::kMultiCheck`: checkmark for each list element can be toggled on/off independently. Same keyboard / mousewheel navigation controls to change highlighted element; enter key / mouse click toggles checkmark.

listselect-types

All three formats are also available in a version for use with dropdowns. When part of a dropdown, list entries are highlighted on mouse hover. Mouse click or enter key picks the dropdown's selected item (or toggles checkmark state in the multiple-select case).

`ListselectLayout::kDropdown` is preserved as a preset for backwards compatibility, but no such shorthand has been implemented for the two new types -- there is no need to configure those directly; such types of listselect are created and set up via the corresponding dropdown types that will utilise them.

[...]

PR #6534, opened 2024-09-10
Decomission Listselect<Entry&>

cake-pie:
**Type of change**
Refactoring

**Issue(s) closed**
`Listselect<Entry&>` was created for backwards compatibility in 9527204f, in which some users of Listselect that used references as entries began to be migrated to use pointers instead.
https://github.com/widelands/widelands/blob/abc27599111b9834a1571ef7482c12edbbb3f74d/src/ui_basic/listselect.h#L242-L249
It has been 17 years since then and `GameObjectivesMenu` is the only place in the codebase still using `Listselect<Entry&>`.

**New behavior / How it works**

Change `GameObjectivesMenu::objective_list_` from `Listselect<const Objective&>` to `Listselect<const Objective*>`. This is what the template specialization was doing under the hood anyway. Note that `Map::objectives_` has ownership of objective objects (it is a `std::map<std::string, std::unique_ptr<Objective>>`); we are handling a non-owning 'dumb' pointer obtained using `std::unique_ptr::get()`

`Listselect<Entry&>` template specialization deleted.

`static_assert` added to "generic " `Listselect<Entry>` prohibiting reference type.

**Possible regressions**
Stuff might break if some other future change alters the lifetime of objectives and leaves us with a dangling pointer.

**Additional context**
This would allow culling some code from #6525

### DIFF
--- a/src/ui_basic/listselect.h
+++ b/src/ui_basic/listselect.h
@@ -198,6 +198,14 @@ private:
 };
 
 template <typename Entry> struct Listselect : public BaseListselect {
+	/**
+	 * Listselect<Entry&> is no longer permitted. Allowing references as
+	 * template parameter is not a good idea (e.g. STL containers don't
+	 * allow it). You should use pointers instead because they are more
+	 * explicit, and that was how Listselect<Entry&> worked internally.
+	 */
+	static_assert(!std::is_reference<Entry>::value, "Listselect does not accept reference types!");
+
 	Listselect(Panel* parent,
 	           const std::string& name,
 	           int32_t x,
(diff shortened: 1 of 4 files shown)


cake-pie:
[Test failure](https://github.com/widelands/widelands/actions/runs/10785528803/job/29911484088?pr=6534#step:7:548) looks unrelated.

bunnybot (commented):
[Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4895#issuecomment-2302420)*
*On Tue Sep 10 08:52:58 CEST 2024, [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese) **approved** this pull request:*

LGTM :)

PR #4648, opened 2021-01-30
3849 win conditions fixes

hessenfarmer:
fixes #3849 
fixes #3190
fixes #2121
reduces the effect of #1037

supersedes #3742

Issue #3849, opened 2020-04-14
Player can be resurrected and has no fog of war.


stonerl:
**Describe the bug**
We had a very interesting bug. A player got defeated because his last warehouse got destroyed and then relegated to spectator. The fog of war was lifted, and he could see everything. He had a few minutes after he got defeated a warehouse that was under construction was finished and the player got back into the match.

This was kind of unexpected. Also, the resurrected player had no fog of war anymore and had an advantage over the others.

Edit:

The player also can look into others buildings but cannot attack.

Noordfrees:
What exactly does "got back into the match" mean – was the player able to build buildings and stuff? What did the end-of-game table say about this player's state?

stonerl:
> was the player able to build buildings and stuff?
Yes.
He got the "You got defeated" message but after the warehouse was finished he got back into the game. 

I'll upload the replay as soon as the game is finished.

hessenfarmer:
This has been reported some times before. Personally I didn't know how to fix that, until #3742. Discussion in forums was unclear whether to fix it at all, cause if you manage to get a constructionsite with all wares to build a warehouse why should you be defeated. 
So for me there are 2 solutions: 
a) just destroy all port and warehouse constructionsites, as well as expedition ships once the last existing warehouse/port has been destroyed. This would be my preferred solution. as it would be consistent.
b) allow ressurection by having a full constructionsite or an expedition ship left. In this case we need to check for this possibilities before lifting the fog of war so a real resurrection can take place.     
  
I believe a) is easier and more consistent, but offers less options. Should be discussed in the forums imho. I'll try to find some old thread for this.

stonerl:
Here is the replay. The player in question is the green one.

[Archive.zip](https://github.com/widelands/widelands/files/4477761/Archive.zip)

PR #6537, opened 2024-09-12
ComputerPlayer: neutering AI for eliminated players

### LINKED ISSUE #1037: AI player continues building activity when defeated
An AI player was defeated but continues activity to build new houses.
Savegame attached, it is the red player to the North of the big island.
Related bug #1644858.

 * date created: 2016-11-25T14:54:23Z
 * owner: 7010622-q-deactivatedaccount
 * the launchpad url was https://bugs.launchpad.net/bugs/1644857

cake-pie:
## Part 1
(commits [5d9f4d93 thru e432e8f5](https://github.com/widelands/widelands/compare/master...e432e8f5))

**Type of change**
Bugfix / Enhancement

**Issue(s) closed**
Fixes #1037

**New behavior**
Result reporting for players has separate category for elimination (a loss in which the player is unable / not allowed to continue playing), differentiated from other losses where player can carry on if "continue playing" option is selected, such as loss declared by wincondition after time limit elapsed (e.g. wood gnome).

AI stops `think`ing if its player has been eliminated from the game.

**How it works**

Result reporting changes:
- new `kEliminated` entry in `PlayerEndResult` enum
- by contrast, `kLost` is for when player can carry on playing -- AI should not deactivate in these cases
- `kWon` can also keep playing; `kResigned` may not
- scripting docu updated, win conditions scripts checked and adjusted as needed

AI changes:
- `think()` in `ComputerPlayer` and its derived classes renamed to `do_think()`
- new `ComputerPlayer::think()` invokes `think_impl_`, a pointer to member function which is initialized to `do_think()`
- `ComputerPlayer::set_thinking()` changes whether `think_impl_` points at `do_think()` or empty function `do_no_think()`

During simulation:
- (in lua) when win condition checks detect player has been eliminated, `wl.game.report_result` is called
- this is routed to `report_result()` in respective `GameController`s; if the player has been eliminated, `ComputerPlayer::set_thinking()` is called to tell the AI that the game is over for it.

Loading from save:
- check if player has an associated `PlayerEndStatus` that contains "eliminated" or "resigned" result; inactivate the AI if so.

Other win conditions tweaks:
- This is subject to the frequency of `check_player_defeated()` tests in win condition lua scripts. Depending on timing, AI can perform actions within the brief duration after destruction of last warehouse has occurred but before getting neutered by `check_player_defeated()`
  - Most win conditions run the check every 1000ms, but couple were per 5000ms -- those have been reduced to 1000ms as well, for consistency, and reducing the window of opportunity in which AI can act.
  - polling frequency could be increased further, but not sure is worth the trade off. Event-driven approach ultimately needed to fully eliminate delay between warehouse destroyed -> player loss detected.
- In `check_player_defeated()`, any remaining warehouses, milsites and constructionsites of the defeated player are being deleted (incl. to prevent #3849) but there was a 400ms sleep added in between each building, with the rationale of the delay "to create drama". This has been removed, as it can add up to several seconds when there are many buildings to be deleted, and causes substantial delay to result reporting (and thus AI deactivation) -- this allows fresh construction sites to get placed in the meantime, that will fail to be removed.

- In #1037 it was suggested to replace the AI with EmptyAI upon defeat. I've opted against this for future-proofing reasons -- it is not inconceivable that some new mechanics or game mode could be introduced in future that might give a defeated player a second chance or otherwise resurrect them, in which case the chosen approach would make it trivial to reactivate the same AI (that has remained in place) vs needing to figure out what kind of AI to (re)instantiate.
- Rather than having separate `think_impl_`, just make `think` the pointer -- but this would a) expose it to being modified from elsewhere, and b) be quite gnarly to call from outside the class -- hence opted to keep `think()` as somewhat like an accessor so that its users can remain unchanged.
- Test `Player::is_defeated‎()` as the criteria. This is unsuitable in several ways. First, the function is not fit for purpose, as the precise definition of defeat has changed since it was implemented. (See Part 2 below for more.) It also doesn't work for the intended purpose when starting a new game, because at time = 0 no buildings have been populated in yet. Furthermore, having the AI code itself make the determination that it has lost and should cease thinking/acting is another downside; determination of victory/defeat should be sole purview of win condition scripts.

**Additional context**
- Has no effect in "Endless Game" mode, which by design doesn't have any definition of defeat -- and thus does not have a main loop that calls `check_player_defeated()` -- see #4648

---

## Part 2
(commits [0c857eeb onwards](https://github.com/widelands/widelands/compare/e432e8f5...5af081ee))

**Type of change**
Refactoring

**Issue(s) closed**
Original intent was cleanup, refactoring out parts of DefaultAI code that contained "have I been defeated?" checks, now rendered redundant. This snowballed into several related changes, all involving `Player::is_defeated‎()`.

`Player::is_defeated‎()` has a misleading name; its behavior only checks if player has zero warehouses -- in the past this alone had been necessary and sufficient conditions for declaring player defeat, but it is no longer correct. Since the addition of expeditions (#3692), the win condition lua scripts have included additional logic to correctly handle shipboard starts, which don't have any warehouses initially. And as of #4648, endless games have no defeat conditions. Hence, its behavior has diverged from the actual definition of defeat for certain start + win condition combinations. Continued existence of this inaccurately-named method poses a risk of incorrect code being written, taking the function name at face value -- in fact, this has aleady happened.

**New behavior**
Function no longer exists.

**How it works**
Changed uses of the function on a case-by-case basis, see details below.

Retain the function but rename appropriately it, perhaps `Player::has_no_warehouse()`. This was what I did at first, but after working through the various uses, almost all of them didn't need it anymore -- the only user left was win conditions scripting, via the similarly misnamed `wl.game.Player.defeated` -- that too had to be addressed (by deprecation) and alternative means have been implemented to replace it.

### 2.1 Automatically invalidate pending diplomatic actions involving eliminated players 

**Type of change**
Refactoring / Enhancement

**Issue(s) closed**
If the computer player has already been defeated (faulty `Player::is_defeated‎()` test), `DefaultAI` performs cleanup of pending diplomatic actions by "manually" rejecting any outstanding team invites or join requests from other players to it. Since AI now gets switched off upon elimination, this needs to happen elsewhere.

No such cleanup previously existed in the case of human players; it should be implemented as well.

**New behavior**
Whenever a player (regardless of human/AI) resigns or is eliminated, any outstanding team invite or join request to/from them is automatically rendered invalid. This is not done if the player has lost but is not eliminated, since they can continue playing.

**How it works**
Per part 1 above, result reporting now differentiates between eliminated players vs those who can continue playing. `report_result()` in various `GameController`s will, in the case of player resignation / elimination, call `Game::invalidate_pending_diplomacy_actions()`, which takes the player number of that player and delete any entries that involves them.

Removed stale code from `DefaultAI`.

### 2.2 Fix faulty check for "other player is defeated" in AI logic

**Type of change**
Bugfix

**Issue(s) closed**
`DefaultAI` keeps track of some stats and a "diplo score" for the other players. If another player is "defeated" (faulty `Player::is_defeated‎()` test), the stats get zeroed and the diplo score set to -20.

**New behavior / How it works**
Stats reset occurs for resigned / eliminated players, as determined by player end status (if any) fetched from `PlayersManager`.

### 2.3 Other stale code in `DefaultAI`

**Type of change**
Refactoring

**Issue(s) closed**
If AI is defeated (faulty `Player::is_defeated‎()` test):  
Some code in `DefaultAI` was gated behind such checks so it wouldn't be run  
Stats and diplo score for all other players reset

**New behavior / How it works**
Stale code removed.

### 2.4 Fix `ParticipantList`

**Type of change**
Bugfix / Enhancement

**Issue(s) closed**
In code added in #4344, related to chat panel, `ParticipantList::get_participant_defeated()` calls `Player::is_defeated‎()`, taking it at face value. This function isn't actually utilised anywhere, but presumably it could be used to add some color or other indicator for chat participants who are already defeated in the game. This would not work correctly due to deviation of `Player::is_defeated‎()` e.g. for players that have picked Discover or New World starts before completion of their first warehouse.

**New behavior**
Checks for presence of reported PlayerEndStatus, obtains result from that. `ParticipantList::get_participant_defeated()` returns true if player has had a reported result of lost, resigned, or eliminated.

New functions added: `ParticipantList::get_participant_end_result()` and `ParticipantList::get_participant_victorious()`

**How it works**
`ParticipantList::get_participant_end_result()` fetches player end status (if any) from `PlayersManager`; the other two methods use it to check the result.

### 2.5 Deprecated `wl.game.Player.defeated`

**Type of change**
Refactoring / Enhancement

**Issue(s) closed**
For lua scripting, `wl.game.Player.defeated` wraps `Player::is_defeated‎()`. It is used as _input_ to win condition scripts, factoring into the declaration of player elimination, but like `Player::is_defeated‎()`, its name could lead to it being misinterpreted as an indicator of whether the player has in fact already been defeated.

**New behavior / How it works**
* New `wl.game.Player.lost` attribute added that actually checks if the player has already been declared to have lost (lost or eliminated); works by retrieving player end status (similar to pre-existing `wl.game.Player.resigned`)
* New methods `wl.game.Player.get_buildings_of_type` and `wl.game.Player.get_constructionsites_of_type` added
* `wl.game.Player.defeated` deprecated, documentation suggests `wl.game.Player.lost` or `wl.game.Player.get_buildings_of_type` instead
* `wl.game.Player.defeated` internally uses `wl.game.Player.get_buildings_of_type("warehouse")` instead of `Player::is_defeated‎()` ([validated](https://github.com/widelands/widelands/actions/runs/10927433308) by testsuite at commit 4530e497)
* win conditions scripts refactored to use `wl.game.Player.get_buildings_of_type("warehouse")` instead of `wl.game.Player.defeated`

**Possible regressions**
scripting addons

### 2.6 Minor driveby refactorings

**Type of change**
Refactoring

**Issue(s) closed**
a0d94202 Add protection against null pointers in `LuaPlayer::get_end_result()` (`wl.game.Player.end_result`).
921cd262 Tidy up ordering in `src/scripting/lua_game.h`


cake-pie:
> "continue playing" option

point taken, will look into it

kmk142789 (approved):

PR #6538, opened 2024-09-12
macos and other CI improvements

bunnybot:
[tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4899)*
*Created on Thu Sep 12 22:17:16 CEST 2024 by [**Tóth András (tothxa)**](https://codeberg.org/tothxa)*

---

**Type of change**
Bugfix / Enhancement

**Issue(s) closed**
 - warnings issued by macos builds for already installed packages
 - try to work around sporadic "resource busy" errors in macos builds
 - move build dep lists from `install_dependencies.sh` to separate files for CI platforms to only run builds for the affected OSes when dependencies change

**New behaviour**
`ccache`, `doxygen` and `graphviz` dependencies are removed.

`doxygen` and `graphviz` are not used by the workflow, but some users may want to build the documentation locally. However, other platforms don't install them either. `graphviz` pulled in lots of additional dependencies too.

`ccache` was supposedly used by the workflow build, but the cache wasn't actually stored, so I doubt it had any advantage. OTOH, `homebrew` kept telling us in the log that they are opposed to it, so will not use it.

**Possible regressions**
CI, dependency installation


hessenfarmer (approved):
changes look reasonable and the ci was green 
so porobably we should have this.

Issue #6557, opened 2024-10-31
several crashes with v1.2.1 rc1


frankystone:
We played a multiplayer game with version 1.2.1-rc1. 

Map: Fellowships
Players: Teayo (release build, windows i guess), hessenfarmer(release build, windows), kaputtnik, host  (debug build, linux)
Win condition: Territorial Lord, 3 hrs
Naval warefare: enabled

After some time the following occurred:
Teayo wasn't able to do create a warship because his ship window didn't provide the "Refit to Warship button". kaputtnik had this button, but as soon he clicked on it the game crashed.
After reloading the save game, we found that the "Configure game" menu lacked the entry "Enable Naval Warefare" below the "Show more options" dropdown.
Starting this save game works fine and Teayo suddenly had the "Refit to warship" button in his ship window and kaputtnik could press this button in his ship window without a crash.

After that from time to time the client hessenfarmer crashed with error "Unreachable code was reached", but after reloading the emergency save game we could continue the game.
After several crashes of this type we decided to end the game.

**Replays**
[2024-10-31T2.zip](https://github.com/user-attachments/files/17594009/2024-10-31T2.zip)

**Savegames**
[wl_autosave_nethost_0.zip](https://github.com/user-attachments/files/17594014/wl_autosave_nethost_0.zip)

**Version:**
 - OS: div
 - Widelands Version: v 1.2.1-rc1
 - Enabled Add-Ons: "Where are my heroes" but i didn't used it (probably because i had no heroes :smile:  )

bunnybot:
Mirrored on Codeberg as [#CB4917](https://codeberg.org/wl/widelands/issues/4917).

hessenfarmer:
the actual error message was this

ERROR:   FATAL EXCEPTION: [D:/a/widelands/widelands/src/logic/map.cc:2191] [D:/a/widelands/widelands/src/logic/map.cc:2191] Unreachable code was reached.
[stdout.txt](https://github.com/user-attachments/files/17594103/stdout.txt)

frankystone:
Ah, tomorrow i try to remember to enable the "write syncstreams" option (if i host again). or should this be enabled by all players?

frankystone:
Hm, i can't find the option "Write Syncstreams" anymore. Was it removed?

bunnybot:
[...]

PR #6563, opened 2024-11-02
Preserve debug symbols in windows debug builds

bunnybot:
[tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4923)*
*Created on Sat Nov 02 21:42:48 CET 2024 by [**Tóth András (tothxa)**](https://codeberg.org/tothxa)*

---

**Type of change**
Bugfix / New feature

**Issue(s) closed**
Missing debug symbols in build artifacts (re https://codeberg.org/wl/widelands/issues/4917#issuecomment-2421298)

**New behavior**
 - exports the executable (and debug symbols where appropriate) as separate artifact too to allow smaller downloads
 - do not fail all builds on first failure (re https://codeberg.org/wl/widelands/pulls/4910/files#issuecomment-2403794)
 - add manually triggered workflow to clean up obsolete release artifacts

**Possible regressions**
Exhausting artifacts quota? Otherwise not much.

**Additional context**
`gdb` cannot use the `.pdb` symbol file from `msvc`, but I gather that there should be freeware debuggers that can.

I was able to attach `gdb` even on linux with wine to the mingw build:
 - on debian or ubuntu the `gdb-mingw-w64` and `gdb-mingw-w64-target` packages are needed
 - run `wine /usr/share/win32/gdbserver.exe localhost:1111 widelands.exe`
 - in another terminal window or tab, run `x86_64-w64-mingw32-gdb widelands.exe`
 - set up in gdb: `target extended-remote localhost:1111`
 - after this, normal `gdb` commands work

I was able to set breakpoints, but the filenames were somehow wrong... `map.cc` was reported not having line 2191, and `get_neighbour()` was shown as being in `pathfield.h` instead of `map.cc`. After figuring this out, the line numbers of `map.cc` worked when applied to `pathfield.h`... (unfortunately I can't reproduce the `get_neighbour()` bug from #6557 with this branch based on current master with the 1.2.1-rc1 savegame from the bug report)


.github/workflows/build_windows.yaml
```diff
@@ -39,7 +40,14 @@ jobs:
...
         if ((Select-String -Quiet REVDETECT-BROKEN VERSION)) { exit 1 }
         cmake.exe --build .
-        strip -sv ./src/widelands.exe
+        if ("${{ matrix.config }}" -Match "Release") {
+          strip -sv ./src/widelands.exe
+        }
+    - name: Uploading executable
+      uses: actions/upload-artifact@v4
+      with:
+        name: Widelands_exe_${{ matrix.config }}_${{ matrix.arch }}_MinGW
```
bunnybot: [Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4923#issuecomment-2422383)*
*On Sat Nov 02 22:36:59 CET 2024, [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese) wrote:*

---

Please give this a human-readable name, e.g.
```
        name: Widelands ${{ matrix.config }} ${{ matrix.arch }} Binary Only (MinGW)
```
and the same for MSVC

bunnybot (commented):
[Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4923#issuecomment-2422384)*
*On Sat Nov 02 22:36:59 CET 2024, [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese) commented.*

.github/workflows/build_windows.yaml
bunnybot: [tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4923#issuecomment-2422654)*
*On Sun Nov 03 01:58:37 CET 2024, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) wrote:*

---

What's not human readable here? As these are actually the filenames for the downloads, I'd much rather convert all the others to underscores instead of spaces too, and drop all parentheses... :( Even "exe" is more than just a (very convenient) shorthand here...

hessenfarmer:
as debug builds are mainly designed for debugging I would appreciate, to remove the debug builds from the pre releases, and I believe we do not need to create installers for debug builds with Inno setup in this case. 
rest LGTM

hessenfarmer:
> [tothxa](https://codeberg.org/tothxa)_[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4923#issuecomment-2498917)_ _On Fri Dec 06 13:48:43 CET 2024, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) wrote:_
> 
> > as debug builds are mainly designed for debugging I would appreciate, to remove the debug builds from the pre releases
> 
> They allow debugging replicable bugs, so they should be very useful.      
     
What I meant was the same as down we do not need to create installers imho for debug. The exe (with symbols will be fine.) 
However for effective debugging probably selfcompiled buods are still preferable.   

> 
> > and I believe we do not need to create installers for debug builds with Inno setup in this case.
> 
> Agreed.
> 
> Since the exe can be simply replaced in an install, would it be OK for you to only create installers from the mingw release builds, and export only the exe from msvc builds, both debug and release (and of course the mingw debug builds)?

it would be ok for me. However I am not using MSVC and therefore I migth not be the best person tom decide on this. So to keep being on the safe side we just could do an installer for the 2 release builds (Mingw and MSVC)

> 
> I'm also thinking about exporting a zip of the datadir for pre-releases. That could be used with any executable on any platform without installation.     

as an addition this might be ok, personally I do use the Datadir in the repository with my self compiled executables anyway.

hessenfarmer (approved):
LGTM.

However I would like to hear the opinion of @CB/Nordfriese on this

hessenfarmer:
@bunnybot update merge

hessenfarmer:
@bunnybot merge

PR #6565, opened 2024-11-05
Mingw dependencies pin architecture

hessenfarmer:
**Type of change**
Enhancement

currently only Packages could be fixed by version, while dependencies fixed versions were ignored.

**New behavior**
now we can take care of fixed versions of dependencies as well.

additionally we can limit some packages only to one architecture.


bunnybot (commented):
[tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4925#issuecomment-2432253)*
*On Thu Nov 07 21:45:15 CET 2024, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) **approved** this pull request:*

Looks OK and worked in the CI.

PR #6570, opened 2024-11-11
Productionsite input queue compatibility with notification

bunnybot:
[tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4930)*
*Created on Mon Nov 11 02:06:22 CET 2024 by [**Tóth András (tothxa)**](https://codeberg.org/tothxa)*

---

**Type of change**
New feature

**Issue(s) closed**
Continuation of #6531

**New behavior**
Updates the input queues of production sites on game loading if the building definition changed between the saving and loading versions. Sends an in-game message to the player for each modified building.

**Possible regressions**
Everything related to in-game messages and message previews in the info panel.
Possibly desyncs in multiplayer?

**Additional context**
This was harder than I expected, but here it is. :) ~I haven't tested it in multiplayer yet.~ *Passed minimal multiplayer test.* ~The message still needs formatting, and probably more detailed info too, at least about lost wares.~

I would have opened this against @cake-pie 's PR branch, but
 - I'm not sure how well bunnybot can handle a PR against a mirror branch
 - if I opened it on github, it wouldn't be visible in the project repo
 - this is based on the `psiq-compat-d` branch, not the PR branch


src/map_io/map_buildingdata_packet.cc
```diff
@@ -819,17 +820,170 @@ void MapBuildingdataPacket::read_productionsite(ProductionSite& productionsite,
...
+				// Probably not worth adding graphic/text_layout as a dependency. It would
+				// require specifying the font styles too, but we already get that through
+				// Building::send_message().
+				// TODO(tothxa): The only problem is the hard-coded vspace gap.
+				static const std::string list_heading("<vspace gap=8><p>%s</p>");
+				static const std::string list_entry("<p> • %s</p>");
+
+				std::string body("<p>");
+				/** TRANSLATORS: %s is a building name */
+				body += format(_("The inputs of %s have changed. Please review the current "
```
hessenfarmer: this won't work in all languages due to e.g. the article, sex , etc. of the buildings name,  
How about 
```suggestion
				body += format(_("The inputs of the building(s): "%s" have changed. Please review the current "
```

src/map_io/map_players_messages_packet.cc
```diff
@@ -119,27 +127,23 @@ void MapPlayersMessagesPacket::read(FileSystem& fs,
 						assert(mol.is_object_loaded(mo));
 						serial = mo.serial();
 					}
-					// Compatibility code needed for map loading.
```
hessenfarmer, line 122: we normally try to keep map compatibility for the longest time possible. so unless it is proven that a map from build 18 won't load anyway we should keep this.

src/wui/game_message_menu.cc
```diff
@@ -487,14 +487,7 @@ void GameMessageMenu::filter_messages(Widelands::Message::Type const msgtype) {
...
 
-	case Widelands::Message::Type::kNoMessages:
-	case Widelands::Message::Type::kAllMessages:
-	case Widelands::Message::Type::kGameLogic:
-	case Widelands::Message::Type::kEconomySiteOccupied:
-	case Widelands::Message::Type::kWarfareSiteDefeated:
-	case Widelands::Message::Type::kWarfareSiteLost:
-	case Widelands::Message::Type::kWarfareUnderAttack:
-	case Widelands::Message::Type::kTradeOfferReceived:
+	default:
```
hessenfarmer: I believe the case list was intended to ensure that only our defined messge types are processed. Other types would throw an error by falling through to default, which is defined as unreachabele code.  So why to skip this safeguarded list. (

hessenfarmer (changes requested):
some inliners.
IMHO putting all the information about the changed wares in the mesages is overkill.  
the important message is from my point of view Something at building y has changed, see and check.   
the rest could simply be logged in verbose mode as it is not relevant for the gameplay but for debugging only .

src/logic/message.h
```diff
@@ -37,11 +37,14 @@ struct Message {
 		kSeafaring,
 		kEconomy,              // economy
 		kEconomySiteOccupied,  // economy
-		kWarfare,              // everything starting from here is warfare
+		kWarfare,              // warfare messages
 		kWarfareSiteDefeated,
 		kWarfareSiteLost,
 		kWarfareUnderAttack,
+		kWarfareEnd = kWarfareUnderAttack,  // end of warfare messages
```
bunnybot, line 44: [hessenfarmer](https://codeberg.org/hessenfarmer)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4930#issuecomment-2485577)*
*On Sat Nov 30 20:31:50 CET 2024, [**Stephan Lutz (hessenfarmer)**](https://codeberg.org/hessenfarmer) wrote:*

---

do we really need this virtual value.

bunnybot (commented):
[hessenfarmer](https://codeberg.org/hessenfarmer)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4930#issuecomment-2485578)*
*On Sat Nov 30 20:31:50 CET 2024, [**Stephan Lutz (hessenfarmer)**](https://codeberg.org/hessenfarmer) commented.*

src/logic/message.h
bunnybot, line 44: [tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4930#issuecomment-2485729)*
*On Sat Nov 30 22:07:28 CET 2024, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) wrote:*

---

I believe it's generally good practice to use such virtual enum members for this kind of comparison. The problem here is though that this enum wasn't designed with extensibility in mind, even though it is stored by value in savegames, so it's not possible to rearrange them. Just look at the other check for the economy types...

hessenfarmer: I just thought that we could have checked <=kWarfareUnderAttack as previously done for the economy. 
However I am fine with this. At least we do not declare the (potential) trade messages as warfare messages anymore.

hessenfarmer:
> [tothxa](https://codeberg.org/tothxa)_[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4930#issuecomment-2485743)_ _On Sat Nov 30 22:19:21 CET 2024, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) wrote:_
> 
> > I just recently had the issue where we fell through to a Never_HERE() which made it much easier to find the bug,. So I really do believe this check together with the definition of NEVER_HERE() does make sense in case the message type gets corrupted.
> 
> In these 2 places we don't do anything with the value afterwards. The only effect is that we toggle some buttons in the message box window / select an image to represent the message type. The `default:` case doesn't hurt in either `switch` even if the value is corrupted.

It is not about whether any harm is done or not. It is about detecting corruption. IMHO NEVER_HERE() was defined for this case to safeguard a strucutured control flow, by detecting when "unreacahable" code is executed.  I'd deem this good coding practise and I am strictly against removing it.

hessenfarmer (approved):
LGTM now.
tested and working

hessenfarmer:
@bunnybot update merge

PR #6580, opened 2024-11-28
Refactor and speedup add-ons manager rebuild

bunnybot:
[Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4940)*
*Created on Thu Nov 28 17:37:52 CET 2024 by [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese)*

---

**Type of change**
Refactoring

**Issue(s) closed**
With more than 500 maps available for download, the add-ons manager was becoming very slow to respond when changing filtering or sorting criteria.

**How it works**
- Split the large `AddOnsCtrl::rebuild()` function into three smaller functions, one for each tab.
- After any change, rebuild only the tab(s) that have changed.
- Cache unchanged rows, which noticeably and significantly speeds up responsiveness.
  The many memory allocations of creating many new rows had by far the biggest performance impact (~95% of the total time).
  So now we only do the whole, costly work once when fetching the remotes, and then only recreate individual rows as needed.

**Possible regressions**
A tab not being updated when it should be; outdated rows not being flushed from the cache.


hessenfarmer (changes requested):
code looks reasonable.
I tested it as well and it is more performant. 
However switching to the browse or maps tab does download both the browse and the maps. So performance could be improved further to just fetch the tab tht is selcted from the server rather then both tabs

hessenfarmer (approved):

hessenfarmer:
@bunnybot merge

PR #6615, opened 2025-01-28
Refactor AI economy accounting

bunnybot:
[tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4975)*
*Created on Tue Jan 28 02:45:34 CET 2025 by [**Tóth András (tothxa)**](https://codeberg.org/tothxa)*

---

### Type of Change
Refactoring / Bugfix

### Issue(s) Closed
Ware economies weren't tracked by `EconomyObserver`s, so `DefaultAI::review_wares_targets()` didn't work. Resolves https://codeberg.org/wl/widelands/pulls/4781#issuecomment-2611018 too.

### New Behavior
 - Raw pointers in `EconomyObserver` were replaced by `OPtr`s and `Serial`s.
 - `check_economies()` does its job in one go, no need to try again in next run of `DefaultAI::think()`

### Possible Regressions
AI flags handling, scheduling of some tasks

### Additional context
One race still remains, because `NoteImmovable::Owneship` is published before the `Flag` gets its serial, so `new_flags` cannot be converted to use `OPtr`s.

-----

Maybe the whole flag accounting should be thrown out altogether, and only the index and serial of the last processed flag in `improve_roads()` should be stored in the `EconomyObserver`. It's not like these observers eliminate much of accessing the game state directly... — alternative PR with this is up at #6616


src/ai/defaultai.cc
```diff
@@ -3663,60 +3650,75 @@ bool DefaultAI::improve_roads(const Time& gametime) {
...
-		economies.push_back(economies.front());
-		economies.pop_front();
+	// TODO(Nordfriese): Someone should update the code since the big economy splitting for the
+	//                   ferries
+	// TODO(tothxa): it has to be done in create_shortcut_road(), then here we can probably
+	//               allow both kinds of economies
+	if (economies_.size() >= 2) {  // rotating economies
+		Widelands::Serial first_econ = economies_.front().economy_serial;
+		do {
+			economies_.push_back(economies_.front());
```
hessenfarmer: shouldn't we use emplace here as well?

hessenfarmer (commented):
one inline comment.
rest of the code looks reasonable.
will test tonight.

PR #6617, opened 2025-01-31
[emscripten] wasm build target

ololoken:
### Type of Change
New feature

Add option to build wasm module, using [emsdk](https://github.com/emscripten-core/emsdk)

### New Behavior
Allows to run widelands in browser. [Run demo.](https://turch.in/widelands/)

### How it Works
For `asio` used this port https://github.com/emscripten-ports/asio
Current demo is build with following command:
```sh
docker run --rm -v $(pwd):/src emscripten/emsdk:4.0.2 sh -c 'git config --global --add safe.directory /src && mkdir -p build && cd build && \
emcmake cmake ../ -DCMAKE_BUILD_PLATFORM=emscripten \
-DCMAKE_BUILD_TYPE=Release \
-DCMAKE_EXE_LINKER_FLAGS="-sMODULARIZE -sEXPORTED_RUNTIME_METHODS=callMain -sEXPORT_ALL -sEXPORT_NAME=widelands" \  
-DOPTION_BUILD_TESTS=OFF \
-DOPTION_BUILD_CODECHECK=OFF \
-DOPTION_BUILD_WEBSITE_TOOLS=OFF && emmake make'
```
For initial configuration of wasm module and data caching used this ems-runner https://github.com/ololoken/ems-runner-widelands

### Possible Regressions
In browser lan game is disabled

### Screenshots
image

### Additional context
First run requires game data caching in indexddb.
Build tested in major desktop browsers, what's going on mobile devices I don't know.
As pthread support in wasm env requires SharedArrayBuffer, which requires custom http headers:
```
'Cross-Origin-Embedder-Policy': 'require-corp';
'Cross-Origin-Opener-Policy': 'same-origin';
```
It's not possible to deploy build to `gh-pages`.

Shaders: In general I created copy existing shaders in `webgl` subfolder, now difference is only in header and precision declaration.


tothxa:
Wow, cool. :)

Of course it's slow, even the mouse movement lags on my laptop.

ololoken:
@tothxa runs pretty smooth on my m1

src/graphic/gl/initialize.cc
```diff
@@ -253,15 +253,17 @@ SDL_GLContext initialize(
 			}
 		}
 	};
-
 	const char* const opengl_version_string = reinterpret_cast<const char*>(glGetString(GL_VERSION));
+#if defined(__EMSCRIPTEN__)  // skip check webgl version anyway it's webgl2
```
bunnybot, line 255: [tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4977#issuecomment-2666900)*
*On Sat Feb 01 15:02:32 CET 2025, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) wrote:*

---

As far as I can tell, the `log_info()` line is the same, so instead of duplicating it, how about 2 `#ifndef` blocks (before and after)?

Same for the shading language version below.

data/shaders/webgl/blit.fp
```diff
@@ -0,0 +1,28 @@
+precision highp float;
```
bunnybot, line 5: [tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4977#issuecomment-2666911)*
*On Sat Feb 01 15:09:39 CET 2025, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) wrote:*

---

> Shaders: In general I created copy existing shaders in webgl subfolder, now difference is only in header and precision declaration.

But then why duplicate them?

I don't know anything about OpenGL, but a quick web search says `hihgp` is supposed to be fine for all but some older mobile devices, whose touch-only interface we don't support anyway.

OTOH it also says that the version header is mandatory... Why did you remove it, did it cause any problems?

bunnybot (commented):
[tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4977#issuecomment-2666914)*
*On Sat Feb 01 15:02:32 CET 2025, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) commented.*

data/shaders/webgl/blit.fp
ololoken: The problem is that GLSL spec says 
```
The #version directive must occur in a shader before anything else, except for comments and white space.
```
And it's not mandatory, if not specified version must be 110. 
So we can't simply change version using ifdefs.
I duplicated shaders as initially expected that will need to rewrite them, but it's rare case of luck when no significant modification were required.

ololoken:
> [tothxa](https://codeberg.org/tothxa)_[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4977#issuecomment-2668032)_ _On Sat Feb 01 23:48:53 CET 2025, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) wrote:_
> 
> > [**ololoken**](https://github.com/ololoken) wrote:
> > The problem is that GLSL spec says
> > ```
> > The #version directive must occur in a shader before anything else, except for comments and white space.
> > ```
> 
> It wouldn't be a problem if we could find a version that works everywhere... Your webgl shader programs (without the version directive) fail to load in my native build ("unexpected new identifier"). Adding `precision` in the originals also fails ("illegal use of precision"). But it looks like if I change the version to 130, then it works. Would that be OK in webgl too?

App uses webgl2 canvas context and WebGL 2 shader the only possible version line is `#version 300 es`. I will try check if it's possible to perform text based search and replace of version/precision options in javascript context.

ololoken:
@tothxa would check `#version 100`?

ololoken:
well, after long consideration(about 15 minutes) I think the best option is to keep two types of shaders for webgl and native. Other option is to override shader sources on the fly in js context, but requires manipulation with generated js for wasm module binding.

ololoken:
@tothxa look shaders sources logically are part of data and not bound with resulting binary. So I can do the magic when preparing data pack for wasm launcher. If you insist that we don't need standalone version of webgl shaders, so be it. But my point is that in future(I understand that it's weak assumption) shader's code can have more differences with native

CMakeLists.txt
```diff
@@ -117,34 +117,82 @@ endif()
...
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${USE_FLAGS}")
+  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${USE_FLAGS} \
+  -sASYNCIFY -sASYNCIFY_STACK_SIZE=32768 \
+  -sALLOW_MEMORY_GROWTH -sINITIAL_MEMORY=128mb -sMEMORY_GROWTH_LINEAR_STEP=32mb \
+  -sENVIRONMENT=web,worker \
+  -lidbfs.js \
+<<<<<<< HEAD
+  -lwebsocket.js \
+  -sPROXY_POSIX_SOCKETS \
+=======
```
bunnybot: [tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4977#issuecomment-4144055)*
*On Tue May 06 00:15:48 CEST 2025, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) wrote:*

---

You have merge errors... Maybe from force pushing? Please don't force push branches that belong to an open PR. It's bad enough on github or codeberg alone, and even worse for our mirror bot.

bunnybot (commented):
[tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4977#issuecomment-4144058)*
*On Tue May 06 00:15:48 CEST 2025, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) commented.*

bunnybot (commented):
[Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4977#issuecomment-5395491)*
*On Sun Jun 15 15:03:27 CEST 2025, [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese) **approved** this pull request:*

Very good :)
Code LGTM now.

I get one warning while building:
```
CMake Warning (dev) at /usr/share/cmake-3.31/Modules/ExternalProject/shared_internal_commands.cmake:1276 (message):
  The DOWNLOAD_EXTRACT_TIMESTAMP option was not given and policy CMP0135 is
  not set.  The policy's OLD behavior will be used.  When using a URL
  download, the timestamps of extracted files should preferably be that of
  the time of extraction, otherwise code that depends on the extracted
  contents might not be rebuilt if the URL changes.  The OLD behavior
  preserves the timestamps from the archive instead, but this is usually not
  what you want.  Update your project to the NEW behavior or specify the
  DOWNLOAD_EXTRACT_TIMESTAMP option with a value of true to avoid this
  robustness issue.
  /usr/share/cmake-3.31/Modules/ExternalProject.cmake:3041 (_ep_add_download_command)
  CMakeLists.txt:189 (ExternalProject_Add)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

Please set that policy and merge master, then this is good to go :)

How do you want to be credited in the developers list: nickname, realname, or both?

bunnybot (commented):
[Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4977#issuecomment-7646854)*
*On Thu Oct 09 13:08:02 CEST 2025, [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese) commented:*

I have reviewed this again and compiled it successfully locally. LGTM, thanks for this feature :)
One more comment inline, then this can be merged.

utils/build_deps.py
```diff
@@ -283,7 +283,8 @@ def main():
             owners_of_src[src] = lib
 
     if (report_unused_sources(srcdir, sources, owners_of_src)):
-        return 1
+        print('UNUSED SOURCES DETECTED! SEE REPORT ABOVE')
+        #return 1
```
bunnybot, line 287: [Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4977#issuecomment-7646851)*
*Created on Thu Oct 09 13:08:02 CEST 2025 by [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese)*

---

I think this should be reverted, this should be an error since it is too easy to overlook otherwise. In my testing Widelands compiled fine with this reverted. If this is absolutely necessary it should be an option used only with Wasm build enabled.

PR #6623, opened 2025-02-05
install-dependencies.sh: Don't error on MacOS when all dependencies already installed.

### LINKED ISSUE #6622: install-dependencies.sh errors when all dependencies are already installed on MacOS
### Describe the Bug

I've just notices that on MacOS, when all dependencies are already installed, install-dependencies.sh tries to install _nothing_, leading to an error and the script exiting with a non-zero code, which seems wrong for everything being fine. I attach a simple commit that fixes this and will be happy to open a PR.

https://github.com/sanj0/widelands/commit/c01ebf9fb78458635443eedaa5300ac089f77f17

### To Reproduce

1. Execute `install-dependencies.sh` on MacOS to install dependencies
2. Execute it again

### Expected Behavior

I would expect the script to not error in the case where everything in (already) fine.

### Crash Log

```text

```

### Screenshots

_No response_

### Replays

_No response_

### Savegames

_No response_

### Operating System

MacOS 13.1 / homebrew 4.4.20

### Widelands Version

current master

### Enabled Add-Ons

_No response_

### Additional Context

_No response_

speicherwerk:
### Type of Change
Bugfix: On MacOS, when all dependencies are already installed, install-dependencies.sh tries to install nothing, leading to an error and the script exiting with a non-zero code, which seems wrong for everything being fine. The fix is as simple as checking if `$PKGS` is empty.

### Issue(s) Closed
Fixes #6622

### To Reproduce
1. Run `install-dependencies.sh` on MacOS
2. Run it again

### New Behavior
The script doesn't error but instead calmly informs the user that all dependencies are already installed

### How it Works
Checks if `$PKGS` is empty and doesn't try to install nothing if it is.

### DIFF
--- a/install-dependencies.sh
+++ b/install-dependencies.sh
@@ -190,7 +190,12 @@ elif [ "$DISTRO" = "homebrew" ]; then
          echo "$PKG is already installed, skipping."
       fi
    done
-   brew install $@ $PKGS
+   if [ -z "$PKGS" ]; then
+       echo
+       echo "All dependencies already installed! Nothing to do ..."
+   else
+       brew install $@ $PKGS
+   fi
 
 elif [ "$DISTRO" = "solus" ]; then
    echo "Installing dependencies for Solus..."
--- a/install-dependencies.sh
+++ b/install-dependencies.sh
@@ -192,7 +192,7 @@ elif [ "$DISTRO" = "homebrew" ]; then
    done
    if [ -z "$PKGS" ]; then
        echo
-       echo "All dependencies already installed! Nothing to do ..."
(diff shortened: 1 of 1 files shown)


bunnybot (commented):
[tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4983#issuecomment-2683786)*
*On Wed Feb 05 15:52:40 CET 2025, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) **approved** this pull request.*

PR #6648, opened 2025-02-26
Update developers

hessenfarmer:
just added @frosch123 to developers list.


hessenfarmer:
@cb/Nordfriese I believe we do not need a milestone here. correct?

frosch123 (approved):

PR #6650, opened 2025-02-27
Fix double shadow of empire mill

bunnybot:
[tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/5010)*
*Created on Fri Feb 28 00:42:21 CET 2025 by [**Tóth András (tothxa)**](https://codeberg.org/tothxa)*

---

### Type of Change
Bugfix

### Issue(s) Closed
Fixes the double shadow of the Empire windmill

### To Reproduce
1. Play with Empire
2. Build a windmill
3. See error

### Possible Regressions
none?

### Screenshots
before:

after:

### Additional context
I figured this out finally after I noticed the different shadows of the market on the structure itelf and on the background. Took me a while to understand how the different lights work.

The main take-away is that if you use the template from the media repo, and the 2 shadow casting lights give you noticeable glitches in the shadows of your building, then constrain the `MainLight` (not its control!) to copy all transformations of the `ShadowGenerator` (again, the light itself). (doing it the other way would give you long sunset/winter shadows)

The double shadows of this one were a slightly different problem with the limitations(?) of the old blender version this model was made in, but given the experience from the market investigation I was able to fix it in blender 2.79. It needed a geometry fix first though...


hessenfarmer (approved):
LGTM.
@bunnybot merge

PR #6681, opened 2025-04-05
Set buildcosts for all tribes’ markets

bunnybot:
[Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/5041)*
*Created on Sat Apr 05 10:23:41 CEST 2025 by [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese)*

---

### Type of Change
Balancing

### Issue(s) Closed
Re https://codeberg.org/wl/widelands/pulls/5036#issuecomment-3491825

Adapted buildcosts for all markets.

These are based on the port buildcosts, with gold and iron (and cloth for those who don't have it on all maps) removed and a few other wares added instead.

This is open to discussion. The branch can be pushed to directly.


hessenfarmer:
@cb/Nordfriese    
I did not have a look yet. However if you and @cb/tothxa agree on it I would be fine.     
If you want my opinion as well I'll try to have look into it next week.

hessenfarmer:
The amounts seem to be balanced.
Only thing probably worth discussing is that amazons and atlanteans do use mined materials. while the other tribes do not so this might be a slight disadvantage for amz and atl, especially in scenarios where marketing might solve shortfalls in the supply. So maybe we could exchange the mined materials of the 2 tribes with some other expensive wares.

bunnybot (commented):
[tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/5041#issuecomment-4045571)*
*On Tue Apr 29 23:39:03 CEST 2025, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) **approved** this pull request:*

I guess that means we're OK?

PR #6694, opened 2025-04-13
Check in tribe loading that buildings don't have more than 3 workareas

bunnybot:
[Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/5054)*
*Created on Sun Apr 13 21:55:29 CEST 2025 by [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese)*

---

### Type of Change
Bugpatch

### Issue(s) Closed
Fixes a bug that recently appeared in the Europeans add-on: When a building has more than 3 workareas, it will load fine but Widelands crashes when trying to show them.

### New Behavior
Fail to load a tribe with such a building.

An alternative would have been to permit this and instead patch the workarea display code to treat all extra workareas as ring 3, but this would be the hackier approach IMHO and make for poor UI.
Tweaking the workarea code to fully support arbitrary higher numbers of workareas would be the best way to fix this but a lot harder.


hessenfarmer (approved):
LGTM
not tested

hessenfarmer:
@bunnybot update

hessenfarmer (approved):
Trested and working
@bunnybot merge

PR #6734, opened 2025-04-28
test checks for correct window

SimonHeimberg:
### Type of Change
Bugfix of a lua test

### Issue(s) Closed
Fixes -

### Old Behavior
Before `windows['sw1_name']` was checked (always same window name), which does never exist.

### New Behavior
Now the variable is used to check the window, so `windows[sw1_name]` is checked as later.


bunnybot (commented):
[tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/5094#issuecomment-4144202)*
*On Tue May 06 00:42:16 CEST 2025, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) **approved** this pull request:*

Thanks for fixing my stupid mistake.

The failure is obviously unrelated, let's hope this run succeeds...

PR #6852, opened 2025-07-23
do not build installers on translation only changes in build.yaml

SimonHeimberg:
Translations are updated automatically frequently. Building installers for this changes is not really necessary.

### Type of Change
Enhancement of build system

### Issue(s) Closed
Fixes -

### New Behavior
Installers are not built if only translation files change. (They are changed by workflow .github/workflows/i18n.yaml)
This reduces load on build system.

This probably means that job "dev_release" copies all the old release binaries. Should it be skipped in this case?

### Possible Regressions
People downloading this installers _might_ miss current translations.

### Additional context
discussion started in https://github.com/widelands/widelands/pull/6314#issuecomment-3097260063
To be continued here.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -43,7 +43,9 @@ jobs:
               # not 'utils/download_asio.sh', called by install-dependencies.sh for some other arch
             - &installer
               - *src_cmake_dependencies
-              - 'data/**'
+              # data/ except of translations, because they change frequently
+              - 'data/!(i18n/translations)/**'
+              - 'data/*' # do not ignore files directly inside
             - &ubuntu_dependencies
               - *src_cmake_dependencies
               - 'utils/ubuntu/packages'


SimonHeimberg:
Start of discussion copied from #6314:

> > [tothxa](https://codeberg.org/tothxa)_[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/4677#issuecomment-5872264)_ _On Fri Jul 18 01:38:03 CEST 2025, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) wrote:_
> > Ideally we shouldn't rebuild w*ndows and mac binaries either for translation updates, but `datadirversion` is a problem... (though we could replace the translations and repackage with the old datadirversion for the sake of the executable, but that could lead to problems for identifying problematic ones)
> 
> I propose to optimise triggering installers in a separate pr. When would the installers be built? Only when the binary will change? Then we could remove triggering on `data/**` altogether. (By the way, we also trigger on cmake/_codecheck_.) I would not worry too much about (data/)datadirversion . A developer will learn quickly how to adapt it from output of `widelands --version`.
> 
> Edit: I see, skipping "data/i18n/translations" for installers would help because translations are updated automatically frequently. And most of the few people using a dev installer will not care about translation changes (getting delivered with the installer). What problem would be with datadirversion when no installer is built when only translation changes? Or maybe problems with the job dev_release?

SimonHeimberg:
The pr ~is~ _was_ marked as draft because we should discuss. It could be applied.

hessenfarmer:
I believe datatdirversion would not be a problem as this is generated during build and can't desync imho. However I am afraid we would have no installers in latest dev release most of the time as we do cleanup files that are from a different commit afaik.    
So if we want to go this way we need to change the prerelase process as well.

SimonHeimberg:
I understand there would be installers from the latest version because missing ones are copied from the previous version (by job "dev_release"):
https://github.com/widelands/widelands/blob/eab012b8a5c34b6eb56bc72c2b530a7bfbb1868e/.github/workflows/build.yaml#L148-L152

They would include the old translations. In "Original filenames", they would show the real version. See "Original filenames" on https://github.com/widelands/widelands/releases/tag/latest

It could be a possibility to skip the job "pre-release" if no installer did run. The old files would remain, as it is this job which deletes the release (with the files) in
https://github.com/widelands/widelands/blob/eab012b8a5c34b6eb56bc72c2b530a7bfbb1868e/.github/workflows/build.yaml#L200-L202

Agree, users of those builds would not have newest translations.

PR #6860, opened 2025-07-30
test: stable_save with lunit warns

SimonHeimberg:
### Type of Change
Enhancement of test suite

### Old Behavior
stable_save() of stable_save.lua fails when called from a lunit test method,
and it hangs in case it is called when saving is disabled.

### New Behavior
stable_save() still fails, but before it prints a warning message in the log.
When it is called when saving is disabled, an assertion is raised instead of hanging till timeout.

### Possible Regressions
There should be none as stable_save() is used in the tests as it is supposed to.

### Additional context
stable_save() and lunit.run() can be used together.
But save may not be called while lunit.run() runs. (Only before and after.) So save is (normally) disabled before lunit.run() is started. In case enabling saving would get forgotten, there is an assertion now. In case stable_save() would get called from a test method, a warning gets printed before the crash. (In case a_game:save() gets called directly, there still is no warning.)

The failure message when saving from a test method looks like:
```
PANIC: unprotected error in call to Lua API (attempt to persist a light C function (0x55f27b133160) (root.__coroutine_registry[1].stack[26]))
```


bunnybot (commented):
[Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/5220#issuecomment-6193297)*
*On Sat Aug 09 13:49:08 CEST 2025, [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese) commented:*

LGTM, one comment inline

test/scripting/stable_save.lua
```diff
@@ -9,6 +9,12 @@ function stable_save(game, savename, desired_speed)
    print("###### stable_save: last save time " .. last_save_time)
 
    game.desired_speed = 1000
+   if lunit and lunit.stats and lunit.stats.run and
+   -- lunit     is loaded and   started running  and ...
```
bunnybot: [Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/5220#issuecomment-6193291)*
*Created on Sat Aug 09 13:49:09 CEST 2025 by [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese)*

---

Breaking the condition across multiple lines with comment lines between makes it hard to read, IMHO. Please keep the condition together and place the comment below it.

```lua
   if lunit and lunit.stats and lunit.stats.run and
         lunit.stats.run + lunit.stats.notrun < lunit.stats.tests then 
         -- lunit is currently running
```

bunnybot (commented):
[Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/5220#issuecomment-6294742)*
*On Wed Aug 13 10:15:21 CEST 2025, [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese) **approved** this pull request.*

PR #6869, opened 2025-08-18
optimize utils/build_deps.py

SimonHeimberg:
### Type of Change
Enhancement of a util

### New Behavior
On my system, the time for the this util goes from 2.5 sec to around 0.4 sec.
This does not help a lot on the build system,
but it helps if you run the script locally before a commit.

### How it Works
Checking for `someText in longText` is faster than matching several regexps which all contain someText. So check this before.

### Possible Regressions
?

### Additional context
Other commits are minor code optimisations.


bunnybot (commented):
[Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/5228#issuecomment-6613756)*
*On Wed Aug 20 19:53:01 CEST 2025, [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese) **approved** this pull request:*

LGTM :)

PR #6870, opened 2025-08-18
optimize utils/find_unused_includes.py

SimonHeimberg:
### Type of Change
Enhancement of a util

### New Behavior
On my system, the time for the this util goes from 5 sec to around 1 sec.
This does not help a lot on the build system,
but it helps if you run the script locally before a commit.

### How it Works
find_unused_includes.py caches results of one function

### Possible Regressions
?

### Additional context
Other commits are enhancements of output


bunnybot (commented):
[Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/5229#issuecomment-6613714)*
*On Wed Aug 20 19:45:57 CEST 2025, [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese) commented:*

Nice! I run this script frequently during feature development to keep my includes sane, so the speedup is much appreciated :)

Regarding non-existent includes, IMO they should be treated like other unused includes, and not as an error in the python util. If an include can't be found by the script then chances are it's the user who is doing something wrong, not the util.

SimonHeimberg:
Having a traceback on a user error is not helpful, agree.

bunnybot (commented):
[Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/5229#issuecomment-6640219)*
*On Thu Aug 21 13:32:05 CEST 2025, [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese) **approved** this pull request:*

LGTM :)

PR #6875, opened 2025-08-21
Remove default quantity of 0 as useless code

bunnybot:
[MarkMcWire](https://codeberg.org/MarkMcWire)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/5234)*
*Created on Fri Aug 22 01:35:18 CEST 2025 by [**MarkMcWire**](https://codeberg.org/MarkMcWire)*

---

This code (74e7ef0009ff2052fe63f59c13c1f7d6b8092269) is useless. It prevents a target value set via the wl.Descriptions():modify_unit("ware", "blackroot_flour", "target_quantity", "atlanteans") from being applied. This script takes effect only if no default target value is specified in units.lua. Otherwise, the default value overrules the modify_unit-use, no matter what modifications wl.Descriptions():modify_unit you make.


hessenfarmer (changes requested):
it is not useless but was set on purpose.

https://github.com/widelands/widelands/pull/3855/files

so this "fix" is not approvable imho. 

it needs to be fixed in our c++ part. so better open an issue on this describing that modify unit is not working as intended.

MarkMcWire:
I dont understand how a default target of 0 for flour enables demant check for blackroot.

0 tells the game, that the economy don't need any blackroot flour at all.

The mill itself has no check for blackroot flour. It produces it endless, even if it is not needed.

So deleting this line would not make any difference in the behavior of the game.

A target value makes only sense, if the mill would check the need for output.

PR #6883, opened 2025-09-04
Remove commas in end lines, changing translation text, adding comment about missing ware check

bunnybot:
[MarkMcWire](https://codeberg.org/MarkMcWire)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/5241)*
*Created on Fri Sep 05 00:53:14 CEST 2025 by [**MarkMcWire**](https://codeberg.org/MarkMcWire)*

---

Only beauty changes on code, no functional changes. I've started with the Amazons and will rework the other tribes as well. The last line in a program should end without a comma. Notes should refer to the current program. Missing product checks should be included as a note to make it clear that this is not a code error but a deliberate decision.


hessenfarmer:
AFAIR this is the second time you included a change that was already declined in a very cluttered big change. 
Please explain why we should not consider this as a malicious attempt. 
Although it can‘t have a Security effect here, this is exactly the method used by criminals to infiltrate Open Source projects

Issue #2020, opened 2019-09-09
Add Rivers and Waterfalls 


bunnybot:
A suggestion to add more realistic river graphics on the map was made by Astuur in the Graphics Forum (https://wl.widelands.org/forum/post/4677/)

"We have neither mountains nor rivers in the map. ... It might be different and more complicated with the rivers, because they need to be fishable and may at some time in the future carry rowing boats, footbridges or even wider bridges. I am not sure that immovables is the best approach here.
Created as immovables, we may have some bolders above the waterline, that would show some water current as a first step.That should be easy. 
Again we need a different set for all the worlds, and all orientations.
A different idea is to create a set of banks and middle-of-the-stream tiles as terrain types, but I unsure whether WL can accomodate them in the current form."

Graphics to indicate water movement such as rapids or whirlpools might be added as immovables.

In the same vein, animated waterfalls might be added to the map to add visual interest.

------------------------------------
Imported from Launchpad using lp2gh.

 * date created: 2015-02-05T03:59:43Z
 * owner: chuckw20
 * the launchpad url was https://bugs.launchpad.net/bugs/1418305

bunnybot:
(by chuckw20)
kapputnik made an attempt to introduce river graphics into the map editor (see forum post https://wl.widelands.org/forum/post/12378/).

His effort identified the obstacles currently faced in 
1) creating the graphics, 
2) positioning them on a map and 
3) adapting them to terrain elevations on a map.

Ideally, creating and placing rivers by map makers should be as easy as "drawing" them with a tool like terrain.  

Perhaps if rivers could be plotted in the Editor like roads in the game this could be achieved.

bunnybot:
(by franku)
> Perhaps if rivers could be plotted in the Editor like roads in the game this could be achieved.

Yes, this comes in my mind too. A nice to have:

1. select the graphic; while moving the mouse a shadow of this graphic follows the mouse so you could imagine where it is placed; after click the real graphic is set
2. a function to flip the graphic clockwise. In that case there has to be e.g. only one graphic for rivercorners. A rightclick flips the corner clockwise about 90 degrees to get each direction you want the river flows.

About Waterfalls: I think this not easy to animate because of the direction the water has to fall. With the actual technic you have to make an animation for waterfalls in each possible cardinal direction.

bunnybot:
(by sirver)
Right now, Widelands only has two concepts: fields (nodes) and edges (roads). Where would these graphics be painted? 

[...]

PR #6900, opened 2025-10-19
Allow bridgeable water type that is both swimmable and walkable

### LINKED ISSUE #5853: Waterway marker indicating impossible directions
Under some circumstances the waterway markers are shown although no waterway is possible in those directions.

Here no waterway can be build at the given waterway markers. The one in the east is impossible at all:

no_waterway

**To reproduce**
Load attached save game.
[waterway_issue.zip](https://github.com/widelands/widelands/files/11235033/waterway_issue.zip)

**Expected behavior**
The markers should show only possible directions.

**Version:**
 - OS: archlinux
 - Widelands Version: 1.2~git26120 (972ae31@watchwindow) Debug

bunnybot:
[tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/5260)*
*Created on Sun Oct 19 12:29:52 CEST 2025 by [**Tóth András (tothxa)**](https://codeberg.org/tothxa)*

---

### Type of Change
New feature

### Issue(s) Closed
https://www.widelands.org/forum/post/43680/
Re: #2020
Fixes: #5853

### New Behavior
It is possible to create areas where both ships and workers can pass. This allows navigable rivers that don't separate the land, as well as straits between large water bodies where land crossing is possible without blocking ship traffic. (just 2 sides of the same coin actually :)

### How it Works
The new `is=bridgeable` terrain type is both `kWater` and new type `kBridgeable`, which allows building roads and becomes walkable where there is a road.

### Possible Regressions
Pathfinding, road building

### Screenshots
oops, should've created some... coming soon

### Additional context
~Currently flags can be placed anywhere on bridgeable water, and thus bridges can join and cross each other, because that happens IRL too, and because I liked it this way.~

~If the consensus is that they shouldn't be allowed, then we also have to decide whether to:~
 - ~use the waterway building logic for bridges too, which requires new field action buttons and options for all combinations of switchig between all 3 types, so I think it would clutter the interface a lot~
 - ~or just simply disallow flags in the middle of bridges and leave it to the players whether they want to place the next flags on the banks or farther away, so it's somewhat inconsistent IMO that bridges on navigable water would be special in one way, but blend so much with plain roads in another way~

Flags are not allowed on bridgeable water, but not required at the banks either.


Fr-Dae:
news ?

Fr-Dae:
I don't know if this will help, but in various games that use a map editor (e.g. Tiberium Sun),

bridges had three parts:

a start/end section and a central section.
So whether the bridge is small or large, you just need to add sections, and it looks good.

Fr-Dae:
max is 12 like a naval road ?

PR #6907, opened 2025-10-28
WIP: Add arm64 architecture to Windows build matrix

sebgod:
### Type of Change
Build for arm64

### DIFF
--- a/.github/workflows/build_windows_msvc.yaml
+++ b/.github/workflows/build_windows_msvc.yaml
@@ -10,6 +10,13 @@ jobs:
       matrix:
         config: [Debug, Release]
         arch: [x64, x86, arm64]
+        include:
+          - arch: x64
+            archset: x64
+          - arch: x86
+            archset: x86
+          - arch: arm64
+            archset: amd64_arm64 
     name: Windows ${{ matrix.config }} ${{ matrix.arch }} Build (MSVC)
     runs-on: windows-2025
     env:
--- a/.github/workflows/build_windows_msvc.yaml
+++ b/.github/workflows/build_windows_msvc.yaml
@@ -71,7 +71,7 @@ jobs:
     - name: Configure MSVC development console
       uses: ilammy/msvc-dev-cmd@v1
       with:
-        arch: ${{ matrix.arch }}
+        arch: ${{ matrix.archset }}
     - name: Compiler
       env:
         VCPKG_ROOT: C:\vcpkg
--- a/.github/workflows/build_windows_msvc.yaml
+++ b/.github/workflows/build_windows_msvc.yaml
@@ -9,7 +9,7 @@ jobs:
       fail-fast: false
       matrix:
         config: [Debug, Release]
-        arch: [x64, x86]
+        arch: [x64, x86, arm64]
     name: Windows ${{ matrix.config }} ${{ matrix.arch }} Build (MSVC)
     runs-on: windows-2022
     env:
--- a/.github/workflows/build_windows_msvc.yaml
+++ b/.github/workflows/build_windows_msvc.yaml
@@ -11,7 +11,7 @@ jobs:
         config: [Debug, Release]
         arch: [x64, x86, arm64]
     name: Windows ${{ matrix.config }} ${{ matrix.arch }} Build (MSVC)
-    runs-on: windows-2022
+    runs-on: windows-2025
     env:
       VCPKG_ROOT: C:\vcpkg
       VCPKG_TARGET_TRIPLET: ${{ matrix.arch }}-windows-static
--- a/.github/workflows/build_windows_msvc.yaml
+++ b/.github/workflows/build_windows_msvc.yaml
@@ -18,7 +18,7 @@ jobs:
           - arch: arm64
             archset: amd64_arm64 
     name: Windows ${{ matrix.config }} ${{ matrix.arch }} Build (MSVC)
-    runs-on: windows-2025
+    runs-on: windows-2022
     env:
       VCPKG_ROOT: C:\vcpkg
       VCPKG_TARGET_TRIPLET: ${{ matrix.arch }}-windows-static


hessenfarmer:
I believe this simple fix won't work, as we need some libraries and we would need arm64 versions of them too.   
AFAIK our MSVC builds do use libraries build with MINGW as well.       
Furthermore we would need to run this on an Win on Arm Image to run our tests, to sucessfully build this.     
@sebgod have you tested this change locally before? On which host image? 
@cb/tothxa any idea on this?

Edit: anyhow I will activate the workflow to give it a try.

hessenfarmer:
As it is failing in the CI ( as expected) I suggest to close the PR and open an issue with the request to provide an windows on ARM version.

sebgod:
> I believe this simple fix won't work, as we need some libraries and we would need arm64 versions of them too. AFAIK our MSVC builds do use libraries build with MINGW as well. Furthermore we would need to run this on an Win on Arm Image to run our tests, to sucessfully build this. @sebgod have you tested this change locally before? On which host image? @cb/tothxa any idea on this?
> 
> Edit: anyhow I will activate the workflow to give it a try.

Hey thanks for looking into this. 
I will try compiling it on my Surface Pro 11 with a Snapdragon X.
Out of the box I am stuck with vcpkg only supporting manifest mode, whereas this project is using classic mode. I might have to use a non-Visual Studio supplied version

hessenfarmer:
well it seems the action "ilammy/msvc-dev-cmd@v1" is called with an invalid architecture alias, it should be amd64_arm64 to signify crosscompiling from x64 to arm64, so I believe the dev console is not correctly configured.       
    
Edit: furthermore from the log it seems that icu is not build for arm64,

sebgod:
Ah yes that would explain why it can't find the conpiler

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

hessenfarmer:
> Ah yes that would explain why it can't find the conpiler Get Outlook for Android<https://aka.ms/AAb9ysg>
> […](#)
>

Please do not provide any spam links here. 
Edit: ok I understand it is from your email, but please remove such signature links

.github/workflows/build_windows_msvc.yaml
```diff
@@ -9,9 +9,9 @@ jobs:
       fail-fast: false
       matrix:
         config: [Debug, Release]
-        arch: [x64, x86]
+        arch: [x64, x86, arm64]
```
hessenfarmer, line 12:
```suggestion
        arch: [x64, x86, arm64]
        include:
          - arch: x64
            archset: x64
          - arch: x86
            archset: x86
          - arch: arm64
            archset: amd64_arm64 
```

hessenfarmer: we could try this and use the archset in calling the "ilammy/msvc-dev-cmd@v1" action

.github/workflows/build_windows_msvc.yaml
```diff
@@ -9,9 +9,9 @@ jobs:
...
     name: Windows ${{ matrix.config }} ${{ matrix.arch }} Build (MSVC)
-    runs-on: windows-2022
+    runs-on: windows-2025
```
hessenfarmer: although this has to happen anyway in the future it does not help for this purpose

hessenfarmer:
@sebgod     
I did not want to insult you, but I am somewhat upset anout links in posts here.  So sorry if my reaction felt harsh to you.        
   
On topic: Did you make any progress? Meanwhile you might consider to add my proposed changes to your PR to see how our CI would react.

sebgod:
Applied those suggestions. Couldn't test this workflow in my fork because it fails to download the non-existent stuff from the cache

hessenfarmer:
> Applied those suggestions. Couldn't test this workflow in my fork because it fails to download the non-existent stuff from the cache

I'll activate the wrkflow now, so let's see what happens

hessenfarmer:
ok we are one step further.    
next peoblem is to have a arm64 tinygettext (and probably eris as well) variant.

hessenfarmer:
I just learned that github is providing a beta "windows-11-arm" runner. so instead of ading it to our normal MSVC workflow we should probably insert a new MSVC_Arm workflow.

hessenfarmer:
I finally succeeded to build an arm64 version with the github windows-11-arm runner unsing MSYS2.     
see https://github.com/widelands/widelands/pull/6911       
it still needs polishing. However I really would appreciate if @sebgod could test the build executable 
(https://github.com/widelands/widelands/actions/runs/19178003152/artifacts/4503135733) (you need to have the actual data directory in same directory as the exe - it doesn't work on my machine being marked as incompatible so I belive it might work on yours)
and the build installer 
(https://github.com/widelands/widelands/actions/runs/19178003152/artifacts/4503159143) (works on my machine so it might not work on arm machines or only in compat mode)

sebgod:
> I finally succeeded to build an arm64 version with the github windows-11-arm runner unsing MSYS2. see #6911 it still needs polishing. However I really would appreciate if @sebgod could test the build executable (https://github.com/widelands/widelands/actions/runs/19178003152/artifacts/4503135733) (you need to have the actual data directory in same directory as the exe - it doesn't work on my machine being marked as incompatible so I belive it might work on yours) and the build installer (https://github.com/widelands/widelands/actions/runs/19178003152/artifacts/4503159143) (works on my machine so it might not work on arm machines or only in compat mode)

Hey this is good news I will try this out later this week, sorry I am a bit held up with `$DAYJOB`

hessenfarmer:
> 
> Hey this is good news I will try this out later this week, sorry I am a bit held up with `$DAYJOB`

@sebgod may I ask if you found some time to test the armbuild?   
in any case would you mind closing this PR?

ArminiusTux:
Thank you for releasing native arm64 builds for Wndows. 
I will update the link at [this repository](https://armrepo.ver.lt/?search=Widelands).

PR #6910, opened 2025-11-05
WIP: add windows arm support to build matrix using native windows arm

hessenfarmer:
### Type of Change
new Architecture / OS support

### New Behavior
Windows Os on arm architecture is natively supported. 

### How it Works
try to compile Widelands natively in an Windows 11 Arm environment

### Possible Regressions
build process


hessenfarmer:
We need to find out why tinygettext is not compiling here correctly. Unfortunately this is too complicated using the CI, so I close this for now.

PR #6929, opened 2025-12-14
Fix Russian translate.

veffhz:
### Type of Change
Translate…

In Russian language this word is wrong - "**рарушении**".
And reads more harmoniously " в развитии, а не в разрушении".

### Screenshots

image


hessenfarmer:
Thanks for the input,
but transalations should be fixed by using our internalization tool Transifex.   
see https://www.widelands.org/wiki/TranslatingWidelands/      
otherwise the translations will be overwritten by our translationsbot.

PR #6941, opened 2025-12-22
Do not disable network-irrelevant add-ons in multiplayer client

### LINKED ISSUE #6940: add-on not available in multiplayer, probably if i am not hosting
[frankystone](https://codeberg.org/frankystone)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/issues/5300)*
*Created on Mon Dec 22 00:40:43 CET 2025 by [**frankystone**](https://codeberg.org/frankystone)*

---

More a question than an issue:

Today i played again with <@>hessenfarmer and <@>janus, hessenfarmer hosted the game. Although i had the add-on "Where are my heroes" enabled the button for add-ons wasn't available for me.

If i start a lan game with some ai players, the button is available, i guess because i am the host. In a game with <@>janus which i hosted, the add-on button was available.

Is this intended? This might gives the hoster in multiplayer some advantage because he can use an add-on while the other players don't.

bunnybot:
[Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/5301)*
*Created on Mon Dec 22 10:55:30 CET 2025 by [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese)*

---

### Type of Change
Bugfix

### Issue(s) Closed
Fixes #6940

### To Reproduce
1. Start a multiplayer game where a client has some plugins enabled (e.g. where are my heroes, UI inspector, etc)
2. Plugins are not loaded

### New Behavior
The game client disables all add-ons and receives the list of add-ons to enable from the host.
The host already excluded non-network-relevant add-ons (i.e. there is no reason why you should be forced to use e.g. the host's Theme add-on).
Now the client also excludes non-network-relevant add-ons from the checks and keeps them enabled.

### Possible Regressions
Add-on mismatches in multiplayer games


hessenfarmer:
ok after some thought I see the following solutions:
1. just send a message about the enabled addons and that using any addon might be unfair so 2 players need to agree about using them or not. 
2. enhance this by hashing the "official" addons from the server, and distribute the hash with the networkl game. So players could agree to use only officail addons.      
3. disable all addons and distribute and enable all addons from the hosts to the clients.
4. keep the current implementation and discard the PR 
5. use the PR without any modification

I would prefer 2.
1. and 5. are to much prone to discussions about cheating.
3. might imply security issues. 
4. has the same issue as 1 without being detectable.

hessenfarmer:
While typing an answer I realized that my original idea with hashing is still not proof to prevent cheating. So I agree to what we did realize meanwhile on a couple of occassions: It is impossible to prevent well crafted cheating in this game.   
so I am happy with option 1 here as well

hessenfarmer (approved):
code LGTM.
tested and working
@bunnybot update merge

PR #6950, opened 2025-12-28
Set MacOS builds deployment target to 11.0

SirVer:
### Type of Change
Bugfix

### Issue(s) Closed

build_app.sh contained a small logic bug: The problem was that when building on newer macOS runners (15.x), the script was incorrectly setting the deployment target to match the runner's OS version instead of keeping it at the intended minimum version.

If the bug/feature does not appear on the issue tracker: a clear and concise description of what your change does *and why it is desirable*.

### To Reproduce / Before

- Script set deployment target to 12.3
- When macOS 12.3 SDK wasn't found on GitHub Actions runners, it reset deployment target to current OS version (15.x)
- Result: binaries required macOS 15.7+

### New Behavior

- Deployment target is set to 11.0 and stays at 11.0
- Even when using a macOS 15 SDK (because that's what's available), the deployment target remains 11.0
- Result: binaries will work on macOS 11.0 and newer

This is replicating the manual fix that worked for me, i.e.  `vtool -set-build-version macos 11.0 15.5`. This changes the deployment target to 11.0, while keeping the SDK at 15.5.

### Possible Regressions / Side Effects

- DMG filenames will now include "11.0" instead of "12.3" (since we are using `widelands_${OSX_MIN_VERSION}_${WLVERSION}.dmg`)
- Binaries will be compatible with macOS 11.0 Big Sur and newer (released in 2020)


SirVer:
Since we care about the changes on GitHub, I could not test this locally. We have to see if building the artifacts works and if it indeed improves the situation.

SirVer:
Downloading the created assets from the builds, then "unquarantiying" them using `xattr -d com.apple.quarantine Widelands.app` as described on https://www.widelands.org/wiki/Download/ worked for me; so I think the fix works and should be merged.

bunnybot (commented):
[Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/5310#issuecomment-9362498)*
*On Tue Dec 30 12:19:42 CET 2025, [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese) **approved** this pull request:*

<@>bunnybot merge force

SirVer:
That is my understanding, though I did only test in my system. Am 30.12.2025 um 12:31 schrieb bunnybot ***@***.***>:﻿
@bunnybot commented on this pull request.


Mirrored from Codeberg
On Tue Dec 30 12:19:42 CET 2025, Benedikt Straub (Nordfriese) approved this pull request:
<@>bunnybot merge force

PR #6955, opened 2025-12-30
Add Shift modifier to default shortcuts for showhide soldiers, buildings, workareas

bunnybot:
[Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/5315)*
*Created on Tue Dec 30 17:14:11 CET 2025 by [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese)*

---

### Type of Change
Accessibility tweak

### Issue(s) Closed
As mentioned in https://www.widelands.org/forum/topic/6403/

These shortcuts are rarely used and too easy to touch by accident.

### New Behavior
Added Shift to the default shortcuts.

This has no effect if your Widelands installation already uses customised shortcuts. So effectively only on clean new installs.

### Possible Regressions
Keyboard shortcuts


hessenfarmer:
especially with the chat in multiplayer I ran into the issue of accidentally invoking the legacy shortcuts quite often, due to wrong focus as @cb/tothxa mentioned, so we shold have a modifier for all of them imho

hessenfarmer:
ok I agree on Alt be better then shift.     
so having all of these shortcuts having an alt modifier might be the compromise, correct?

PR #6985, opened 2026-02-19
Handle errors in add-on tribe init gracefully

carli2:
## Summary
- Wraps individual add-on tribe loading in `get_all_tribeinfos()` with a try-catch
- A broken add-on tribe (e.g. missing dependency) now logs a warning instead of crashing the entire tribe discovery
- Other working add-on tribes still appear in the tribe selection

## Test plan
- [ ] Install two tribe add-ons, one with a deliberate error in its init.lua
- [ ] Verify the broken add-on is skipped with a log warning
- [ ] Verify the working add-on still appears in tribe selection

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/logic/map_objects/tribes/tribe_basic_info.cc
+++ b/src/logic/map_objects/tribes/tribe_basic_info.cc
@@ -142,7 +142,12 @@ AllTribes get_all_tribeinfos(const AddOns::AddOnsList* addons_to_consider) {
 			for (const std::string& tribe : g_fs->list_directory(dirname)) {
 				const std::string script_path = tribe + FileSystem::file_separator() + "init.lua";
 				if (g_fs->file_exists(script_path)) {
-					tribeinfos.emplace_back(lua.run_script(script_path));
+					try {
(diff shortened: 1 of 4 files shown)


hessenfarmer:
how is it safeguarded that nothing has changed in the data when the script is failing.

hessenfarmer:
@wl/Nordfriese no response on that from @carli2      
how shall we proceed ?

carli2:
Sorry for the late response, and thanks for pointing this out. This is now safeguarded in 7609482. Each add-on tribe is loaded using a separate Lua interface, so state changes made before an exception cannot affect subsequently loaded tribes. The insertion into tribeinfos is also exception-safe, so only fully constructed tribes are added.

hessenfarmer:
Ok I tested it a bit now:
1. this is masking an error that is clearly indicated to the user in current master.
2. If you have more then one addon tribe activated this leads to an error when starting the game, which might be misleading now (had europeans and gauls activated and added a non existent starting condition to the gauls) see screenshot.
so it does not improve anything in the current state
grafik

Issue #895, opened 2019-09-09
FPS and rendering speed affect the simulation speed of widelands


bunnybot:
I just found out that gamespeed and fps have an impact on the logic of Widelands. The following scripts should behave exactly the same.

wl.Game().desired_speed = 1000 * 1000   -- 1000x game speed
sleep(2000)

wl.Game().desired_speed = 1000   -- 1x game speed
sleep(2000)

but they do not. Depending on fps value and the power of your machine the game while yield a little bit sooner or a little bit later. This is not ideal, we should somehow make sure that the game always simulates the same - I have no good idea how to do this though. Any suggestions?

ps: the reason why this matters is for automated tests like the scripting framework I am currently building. Some tests run for a long time and you have to add in a bunch of thresholds if you want to speed up the game and still have realiable tests.

------------------------------------
Imported from Launchpad using lp2gh.

 * date created: 2013-10-07T07:19:34Z
 * owner: sirver
 * the launchpad url was https://bugs.launchpad.net/bugs/1236199

bunnybot:
(by ad-simplypeachy)
When playing some of the 2013 tournament replays I came across several different crashes but none were reproducible. All were at >10x speed, I believe. If it will help I will spend some more time seeing if I can find more info on them.

bunnybot:
(by sirver)
Setting to incomplete for bug sweeping.

bunnybot:
(by janitor)
[Expired for widelands because there has been no activity for 60 days.]

frankystone:
Isn't this been solved by #4319 ?

PR #7009, opened 2026-04-15
Make regression testsuite `stable_save` sleeptime more deterministic

bunnybot:
[Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/5368)*
*Created on Wed Apr 15 19:17:59 CEST 2026 by [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese)*

---

### Type of Change
Bug mitigation

### Issue(s) Closed
This should reduce the frequency of the regression testsuite's random transient failures.

### New Behavior
Many of the tests, especially shipping tests, rely heavily on exact timing. However, `stable_save` has two uncertainties: First it waits an unspecified amount of time for the save to happen and then another unspecified amount of time for the  framerate to stabilise. The latter especially can be anywhere from 0 to 10,000 milliseconds, which makes a huge difference in some tests.

Now we always sleep a fixed amount of time (10 seconds total, can be overridden by the caller) and only append extra sleep time if needed. FPS is purely informational and should hopefully no longer have a serious impact on the regression test logic at least in most cases. (FPS still has a serious impact in UI-related tests, that's bug #895 which is *not* fixed by this.)

One test (destroy ship before picking up ware) needed adapting because it reproducibly always fails now when left unchanged.

### Possible Regressions
The regression tests :)
Let's see what the CI says…


bunnybot (commented):
[tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/5368#issuecomment-13900919)*
*On Mon Apr 27 17:35:55 CEST 2026, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) **approved** this pull request:*

Looks OK. Thanks for fixing!

bunnybot (commented):
[tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/5368#issuecomment-13909580)*
*On Mon Apr 27 23:10:47 CEST 2026, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) commented.*

test/scripting/stable_save.lua
```diff
@@ -23,23 +24,17 @@ function stable_save(game, savename, desired_speed)
...
-   local mapview = wl.ui.MapView()
-   local counter = 0
-   while mapview.average_fps < 20 and counter < 100 do
-      sleep(200)
-      counter = counter + 1
+   -- Wait until save was finished and give the game some time to catch up
+   wake_me(initial_gametime + (desired_wait_time or 10000))
+   while game.last_save_time == last_save_time do
+      print("WARNING: Game save was delayed! Adding extra sleep time")
+      sleep(1000)
```
bunnybot, line 31: [tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/5368#issuecomment-13909577)*
*Created on Mon Apr 27 23:10:47 CEST 2026 by [**Tóth András (tothxa)**](https://codeberg.org/tothxa)*

---

Now I remember, this was the part where I had doubts after the previous timeout:

Can this sleep actually achieve anything? Isn't the save supposed to happen on next game tick after `game:save()`? (i.e. after yielding) And any `sleep()` or `wake_me()` calls should only resume after the game had actually advanced?

If that's so, then isn't all we need is to store the gametime in the same coroutine timeslot where we have `game:save()`, and after some `sleep()` check that the gametime did advance?

bunnybot: [Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/5368#issuecomment-14132957)*
*Created on Fri May 01 09:58:19 CEST 2026 by [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese)*

---

`last_save_time` is a realtimestamp, and saving is handled by the UI thread, so there is an undefined realtime delay between the save command from Lua (in the logic thread) until the UI thread handles the save. Saving can only be done at the end of the current logic frame, and each logic frame's duration (in both realtime and gametime) depends on the current framerate. So yes, if the game is laggy it can take much longer in gametime until the save actually happens… it's a headache :)

bunnybot: [tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/5368#issuecomment-14142044)*
*Created on Fri May 01 15:48:48 CEST 2026 by [**Tóth András (tothxa)**](https://codeberg.org/tothxa)*

---

Oops, sorry.

So how about storing `last_save_gametime_` too, and reporting that to lua instead of the realtime? (I don't think it's used anywhere else)

BTW I did some testing, and it looks like `last_save_realtime_` is not set properly on game loading.

bunnybot: [tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/5368#issuecomment-14143268)*
*Created on Fri May 01 16:04:42 CEST 2026 by [**Tóth András (tothxa)**](https://codeberg.org/tothxa)*

---

Maybe a stupid question, but it looks to me at a quick glance that the gametime is incremented by `CmdQueue::run_queue()`. AFAIK we can save commands (that are due later), so why can't we handle saving before the queue is run?

bunnybot: [Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/5368#issuecomment-14152259)*
*Created on Fri May 01 18:34:28 CEST 2026 by [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese)*

---

> So how about storing `last_save_gametime_` too, and reporting that to lua instead of the realtime? (I don't think it's used anywhere else)

:+1: Good idea, done. Let's see what the CI says.

> Maybe a stupid question, but it looks to me at a quick glance that the gametime is incremented by `CmdQueue::run_queue()`. AFAIK we can save commands (that are due later), so why can't we handle saving before the queue is run?

I don't think it would make any real difference because `Game::think()` is called in a loop, directly from the top-level logic thread loop (`Panel::logic_thread()` → `InteractiveBase::game_logic_think()`). The control flow spends like 99% of the time in `CmdQueue::run_queue()` which does all the actual work of advancing game state, everything else in `Game::think()` and its callees and its caller therefore runs only once between each logic frame (more specifically: at the start of the logic frame if it comes before `run_queue` and at the end of the logic frame if it comes after `run_queue`). Moving the `SaveHandler::think` up before would move the save check from the end of one logic frame to the start of the next one, which is a negligible difference (and the same gametime, since the gametime advances *only* in `CmdQueue::run_queue`). However in the current way this means the save requested from Lua actually happens a bit earlier than it would otherwise, since the Lua script itself is called inside `run_queue` (via a `CmdLuaCoroutine`), so that when `run_queue` returns, any save requested in the current frame is then carried out immediately.

bunnybot (commented):
[tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/5368#issuecomment-15079056)*
*On Sat May 16 21:23:28 CEST 2026, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) submitted a since-dismissed review:*

Thank you. This should be OK now.

However I noticed that manual saving doesn't update the times, see code comment for my findings. But that's not a regression, so you may ignore it in this PR.

src/logic/save_handler.cc
```diff
@@ -148,6 +148,7 @@ void SaveHandler::think(Widelands::Game& game) {
 			if (skip_when_inactive_ && UI::Panel::time_of_last_user_activity() < last_save_realtime_) {
 				verb_log_info_time(game.get_gametime(), "Autosave: Skipping due to user inactivity");
 				last_save_realtime_ = realtime;
+				last_save_gametime_ = game.get_gametime();
```
bunnybot, line 151: [tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/5368#issuecomment-15078927)*
*Created on Sat May 16 21:23:29 CEST 2026 by [**Tóth András (tothxa)**](https://codeberg.org/tothxa)*

---

I think these are redundant here, all cases are handled in `save_game()` in lines 259-260.

However `ui/game/main_menu_save_game.cc` uses the `GenericSaveHandler` directly, so manual saves don't update these times. I wonder whether that could be switched to use `game.save_handler().save_game()`.

bunnybot (commented):
[tothxa](https://codeberg.org/tothxa)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/5368#issuecomment-15082494)*
*On Sat May 16 23:32:23 CEST 2026, [**Tóth András (tothxa)**](https://codeberg.org/tothxa) **approved** this pull request.*

PR #7045, opened 2026-06-13
Validate server-supplied paths in add-on downloads

qoole:
### Type of Change
Bugfix (security hardening)

### Issue(s) Closed
Not currently on the issue tracker.

The add-on download routines in `src/network/net_addons.cc` write files using directory and file names received verbatim from the add-on server, without validating them. Because the connection is not encrypted, a malicious or compromised server — or a man-in-the-middle — can send names containing `..` or path separators and cause files to be written outside the intended destination directory (path traversal / arbitrary file write). The upload path already validates every name via `name_valid()` in `gather_addon_content()`; the download path did not.

### To Reproduce
1. Point the client at an add-on server (or intercept the unencrypted connection) that returns a directory or file name such as `../../foo` for an add-on, translation set, or screenshot.
2. Download it via `download_addon()`, `download_i18n()`, or `download_screenshot()`.
3. The file is written outside the add-ons directory.

### New Behavior
Every server-supplied path component is validated before it is used to build a filesystem path. If any component is invalid the download is aborted: `download_addon()` and `download_i18n()` raise a warning, and `download_screenshot()` returns an empty string (matching its existing error contract). Nothing is written outside the destination directory.

### How it Works
- `name_valid()` (already used on the upload side) is now applied to every server-supplied file name in `download_addon()`, `download_i18n()` and `download_screenshot()`.
- A new helper `relative_path_valid()` validates multi-component directory names one component at a time, so legitimately nested add-on directories keep working while `..`, absolute paths and embedded separators are rejected.

### Possible Regressions
None expected. Uploads reject any name that fails `name_valid()` (`IllegalFilenamesException`), so every add-on stored on the server consists solely of valid names. The new download-side validation therefore cannot reject an add-on that was legitimately uploaded.


bunnybot (commented):
[Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/5404#issuecomment-17411303)*
*On Sat Jun 13 22:01:09 CEST 2026, [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese) commented:*

Hi and thank you for this patch! Great catch :)

LGTM, tested and working, one suggestion inline.

<@>bunnybot trust all

src/network/net_addons.cc
```diff
@@ -108,6 +108,25 @@ inline int portable_read(const int socket, char* buffer, const size_t length) {
...
+
+// A relative path supplied by the server (e.g. an add-on subdirectory) is
+// accepted only if each of its components is a valid name. This also rejects
+// empty components, absolute paths and parent-directory references, and thus
+// prevents writing outside the destination directory via path traversal.
+[[nodiscard]] inline bool relative_path_valid(const std::string& path) {
+	if (path.empty()) {
+		return false;
+	}
+	for (size_t start = 0, end = 0; end <= path.size(); ++end) {
```
bunnybot: [Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/5404#issuecomment-17411240)*
*Created on Sat Jun 13 22:01:09 CEST 2026 by [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese)*

---

I think this could be done more readable using the `split()` function from `base/string.h`:
```cc
/** Split a string into substrings at all occurrences of any of the given delimiters. */
void split(std::vector<std::string>& result, const std::string&, const std::set<char>&);
```

hessenfarmer:
@qoole do you plan to implement the suggested change, so we could merge this?

src/network/net_addons.cc
qoole: Implemented in 28879246 — `relative_path_valid()` now uses `split()`.

One detail: `split()` iterates via `c_str()` and stops at an embedded NUL byte, whereas the previous loop validated the full string (a NUL fails `name_valid()`, so such paths were rejected). Since `read_line()` can return strings containing NUL bytes, I kept an explicit rejection of paths containing `'\0'` so the part of the string after a NUL cannot go unvalidated. Behaviour is otherwise identical — I differential-tested the two implementations against each other on directed edge cases (absolute paths, `a//b`, trailing separators, `..` variants, backslashes, over-long components) plus a 2M-case random fuzz; they agree on all inputs.

qoole:
@hessenfarmer Done — the suggested change is implemented in 28879246 (see the inline thread for one note about embedded NUL bytes).

bunnybot (commented):
[Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/5404#issuecomment-18592619)*
*On Sun Jul 05 11:04:05 CEST 2026, [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese) **approved** this pull request:*

Thank you :)

<@>bunnybot update merge

hessenfarmer:
thank you 
@qoole unfortunately our clang-tidy tool suggests to replace the for loop in line 125 by 

for (const std::string& component : components) {
		if (!name_valid(component)) {
			return false;
		}
	}
	return true;

to be replaced by something like

return std::all_of(components.begin(), componentsend(),
	                   [this](const auto& component) { return name_valid(component); });

hessenfarmer:
furthermore @qoole how do you want to be credited nickname, realname or both

hessenfarmer:
@qoole would you be so kind to fix the clang_tidy issue so this could go in?
and please answer my question about your credits

hessenfarmer:
@qoole 
Any intent to further contribute on this PR?

qoole:
@hessenfarmer Apologies for the delay - both points are addressed now.

**clang-tidy:** the loop is replaced with `std::all_of` in d60086e2, plus the `<algorithm>` include. Two small deviations from your snippet: `components.end()` (yours had a typo), and no capture - `relative_path_valid` is a free function in the anonymous namespace, so `[this]` wouldn't compile there.

```cpp
return std::all_of(components.begin(), components.end(),
                   [](const std::string& component) { return name_valid(component); });
```

I couldn't do a full build locally (missing SDL2_image/mixer/ttf and glew here), so I checked it separately: compiled the new expression against a verbatim copy of `split()` from `base/string.h` under `-std=c++17 -Wall -Wextra`, and diffed old vs new over a 33-path corpus - traversal attempts, embedded NULs, mixed separators, the 80-char boundary. Identical results, and the traversal cases are still rejected. Worth a CI confirmation on your side all the same.

**Credits:** nickname only, please - `qoole`.

hessenfarmer:
@bunnybot trust all

bunnybot (commented):
[Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/5404#issuecomment-22416631)*
*On Sun Sep 06 11:44:52 CEST 2026, [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese) **approved** this pull request.*

hessenfarmer:
@bunnybot merge force

PR #7064, opened 2026-09-10
Escape pinned note text before rendering as richtext

### LINKED ISSUE #7063: Crash when pinned note includes less than sign
### Describe the Bug

I went to add a Pinned Note to a place where my engineer had found marble, but instead of 'M' I entered '<'.  Widelands immediately crashed and won't reload the emergency save.

```
[05:20:32.320 real] ERROR: ##############################
[05:20:32.320 real] ERROR:   FATAL EXCEPTION: Syntax error at 1:72: expected an allowed tag, got '</font'. String continues with: '<</font></p></rt>'
[05:20:32.320 real] ERROR: ##############################
[05:20:32.330 real] ERROR:   Please report this problem to help us improve Widelands.
[05:20:32.330 real] ERROR:   You will find related messages in the standard output (stdout.txt on Windows).
[05:20:32.330 real] ERROR:   You are using version 1.4~git27301 (a727165@master) Release.
[05:20:32.330 real] ERROR:   Please add this information to your report.
[05:20:32.330 real] ERROR:   If desired, Widelands attempts to create an emergency savegame.
[05:20:32.330 real] ERROR:   It is often ? though not always ? possible to load it and continue playing.
[05:20:32.330 real] ERROR: ##############################
[05:20:32.333 real] ERROR: Invalid richtext: Syntax error at 3:58: expected '>', got '''. String continues with: ''. String continues with: '<</font></p></rt>'
[05:20:32.333 real] ERROR: Please report this problem to help us improve Widelan'
```

### To Reproduce

I went to add a Pinned Note to a place where my engineer had found marble, but instead of 'M' I entered '<'.  Looks like the location is irrelevant, the text isn't.

### Expected Behavior

Not a crash.

### Crash Log

```text

```

### Screenshots

_No response_

### Replays

_No response_

### Savegames

The crash save:
[2026-09-09T12.38.57.wgf.gz](https://github.com/user-attachments/files/32018395/2026-09-09T12.38.57.wgf.gz)

A save which can easily (for me) reproduce:
[wl_autosave_00.wgf.gz](https://github.com/user-attachments/files/32018741/wl_autosave_00.wgf.gz)

### Operating System

Devuan 6

### Widelands Version

1.4~git27301 (a727165@master)

### Enabled Add-Ons

I was playing Australia-v2.wmf  

```
spartan$ ls maps/Downloaded/
Australia-v2.wmf  Flower_Power.wmf  Mountain_Labyrinth.wmf  Strangers_Pass.wmf
spartan$ ls addons
Coastal_Mountainsides.wad            foreign_planet.wad
Down_the_Mountain.wad                Planetae_Numero_Duos_Mk_2_1.wad
Fjords_between_wooded_mountains.wad  Volcania_final.wad

```

### Additional Context

I also tried '<N>' for kicks and that crashed too.

jadhavgaurav:
### Type of Change
Bugfix

### Issue(s) Closed
Fixes #7063

### To Reproduce
1. Start a game and place a pinned note.
2. Edit the note's text so it contains a character with special meaning in richtext, for example `<`.
3. Enable the census overlay (or otherwise cause the note's label to be drawn on the map).
4. Widelands throws an uncaught richtext parse error and crashes.

### New Behavior
The note's text is escaped with the existing `richtext_escape()` helper before it is handed to the richtext renderer, so characters such as `<`, `>` and `&` are displayed literally instead of being interpreted as markup. Notes with such text no longer crash the game.

### How it Works
`PinnedNote::draw()` builds the on-map label by wrapping the note's raw text in richtext font tags. The raw player-typed text was passed through unescaped, so a stray `<` broke the surrounding markup and the renderer threw. This mirrors how other free-form strings that get rendered as richtext elsewhere in the codebase (list entries, chat messages, tooltips) are already escaped before rendering.

### Possible Regressions
None expected. The note's stored text is unchanged; only the string handed to the renderer is escaped, so editing an existing note still shows and round-trips the original text.

### Additional context
Extended the existing `test/maps/plain.wmf/scripting/test_pinned_notes.lua` test to turn on the census overlay before setting a note's text to an unescaped string containing `<`, so the crash is actually exercised. Confirmed the test fails with the exact reported crash (`FATAL EXCEPTION`, exit code -6) when only the source fix is reverted, and passes with the fix applied. Also ran the full `plain.wmf` scenario test suite (20 tests, including this one and the pre-existing `test_text_renderer_does_not_crash.lua`) and the C++ unit test suite (`ctest`), all passing.

### DIFF
--- a/src/logic/map_objects/pinned_note.cc
+++ b/src/logic/map_objects/pinned_note.cc
@@ -21,6 +21,7 @@
 #include <memory>
 
 #include "graphic/rendertarget.h"
+#include "graphic/text_layout.h"
 #include "logic/game_data_error.h"
 #include "logic/player.h"
 #include "ui/wui/interactive_player.h"
@@ -110,7 +111,11 @@ void PinnedNote::draw(const EditorGameBase& egbase,
 
 	dst->blit_animation(field_on_dst, coords, scale, owner().tribe().pinned_note_animation(),
 	                    egbase.get_gametime(), &rgb_);
-	do_draw_info(info_to_draw, text_, std::string(), field_on_dst, scale, dst);
+	// The note text is free-form player input and is rendered as richtext below,
+	// so any characters that richtext treats specially (e.g. '<') must be escaped
+	// first. Without this, a note containing such a character throws an uncaught
+	// richtext parse error and crashes the game.
+	do_draw_info(info_to_draw, richtext_escape(text_), std::string(), field_on_dst, scale, dst);
 }
 
 constexpr uint8_t kCurrentPacketVersion = 1;
--- a/test/maps/plain.wmf/scripting/test_pinned_notes.lua
+++ b/test/maps/plain.wmf/scripting/test_pinned_notes.lua
@@ -11,6 +11,11 @@ run(function()
    assert_equal("Hello World", note.text)
    assert_equal(123, note.color[3])
 
+   -- Show the census overlay so the note's text actually gets rendered as
+   -- richtext on the map. Regression test for a crash when the note text
+   -- contains a character with special meaning in richtext, e.g. '<'.
+   wl.ui.MapView().census = true
+
    note.text = "<Richtext injection attempt>"
    note.color = {50, 200, 250}
    sleep(5000)
@@ -18,6 +23,8 @@ run(function()
    assert_equal("<Richtext injection attempt>", note.text)
    assert_equal(50, note.color[1])
    assert_equal(1, #f.bobs)
+
+   wl.ui.MapView().census = false
 
    sleep(1000)
    note:remove()


bunnybot (commented):
[Nordfriese](https://codeberg.org/Nordfriese)*[Mirrored from Codeberg](https://codeberg.org/wl/widelands/pulls/5423#issuecomment-22635481)*
*On Thu Sep 10 17:50:43 CEST 2026, [**Benedikt Straub (Nordfriese)**](https://codeberg.org/Nordfriese) **approved** this pull request:*

LGTM, thanks for the patch :)

hessenfarmer:
@jadhavgaurav as you contributed to widelands how do you want to be credited:
Nickname, realname or both

